PR #12, opened 2017-12-11
Update Tile.java

Raotor123:
I tested this on my desktop and it seems to fix the problem. Found out because you commented that right is the dump, but you do a left-right.

### DIFF
--- a/core/src/io/anuke/mindustry/world/Tile.java
+++ b/core/src/io/anuke/mindustry/world/Tile.java
@@ -143,7 +143,7 @@ public class Tile{
 	}
 	
 	public byte getDump(){
-		return Bits.getLeftByte(Bits.getRightByte(data));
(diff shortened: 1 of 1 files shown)


Anuken:
Sorry, that comment was slightly outdated and misleading. After it was written, I had changed it to have the 'block dump' be saved in the left byte _of the rightmost byte of data_. I don't think your change is correct; now, it performs the same thing as _getExtra_ the line below it, which shouldn't happen.

I also can't seem to reproduce your problem; can you post a screenshot of the scene where it happens?

Raotor123:
When I place a router there i get the ArrayOutOfBounds, but it did not happen when i placed the previous router to the left. I also discovered that a tunnel without a gap does not work (this could be intentional?) like in the screenshot. And if you place a router infront of a tunnel nothing goes through.
image

PR #22, opened 2017-12-16
Made crafter support multiple recipes

TheChosenEvilOne:
Pretty self explanatory

This might be unoptimized, will look into optimizing it if possible


Anuken:
I'm not entirely sure why this is necessary. Why not just add another block if multiple recipes are needed? Funneling 3+ types of resources into one multi-recipe crafter would be a hassle, and so would processing the outputs.

TheChosenEvilOne:
I never meant that this would do multiple recipes in one block (as it is not made for that), just that you could use one block type (note the word type) to do multiple recipes.

PR #102, opened 2018-03-16
add an info button for descriptions of the difficulty mode.

rangzen:
I refactor the function but maybe you don't like it like that.


Anuken:
- The refactoring is fine, but you should use `io.anuke.ucore.function.Consumer` instead of `java.util.function.Consumer`, as older Android versions don't support anything in that package.
- The information listed doesn't seem very useful: it would be preferable to see exact numbers of time between waves and extra blocks targeted for each difficulty.

rangzen:
Using the Consumer of ucore is easy to fix :) but to be precise about difficulty is way more complex:
- values are in multiple File (Logic, Vars, etc.)
- functions too and can be private, I mean the results of calculus.
- heuristic is a Predicate<Block> so you have to know every child of Block (introspection yerk) to check with a predicate formed of instanceof (yerk 2 :) )
I you want to print the actual data with current calculus, you can't, if you want to print it from static calculus, great chances that it will be false quick :/

Anuken:
I won't be changing anything related to difficulty until the 4.0 release, which won't be public for at least 2 
 more months, so I don't think having hard-coded difficulty information is a bad thing. When 4.0 comes, everything related to waves, difficulty and gamemode will be completely re-coded anyway, so there's not much use making it use the predicate and wave time data anyway.

Anuken:
Closing as dead.

PR #110, opened 2018-03-28
File choosing assistance

cpaca:
When choosing files, what used to happen is if I, say, had a file named "A" and file named "a" they would not appear together because that is *technically* what alphabetical order says. Other than that, it would work just fine in "alphabetical" order, ABCDEFabcdef.  
However, in Mindustry font, caps and lower looks the same. So it looks really weird: ABCDEFABCDEF.  
  
This change makes it so it doesn't matter the caps/un-caps, it will sort the above files as AABBCCDDEEFF.  
(The slight problem comes when someone has two files with the same name but different caps at the start, but that seems like a less common problem, I feel.)  
  
That's the change that has been made here.  
I feel like it's fairly inefficient, so I hope you find a better source in the future.  
It also replaces a foreach loop with a fori loop, but there (might) be a better way to do it in the future.

### DIFF
--- a/core/src/io/anuke/mindustry/ui/dialogs/FileChooser.java
+++ b/core/src/io/anuke/mindustry/ui/dialogs/FileChooser.java
@@ -201,13 +201,39 @@ public class FileChooser extends FloatingDialog {
 		
 		ButtonGroup<TextButton> group = new ButtonGroup<TextButton>();
 		group.setMinCheckCount(0);
-
-		for(FileHandle file : names){
+		
+		////////////////////////////////////////////////////////////////
+		// Contributer's name (On discord)://///////////////////////////
+		// we need moar missiles#6435 //////////////////////////////////
+		/*Sorting time!*/
+		String[] sortedNames = new String[names.length];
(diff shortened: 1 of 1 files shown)


TheChosenEvilOne:
Just make this force uppercase or lowercase on everythinv

cpaca:
TheChosenEvilOne I tried doing that, but it didn't change the files at all. Some digging found it finds the files using badlogic which I can't move around, so I just had to make a huge code mess to fix the problem.  
  
In all likelihood, there's a better way. But it's what I came up with in one day.

TheChosenEvilOne:
I can look into a better way when I come home

Anuken:
Why are you sorting it twice? The sorting comparator in `FileChooser#getFileNames` already does this for you, and that's what should be edited. See https://github.com/Anuken/Mindustry/commit/f0302ca7dc41614135bdd4541e4b049054b0a362.

*Side note: In the future, please don't add your name as a giant comment like that. Create a new method and use the javadoc `@author` attribute; or, if creating a new method doesn't work well, add yourself as one of the `@author`s of the class.*

PR #112, opened 2018-04-03
Rotational replace, noclip & infinity reach in desktop sandbox

juh9870:


TheChosenEvilOne:
Noclip feels like something that should not be added to sandbox. Infinite reach could fit

Anuken:
- Noclip is not something I believe should be added to the game, at least not while the player is still a *mech*. Most users will probably regard it as a bug rather than a feature when they see it.
- By changing how replacing works, you've broken replacement for all other blocks (e.g. drills can can no longer be replaced by omnidrills, conveyors can't be replaced by routers)
- Most of the replacement and place-checking code has been rewritten in 4.0 to be cleaner, and merging this would cause conflicts later in that branch. Adding rotation-replacement for *all* rotating blocks there would only require changing 2 lines in 2 different files, so I would rather implement it myself in the 4.0 branch.
- Infinite reach is fine.

TheChosenEvilOne:
any news on this?

juh9870:
Go on, I don't know when I will have time to make another pull request

TheChosenEvilOne:
ill implement this tomorrow

TheChosenEvilOne:
Ill do it tomorrow probably

Anuken:
Implemented infinite reach as of 6e9d833c77a568db1857e6beba4ba979a820f9e0.

PR #121, opened 2018-05-04
Update translate file to Mindustry 4.0

cloud9350:
Translate by Mine5831 -> Gureumi (Full re-translated)


Anuken:
I think you meant _build 38_, not 4.0?

PR #125, opened 2018-05-12
Added edit logging and whole world rollback

Commodore64x:
As the title suggests, this PR adds blocks logs accessible to everyone in a multiplayer game by holding down `i` and clicking the tile you wish to display edit information about. From console, using the `rollback <amount>` command you can rollback the changes to each individual tile the amount of times you specified


TheChosenEvilOne:
this looks good but how much memory will this eat?

iczero:
You could always store only the last ~1000 changes or something if you are afraid of memory usage. I don't see how changelogs will eat that much RAM

Commodore64x:
Ok, @Anuken I've implemented basically everything, mind having another look over?

core/src/io/anuke/mindustry/net/EditLog.java
```diff
@@ -30,6 +34,15 @@ public static EditLog valueOf(String string) {
 						   EditAction.valueOf(parts[3]));
 	}
 	
+	public static Array<EditLog> logsFromTile(int x, int y) {
```
Anuken: This whole method can be changed to `netServer.admins.getEditLogs().get(x + y * world.width(), new Array<>());`.  (Why are you checking each entry sequentially? This defeats the whole purpose of using a map.)

core/src/io/anuke/mindustry/net/Packets.java
```diff
@@ -140,51 +140,36 @@ public void read(ByteBuffer buffer) {
...
 				
-				ArrayList<EditLog> values = editlog.value;
-				buffer.putInt(values.size());
-				for(EditLog value : values) {
-				 
-					String rawValue = value.toString();
-					buffer.put((byte) rawValue.getBytes().length);
-					buffer.put(rawValue.getBytes());
-				}
+				String rawValue = value.toString();
```
Anuken: You're still writing the edit log as a string, which is space-inefficient. You should write two ints for player ID and block ID and two bytes for rotation and action ordinal instead.

core/src/io/anuke/mindustry/net/Packets.java
```diff
@@ -140,51 +140,36 @@ public void read(ByteBuffer buffer) {
...
 		@Override
 		public void write(ByteBuffer buffer) {
-			int size = editlogs.size;
-			buffer.putInt(size);
-			for(IntMap.Entry<ArrayList<EditLog>> editlog :editlogs.entries()) {
-				
-				String key = String.valueOf(editlog.key);
-				buffer.put((byte) key.getBytes().length);
-				buffer.put(key.getBytes());
+			buffer.putInt(x);
```
Anuken: Not much of an issue, but you can write the position as a `short` instead of an `int`, since map sizes never get near `short_max`.

Anuken:
Due to potential compatibility breaks, I won't be merging this until the iOS version is out. The new key-based admin system needs to be implemented as well.

Could you look into implementing a simple number-input rollback menu for server admins? I don't see it being very useful on most servers otherwise, as the owner would need to have constant console access for it to be used against griefers.

Commodore64x:
@Anuken I've commited but the spacing is still off, I have no idea why it is like that.

PR #127, opened 2018-05-31
Update bundle_ko.properties

cloud9350:
Early update language file!
may be? :smile:


Anuken:
I just want to point out that all of the block descriptions and stats are currently unfinished, so it might be best to wait next time.

PR #133, opened 2018-06-19
fix fatal language translate (bundle ko)

cloud9350:
fix no show server version + host user name on list


cloud9350:
It's not critical language translate missing (line 67)
It's ok :worried:

JaiunL:
I understand that this is not a critical point. I am merely highlighting a place where improvements are possible. No offense was intended, if you did feel any.

PR #134, opened 2018-07-04
Change Korean font & full support Korean texts

cloud9350:
### changes
- Replacing fonts for version 4.0 (NanumGothic ExtraBold -> NanumBarunGothic Normal)
- ALL Korean texts support (Korean text **11,724**, English text 94 and extra texts)


Anuken:
Have you tested whether this affects load times? These font files seem to be massive.

cloud9350:
I didn't notice the difference in loading time before and after this patch was applied.
~~using NVMe SSD~~

PR #139, opened 2018-07-12
Update .travis.yml

cloud9350:
not tested
Maybe it can be fix travis build.


Anuken:
Looks like I needed to add build-tools to the Travis android components for it to automatically accept the license. Thanks for trying, though.

PR #157, opened 2018-07-28
Introduce workaround for libgdx issue by @codegist

Ameb:
libgdx has a [buggy piece of code](https://github.com/libgdx/libgdx/blob/9e46b2768e06292f4c9c859ad95ae37bf52bd0ac/backends/gdx-backend-android/src/com/badlogic/gdx/backends/android/AndroidGraphics.java#L392) which introduces crashes into some devices. This happens in some devices as commented in libgdx/libgdx#4626.

Basically if the game is in foreground and you lock the screen the application gets killed by libgdx `AndroidGraphics` when resuming.
```
AndroidGraphics: waiting for pause synchronization took too long; assuming deadlock and killing
```
Luckily @codegist discussed a workaround in libgdx/libgdx#3861. This pull request adds the workaround to Mindustry.

### DIFF
--- a/android/src/io/anuke/mindustry/AndroidLauncher.java
+++ b/android/src/io/anuke/mindustry/AndroidLauncher.java
@@ -12,7 +12,6 @@ import android.provider.Settings.Secure;
 import android.telephony.TelephonyManager;
 import android.util.Log;
 import com.badlogic.gdx.Gdx;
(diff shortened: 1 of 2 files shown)


Anuken:
Just tested, and there don't seem to be any issues with 2 different Android devices.

PR #167, opened 2018-08-01
Update bundle.properties

Prosta4okua:


Anuken:
I'd rather just change it myself at this point.

PR #181, opened 2018-08-11
Super Duper Hard add korean chat fonts with keep orignal english fonts

cloud9350:
ezgif com-optimize
im dead

P.S - Doesn't work korean chat in Linux dedicated server, but work Windows dedicated server. 🤔

left - Linux OS
right - Windows OS


cloud9350:
@Anuken Found many typo mistake + smail english chat font error.
After solving this problem, I will pull request again.

PR #194, opened 2018-08-19
Fix 'has connected' message + Add chat font

cloud9350:
before
Before

after
After

Add chat font (English font work now! --but not work Ukrainian dot font 🤔)


Anuken:
The connected message color tag is necessary to prevent users with colored names from re-coloring the rest of the message. I know a workaround, however.

PR #216, opened 2018-09-06
smail translate

cloud9350:
not support donate description translate in android app


Anuken:
Donation descriptions are part of [another library](https://github.com/PrivacyApps/donations), so translations need to be done by someone over there.

PR #226, opened 2018-09-24
Update bundle.properties

Prosta4okua:


Prosta4okua:
block.liquid-router.name=Liquid Router
block.liquid-tank.name=Liquid Tank
block.liquid-junction.name=Liquid Junction
block.bridge-conduit.name=Bridge Conduit

PR #236, opened 2018-10-13
keep the saved width and height of sector

islq:
width,height changed after expandSector mission, It will reset to original value in initSector()


Anuken:
Nice catch. However, looking at the `generate()` code, sector X and Y may be shifted after loading as well. You'll need to save that with the width and height before `initSector()`.

PR #237, opened 2018-10-13
add BlockBuildCompletEvent

islq:
Missions.blockRecipe can not complete on BlockBuildEvent.
add BlockBuildCompletEvent  and listen on it for content mission


TheChosenEvilOne:
why is it not called BlockBuildCompleteEvent?

Anuken:
Instead of making a new event type, I've made `Control` use `TileChangeEvent` to handle building.

PR #238, opened 2018-10-13
set charset

islq:
default file charset is not utf8 when run from win10


Anuken:
Any particular reason why you needed to run the bundle updater locally..?

PR #245, opened 2018-10-16
adjust buttons width in sector dialog

islq:
the total width of 3 buttons exceed the width of mobile device


Anuken:
The issue with mobile device scaling is that the screen could be both tiny or massive in width, leading to out-of-bounds text on mobile or extremely wide buttons on desktop.

I've fixed this in 82577b91d5c2946e1ddf2146ab77a2b52bceb6b6 by using a 2-row button layout.

PR #258, opened 2018-10-21
Update and standardize translations (POEditor.com)

xgamezs:
Close about an hour before release 59


xgamezs:
Fixed some translation errors.

PR #261, opened 2018-10-23
Simplified power logic

baltitenger:
I've just looked in the code to know what was going on under the hood because I had issues with power, stared at it for a while. PowerGraph seemed oddly complex to me so I made this. There may have been something that I didn't understand or overlooked, so if the complexity was intentional, please tell me why. Tested it and it seems to work just fine.

### DIFF
--- a/core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
+++ b/core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
@@ -39,40 +39,26 @@ public class PowerGraph{
 
         float totalInput = 0f;
         for(Tile producer : producers){
-            float accumulator = producer.entity.power.amount;
-            if (accumulator > 0.0001f) {
-                totalInput += producer.entity.power.amount;
-            }
+            totalInput += producer.entity.power.amount;
         }
 
         float maxOutput = 0f;
         for(Tile consumer : consumers){
-            float canHold = consumer.block().powerCapacity - consumer.entity.power.amount;
-            if (canHold > 0.0001f) {
-                maxOutput += canHold;
-            }
+            maxOutput += consumer.block().powerCapacity - consumer.entity.power.amount;
         }
 
-        if (totalInput == 0 || maxOutput == 0) {
+        if (totalInput <= 0.0001f || maxOutput <= 0.0001f) {
             return;
         }
 
         float inputUsed = Math.min(maxOutput / totalInput, 1f);
         for(Tile producer : producers){
-            float accumulator = producer.entity.power.amount;
-
-            if (accumulator > 0.0001f) {
-                producer.entity.power.amount -= accumulator * inputUsed;
-            }
+            producer.entity.power.amount -= producer.entity.power.amount * inputUsed;
         }
 
         float outputSatisfied = Math.min(totalInput / maxOutput, 1f);
         for(Tile consumer : consumers){
-            float canHold = consumer.block().powerCapacity - consumer.entity.power.amount;
-
-            if (canHold > 0.0001f) {
-                consumer.entity.power.amount += canHold * outputSatisfied;
-            }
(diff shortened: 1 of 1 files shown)


core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -39,37 +38,41 @@ public synchronized void update(){
         lastFrameUpdated = threads.getFrameID();
 
         float totalInput = 0f;
-
         for(Tile producer : producers){
-            totalInput += producer.entity.power.amount;
+            float accumulator = producer.entity.power.amount;
+            if (accumulator > 0.0001f) {
```
Anuken: The general logic seems to be fine here, but most of the if-statements and extra variable declarations are redundant. `totalInput += producer.entity.power.amount` should happen regardless of whether `accumulator > 0.0001f`.

I suggest replacing the entire contents of this for-loop with just `totalInput += producer.entity.power.amount;`.

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -39,37 +38,41 @@ public synchronized void update(){
...
+                totalInput += producer.entity.power.amount;
+            }
         }
 
-        for(Tile producer : producers){
-            float accumulator = producer.entity.power.amount;
+        float maxOutput = 0f;
+        for(Tile consumer : consumers){
+            float canHold = consumer.block().powerCapacity - consumer.entity.power.amount;
```
Anuken: Same here. Redundant variable and if statement.

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -39,37 +38,41 @@ public synchronized void update(){
...
+            if (canHold > 0.0001f) {
+                maxOutput += canHold;
+            }
+        }
 
-            if(accumulator <= 0.0001f) continue;
+        if (totalInput == 0 || maxOutput == 0) {
```
Anuken: While this is enough to catch NaN errors, it should be `<= 0.0001f` to catch situations where the input or output is very small and there's not enough power to warrant updating the flow.

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -39,37 +38,41 @@ public synchronized void update(){
...
 
-            float toEach = accumulator / consumers.size;
-            float outputs = 0f;
+        float inputUsed = Math.min(maxOutput / totalInput, 1f);
+        for(Tile producer : producers){
+            float accumulator = producer.entity.power.amount;
 
-            for(Tile tile : consumers){
-                outputs += Math.min(tile.block().powerCapacity - tile.entity.power.amount, toEach) / toEach;
+            if (accumulator > 0.0001f) {
```
Anuken: Redundant variable and if statement.

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -39,37 +38,41 @@ public synchronized void update(){
...
-            }
+        float outputSatisfied = Math.min(totalInput / maxOutput, 1f);
+        for(Tile consumer : consumers){
+            float canHold = consumer.block().powerCapacity - consumer.entity.power.amount;
 
-            for(Tile tile : consumers){
-                float used = Math.min(tile.block().powerCapacity - tile.entity.power.amount, finalEach) * accumulator / totalInput;
-                buffer += used;
-                tile.entity.power.amount += used;
+            if (canHold > 0.0001f) {
```
Anuken: Same here. Redundant variable and if statement.

PR #264, opened 2018-10-23
Add Scale settings in a rough way...(with one change in uCore)

arnesacnussem:
And Bundle load before Vars.init()  <-  tail now i do not find any problem,but i have not read though the code,if anything wrong,please fix it yourself or tell me :)


Anuken:
- Please fix your formatter, you've added extra spaces next to braces all over the codebase.
- Any particular reason why you're loading the bundle before `Vars.init` now?
- From what I can tell, your problem lies with the base scale of the Chinese font(s), not the scale of all the fonts or UI. Perhaps instead of implementing an option, it would be preferable to just increase the scale of the Chinese font by editing the file?

arnesacnussem:
i'll check my fomatter,i just use the oiginal formate of IDEA

i had to load settings before Vars.init().

PR #266, opened 2018-10-24
Add texts in font file

Kieaer:


Kieaer:
Prosta4ok_ua has requested add more texts.

PR #267, opened 2018-10-24
Add UI,font and base camera scale settings(with one change in uCore)

arnesacnussem:
- i'll keep that font scale opinion,cause maybe someone really need that(i'm one of them).
- bundle did not need load before `Vars.init()` ,i don't know why last time need,maybe something like "try reboot".
- reformat code to "project style".
- uCore change: 
 https://github.com/Anuken/uCore/pull/17


Anuken:
My main issue with this PR is that the user can easily screw up their scale settings (e.g. make the UI so big it doesn't fit on-screen) and lock their game in an unplayable state. It would be preferable to have some way to scale the UI while the game is still running and have a reset countdown, but unfortunately that might be difficulty to implement.

It's also difficult for users to set up the various scales using a slider bar without knowing how they look. I would like it if this was implemented with a spinner box of some sort with preset scales (e.g. "normal", "small", "large", "extra large")

I'm also not sure if camera scale should be an option at all.

arnesacnussem:
- there should be a way to reload all thing while it's running,but i did not find how to and it's easy to make some mistake.

- can we add an command line to reset those settings? such as`java -jar desktop-release.jar -reset`.
or use an command to let those opinions show in game,or just use command to set scale.

- i only implemented desktop's scale(but forget to remove those opinion on mobile,while it's also needed on my android phone)

- camera scale only affect the default status.I was too lazy to use Ctrl + mouse wheel when i back to game.

arnesacnussem:
thought i can add a HUGE dialog on next startup when changed scale.

should have a _confirm_ button that he can save new setting.
or with another _cancel_ button to cancel those changes.
those functions should add to the restart warn.

Anuken:
@Kieaer It's just that I don't like how some things are implemented in this PR, so I'll probably be doing it myself at some point.

arnesacnussem:
> Is this dead? 🤔

quite possible...

****
- i found font scale has been removed from all code,this is one of most important sad thing for me,cause i'm struggling on read all those "nano bricks".
- Anuken seems don't want to implement that "baseCam scale"  i need this cause i'm lazy ,or we can add an implement to store that value in saves(never noticed if it already done....).

- ... and another change in uCore
****

Anuken:
Closing this PR, as Mindustry doesn't use uCore anymore and most of the code is incompatible now.

*I will be implementing this myself, later.*

PR #272, opened 2018-10-25
Incomplete update for turkish translation

CrazyBearTR:
Its incomplete. Still needed to work on it.


Anuken:
Merging this, as an incomplete translation is better than no translation at all. Submit another PR if needed.

PR #274, opened 2018-10-26
Fix typo

Kieaer:
Need sync another bundle files


Anuken:
Fixed it with latest commit across all bundles with project-wide replace.

PR #275, opened 2018-10-26
Add texts in font

Kieaer:
request by Prosta4ok_ua


Kieaer:
And bundle files

Prosta4okua:
I don't know what it is, but it should correct the letters.

PR #281, opened 2018-10-28
Rename bundle_ko.properties to bundle_ko_KR.properties

JaiunL:


Anuken:
I don't think I'll merge this one, as I don't see any problems with the current bundle locale name. This might lead to unnecessary conflicts

PR #285, opened 2018-10-30
Cultivator placement fix

Timmeey86:
This change makes sure that when trying to replace a block by another block of the same type (e.g. replacing any drill by a Cultivator), the replacement is not possible if the new block may not be built in this spot.
Otherwise, it would be possible to e.g. place a Cultivator on a non-grass resource field. That cultivator would then not produce anything since it is not actually placed on a grass tile.

### DIFF
--- a/core/src/io/anuke/mindustry/world/Build.java
+++ b/core/src/io/anuke/mindustry/world/Build.java
@@ -132,7 +132,7 @@ public class Build{
         if(tile == null) return false;
 
         if(type.isMultiblock()){
-            // If a block shall replace another block we need to make sure the drills have the same size and the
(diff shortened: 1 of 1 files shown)


TheChosenEvilOne:
Already fixed, see:
https://github.com/Anuken/Mindustry/commit/5aba065413e886318c52778d91f25b2e18fa9280#diff-9df5a77ba78ad2e1b4f5d1195edfd37bR135

Timmeey86:
Looks like you added the same fix while I was trying to figure out how to create branches & pull requests ;)

Timmeey86:
Closing the pull request since it's already fixed

PR #287, opened 2018-10-31
Update translations and add some languages

xgamezs:


Anuken:
You've moved the new bundle entries to the end of `bundle.properties`, breaking organization and causing a conflict. Please configure your editor to keep the correct order.

PR #288, opened 2018-10-31
Rewritten a lot of power related stuff

baltitenger:
Power generators don't use their buffer now, they are outputting there power directly to the `PowerGraph`.
Also fixed batteries making power exponentially.


core/src/io/anuke/mindustry/world/blocks/power/SolarGenerator.java
```diff
@@ -7,28 +7,16 @@
...
 public class SolarGenerator extends PowerGenerator{
-    /**
-     * power generated per frame
-     */
-    protected float generation = 0.005f;
-
     public SolarGenerator(String name){
         super(name);
         flags = EnumSet.of();
+        powerProduction = 0.005f; //power generated per frame
```
Anuken: This shouldn't be needed anymore, since power is defined in the other solar panel definitions.
Not a problem with your code specifically, but still something that shouldn't be here.

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -107,11 +161,11 @@ public void add(Tile tile){
...
 
-        if(tile.block().outputsPower){
+        if (tile.block().outputsPower && tile.block().consumesPower) {
+            batteries.add(tile);
+        } else if (tile.block().outputsPower) {
             producers.add(tile);
-        }
-
-        if(tile.block().consumesPower){
+        } else if (tile.block().consumesPower) {
```
Anuken: Please fix your formatter; statements should have no extra spaces, e.g. `} else if (` -> `}else if(`.
This occurs throughout the PR.

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -31,12 +35,62 @@ public int getID(){
...
+            return (a.block().powerCapacity - a.entity.power.buffer) < (b.block().powerCapacity - b.entity.power.buffer) ? -1 : 1;
+        });
+
+        int yetToServe = consumers.size;
+        for (Tile consumer: sortedConsumers) {
+            float forEach = totalProduction / yetToServe;
+            float used = Math.min(consumer.block().powerCapacity - consumer.entity.power.buffer, forEach);
+            consumer.entity.power.buffer += used;
+            totalProduction -= used;
+            --yetToServe;
```
Anuken: Should be `yetToServe--` to stay consistent with the style of other code.

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -31,12 +35,62 @@ public int getID(){
...
+
+        if (totalProduction < requiredPower && totalAccumulator > 0.0001f) {
+            float accumulatorUsed = Math.min(requiredPower - totalProduction, totalAccumulator);
+            totalProduction += accumulatorUsed;
+            for (Tile battery: batteries) {
+                battery.entity.power.buffer *= 1f - (accumulatorUsed / totalAccumulator);
+            }
+        }
+
+        Array<Tile> sortedConsumers = consumers.iterator().toArray();
```
Anuken: If you're calling this *every frame*, it's going to pollute the heap with new arrays.
Use an `OrderedSet<Tile>` for consumers, iterate through it using `consumers.orderedItems()` and sort it with `consumers.orderedItems().sort(...)` *only when a new consumer is added*.

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -31,12 +35,62 @@ public int getID(){
...
+        sortedConsumers.sort((a, b) -> {
+            return (a.block().powerCapacity - a.entity.power.buffer) < (b.block().powerCapacity - b.entity.power.buffer) ? -1 : 1;
```
Anuken: Should be an expression lambda instead of a statement.
Use `Float.compare((a.block().powerCapacity - a.entity.power.buffer), (b.block().powerCapacity - b.entity.power.buffer))` for clearer code.

core/src/io/anuke/mindustry/world/BaseBlock.java
```diff
@@ -33,6 +33,8 @@
     public float liquidFlowFactor = 4.9f;
     public float powerCapacity = 10f;
 
+    public float powerProduction;
```
Anuken: Why is this in `BaseBlock` when it's generator-specific?

core/src/io/anuke/mindustry/world/modules/PowerModule.java
```diff
@@ -8,13 +8,14 @@
 import java.io.IOException;
 
 public class PowerModule extends BlockModule{
-    public float amount;
+    public float buffer;
+    public float productionEfficiency = 1;
```
Anuken: Why is this in `PowerModule` when it's generator-specific?

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
baltitenger: The order has to be updated every frame because consumers still use buffers so they can change every frame, so I think I'll revert to my previous algorithm.

core/src/io/anuke/mindustry/world/BaseBlock.java
baltitenger: Where should I put it then?

core/src/io/anuke/mindustry/world/modules/PowerModule.java
baltitenger: And buffer is only used by consumers. Where should it go then?

Anuken: In your code, the buffer is used by generators as well as consumers and batteries. It's a general value that all power blocks should have.

Production efficiency is a property of specific generators only, so it should go in `GeneratorEntity`

core/src/io/anuke/mindustry/world/BaseBlock.java
Anuken: Put it in `GeneratorBlock`, with a `float getPowerProduction(Tile tile)` in `BaseBlock` that calculates the current block power production, returning 0 by default and `powerProduction * entity.productionEfficiency` for generators.

core/src/io/anuke/mindustry/world/blocks/power/LiquidGenerator.java
```diff
@@ -11,6 +11,7 @@
 import io.anuke.ucore.graphics.Draw;
 import io.anuke.ucore.util.Mathf;
 
+// TODO figure out what to do with this
```
Anuken: Does this refer to the fact that these generators output to buffers, or something else?

baltitenger: Yes. `PowerGraph` currently ignores generators' buffers, so these don't generate any power now.

core/src/io/anuke/mindustry/world/BaseBlock.java
baltitenger: Makes sense, but the problem with that is now I'll have to change `DebugBlocks` because `powerInfinite` inherits from `PowerNode` which is not a `PowerGenerator`, thus doesn't have the `powerProduction` attribute. Is it ok if I change it to a `PowerGenerator`?

baltitenger: There's also a problem with `FusionReactor`. Its entity is a `GenericCrafterEntity`. I'd have to turn `GenericCrafterEntity` and `GeneratorEntity` into interfaces or something.

Anuken: This is why I suggested making power production a method in `BaseBlock`. `FusionReactor` and `powerInfinite` can override it.

core/src/io/anuke/mindustry/world/blocks/power/LiquidGenerator.java
Anuken: This could be fixed by assigning an entity `currentOutput` variable each update(), then returning that in `getPowerProduction()`

core/src/io/anuke/mindustry/world/BaseBlock.java
```diff
@@ -40,6 +44,13 @@ public boolean shouldConsume(Tile tile){
         return true;
     }
 
+    public float getPowerProduction(Tile tile){
+        if(tile.block().hasPower && tile.block().outputsPower){
```
Anuken: Why... are you doing all these checks specifically for power generators when the point is for `PowerGenerator` to override this method?

baltitenger: ok turns out I'm dumb

Anuken:
Just did some testing.
- Power generators still have their power bar, only it's always empty.
- Crafting blocks still have buffers. I thought the point was for them to not have *any* buffer and display % of power requirements met with their power bar.
- What should happen when requirements of a block aren't met? In the master branch, it just flashes on and off. What are your plans to handle this?

baltitenger:
Power generators don't have a power bar (at least for me).
Consumers have buffers because I don't know yet what to do with them and how to handle the lack of power. A lot of them don't have a constant power intake but large usage then nothing/constant for a while.
To be honest, I kinda want to remove power bars from consumers in the end, because with the current algorithm, everything gets filled up to the same percentage. I'd rather add something like a power network analyzer which could display the power generation and consumption per second so you could get a sense of what can you connect to the same network and have enough power.

baltitenger:
One possible solution would be to just have an efficiency that is 1 if there is enough power and less if not. the _not enough power_ indicator should be visible if efficiency is less then 1 or maybe some other threshold. But with the current algorithm, that would happen at the same time for everything. Also there are shield projectors. Should they just break instantly or what? I don't know how you want things to work.

Anuken:
Shield projectors may need to be reworked, but I can handle that later. Most blocks already operate on an 'efficiency' system, although it is implemented separately for each block. You could make efficiency a property of power modules for now.

The difficult part may be in implementing batteries and battery-like consumers like mass drivers and unit pads... although the former one could be reworked to an efficiency model.

baltitenger:
So... I've got problems. This won't work out the way I thought it would, at least not with the consumers. I have two implementations (`ConsumePower` and `ConsumePowerBuffered`) for a power consumer but that does not work with the consumes system. I think the best way would be to create a `PowerConsumer` and `PowerConsumerBuffered` class and all blocks that consume power could just inherit from those. If you have anything better in mind, please tell me.

Anuken:
Creating subclasses that must be inherited is unacceptable; that defeats the whole purpose of the consumption system.

Why do you say consumers won't work?

baltitenger:
Eg in `PowerGraph`, you need to get the power consumption of every consumer. If I use `tile.consumes.get(ConsumePower.class)` it will only work with unbuffered consumers. I couldn't really think of a non-janky way to solve this issue

Anuken:
Unbuffered consumers don't actually use up any power until a certain method is called; can't their logic remain the same as the old consumption logic?

In code, they should be equivalent to charge-only batteries.

baltitenger:
Ok, I see what you mean. I'll do that but gotta sleep now.

Timmeey86:
If you push your current changes to your fork I might be able to add some unit tests, in case the consumer and producer interfaces you're going to use are somewhat stable.

Timmeey86:
I started collecting some rather simple test cases (currently testing manually ingame), which fail in v59. They work again in master since https://github.com/Anuken/Mindustry/commit/de6bca424f04a55f28f7b5a64bbe03eca9354353

Anuken:
Why did you remove the power consumption definitions and make it a variable in `BaseBlock`?

Didn't you mention that you were going to continue to use consumers?

Timmeey86:
https://cdn.discordapp.com/attachments/513811720737914930/514162241902346247/PowerGraphActivity_Root.png
This is more or less what we're trying to do. Ignore the notes, they are outdated.
The idea is to get a stable implementation first, write unit tests for them and then refactor it to be more efficient or more sophisticated.
Either way, I'll have a look at if we can't achieve the same thing through the 'Consumes' interface this evening, maybe by introducing a `hasBuffer` method or similar

Timmeey86:
@Anuken `ConsumePower` has been restored and adapted to the new system
(The PR is still far from done though)

tests/src/test/java/PowerTests.java
```diff
@@ -101,5 +101,47 @@ void test_bufferedConsumptionCalculation(float producedPower, float maxBuffer, f
...
+                dynamicTest("04", () -> test_directConsumptionWithBattery(0.0f, 0.0f, 0.0f, 0.0f, 0.0f, "No producer, no consumer, empty battery")),
+                dynamicTest("05", () -> test_directConsumptionWithBattery(0.0f, 0.0f, 100.0f, 100.0f, 0.0f, "No producer, no consumer, full battery")),
+                dynamicTest("06", () -> test_directConsumptionWithBattery(0.0f, 10.0f, 0.0f, 0.0f, 0.0f, "No producer, empty battery")),
+                dynamicTest("07", () -> test_directConsumptionWithBattery(0.0f, 10.0f, 100.0f, 90.0f, 1.0f, "No producer, full battery")),
+                dynamicTest("08", () -> test_directConsumptionWithBattery(0.0f, 10.0f, 5.0f, 0.0f, 0.5f, "No producer, low battery")),
+                dynamicTest("09", () -> test_directConsumptionWithBattery(5.0f, 10.0f, 5.0f, 0.0f, 1.0f, "Producer + Battery = Consumed")),
+            };
+        }
+        void test_directConsumptionWithBattery(float producedPower, float requestedPower, float initialBatteryCapacity, float expectedBatteryCapacity, float expectedSatisfaction, String parameterDescription){
```
Anuken: Just a comment about the `test_xxx` naming convention: I don't think it's necessary. Java doesn't use `snake_case`, and it's usually fairly clear when a test is a test. If it's really necessary, add a comment to the method instead of adding the `test_` prefix.

Timmeey86: Noted

Anuken:
Since I've been told that this is ready for testing:
- Nuclear, turbine and combustion generator's power bars only display as 3/4 full regardless of whether there's anything connected.
- Solar panels' power bars are likewise always full regardless of what's connected. I'm not sure what you're trying to make them display, but either way it probably should not function like this.
- The force projector doesn't accept power properly and doesn't display the 'power lacking' symbol next to it when it is lacking power.
- Blocks such as the dagger mech factory put strain on the power system even if there's not enough item resources for them to function. As an example, I can connect some solar panels to a mend projector to make it function, but if I add an empty dagger mech factory, it stops functioning. Power should not be drawn until all other consume requirements are met.
- When the mend projector is at 3/4 power met, it doesn't work at all, but when it is at about 2/3, it does. This feels very strange.
- I think power bars should be removed from everything without a buffer (as it is confusing) and moved to the consume fragment. When power requirements are met, the 'power' icon could display relevant power usage statistics when hovered over.

Timmeey86:
The power bars were an attempt to display efficiency where a half filled bar means 100% (how would one display 200% if 100% is 0.5/1.0?)
I do agree that using a bar to display that never felt right, however being able to display the efficiency in any way feels rather important.
I'll remove the bars and look into the rest this evening. We will probably require assistance with the power icon hover info

baltitenger:
First of all, ^
For the unnecessary power consumption: Every consumer does that, and they sorta did before the patch, too. I believe they consumed power if they had enough items but couldn't output. This could be solved by an `isActive` flag or add a `conditionsMet` method.
For strange working-or-not issue: I think you switched up those numbers, they don't work until they have at least 60% power, after which they work at an efficiency proportional to the charge. The threshold can be changed if that's the problem.
For removing bars: I can agree, they can be confusing. Displaying the stats would be pretty useful. If they have enough power, should the icon move to the other side as the items do and be always displayed while hovering over the block or do you have a different plan?

Timmeey86:
Producers no longer have power bars; Consumers will not receive power while they are still missing required inputs. There are still cases where power is wasted, however, e.g. when a production building stopped production due to items piling up

Timmeey86:
@Anuken How would you want to have minimum power requirements / power efficiency in general? Would you want to have a different number than 60% for most blocks, or a completely different approach?
Which of the direct consumers shall only work at 100%? 
Here's a potential list for the 100% Blocks:
- Mend Projector
- Overdrive Projector
- Incinerator
- Phase Conveyor
- Phase Conduit

Timmeey86:
@Anuken your requested changes should be handled now, except for an actual efficiency/satisfaction display

Timmeey86:
I noticed with the current mechanics, where power is not consumed as long as the other requirements are not satisfied (e.g. there are missing items), the power missing symbol will be displayed as long as items are missing. I guess the symbol should be independent instead?

core/src/io/anuke/mindustry/world/consumers/Consumers.java
```diff
@@ -87,6 +121,15 @@ public boolean has(Class<? extends Consume> type){
         return map.containsKey(type);
     }
 
+    public boolean hasSubtypeOf(Class<? extends Consume> type){
```
Anuken: I would highly recommend avoiding class reflection (`isAssignableFrom()`) and O(n) search in this method. If a block has a lot of consumers, it's going to need to search through every single one for every tile in the graph.

All this searching is here to accommodate different types of power consumers, and I don't think it's worth it. The only class that extends `ConsumePower` is `ConsumeForceProjectorPower`, and there are ways to get around this.

A quick fix would involve changing:
```java
    public Consume add(Consume consume){
        map.put(consume.getClass(), consume);
        return consume;
    }
```
to:
```java
    public Consume add(Consume consume){
        map.put((consume instanceof ConsumePower ? ConsumePower.class : consume.getClass()), consume);
        return consume;
    }
```

This would ensure that all types of power consumers are registered under the same class type (ConsumePower).

baltitenger: To be honest, I don't think there should be searching at all. Just create an `enum` for the possible types of consumers (6 at most) and use a regular array to store the consumers, `null` for the non-existent ones. The consumers which do basically the same thing (and have the same interface) but have different implementations could be stored in the same field of the array.

Anuken: Using classes as keys in the map has the benefit of automatically casting to the right type, e.g. `get(ConsumeItem.class) -> returns ConsumeItem`. But I don't suppose it matters too much, since you could still do `<ConsumeItem>get(ConsumeType.item)` or just use the utility methods for that purpose. Performance would be equivalent; it's a matter of design.

Timmeey86: I would suggest doing the quick fix for now and implementing the enum thing a new branch if desired. The hasSubtypeOf was introduced since tests failed for the force projector, but doing what you suggested in the fix would be a cleaner and more efficient solution indeed.

PR #289, opened 2018-10-31
Info Pages will now display tool tips for any item icons

Timmeey86:
This is especially useful for beginners who are unable to distinguish all the item icons from one another.
Note that I added several dependencies which might include undesired ones. Feel free to point out better locations for the static functions, for example.


Anuken:
This seems good, but in the future I would like to integrate all the items, liquids and blocks into the font to streamline the process of creating stat descriptions. Tooltips don't work on mobile (at least, not very well), so this is will be a temporary solution until the font implementation.

PR #305, opened 2018-11-05
Proposal for configurable starting core position and fixing tutorial sector

Timmeey86:
Before this change, sectors with wave missions would always spawn their core in the center of the map. This however breaks the tutorial sector which has all its tutorials based on a core at (60, 60). 
This pull request is a proposal on how the tutorial sector could override the core location through lambda functions while maintaining the current behavior of default Wave and Battle missions.
The MissionWithStartingCore class was introduced in order to avoid modifying the constructors in other Mission subclasses which are not interested in Core positioning.
An alternative implementation would be to create a regular virtual function and create a special subclass for the Tutorial sector which overrides this function and provides a different tile.
A third, but more cumbersome implementation could leave the tutorial sector core in the center of the map, but shift all other missions to the new location.


Anuken:
I don't think it's a good idea to keep the tutorial dependent on placing blocks in specific locations. Many people have reported frustration at needing to overwrite their old builds with mandatory conveyors or drills.

Instead of moving the spawnpoint, the idea was to make all the missions have loose (not location based) tasks such as "place drill on coal", "link smelter", etc.

However, placing cores at a specific position could still be useful for things like wave missions and defending the core at specific points on the map.
Is there any particular reason why you chose to make the constructor accept a lambda expression instead of a raw (x,y) position for the core?

Timmeey86:
I could have done an array which is set to null for the default case, and then calculate the position when null, or use the stored position. I guess both solutions have their advantages and disadvantages

PR #306, opened 2018-11-05
change to game/screen mode of libgdx

islq:
I think a simple splash screen is better than black only when lunch.
Can we change to game/screen mode and add a splash screen?
I have modified ModuleCore to implement screen interface of libgdx, and send a pull request too.
Please check it. Thanks


Kieaer:
Can't build it :thinking:

Anuken:
This isn't a bad solution for a single splash screen, but it still freezes the game loop while assets are loaded. I was planning to have a visual loading animation at the start while assets are loaded asynchronously with an `AssetLoader`.

If possible, I would also like to avoid using libGDX's `Screen`/`Game` classes, as they're... not really designed for integration into my module system. I would like to do this loading system myself, since these are some fairly subjective architectural issues.

islq:
Thanks for check it.
Nevermind, It's only a cheap option when you have no time to apply AssetLoader

PR #307, opened 2018-11-05
Adjustable core position

Timmeey86:
Refactored the previous pull request so it uses `int`s in the public interface. I opened a new pull request since I messed up the old branch. To sum it up again:
- Wave missions spawn the core in the center of the map by default
- Battle missions spawn the core at (50|50) by default
- Both mentioned mission types allow configuring a custom location.
- The tutorial sector spawns the core at (60|60) again like in previous versions (but is still disabled)

### DIFF
--- a/core/src/io/anuke/mindustry/maps/missions/BattleMission.java
+++ b/core/src/io/anuke/mindustry/maps/missions/BattleMission.java
@@ -13,8 +13,20 @@ import static io.anuke.mindustry.Vars.*;
 public class BattleMission extends MissionWithStartingCore{
     final int spacing = 30;
 
+    /**
+     * Creates a battle mission with the player core being in the center of the map.
+     */
     public BattleMission(){
-        super((gen, team) -> gen.tiles[50][50]);
+        super();
+    }
+
+    /**
+     * Creates a wave survival with the player core being at a custom location.
+     * @param xCorePos The X coordinate of the custom core position.
+     * @param yCorePos The Y coordinate of the custom core position.
+     */
+    public BattleMission(int xCorePos, int yCorePos){
+        super(xCorePos, yCorePos);
     }
 
     @Override
--- a/core/src/io/anuke/mindustry/maps/missions/BattleMission.java
+++ b/core/src/io/anuke/mindustry/maps/missions/BattleMission.java
@@ -10,8 +10,12 @@ import io.anuke.ucore.util.Bundles;
 
 import static io.anuke.mindustry.Vars.*;
 
-public class BattleMission extends Mission{
+public class BattleMission extends MissionWithStartingCore{
     final int spacing = 30;
+
+    public BattleMission(){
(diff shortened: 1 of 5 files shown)


core/src/io/anuke/mindustry/maps/missions/MissionWithStartingCore.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.badlogic.gdx.math.GridPoint2;
+import io.anuke.mindustry.content.blocks.StorageBlocks;
+import io.anuke.mindustry.game.Team;
+import io.anuke.mindustry.maps.generation.Generation;
+import io.anuke.mindustry.world.Tile;
+
+import static io.anuke.mindustry.Vars.state;
+
+public abstract class MissionWithStartingCore extends Mission{
```
Anuken: This class needs to override `getSpawnPoints()` and return this point in the list so the area around them is generated properly.

Now that I think about it, generating cores should *only* be done using the points from `getSpawnPoints()` to streamline things. I could implement this myself later, or you could include it in this PR.

core/src/io/anuke/mindustry/maps/missions/BattleMission.java
```diff
@@ -10,8 +10,26 @@
 
 import static io.anuke.mindustry.Vars.*;
 
-public class BattleMission extends Mission{
+public class BattleMission extends MissionWithStartingCore{
     final int spacing = 30;
+    public static final int defaultXCorePos = 50;
+    public static final int defaultYCorePos = 50;
+
+    /**
```
Anuken: Please use single-line javadoc comments whenever possible (e.g. `/** comment */`).

core/src/io/anuke/mindustry/maps/missions/MissionWithStartingCore.java
Timmeey86: Makes sense; I included the change in the PR

core/src/io/anuke/mindustry/maps/missions/MissionWithStartingCore.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Generates a player core based on generation parameters.
+     * @param gen  The generation parameters which provide the map size.
+     * @param team The team to generate the core for.
+     */
+    public void generateCoreAtFirstSpawnPoint(Generation gen, Team team){
+        Array<GridPoint2> spawnPoints = getSpawnPoints(gen);
+        if(spawnPoints == null || spawnPoints.size == 0){
+            throw new IllegalArgumentException("A MissionWithStartingCore subclass did not provide a spawn point in getSpawnPoints(). However, at least one point must always be provided.");
```
Timmeey86: Reason for this decision: If a mission provides no spawn point, it should not derive from this class in the first place.

PR #319, opened 2018-11-09
Enable/Disable option for blocks

etaoinshrdlcumwfgypbvkjxqz:
**Overview**
An enable/disable option is added to every block except
> Blocks that are either distributors (Conveyor, Conduct, Power Node, etc.) or not meant to be disabled (Core, Command Center, etc.) do not provide an option to be disabled.

**Functions**
- It can be used to not let a system run before it is finished, which prevents item-clogging and then rebuilding the clogged parts.
- It can be used to save _items_, _liquid_ and _power_ when needed.
- _etc._

**Changes**

- When `configurable` is `true`, it will add enable/disable option.
To disable this behavior, override `Block.buildTable(Tile, Table)`.

- `Block.enable(Tile)` and `Block.disable(Tile)` are added.
`Block.enable(Tile)` is called when the block is being enabled (not on placing it).
`Block.disable(Tile)` is called when the block is being disabled.

- `TileEntity.enabled` is the `boolean` which shows whether it is enabled or disabled.

- This will break existing saves as of whether blocks are enabled is saved to saves now.


etaoinshrdlcumwfgypbvkjxqz:
You could add a switch for *items*, *liquid* and *power*, but since I'm not so sure what you would want the graphics to look like, so I didn't include them here.

etaoinshrdlcumwfgypbvkjxqz:
> Rename `enabled(Tile)` and `disabled(Tile)` respectively to `onEnable(Tile)` and `onDisable(Tile)`

Anuken:
In my opinion, this adds unnecessary complexity to a system that is already confusing for many players. If I were to add something like this, it would be in conjunction with [logic systems](https://trello.com/c/jonIx24W/90-logic-systems) and not be linked to the block configuration system. It bloats UI by making a checkbox pop up nearly every time you click a block, using up valuable screen space on mobile.

etaoinshrdlcumwfgypbvkjxqz:
After thinking about it, complexity is **inevitable** as new features are added, like the planned _logic system_ you have mentioned.
One way to help that is by hiding the more complex stuff and only shows it when a _config button_ is clicked.
This way, new players won't be presented with overwhelmingly UI, while advanced players can still use the complex stuff.

etaoinshrdlcumwfgypbvkjxqz:
Any update?

etaoinshrdlcumwfgypbvkjxqz:
Integrated this to #362.

PR #334, opened 2018-11-15
Update ServerControl.java

Milinai:
Made fillcore have an argument for team.


Milinai:
I forgot to make it not log "Core filled." when no team is provided... Oh no.

Anuken:
Use `Team.valueOf(name)` with a try/catch to get the team, or a for-loop. This is verbose and breaks when new teams are added

PR #336, opened 2018-11-16
Power can no longer go negative or exceed capacity

Timmeey86:
This should prevent a number of weird effects like infinite power drills or consumers which cannot be powered at all. Distribution issues are unaffected by this.

### DIFF
--- a/core/src/io/anuke/mindustry/content/blocks/DebugBlocks.java
+++ b/core/src/io/anuke/mindustry/content/blocks/DebugBlocks.java
@@ -69,7 +69,7 @@ public class DebugBlocks extends BlockList implements ContentList{
             @Override
             public void update(Tile tile){
                 super.update(tile);
-                tile.entity.power.amount = powerCapacity;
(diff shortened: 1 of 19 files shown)


Anuken:
I don't think this is worth the mess caused by switching to getters/setters. This doesn't actually fix the power issue, which is caused by overdraw of power from producers by the power graph-- it's the reason power went negative in build 59, and not in any other builds.

Timmeey86:
I went for getter/setter protection in order to have a single piece of code which makes sure power does not get out of bounds rather than having to check that in multiple classes. Checking that all over the place, like in the current v59 implementation is prone to bugs which happen in addition to what's actually broken - This PR would fix only that, so it is meant as a safety net until baltitenger finished reworking the power system

Anuken:
The issue, in my opinion, is that protecting power with getters/setters doesn't actually fix the problem; it just hides bad values when something goes wrong, letting the problematic code keep running as it is. If power goes negative, that is indicative of some core flaw in the way power is drained, and we should know about it.

PR #338, opened 2018-11-16
Complete most of the Chinese translations

etaoinshrdlcumwfgypbvkjxqz:
Fix color codes, punctuations, and existing translations.
Also, translate untranslated strings.

**THE WORDS BELOW ARE NO LONGER TRUE.**
**Do not accept this yet,**
_because I am still working on it..._


etaoinshrdlcumwfgypbvkjxqz:
**Well, it's okay to accept it now.**
I will work on the rest later...

PR #340, opened 2018-11-17
Fixed BlocksFragment categories row limit

Sonnicon:
After the first row, the row limit for categories wouldn't affect anything.


Sonnicon:
> I assume this is a problem that only appears when modding in more categories?

That's how I discovered it, yes, but it could cause issues in the future for the base game.

Anuken:
I won't be adding any new rows to the current placement dialog, but I suppose there's no harm done with by fixing it for now.

PR #348, opened 2018-11-21
Power info

islq:
show powergraph info when tap a battery

### DIFF
--- a/core/src/io/anuke/mindustry/ui/fragments/PowerInfoFragment.java
+++ b/core/src/io/anuke/mindustry/ui/fragments/PowerInfoFragment.java
@@ -1,0 +1,94 @@
+package io.anuke.mindustry.ui.fragments;
+
+import com.badlogic.gdx.Gdx;
+import com.badlogic.gdx.math.Interpolation;
+import com.badlogic.gdx.math.Vector2;
+import com.badlogic.gdx.utils.Align;
+
+import io.anuke.mindustry.core.GameState.State;
+import io.anuke.mindustry.world.Tile;
+import io.anuke.mindustry.world.blocks.power.PowerGraph;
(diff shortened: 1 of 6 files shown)


core/src/io/anuke/mindustry/ui/fragments/PowerInfoFragment.java
```diff
@@ -0,0 +1,94 @@
...
+        if (!visible) return;
+        table.clear();
+        table.update(() -> {
+        });
+        visible = false;
+    }
+
+    private String round(float f){
+        if(f >= 1000000){
+            return Strings.toFixed(f / 1000000f, 2) + "[gray]mil[]";
```
sparr: why "mil" instead of "M"?

also, gray makes me think less, not more. would a different color be better for highlight?

Anuken:
Closing this PR, as power graph info is now planned to be displayed in the bottom right menu (when hovering over blocks).

(Next time, _please ask me_ (either through Discord or a new issue) before submitting a new feature like this. It may already be planned, or I may have different ideas on how it should be implemented.)

core/src/io/anuke/mindustry/ui/fragments/PowerInfoFragment.java
islq: It was copyed from BlockInventoryFragment.java, just keep the same pattern

islq:
Anuken, thank you for reply it. I'm sorry if it wast your time.
It's a great game, I can't stop to play it, power info is a big issue for me, so I made this temporary patch to check power info. When discussed with VXF2016, he ask me to share it, so I send a pull request at here.
It's much better to displayed in the bottom right when hover any power block. I can't wait to see it in game now. :)

PR #367, opened 2018-12-05
Check client version type against server version type

iczero:
Previously, the server would reject all clients with a Version.type that is not "official". This allows for clients compiled with the same Version.type as the server to connect.


TheChosenEvilOne:
I just want to note something here:
Instead of using this custom version bs the whole system should be remade with a blacklist and whitelist system for version types with default whitelisting official

PR #371, opened 2018-12-09
Gridmode

indielm:
Adds keybindings to category and block selection


TheChosenEvilOne:
The code style is inconsistent and the keybinds should be configurable

PR #372, opened 2018-12-09
Gridmode and eyedropper select

indielm:
Added keybindings for grid-style selection of build categories and blocks with "gridMode"+"gridModeShift" keys.
Also eyedrop select the block under mouse with "gridMode"+"select" keys.

Like last time, but better


TheChosenEvilOne:
Again code style is inconsistent and the keybinds should be configurable

TheChosenEvilOne:
Muuch better, thank you

Anuken:
That's some weird formatting, but IntelliJ doesn't provide any options for making it better than that. I guess it'll have to do.

PR #375, opened 2018-12-13
Center sectors map on most recently saved (played and not abandoned) sector

sparr:
Also minor refactor, moving date formatting from SaveMeta to SaveSlot


sparr:
Also minor refactor, moving date formatting from SaveMeta to SaveSlot

PR #376, opened 2018-12-14
Update block config every frame

etaoinshrdlcumwfgypbvkjxqz:
---
### Description
Makes config every frame using `Element.update(Runnable)`,
which means when a config of a block is opened,
if somebody else changes the settings of the block,
the change is reflected instantly instead of reopening the config.
### Changes
Buttons are updated every frame.
~~`configurable` is replaced with `isConfigurable(Tile)`.~~
~~`buildTable(Tile, Table)` returns a `boolean` value.~~

---
> Email: <williamso26144225@gmail.com>
> Discord: `[Insert]#6021`

### DIFF
--- a/core/src/io/anuke/mindustry/ui/fragments/BlockConfigFragment.java
+++ b/core/src/io/anuke/mindustry/ui/fragments/BlockConfigFragment.java
@@ -79,8 +79,10 @@ public class BlockConfigFragment extends Fragment{
         return e != null && (e == table || e.isDescendantOf(table));
     }
 
+    private static final Runnable empty = () -> {};
     public void hideConfig(){
         configTile = null;
+        for(Element e : table.getChildren()) e.update(empty);
         table.actions(Actions.scaleTo(0f, 1f, 0.06f, Interpolation.pow3Out), Actions.visible(false));
     }
 }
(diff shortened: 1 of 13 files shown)


Anuken:
> `configurable` is replaced with `isConfigurable(Tile)`.
> `buildTable(Tile, Table)` returns a `boolean` value.

Could you explain why these changes are necessary?

etaoinshrdlcumwfgypbvkjxqz:
They are not necessary.
The changes are reverted.

core/src/io/anuke/mindustry/ui/fragments/BlockConfigFragment.java
```diff
@@ -79,8 +79,10 @@ public boolean hasConfigMouse(){
         return e != null && (e == table || e.isDescendantOf(table));
     }
 
+    private static final Runnable empty = () -> {};
```
Anuken: This is unnecessary. First of all, you can do `e.update(null)`, and second of all, there's no need for the elements to stop updating at all. When a table is not visible, its elements do not update.

etaoinshrdlcumwfgypbvkjxqz: If the block config UI is opened, and then it is deconstructed, the game will crash because of `NullPointerException`.
For some reason, after deconstructing it, the update function is called once more, and when `tile` is accessed in the update function, and the block is being deconstructed so `tile` is `null`, `NullPointerException` is raised.

And for the part about providing `null` instead of `() -> {}`, I'll look into it.

Anuken:
I ended up implementing this manually without any `update(null)` calls. After some testing, I could not reproduce any `NullPointerException`s being thrown from the button update loop. As far as I can tell, the code in the `update()` lambda never accesses objects that could become null.

PR #384, opened 2018-12-21
Create  strings.xml

Prosta4okua:
uk or uk-ua or ua?


Anuken:
This should be `values-uk`, I believe.

Prosta4okua:
I fixed this.

PR #386, opened 2018-12-28
Update bundle_pl.properties

ghost:
sorry for translating so little, i will translate more now but I've gotten unknown for me function that was deleting text instead of pushing it


latekvo:
I think this was my very first PR ever :)

PR #436, opened 2019-03-24

laohuaji233:
more translations


laohuaji233:
> Line 380
> Oh and more but I know this is hard, right?

Yes,through it’s hard,but I will translate all of them into Chinese

laohuaji233:
> Line 380
> Oh and more but I know this is hard, right?

Yeah,a little

PR #440, opened 2019-03-29

laohuaji233:
More translations and fixed some wrong words


Anuken:
Some of the merged/conflicted text may be incorrect. If you have time, look over the merged result and see if there are any errors. Thanks.

PR #465, opened 2019-04-13
Add string conversion to utf8 for discord rpc

swiftail:
Fixing discord rpc for non-ascii map names. I hope my first commit will not break the whole repo


swiftail:
Seems like in new version of rpc they're decoding string in UTF8 so this PR is useless now 😃

PR #466, opened 2019-04-15
Update bundle_ita.properties

Leone25:
fixed flew grammatics error and added flew translations


Anuken:
This seems to be referencing an ancient file that has since been renamed; all of the translated text is no longer usable. Thanks anyway.

PR #470, opened 2019-04-23
Improved sandbox

Natchuz:
- Added special dialog where you can spawn certain enemies directly trough game.
- Added special sandbox tools, used to modify in-game rules, and change Mech on-demand.

Needed to change `CustomRulesDialog` a little to make possible to properly change rules in-game.

### DIFF
--- a/core/src/io/anuke/mindustry/game/Rules.java
+++ b/core/src/io/anuke/mindustry/game/Rules.java
@@ -55,6 +55,6 @@ public class Rules{
     public boolean waitForWaveToEnd = false;
     /** Determinates if gamemode is attack mode */
     public boolean attackMode = false;
-    /** Enables sanbox tools */
+    /** Enables sandbox tools */
     public boolean sandobx = false;
 }
--- a/core/src/io/anuke/mindustry/game/Rules.java
+++ b/core/src/io/anuke/mindustry/game/Rules.java
@@ -55,4 +55,6 @@ public class Rules{
     public boolean waitForWaveToEnd = false;
     /** Determinates if gamemode is attack mode */
     public boolean attackMode = false;
+    /** Enables sanbox tools */
+    public boolean sandobx = false;
 }
--- a/core/src/io/anuke/mindustry/ui/dialogs/CustomRulesDialog.java
+++ b/core/src/io/anuke/mindustry/ui/dialogs/CustomRulesDialog.java
@@ -2,6 +2,7 @@ package io.anuke.mindustry.ui.dialogs;
 
 import io.anuke.arc.function.*;
 import io.anuke.arc.graphics.Color;
+import io.anuke.arc.scene.ui.Dialog;
 import io.anuke.arc.scene.ui.layout.Table;
 import io.anuke.arc.util.Strings;
 import io.anuke.mindustry.core.Platform;
@@ -9,6 +10,7 @@ import io.anuke.mindustry.game.Gamemode;
 import io.anuke.mindustry.game.Rules;
 import io.anuke.mindustry.graphics.Pal;
 
+import static io.anuke.mindustry.Vars.state;
 import static io.anuke.mindustry.Vars.tilesize;
 
 public class CustomRulesDialog extends FloatingDialog{
@@ -27,7 +29,13 @@ public class CustomRulesDialog extends FloatingDialog{
     public void show(Rules rules, Gamemode gamemode){
         this.rules = rules;
         this.selectedGamemode = gamemode;
-        show();
+        super.show();
+    }
+
+    @Override
+    public Dialog show(){
+        this.rules = state.rules;
+        return super.show();
     }
 
     void setup(){
@@ -35,7 +43,8 @@ public class CustomRulesDialog extends FloatingDialog{
         cont.pane(m -> main = m);
         main.margin(10f);
         main.addButton("$settings.reset", () -> {
-            rules = selectedGamemode.get(); 
+            state.rules = selectedGamemode.get();
+            rules = selectedGamemode.get();
             setup();
         }).size(300f, 50f);
         main.left().defaults().fillX().left().pad(5);
--- a/core/src/io/anuke/mindustry/ui/dialogs/SandboxDialog.java
+++ b/core/src/io/anuke/mindustry/ui/dialogs/SandboxDialog.java
@@ -1,0 +1,48 @@
+package io.anuke.mindustry.ui.dialogs;
+
+import io.anuke.arc.scene.ui.Image;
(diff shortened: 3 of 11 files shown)


core/src/io/anuke/mindustry/game/Rules.java
```diff
@@ -55,4 +55,6 @@
     public boolean waitForWaveToEnd = false;
     /** Determinates if gamemode is attack mode */
     public boolean attackMode = false;
+    /** Enables sandbox tools */
+    public boolean sandobx = false;
```
Anuken: "sandobx" -> "sandbox"

core/src/io/anuke/mindustry/ui/dialogs/CustomRulesDialog.java
```diff
@@ -27,15 +29,22 @@ public CustomRulesDialog(){
     public void show(Rules rules, Gamemode gamemode){
         this.rules = rules;
         this.selectedGamemode = gamemode;
-        show();
+        super.show();
+    }
+
+    @Override
+    public Dialog show(){
```
Anuken: Don't override `show()`, register a listener for it instead, e.g. `shown(() -> this.rules = state.rules);`.

core/src/io/anuke/mindustry/ui/fragments/HudFragment.java
```diff
@@ -21,6 +21,7 @@
 import io.anuke.mindustry.gen.Call;
 import io.anuke.mindustry.graphics.Pal;
 import io.anuke.mindustry.input.Binding;
+import io.anuke.mindustry.io.BundleLoader;
```
Anuken: Unnecessary import.

core/src/io/anuke/mindustry/ui/dialogs/SandboxDialog.java
```diff
@@ -0,0 +1,48 @@
...
+        onResize(this::setup);
+        addCloseButton();
+    }
+
+    private void setup(){
+        cont.clear();
+        cont.pane(m->table=m);
+        table.top();
+        table.margin(10f);
+        table.addButton("$sandbox.rules", ()->ui.custom.dialog.show()).width(300f).center();
```
Anuken: Whenever you use a lambda expression, add spaces before and after the parameters or `()`. This should be
 `() -> ui.custom.dialog.show()`. The spaces are missing in many different places here.

core/src/io/anuke/mindustry/ui/dialogs/WaveSpawnDialog.java
```diff
@@ -0,0 +1,107 @@
...
+        cont.pane(m->table=m);
+        table.top();
+        table.margin(10f);
+        for(SpawnGroup group : groups) {
+            table.table("clear", t->{
+                t.margin(6f).defaults().pad(2).padLeft(5f).growX().left();
+                t.addButton(b -> {
+                    b.left();
+                    b.addImage(group.type.iconRegion).size(30f).padRight(3);
+                    b.add(group.type.localizedName).color(Pal.accent); }, () -> showUpdate(group)).pad(-6f).padBottom(0f);
```
Anuken: Bad formatting, curly brace should be on a new line.

Natchuz:
### Spawning pre-made waves
When we were discussing I didn't think of pre-made waves as you did (I thought of these which are defined for map). You have totally right here. 

### Mechs...
First of all, I would like to emphasize that the point of this SandboxDialog isn't only to change Mech. This dialog is specially made for the sandbox to easily navigate and operate things and make them simpler. 

> It adds more code for me to maintain. If (when) I change how mechs or UI work, I may have to rewrite this dialog class.

Well, I don't think 3 (basically 1) lines of code are hard to maintain. The only thing you would change there is the way how player transforms into the mech. I don't want to be nosy, but you could use this type of argument every time someone opens any type of PR. ~~_Cause the only constant is change_~~

> It adds unnecessary UI to something that can already be done in-game without much effort.

As I mentioned before, the purpose of this dialog and isn't only to change Mech. We can use it in many ways. I'll move the "spawn enemies" button there. You can change rules there. Changing Mech is only one of the quick-actions we can put there. (I just thought I can make checkbox which would turn infinite resources in storage blocks or power any block instantly). Also I don't really think placing Mech pad and power source is totally simple. At least, not as simple as it can be.

> This is bloat that I would like to avoid, especially in a mobile game.

I guess one more button on the screen won't make Hud too complex, and there is a place to fit it good.

Anuken:
> First of all, I would like to emphasize that the point of this SandboxDialog isn't only to change Mech.

I understand this, but it does not change my position. I don't think there should be any UI (in the dialog) to perform actions that are simple to do in-game.

> you could use this type of argument every time someone opens any type of PR.

While this is true, I'm only saying this to explain why extra code should be avoided. If I were referring to code that added a necessary feature, I wouldn't be pointing that out.

> The only thing you would change there is the way how player transforms into the mech.

I may very well be changing the way content icons work, as well as the base UI skin and general style. Either of those changes would require refactoring the mech dialog.

> I guess one more button on the screen won't make Hud too complex

I was not referring to the main HUD when I wrote that message. That being said, one more button *will* make the HUD more complex, even if only by a tiny amount.

Natchuz:
image
Let's talk about UI design for spawning enemy's units. To access this dialog/fragment we can't avoid placing a button on Hud (or you have other ideas on how to do that). On the picture, the button doesn't seem to be invasive. I thought of this button as a toggle button, which will change blocks menu into the same but with ~Mechs~ units. You could select an enemy and place him. Another change I would like to discuss is my idea of a button that would power any block. My idea was to create fragment like that one which appears in StorageBlocks, but with a button that would toggle infinity power. I also thought of placing a button in the mentioned BlockInventoryFragment which would fill the storage on demand. I know what you think about simple actions. Fill button would be useful I guess. But infinity power button depends on you (since placing the solar panel in the sandbox isn't hard (following your way of thinking))

Natchuz:
Off there is infinty power source block. Nevermind

Anuken:
>  I thought of this button as a toggle button, which will change blocks menu into the same but with Mechs units. You could select an enemy and place him.

This would be *very* annoying and messy to implement. A dialog would be preferable. Even something as simple as this would work:

Screenshot from 2019-04-30 12-31-56

(This is about 10 lines or less to implement, and works for my purposes.)

> On the picture, the button doesn't seem to be invasive.

It's not, but there may still be some text bounds issues in certain situations. I would need to test it.

Natchuz:
I'll hold it for now

LeRoiDesKiwis:
Why you deleted the branch ? I wanted to test that feature, but you deleted...

PR #476, opened 2019-04-28
Create bundle_af.properties

ha6000:
I think i have proposed an other one of af but i then went offline and worked on this one


ha6000:
im 15.2 percent done

PR #501, opened 2019-05-27
Reduce the cyclomatic complexity of render method in MapRenderer.java 

fenrir94:
# Method refactoring reason

I have checked the cyclomatic complexity of `MapRenderer.java` file through the analysis program which is a plugin of IntelliJ. The result shows the below table.

method | ev(G)<sup>[1](#f1)</sup> | iv(G)<sup>[2](#f2)</sup> | v(G)<sup>[3](#f3)</sup>
-- | -- | -- | --
draw(float,float,float,float) | **4**| 4 | 5
dispose() | 2 | 4 | 5
resize(int,int) | 1 | 4 | 6
MapRenderer.updateAll() | 1 | 3 | 3
MapRenderer.MapRenderer(MapEditor) | 1 | 1 | 1
**render(int,int)** | 1 | **12** | **13**
updatePoint(int,int) | 1 | 1 | 1
-- | -- | -- | --
*Total* | 11 | 29 | 34
*Average* | 1.571428571 | 4.142857143 | 4.857142857

Summarize is looks like the below graph.

PieChart

As you can see only one `render` method has **almost 40%** of cyclomatic complexity in this file. Therefore, I refactor this method and reduce the cyclomatic complexity.

# Changed list

Unfortunately, I can't fully understand your code. So, I think that you can think the some of the names are really odd. I ask for your understanding regarding this matter. Following contents are related in essential changes of my pull requests.

**1. I packed some formula**
    - Your code is really compact and reasonable. But some of the formulae was hard to read. So I pack that formula. An example is like below.

```java
// BEFORE
float offsetX = -(wall.size / 3) * tilesize, offsetY = -(wall.size / 3) * tilesize;

// AFTER
float offsetSize = -Mathf.floor((float)(wall.size / 3.0)) * tilesize;
FloatPoint offset = new FloatPoint(offsetSize, offsetSize);
```

**2. Add some `Point` and `Rectangle` class**
   - `Point` class separated built-in and nested class `FloatPoint`. Because of the former doesn't handle the float value. So, I added the handle the float value class like the latter. Furthermore, I don't want to interfere with your overall system. As a result, I made the latter with nested class.
   - `Rectangle` class is related in the region. I checked that `mesh.draw` has `x, y, width, height`. So, I added the appropriate class to make more efficient to map `mesh.draw` parameters.

**3. Separate two-part**
   - I separate the inner 'render' method to two-part(`renderMeshRegion`, `renderTileRegion`). I thought that these changes are necessary. But I don't feel that this naming is true. **With all due respect, please confirm this naming.**

# Result

After I did this, the metric changes like below.

method | ev(G)<sup>[1](#f1)</sup> | iv(G)<sup>[2](#f2)</sup> | v(G)<sup>[3](#f3)</sup>
-- | -- | -- | --
dispose() | 2 | 4 | 5
draw(float,float,float,float) | 4 | 4 | 5
drawMesh(...) | 1 | 2 | 2
resize(int,int) | 1 | 4 | 6
updateAll() | 1 | 3 | 3
updatePoint(int,int) | 1 | 1 | 1
**FloatPoint.FloatPoint(float,float)** | 1 | 1 | 1
**getChunkStartPoint(Point)** | 1 | 1 | 1
**getEditorIcon(Block,boolean)** | 1 | 2 | 2
**getMesh(Point)** | 1 | 1 | 1
**getMeshOffset(float)** | 1 | 1 | 1
**getRandomValue(int,int)** | 1 | 1 | 1
**getRegionInstance(String)** | 1 | 1 | 1
**getTile(Point)** | 1 | 1 | 1
**isAirBlock(Block)** | 1 | 1 | 1
**MapRenderer(MapEditor)** | 1 | 1 | 1
**render(int,int)** | 1 | 1 | 1
**renderMeshRegion(...)** | 1 | 5 | 5
**renderTileRegion(...)** | 1 | 6 | 7
-- | -- | -- | --
*Total* | 23 | 41 | 46
*Average* | 1.210526316 | 2.157894737 | 2.421052632

Summarize is looks like the below graph.

PieChart2

Unlike before, you can see that the values are evenly distributed. And I expect that my request will have a positive impact on your future changes in the `editor`.

*I am looking forward to your positive response.*

---

<a name="f1">1</a>: Essential Cyclomatic Complexity
<a name="f2">2</a>: Design Complexity
<a name="f2">3</a>: Cyclomatic Complexity

### DIFF
--- a/core/src/io/anuke/mindustry/editor/MapRenderer.java
+++ b/core/src/io/anuke/mindustry/editor/MapRenderer.java
@@ -6,6 +6,7 @@ import io.anuke.arc.collection.IntSet.IntSetIterator;
 import io.anuke.arc.graphics.Color;
 import io.anuke.arc.graphics.Texture;
 import io.anuke.arc.graphics.g2d.Draw;
+import io.anuke.arc.graphics.g2d.TextureAtlas.AtlasRegion;
 import io.anuke.arc.graphics.g2d.TextureRegion;
 import io.anuke.arc.math.Mathf;
 import io.anuke.arc.util.Disposable;
@@ -15,6 +16,9 @@ import io.anuke.mindustry.graphics.IndexedRenderer;
 import io.anuke.mindustry.world.Block;
 import io.anuke.mindustry.world.Tile;
 import io.anuke.mindustry.world.blocks.BlockPart;
+import java.awt.Point;
+import java.awt.geom.Rectangle2D;
+import java.awt.geom.Rectangle2D.Float;
 
 import static io.anuke.mindustry.Vars.tilesize;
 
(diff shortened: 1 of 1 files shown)


core/src/io/anuke/mindustry/editor/MapRenderer.java
```diff
@@ -15,6 +16,9 @@
 import io.anuke.mindustry.world.Block;
 import io.anuke.mindustry.world.Tile;
 import io.anuke.mindustry.world.blocks.BlockPart;
+import java.awt.Point;
```
Anuken: *Never* use any classes in `java.awt` in projects that target Android, such as Mindustry. The only main packages that should be used are `java.util` and `java.io`. `java.awt` is not available on Android, and is stripped out of the core library on desktop to reduce packaged size.

If you look around the codebase, you'll see that I use `io.anuke.arc.math.geom.Point2` when a point object is absolutely necessary. However, this is not the case (see other comments below).

core/src/io/anuke/mindustry/editor/MapRenderer.java
```diff
@@ -97,59 +101,91 @@ public void updateAll(){
     }
 
     private void render(int wx, int wy){
-        int x = wx / chunkSize, y = wy / chunkSize;
-        IndexedRenderer mesh = chunks[x][y];
-        Tile tile = editor.tiles()[wx][wy];
+        final Point windowPoint = new Point(wx, wy);
```
Anuken: You should *never* create objects in a method that may be called hundreds of times per frame. This makes massive amounts of garbage, which is a significant issue for Android phones. Objects should only be created during game initialization or world loading periods.

Even if performance were not a problem, I don't see how using objects here improves readability.

core/src/io/anuke/mindustry/editor/MapRenderer.java
```diff
@@ -97,59 +101,91 @@ public void updateAll(){
...
+        final Point chunkStartPoint = getChunkStartPoint(windowPoint);
+        final int chunkArea = chunkSize * chunkSize;
+
+
+        IndexedRenderer mesh = getMesh(gridPoint);
+        Tile tile = getTile(windowPoint);
         Block wall = tile.block();
 
-        TextureRegion region;
+        int indexWall = chunkStartPoint.x + chunkStartPoint.y;
```
Anuken: Why rename `idx` to `index`?

core/src/io/anuke/mindustry/editor/MapRenderer.java
```diff
@@ -97,59 +101,91 @@ public void updateAll(){
     }
 
     private void render(int wx, int wy){
-        int x = wx / chunkSize, y = wy / chunkSize;
-        IndexedRenderer mesh = chunks[x][y];
-        Tile tile = editor.tiles()[wx][wy];
+        final Point windowPoint = new Point(wx, wy);
+        final Point gridPoint = new Point(wx/chunkSize, wy/chunkSize);
```
Anuken: Don't mark variables as `final` unless they're static/instance variables. It doesn't increase performance, and IntelliJ marks non-implicitly-final variables with different colors anyway.

core/src/io/anuke/mindustry/editor/MapRenderer.java
```diff
@@ -97,59 +101,91 @@ public void updateAll(){
...
+    }
 
-            mesh.draw(idxWall, region, wx * tilesize, wy * tilesize, 8, 8);
-        }
+    private void renderTileRegion(Point windowPoint, IndexedRenderer mesh, Tile tile, Block wall, int indexWall, int indexDecal) {
+        TextureRegion textureRegion;
+        final boolean isEditorIconExist = Core.atlas.isFound(wall.editorIcon());
 
-        float offsetX = -(wall.size / 3) * tilesize, offsetY = -(wall.size / 3) * tilesize;
+        Float tileRegion = new Float();
```
Anuken: This is extremely confusing instantiation syntax. It looks exactly like you're creating a new wrapper `Float`, and not a `Rectangle2D.Float`. Only after looking at the imports and usage can you figure out what's going on.

core/src/io/anuke/mindustry/editor/MapRenderer.java
```diff
@@ -97,59 +101,91 @@ public void updateAll(){
...
+        Tile tile = getTile(windowPoint);
         Block wall = tile.block();
 
-        TextureRegion region;
+        int indexWall = chunkStartPoint.x + chunkStartPoint.y;
+        int indexDecal = chunkStartPoint.x  + chunkStartPoint.y + chunkArea;
 
-        int idxWall = (wx % chunkSize) + (wy % chunkSize) * chunkSize;
-        int idxDecal = (wx % chunkSize) + (wy % chunkSize) * chunkSize + chunkSize * chunkSize;
+        renderMeshRegion(windowPoint, mesh, tile, wall, indexWall);
```
Anuken: I don't see why splitting up this method is necessary at all. The new functions you're creating are only ever called from `render()` anyway; all this does is add another layer of method calls, making it more difficult to figure out what's going on.

Having lower cyclomatic complexity does not necessarily make a method more readable, especially in this case.

core/src/io/anuke/mindustry/editor/MapRenderer.java
```diff
@@ -163,4 +199,66 @@ public void dispose(){
...
+        }else{
+            mesh.draw(index, textureRegion,
+                      meshRegion.x,
+                      meshRegion.y,
+                      meshRegion.width,
+                      meshRegion.height);
+        }
+    }
+
+    private int getRandomValue(int index, int length) {
```
Anuken: To re-iterate my point from before: None of these methods should exist. `getRegionInstance` is simply more characters than just writing `Core.atlas.find`, while the rest of these methods (and classes) are just clutter. Tiny helper methods like these only harm readability and waste space.

core/src/io/anuke/mindustry/editor/MapRenderer.java
```diff
@@ -97,59 +101,91 @@ public void updateAll(){
...
+        tileRegion.y = windowPoint.y * tilesize + offset.y;
+        tileRegion.width = textureRegion.getWidth() * Draw.scl;
+        tileRegion.height = textureRegion.getHeight() * Draw.scl;
+
+        drawMesh(mesh, tile, textureRegion, indexDecal, tileRegion, false);
         mesh.setColor(Color.WHITE);
     }
 
+
+    private void renderMeshRegion(Point windowPoint, IndexedRenderer mesh, Tile tile, Block wall, int indexWall) {
```
Anuken: Your methods have an unnecessary space after the closing parenthesis.

BlaCkinkGJ:
A lot of your advice is going to help me a lot. I think I made a lot of mistakes because it was the first pull request related in game project. I will consider your advice when posting a pull request later. Thank you for your advice. Have a nice day!

core/src/io/anuke/mindustry/editor/MapRenderer.java
BlaCkinkGJ: Originally, I used the `Rectangle2D.Float something = Rectangle2D.Float(x, y, w, h)` like that. But I ran my code formatting program and this was happened :( I'm really sorry about this to make you confusing.

PR #510, opened 2019-06-03
Fix call to method with a wrong name (Scope::getMembers() instead of Scope::getSymbols())

Paragoumba:


Anuken:
Can you provide an example of incorrect behavior of this code and explain how this PR fixes it?

Paragoumba:
In fact the error comes from my JDK version. I think Mindustry uses JDK 8 but I used JDK11 to build. There might be a change in the method name that causes a build error. I'll check further to see if my guess is right.

PR #522, opened 2019-06-10
Update README.md

ZacBytes:
fixed the google play store image L0L uWu


Anuken:
thanks but please don't say "L0L uWu" ever again

PR #528, opened 2019-06-14
English bundle graphite description

Arkanic:
graphite now has a description
its not much
Yay. my fist commit


core/assets/bundles/bundle.properties
```diff
@@ -570,6 +570,7 @@ item.lead.description = A basic starter material. Used extensively in electronic
 item.coal.name = Coal
 item.coal.description = A common and readily available fuel.
 item.graphite.name = Graphite
+item.graphite.description = Mineralized carbon, used as ammunition as well as electrical insulation.
```
Anuken: Should be "used *for* ammunition"; don't use "as" three times in a sentence.

Arkanic: oof
ill fix it

Arkanic: @anuken i might start doing this more often
i hope ill get better eventually

Anuken:
:+1:

PR #536, opened 2019-06-25
formatting en bundle with backslash apostrophe

Arkanic:
just cleaning the properties code


Anuken:
Why? This doesn't change any of the displayed text.

Arkanic:
@Anuken 
just tidying the code

Anuken:
If the file already works fine without backslashes, how does adding them make it more tidy?

Arkanic:
@Anuken you can close the pull request, i just made it because it makes the file look strange in code editors because section of it appear as strings, but you are right, it is quite pointless.

Anuken:
I haven't encountered that issue in IntelliJ or the Github source preview, so whatever incorrect syntax highlighting you could potentially be getting is probably a bug in the editor. Properties files don't have quoted strings.

Arkanic:
@Anuken this is probably to late, but yeh, one of my eclipse plugins had a setting with "display strings on .properties"

Arkanic:
looking back at me i did a lot of ping spam

Anuken:
image

Well, commenting on any PR does the same.

PR #553, opened 2019-07-08
Update bundle.properties

Prosta4okua:


core/assets/bundles/bundle.properties
```diff
@@ -158,14 +157,6 @@ cancel = Cancel
 openlink = Open Link
 copylink = Copy Link
 back = Back
-quit.confirm = Are you sure you want to quit?
```
Anuken: This text is still used [when quitting on mobile](https://github.com/Anuken/Mindustry/blob/12bff75586576ed78bcb1b1dff97bdbf0f93a321/core/src/io/anuke/mindustry/ui/dialogs/PausedDialog.java#L63).

PR #559, opened 2019-07-13
Resources war - new pvp gamemode

Natchuz:
# Resources war
The general idea of this gamemode is to feed a special block with items, and gain points. Some items are more valuable and collect more points. When the condition is met, a team with the fewest points gets eliminated.

## 2 Submodes
- Regular - The weakest team gets eliminated after a specified time
- Rushgame - The weakest team gets eliminated after someone reaches specified points threshold

## Mechanics
- Remember to protect your Item Eater at all cost. When destroyed, points are also gone.
- Core is immutable, but when an enemy team damages it, they steal points from Eaters.
- Buffs. They appear from time to time and makes specified item 2 times more valuable.
- Lifes. Team may have many chances before final elimination.

## Todo & to discuss
- [x] Item Eater. Any idea of a better name?
- [x] Item Eater. You know, my graphics skills end on drawing stickmen. And texture... Could someone help?
- [x] ~~Tag submode. I thought of this, but I'm not sure if this should be a thing. Like in regular mode, but time resets after the weakest team is changed.~~
- [x] ~~Resources war. Maybe this should be Economic Wars, or sth?~~
- [x] Balancing
- [x] This may collide with my other PR, especially eliminated dialog (I made this before)
- [x] Some strings, and descriptions
- [x] I implemented limited respawns for this mode. We can disable it only for regular pvp mode
- [ ] UI. It's too complex. And this is problem


BasedUser:
Item Eater -> Item Bin/Void

Natchuz:
@BasedUser Item void would be confused with that one in the sandbox

Prosta4okua:
java.lang.NullPointerException
	at io.anuke.mindustry.core.Logic.eliminateTeam(Logic.java:165)
	at io.anuke.mindustry.gen.Call.eliminateTeam(Call.java:158)
	at io.anuke.mindustry.core.Logic.eliminateWeakest(Logic.java:129)
	at io.anuke.mindustry.core.Logic.update(Logic.java:283)
	at io.anuke.arc.backends.headless.HeadlessApplication.mainLoop(HeadlessApplication.java:98)
	at io.anuke.arc.backends.headless.HeadlessApplication$1.run(HeadlessApplication.java:63)

Natchuz:
@Prosta4okua Except auto hosting doesn't work (at least for me). I have to hold timer until there are enough players. I missed it.

Arkanic:
@Prosta4okua 
io.anuke.mindustry.core.Logic.eliminateTeam may require a bit of editing to get this to work
(basic rulesets may not always be correct, and the eliminateTeam call seems to be hardwired to the "core destroyed, u destroyed" logic. This seems like it may be rather easy to fix though.

Natchuz:
@Arkanic what do you mean? D:

Arkanic:
@Natchuz you already fixed it
The rules class looked a bit hardwired to "if your core dies you die" 
But I may be wrong

Id like to help with this

Natchuz:
@Arkanic the simplest way is to destory cores, so core won't become too complex.

Arkanic:
@Natchuz ok

Is there a way I can help with this?

Natchuz:
@Arkanic See todo section. Graphic especially. Then PR into my repo

Arkanic:
@Natchuz ok. I may not be too much help, but I'll see what I can do.

Arkanic:
@Natchuz I can't seem to find the text you've added to the bundle in your repo?
 halp
i was going to edit some stuff

Prosta4okua:
@arcanic, second file is a bundle in changes.

Arkanic:
@Prosta4okua found it.
https://github.com/Natchuz/Mindustry/pull/1

Arkanic:
@Natchuz do you think "Item consumer" or "Large item void" sounds a bit more in the theme? at this point, you may only want to change it's superficial name because of all of the objects and variables named after the original name.

Natchuz:
@Arkanic I still think, that its role is going to be confused with that one in sandbox

Arkanic:
@Natchuz I think if the "item eater" is relatively core-sized, you will be able to distinguish it from other blocks :D 
I made a pull, it's some edits to the bundle.properties

Arkanic:
@Natchuz thanks

Arkanic:
@Natchuz some name ideas:
Item economizer - my favorite, but probably also the geekiest one
Item consumer

core/assets/bundles/bundle.properties
```diff
@@ -852,8 +852,8 @@ block.container.name = Container
 block.launch-pad.name = Launch Pad
 block.launch-pad.description = Launches batches of items without any need for a core launch.
 block.launch-pad-large.name = Large Launch Pad
-block.items-eater.name = Items Eater
-block.items-eater.description = He wants to EAT.
+block.item-eater.name = Item Eater
+block.item-eater.description = He wants to EAT.
```
Arkanic: @Natchuz yes

Natchuz: yes

Natchuz:
Look's like I'm done

Arkanic:
@Natchuz
You have your very own game mode now

CKHammer:
name idea: Natchuz Mode
inspiration: Trigg Mode

Arkanic:
@CKHammer that's good

Anuken:
I don't think Mindustry, in its current state, can support advanced gamemodes like this properly. Changing `State`, `Logic` *and* `Rules`, all to support one mode, is not what I would call 'good organization'-- not that there any alternatives right now. It would be extremely to 'fix' some logic or state in a future update and break this entire gamemode, unintentionally. 

Ideally, each mode should have its own compartmentalized state, rules and logic. It hasn't been a problem until now, as there are only 2-3 unique types of modes, but this PR really lays bare the flaws in the mode system.

Anuken:
I'll shelve this for now.

PR #565, opened 2019-07-28
Update bundle_ko.properties

Alexjimin:
Fixed some unnatural words.
Founded more but will fix later.


Anuken:
Seems to have been fixed by other pull requests; causes conflicts.

PR #578, opened 2019-08-09
pls delete it

AmateurPotion:
Because of this, co-op game mode(use 2+teams map) is not implemented. Please remove it.

### DIFF
--- a/core/src/io/anuke/mindustry/game/Gamemode.java
+++ b/core/src/io/anuke/mindustry/game/Gamemode.java
@@ -28,7 +28,6 @@ public enum Gamemode{
     }, map -> map.teams.contains(waveTeam.ordinal())),
     pvp(rules -> {
         rules.pvp = true;
-        rules.enemyCoreBuildRadius = 600f;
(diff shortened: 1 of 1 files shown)


AmateurPotion:
If the size of the map is 200 × 200 or less, and there are multiple cores in a map, there is a problem with the play because the overlapping range is impossible.

Prosta4okua:
You can customize this rule when you are creating map.

Anuken:
You can just change it in the map-specific rules, as said above. No need to change it for everyone.

AmateurPotion:
I tried to modify and run the map's custom rules for this, but there was a bug where the game rules-never install core zones apply on the client but not on the server.

Anuken:
Then you should specify what the bug is, which map it happened on, and how to reproduce it
.

AmateurPotion:
The gamerule- enemy no build zone was applied correctly when the server was opened as a game client, but this rule was not applied properly when opened as a server (like server-release.jar).

PR #580, opened 2019-08-15
Update bundle.properties

AmateurPotion:


Anuken:
I'll add my own descriptions for these when the time is right.

PR #581, opened 2019-08-15
add co_op

AmateurPotion:


AmateurPotion:
I think this gamemode need rule { rules.enemy = "crux" }

Anuken:
What is the point of this mode? If you want to cooperate, just play on the same team.

Arkanic:
isn't this just TeamVTeam?
correct me if i'm wrong

Anuken:
It's like PvP, only the teams can't directly attack each other, since each player does 0 damage. I believe the idea is that they work together to defeat waves.

AmateurPotion:
Mindustry already has a great cooperative game mode called Survival Mode. However, survival mode feels more like a single player than a multiplayer because it shares all the resources, what to protect and energy. 
So I thought about co-op game mode. To illustrate this game mode, similar to pvp, teams have different cores, start in different places, use different energy synchronizations, but only one enemy, the crux (red) team. 
In addition, if resources are needed to be shared, they can be shared through the gray team's building, no core build radius is 0 to defend each other's cores, and other than the red team, turrets or units can't attack each other's buildings or units.

AmateurPotion:
It's like PvP, only the teams can't directly attack each other, since each player does 0 damage. I believe the idea is that they work together to defeat waves.) That's a right understanding.  Thank you.

Dexapnow:
I would like to see gamemode like that.
Anuke, players sometimes would like to play together, but there is a point when having the same "inventory" is anoying, I would like to play with my "friends" together but with separate core/inventory.

Anuken:
Closing, as the game doesn't support "co-op" properly right now, and this implementation doesn't really do it properly. It might be implemented someday, but there are higher priority things to take care of.

PR #588, opened 2019-08-17
Update bundle_ko.properties

AmateurPotion:


AmateurPotion:
oh no internet error xd

PR #593, opened 2019-08-18
fixed scrolling with mouse (2)

Kylof61034:
Hey,

I noticed a little bug, which made it impossible to scroll through some dialogs with the mouse on a desktop machine.

Affected Dialogs:
- ControlsDialog
- AboutDialog
- SettingsDialog (Graphics)
- etc.

Note: This is part two of the fix. Part one is on the Arc repository.


Anuken:
I did some investigating, and the issue appears to be caused by two things.

1. Not all dialogs inherit from FloatingDialog, so the scroll focus isn't set for normal Dialogs. You fixed this by moving the focusing code into the Dialog class.
2. When the game resizes, a dialog may be rebuilt, which causes the scroll focus to get lost. You solved this by manually setting the scroll focus to a specific scroll pane on a rebuild event for every dialog that has one.

This second part is an issue. While it does function correctly, it may be prone to future bugs, as every dialog with a scrollpane now needs to set its focus manually. To solve this, I've pushed a workaround that implements the first section of this PR, as well as automatically sets the scroll focus once a dialog is rebuilt after a resize event.

PR #606, opened 2019-08-23
Update to Bundle_tr

SpicoTheSpiritGuy:
Updated lines 1 to 115 (a few lines havent beed updated, since I couldnt translate them very well, so I will update them at the probably next update)


SpicoTheSpiritGuy:
there were some conflicting lines so I fixed them now

PR #619, opened 2019-08-26
Real CHAOS

AmateurPotion:


Arkanic:
that is 
OP
*insert accurate lightning here*

Anuken:
nah, I think a new enemy with lightining-based attacks would be preferable

Issue #545, opened 2019-07-03
Suggestions, Ideas & Feature Requests


Arkanic:
i've noticed a sudden influx of ideas, and just to clean things up, please comment them here.

Arkanic:
Just thinking about some stuff here...

I have noticed that there is a support mech (the Tau) but what about some kind of defense mech?
something like an aegis with a "Shield project" ability (Player hits key/double taps, shield similar to the force projector appears, effectively temporarily shielding the surrounding area (assuming the player gets locked in place). On second thought, the Fortress mech could play this role, (it may make it to OP tho) seeing as it would fit with the Mech's description.

~~~Totally didn't get the idea from the SW battlefront 2 personal shield~~~

Anuken:
Moving older issues here.

*Luxray5474:*
>What if there was an option that toggles the camera between following the unit like normal and the camera controlling the unit like in mobile? That would be useful for quickly glancing at a particular area in advance before a slow ground mech reaches it.

Arkanic:
*Olo123pl:*
>I think about the slag and mass driver and invented a weapon that would shoot like a catapult and create a temporary slosh of slag on the path

Arkanic:
*xslendix:*
> Here are some stuff that i wanna see in mindustry (please read it took me a lomg time to make):
> 
> * Material counter
>   
>   * When clicked, a GUI will appear where it allows you to set how many materials you want, what type of material you want, the operation (<, >, =<, =>, ==).
>   * If the condition is met, the materials are sent forward and the rest are either checked again or sent on the side (option in the GUI).
>   * The delay between checks is a minimum of 1 material(s) per second and the maximum of like idk 10 maybe?
>   * Creates power when condition is met (needs to be powered, can act like a power node i guess)
>   * This sounds like a bad idea but at least try it. Would make us able to create advanced stuff
> * LEDs. Yes.
>   
>   * The color can be chosen from a GUI
>   * Will lit up if has power
>   * Brightness is determined by the amount of power it receives
> * Material blocker
>   
>   * It can be powered (toggle or minecraft door mode, this can be changed from a GUI) or toggled by a player.
> * Electricity power level changer (change the name is more or less like a resistor)
>   
>   * When fed power, the nearest node will get that power that has been received reduced my an amount set in a GUI
> * Inverters
>   
>   * Kinda like redstone torches in mintcraft is fed power, it outputs no power, it gets no power, emits a small amount of power.
> * Power amplifier
>   
[...]

PR #624, opened 2019-08-30
[Ready to merge] add leds

LeRoiDesKiwis:
I made the idea of @Arkanic from #545 👍 
image

## Todo

- [x] Support save and load (actually you lost the color when you reload the world)

- [x] Fix hitbox bug
- [x] remove the texture "oh no" in the background

## What does it looks like
gif


Luxray5474:
Awesome execution! Can't wait for the texture :+1:

LeRoiDesKiwis:
> Awesome execution! Can't wait for the texture 👍

You want a texture ? I think it's better to don't use a texture and to totally fill the block with the color

LeRoiDesKiwis:
I think the conflicts is not a problem because i added the led texture. You can change the texture if you want, i just put a white texture, i'm a very bad graphist ^^.

LeRoiDesKiwis:
Just a preview : 
gif

Arkanic:
@LeRoiDesKiwis this wasn't my idea, but @xslendix 's. i merely referenced his thoughts from an older issue.
:)

J-VdS:
I don't know Anuke is going to merge. But you can adapt you code and make a small mod (if possible). https://github.com/Anuken/TestMod He would probably use this method for the steamworkshop. But it is still early alpha.

Arkanic:
@LeRoiDesKiwis i think you may need a texture for it, even if it's only a border, as a solid colour makes it look kind of unfinished.
:)

Luxray5474:
If you could add a bloom effect for blocks and not effects these would be 2x better :)

Anuken:
To be honest, I... don't really understand what the point of LEDs is. Since there is no day-night cycle, they can't provide any real light, and as they consume power, making pixel art isn't viable-- the power beams would make a mess. Even if the power requirement were removed, couldn't you just use sorters for a 16-color palette...?

PR #630, opened 2019-09-02
Extra events

J-VdS:
I added a couple of events. 
* PlayerJoin
* PlayerLeave

(not tested)


core/src/io/anuke/mindustry/core/NetServer.java
```diff
@@ -74,8 +74,9 @@ public NetServer(){
         });
 
         Net.handleServer(Disconnect.class, (id, packet) -> {
-            Player player = connections.get(id);
+            Player player = connections.get(id);           
             if(player != null){
+                Events.fire(new PlayerLeave(player));
```
Anuken: This won't accurately fire for all player leaving events. Instead, fire the event in `onDisconnect(...)` inside the `hasConnected` if-statement, before the message is sent.

PR #634, opened 2019-09-04
Coherent help message for a button (minor).

amrsoll:
Replaces "Import Legacy Image" with "Import Legacy Map" as per the description. 

Should it keep "Import Legacy Image"?


Anuken:
I opted for a slightly different text change in 168011586bd35071cd4a820a929444891c250f8c.

Issue #18, opened 2017-12-14
Localization Checklist


Anuken:
This will be a separate checklist for setting up localization.

LibGDX I18Bundle documentation: [Link](https://github.com/libgdx/libgdx/wiki/Internationalization-and-Localization)

- [x] Add I18Bundle variable to Control/Vars class, set up appropriate way to fetch text easily
- [x] Set up assets directory for text files
- [x] Create basic test bundle and get 'fetching' working
- [x] Make I18Bundle work with all UI
- [x] Make I18Bundle work with all block descriptions and extended descriptions
- [x] Make I18Bundle work with map names (?)
- [x] Remove "short descriptions", if possible, and replace them with something else
- [x] Transcribe all ingame text to I18Bundle strings
- [x] Translate to French (contact Omicron translation team)
- [x] Translate to Russian (already found a translator who is willing, or I can even do it myself)
- [x] Translate to German

Luxray5474:
...can't we just use JSON? What are the advantages or benefits in using I18Bundle instead of JSON? I've already gone through about half of the total game text and put it in JSON, and was thinking we'd be able to import the JSON as an object in the files that the game has text in, and then reference them wherever it is needed.

Anuken:
Storing the text and fetching it isn't the hard part; the hard part is making it load the right text bundle for the locale, and having it fallback when necessary. 
Advantages:

- simpler format, no quotes, easy to edit
- automatically format strings you pass in with specific arguments
- automatically load the right bundle for the locale of the user
- automatically fall back to the right bundles when the optimal one is not found
- works across all platforms, including HTML5 (a lot of standard Java stuff doesn't work there, so that's a big deal)
- probably faster to load and use than JSON, as the format is much simpler

If I use JSON, I'll have to re-implement all of these features, probably for each platform, too. The formats aren't _too_ different; can't you do something like remove all quotes and replace `:` with `=`, using the appropriate tools in your editor?

Luxray5474:
Ah, gotcha. I'll work on porting all the gametext i stored onto the JSON into I18Bundle. Starting pull request when done.

Epowerj:
I just want to note that there is likely an automatic JSON to l18Bundle converter out there somewhere that you could use.

Luxray5474:
...I wish I had known that... *earlier*... 😒

Anuken:
If you give me the files, I can probably write a script that can convert them very quickly. The l18Bundle format is very simple.

[...]

PR #635, opened 2019-09-04
Add French app description for the Play Store.

amrsoll:
I hope this is the right way to do it. I asked about this recently in #18


Anuken:
Yep, this is correct. I'll add some info about it to the translation guide soon.

PR #652, opened 2019-09-11
Show position of player

mayli:
Useful in multi-player games

screenshot
image


Anuken:
I don't think extra information like position should be displayed along with FPS. This should be a separate setting or keybind toggle.

mayli:
Maybe display it with the minimap? Minecraft's minimap mod usually does that.
How about make it a separate setting, and put it under minimap?

image

Anuken:
That would work.

PR #662, opened 2019-09-14
Plugin YML

BagayRodolfo:
Simple edit on Pluigns.java

This creates new folder for each plugin before load.

(Will use fork next time)

### DIFF
--- a/core/src/io/anuke/mindustry/plugin/Plugins.java
+++ b/core/src/io/anuke/mindustry/plugin/Plugins.java
@@ -1,6 +1,5 @@
 package io.anuke.mindustry.plugin;
 
-import io.anuke.arc.Core;
 import io.anuke.arc.collection.*;
 import io.anuke.arc.files.*;
 import io.anuke.arc.function.*;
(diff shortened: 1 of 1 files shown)


Anuken:
I don't think YML should be used as a standard config format; JSON would be preferable.

BagayRodolfo:
Okay, Use JSON file named as <plugin_name>.json then put it on the same level of the jar file?

Milinai:
How would this be useful?

BagayRodolfo:
Some plugins requires configuration based on the owner of the server. This will offer more option ,flexibility and variety of each plugin.

For now, plugins cannot offer much, But i know someday it will be more useful.

Kieaer:
My Essentialsis the only plugin I have that uses the settings file.
Plugins need to be very complex to be used...

Anuken:
Implemented in f65ab002500258269abdf48a811569597dc7fe83 with a `getConfig()` method that can be used in a plugin's constructor, e.g.

`getConfig().writeString("{}");`

...would create a 'config.json' in a plugin-specific config directory with the contents "{}".

PR #668, opened 2019-09-16
Fixes minimap zoom bug

dheld:
This bug appears to primarily affect desktop users, since mobile allows for full map scrolling during pause.  The number of listeners increased every time the minimap was opened, which would cause each scroll event to get multiplied by the number of listeners.  Clearing the listeners fixes the problem.


Anuken:
I fixed this in d49b15b2a0fdfecf05360895ea48986cc7847758 before I saw this PR; I should've called `clear()` instead of `clearChildren()`.

PR #669, opened 2019-09-16
Update bundle_cs - Czech translation

firdacz:
I will keep working on the translation when I have the time, squashing it when I see it not being merged yet. (To keep it clean, hoping that is desired.)

BTW: Is there a way to reference key-bindings from the documentation / tutorial?
Would be nice to instruct the starter to actually move the ship next to the deposit before drilling it,
totally did not expect that (expected the ship to move on its own, like in Star Craft, Total Anihilation etc.)

Don't know what this is used for:
server.kicked.vote = You have been vote-kicked. Goodbye.
(similar to server.kicked.kick = Byl jsi vykopnut ze serveru!
but does it mean kicked FROM voting or BY voting or what?)


firdacz:
I will squash it into single commit if desired.

Is there a way to reference key-bindings from the documentation / tutorial?
Would be nice to instruct the starter to actually move the ship next to the deposit before drilling it,
totally did not expect that (expected the ship to move on its own, like in Star Craft, Total Anihilation etc.)

Anuken:
'server.kicked.vote' means that a group of people voted to kick that person. You can mention that WASD can be used to vote in the tutorial, if you think it's aproppriate.

PR #671, opened 2019-09-16
Minor code cleanup and fail-fast NPE validation for status effects

dheld:
This PR undoes your band-aid for null status effects, but adds earlier validation to help debugging.  Alternatively, when a null effect is requested, we can set it to none instead and dump a stack trace, but then it will be less obvious when something blows up, and we will have to rely on users being diligent and sending traces.

### DIFF
--- a/core/src/io/anuke/mindustry/Validate.java
+++ b/core/src/io/anuke/mindustry/Validate.java
@@ -1,0 +1,21 @@
+package io.anuke.mindustry;
+
+/**
+ * Utility methods for parameter and expression validation.  API copied from Apache Lang Validate
+ * https://commons.apache.org/proper/commons-lang/javadocs/api-3.9/org/apache/commons/lang3/Validate.html
+ */
+public class Validate{
+    /**
+     * Returns value or throws NPE if it is null
+     * @param value the variable to validate
+     * @param name  the variable name
+     * @param <T>   the type of value
+     * @return      value
+     */
+    public static <T> T notNull(T value, String name){
+        if(value == null){
+            throw new NullPointerException(name);
+        }
+        return value;
+    }
+}
--- a/core/src/io/anuke/mindustry/entities/units/Statuses.java
+++ b/core/src/io/anuke/mindustry/entities/units/Statuses.java
@@ -124,7 +124,7 @@ public class Statuses implements Saveable{
 
     @Override
     public void readSave(DataInput stream, byte version) throws IOException{
-        statuses.forEach(Pools::free);
+        statuses.each(Pools::free);
         statuses.clear();
 
         byte amount = stream.readByte();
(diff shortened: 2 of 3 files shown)


core/src/io/anuke/mindustry/entities/units/Statuses.java
```diff
@@ -134,28 +124,34 @@ public void writeSave(DataOutput stream) throws IOException{
 
     @Override
     public void readSave(DataInput stream, byte version) throws IOException{
-        for(StatusEntry effect : statuses){
-            Pools.free(effect);
-        }
-
+        statuses.forEach(Pools::free);
```
Anuken: forEach isn't compatible with Android, use `each`.

core/src/io/anuke/mindustry/Validate.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public class Validate{
+    /**
+     * Returns value or throws NPE if it is null
+     * @param value the variable to validate
+     * @param name  the variable name
+     * @param <T>   the type of value
+     * @return      value
+     */
+    public static <T> T notNull(T value, String name){
```
Anuken: I am against adding an entire new class for a single method that could be inlined. `Validate.notNull(x, "abc")` isn't much shorter than  `if(x == null) throw new NullPointerException("abc")`. It's also worth mentioning that utilities like this should not be added into the main package, which is for core things like Vars only.

core/src/io/anuke/mindustry/type/StatusEffect.java
```diff
@@ -60,8 +60,8 @@ protected void trans(Supplier<StatusEffect> effect, TransitionHandler handler){
         transInit.add(new Object[]{effect, handler});
     }
 
-    @SuppressWarnings("unchecked")
-    protected void opposite(Supplier... effect){
+    @SafeVarargs
```
Anuken: This annotation can't be used in Android 19-14, which Mindustry supports.

core/src/io/anuke/mindustry/entities/units/Statuses.java
```diff
@@ -134,28 +124,34 @@ public void writeSave(DataOutput stream) throws IOException{
...
         }
     }
 
+    /**
+     * Creates a StatusEntry from {@code effect} and {@code time} and adds it to {@code statuses}
+     * @param effect    the status effect
+     * @param time      the effect duration
+     */
+    private void addStatus(StatusEffect effect, float time)
+    {
```
Anuken: newline brace

core/src/io/anuke/mindustry/Validate.java
dheld: Well, if you look at the Apache Commons Lang Validate class, you can see that there are quite a number of useful validations.  I just didn't want to add them until I used them.  But having a helper method is much nicer because you can inline the validation, which you can't do with the throw statement.  I could add the helper method to the class where I used it, but the idea is to use it more broadly to help stop NPEs.  So having it in its own class makes more sense.  What do you think?

core/src/io/anuke/mindustry/type/StatusEffect.java
dheld: Bummer.

Anuken:
I did some more investigating later, and apparently the cause of the crash wasn't related to null effects at all, but rather was caused by an incorrect transition setting an internal effect to null. Then, validation would not have fixed or even detected this issue at all.

PR #672, opened 2019-09-16
Fixes bug which let phase conveyor operate with any non-zero power; now requires full power to work

dheld:
AnnaLaura reported this bug a few weeks ago.

Instead of making it binary, we could degrade the throughput by scaling the transport by the `satisfaction`, but this is the easiest fix.

I noticed that Phased Conduit has the same problem, but the code looks like it does scale by `satisfaction`.  However, I can't measure any perf difference betwen a barely-powered and fully powered liquid bridge, so I'm not sure how the code works.

### DIFF
--- a/core/src/io/anuke/mindustry/plugin/Plugins.java
+++ b/core/src/io/anuke/mindustry/plugin/Plugins.java
@@ -67,7 +67,7 @@ public class Plugins{
         URLClassLoader classLoader = new URLClassLoader(new URL[]{jar.file().toURI().toURL()}, ClassLoader.getSystemClassLoader());
         Class<?> main = classLoader.loadClass(meta.main);
         metas.put(main, meta);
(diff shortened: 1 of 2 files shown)


core/src/io/anuke/mindustry/plugin/Plugins.java
```diff
@@ -67,7 +67,7 @@ private LoadedPlugin loadPlugin(FileHandle jar) throws Exception{
         URLClassLoader classLoader = new URLClassLoader(new URL[]{jar.file().toURI().toURL()}, ClassLoader.getSystemClassLoader());
         Class<?> main = classLoader.loadClass(meta.main);
         metas.put(main, meta);
-        return new LoadedPlugin(jar, zip, (Plugin)main.newInstance(), meta);
+        return new LoadedPlugin(jar, zip, (Plugin)main.getDeclaredConstructor().newInstance(), meta);
```
dheld: Fixes a build warning.

PR #673, opened 2019-09-17
Translate to Japanese by Eru

EruJapan:
I was translated to Japanese.
Discord: eru_java#9010


EruJapan:
I translated it. Please check.
I have checked it many times, but there may be some sorry mistakes.

Nerahikada:
This translation is not good because the file is old and contains a literal translation.
I'm working on a better translation.

https://github.com/Nerahikada/Mindustry

PR #688, opened 2019-09-21
Add ability to rotate each block in a line

Synray:
This PR lets you rotate each block in a straight line placement.

rotateLine

It's very handy for situations like the one above.

~This isn't ready to merge, since as it is now, drawing a straight line won't automatically set the rotation.~

I have a few ideas for improving it, but since it's my first PR here, I wanted to check first if this kind of feature is okay.

For one thing, should I move this to DesktopInput, or would it make sense to have this feature on mobile as well?

### DIFF
--- a/core/src/io/anuke/mindustry/input/DesktopInput.java
+++ b/core/src/io/anuke/mindustry/input/DesktopInput.java
@@ -237,11 +237,9 @@ public class DesktopInput extends InputHandler{
         }
 
         if (mode == placing && block != null){
-            if (!overrideLineRotation && !Core.input.keyDown(Binding.diagonal_placement) && (selectX != cursorX || selectY != cursorY)){
-                if ((int) Core.input.axisTap(Binding.rotate) != 0){
-                    rotation = ((int)((Angles.angle(selectX, selectY, cursorX, cursorY) + 45) / 90f)) % 4;
-                    overrideLineRotation = true;
-                }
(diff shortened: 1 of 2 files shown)


Synray:
I fleshed it out, let me know if this is good or I should change something.

Anuken:
Are there any other important use cases for this besides that specific design?

BasedUser:
Yes, many modular designs in fact. Not just this.

Synray:
Yeah, especially designs that feed multiple materials into machines, or ones that separate one material into many different conveyers. It also works with designs that use conduits.

It's a small feature, but when I tested it in an online game I found I used it for things I hadn't intended it for. For example, when placing long conveyers like this, where it's hard to judge horizontal distance, you can just start from the destination:
rorateLine2

core/src/io/anuke/mindustry/input/DesktopInput.java
```diff
@@ -236,6 +236,17 @@ void pollInput(){
             selectY = tileY(Core.input.mouseY());
         }
 
+        if (mode == placing && block != null){
+            if (!overrideLineRotation && !Core.input.keyDown(Binding.diagonal_placement) && (selectX != cursorX || selectY != cursorY)){
+                if ((int) Core.input.axisTap(Binding.rotate) != 0){
```
Anuken: This if-statement can be merged with the one above it.

Anuken:
Looks like a good idea, then. This probably won't be so simple or intuitive to implement for mobile, so let's not think about that yet.

Synray:
Thanks!

PR #689, opened 2019-09-21
Update bundle_it.properties

Maxwelllondon92:
Minor corrections.
Fixed douled name for wave tower (was the same as lancer)
Changed spore pod name (translation was wrong)
Changed Launch pad name (now translate as space elevator)


Maxwelllondon92:
updated:

PR #707, opened 2019-09-24
Can I add it?

AmateurPotion:
My real name


Anuken:
Instead of putting yourself there twice, you could do `Potion / 김정인` (if that's valid).

PR #717, opened 2019-09-26
hud: show player position

mayli:
Now it looks like

image


core/src/io/anuke/mindustry/ui/fragments/HudFragment.java
```diff
@@ -249,8 +250,15 @@ public void build(Group parent){
             }).top().left();
         });
 
+        Table map_table = new Table();
```
Anuken: You shouldn't create a table variable here. Use the table lambda syntax instead, e.g. `.table(t -> /**stuff*/)`.

core/src/io/anuke/mindustry/ui/fragments/HudFragment.java
```diff
@@ -249,8 +250,15 @@ public void build(Group parent){
...
         //minimap
-        parent.fill(t -> t.top().right().add(new Minimap()).visible(() -> Core.settings.getBool("minimap") && !state.rules.tutorial));
+        map_table.add(new Minimap().visible(() -> Core.settings.getBool("minimap") && !state.rules.tutorial));
+        map_table.row();
+        //position
+        map_table.label(() -> String.format("%.0f,%.0f", Vars.player.getX(), Vars.player.getY())).visible(
```
Anuken: Use player.x/y instead of the getters, and don't use `String.format`, as that is very slow. Instead, just do `(int)player.x + "." + (int)player.y`.

core/assets/bundles/bundle.properties
```diff
@@ -48,6 +48,7 @@ customgame = Custom Game
 newgame = New Game
 none = <none>
 minimap = Minimap
+position = Position
```
Anuken: Unless I'm missing something, this is unnecessary.

Luxray5474: variable name consistency

Anuken: What? I'm saying this string isn't used anywhere in the PR, or in the base game, as far as I'm aware.

PR #722, opened 2019-09-26
Partial Cross Team Interaction Exploit Fix

CKHammer:
I still have no idea how GitHub works after all these days.
But here are some commits that prevent hackers from interacting with other teams' blocks.


Anuken:
Thanks.

CKHammer:
np

PR #738, opened 2019-09-28
Node edit

AmateurPotion:
Many connections of nodes make the game too easy.


Anuken:
I don't think easy node connection is a bad thing.

PR #741, opened 2019-09-28
Made unloaders able to take items out of bridge conveyors / phase conveyors

ado1928:


Anuken:
Is this really a good idea? I could see someone trying to unload items out of something, then having it go directly into a bridge. This could potentially break designs.

ado1928:
It would add a some new designs, but break some too. 

I don't think the issue you mentioned would be a problem.
image

image

Anuken:
>  but break some too

I'd rather not break designs unless there's a very good reason for it. This would require a lot of testing before merging as well.

PR #750, opened 2019-09-30
Update bundle_pt_BR.properties

eirr1:


YaFou:
The pull request looks like stale. I think you should close it because the author (@) is inactive for a few months. And others pull requests are merged like #1122 or #1128.

PR #756, opened 2019-10-01
Change item amount display of Block Inventories to show decimals

Micky5991:
To improve a more precise understanding how many items are in certain
blocks, this change will alter the rounding of those item amounts.

It will display the decimals if the amount is between 1000 and 9999
also between 1 million and 9,9 million.

java_2019-10-01_13-13-11

### DIFF
--- a/core/src/io/anuke/mindustry/ui/fragments/BlockInventoryFragment.java
+++ b/core/src/io/anuke/mindustry/ui/fragments/BlockInventoryFragment.java
@@ -196,11 +196,11 @@ public class BlockInventoryFragment extends Fragment{
         if(f >= 10000000){
             return (int)(f / 1000000f) + "[gray]mil[]";
         }else if(f >= 1000000){
-            return (Math.floor(f / 100000) / 10f) + "[gray]mil[]";
+            return Strings.fixed((float) Math.floor(f / 100000) / 10f, 1) + "[gray]mil[]";
         }else if(f >= 10000){
             return (int)(f / 1000) + "k";
         }else if(f >= 1000){
-            return (Math.floor(f / 100) / 10) + "k";
+            return Strings.fixed((float) Math.floor(f / 100) / 10f, 1) + "k";
         }else{
             return (int)f + "";
         }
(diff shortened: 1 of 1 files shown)


core/src/io/anuke/mindustry/ui/fragments/BlockInventoryFragment.java
```diff
@@ -193,10 +193,14 @@ public void touchUp(InputEvent event, float x, float y, int pointer, KeyCode but
...
-        if(f >= 1000000){
+        if(f >= 10000000){
             return (int)(f / 1000000f) + "[gray]mil[]";
-        }else if(f >= 1000){
+        }else if(f >= 1000000){
+            return (Math.floor(f / 100000) / 10f) + "[gray]mil[]";
+        }else if(f >= 10000){
             return (int)(f / 1000) + "k";
+        }else if(f >= 1000){
+            return (Math.floor(f / 100) / 10) + "k";
```
Anuken: This should be formatted. Even if a potential result here looks like `10.2`, it *may* show up as `10.200003` or something similar when you turn it into a string.

core/src/io/anuke/mindustry/ui/fragments/BlockInventoryFragment.java
```diff
@@ -193,10 +193,14 @@ public void touchUp(InputEvent event, float x, float y, int pointer, KeyCode but
 
     private String round(float f){
         f = (int)f;
-        if(f >= 1000000){
+        if(f >= 10000000){
             return (int)(f / 1000000f) + "[gray]mil[]";
-        }else if(f >= 1000){
+        }else if(f >= 1000000){
+            return (Math.floor(f / 100000) / 10f) + "[gray]mil[]";
```
Anuken: See other comment.

Micky5991:
I kept `Math.floor` where it is, to achieve a flooring of the value instead of a round that may increase the decimals to unwanted values.

Anuken:
Due to space issues with other translations, I'm going to close this:

image

With another decimal place, this would become a mess.

PR #764, opened 2019-10-02
"Fixed" bloom being dark when zooming out

ado1928:
My attempt at making the bloom better

### DIFF
--- a/core/src/io/anuke/mindustry/core/Renderer.java
+++ b/core/src/io/anuke/mindustry/core/Renderer.java
@@ -104,7 +104,9 @@ public class Renderer implements ApplicationListener{
         Color.white.set(1f, 1f, 1f, 1f);
 
         camerascale = Mathf.lerpDelta(camerascale, targetscale, 0.1f);
-
+        
+        Bloom.setBloomIntesity(12f / camerascale);
+        
         if(landTime > 0){
             landTime -= Time.delta();
             landscale = Interpolation.pow5In.apply(minZoomScl, Scl.scl(4f), 1f - landTime / Fx.coreLand.lifetime);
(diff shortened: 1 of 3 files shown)


core/src/io/anuke/mindustry/graphics/Bloom.java
```diff
@@ -331,7 +331,7 @@ private void gaussianBlur(){
      * @param intensity multiplier for blurred texture in combining phase. must be
      * positive.
      */
-    public void setBloomIntesity(float intensity){
+    public static void setBloomIntesity(float intensity){
```
Anuken: This is a very bad idea. **Never** make things static unless they need to be, especially in a case like this where the bloom can be uninitialized.

core/src/io/anuke/mindustry/core/Renderer.java
```diff
@@ -104,7 +104,9 @@ public void update(){
         Color.white.set(1f, 1f, 1f, 1f);
 
         camerascale = Mathf.lerpDelta(camerascale, targetscale, 0.1f);
-
+        
+        Bloom.setBloomIntesity(12f / camerascale);
```
Anuken: What if bloom is disabled, and/or bloom doesn't work on someone's machine?

ado1928: I think that this whole issue about bloom could be fixed by adjusting Gaussian blur size based on zoom, but that would need a different code for the Gaussian shader. Also, by disabling the original layer and displaying only the Gaussian layer, I've noticed that its much darker than the original, is this intentional or just an issue?

Anuken: That's either a side effect of the blurring taking into account nearby black clear pixels, or just how the threshold filter outputs the result.

Anuken:
Closing until a better solution is found.

PR #777, opened 2019-10-05
Hybrid grid hotkey mode.

eldstal:
Use number keys to select a building category, then the 4x4 grid of keys (1234 through zxcv) to select a recipe from that category.

No modifier keys needed, makes it easier to play with a keyboard.


1RedOne:
Could you provide some testing instructions for this?  Is this something I need to enable in settings first?

eldstal:
No configuration needed. It adds 10 additional keybinds to the settings menu (by default set to the numbers 1-9 and 0) and is always enabled.

While playing, try this:
Press 3 to select the "conveyors" category
Press W to select the second thing in the second row (corresponds to the graphical menu locations)

The result is that you enter place mode and can build that block.

core/src/io/anuke/mindustry/ui/fragments/PlacementFragment.java
```diff
@@ -98,18 +132,32 @@ boolean gridUpdate(InputHandler input){
...
             Array<Block> recipes = getByCategory(currentCategory);
             for(KeyCode key : inputGrid){
                 if(Core.input.keyDown(key))
                     input.block = (i < recipes.size && unlocked(recipes.get(i))) ? recipes.get(i) : null;
                 i++;
             }
+            return true;
         }
+
+        for (int i=0; i<categoryKey.length; i++) {
```
Anuken: Unnecessary spaces around parentheses, should be `for(){` / `}else{`
This occurs throughout the PR.

core/src/io/anuke/mindustry/input/Binding.java
```diff
@@ -19,6 +19,16 @@
     dash(KeyCode.SHIFT_LEFT),
     gridMode(KeyCode.BACKTICK),
     gridModeShift(KeyCode.ALT_LEFT),
+    build_category_1(KeyCode.NUM_1),
```
Anuken: All of these need to be put in `bundle.properties` to show up properly.

eldstal: I've added it to the base bundles.properties, does it also need to be added to the other region-specific translations or do they fall back to english until properly translated?

core/src/io/anuke/mindustry/ui/fragments/PlacementFragment.java
eldstal: Fixed.

eldstal:
If this is to be merged, it would probably be good to add some sort of visualization showing that the next press is going to select a recipe (i.e. when `PlacementFragment.selectingRecipe == true`).

For example, overlaying the recipe grid with the keys that would select them as a hint.
I wasn't sure how to properly implement this or what is a UI-suitable style so I've left it.

eldstal:
It might also be beneficial to have each recipe in an assigned place in the UI grid, so that they don't move around when you unlock new recipes. That way, you don't have to retrain yourself (or search among the rather similar icons) while playing. That's probably a bigger change though, so it doesn't belong here.

core/src/io/anuke/mindustry/input/Binding.java
Anuken: It falls back, and there's a script to add them to other translations that I run occasionally.

Anuken:
This doesn't seem to work well with the grid mode system. I've just tested it out, and it seems to conflict with block selection. Switching to section 1, then trying to switch to category 1 with the '1' key just selects the first block instead. Also, if you're going to be replacing the old category grid system (1, 2, Q, W, etc), it would be a good idea to remove it completely, instead of adding another layer on top.

eldstal:
If I understand you correctly (That pressing `1` and then `1` again selects the first block in the first category), that is as intended.
Using this scheme, selecting a block is always a pair of `<category, block>` presses, which is why an indicator in the UI of being 'half-way' might be useful.
Although this uses the same grid of keys, it should not conflict with the existing mechanism since that requires modifier keys.

Perhaps have it "finish" if the user uses the keys to select a category but then clicks a recipe instead of using grid keys to select it?

Anuken:
Sorry, typo. I meant pressing category 2, then trying to go to category 1 selects the first block. I see what you mean by pairs of presses, but that doesn't seem very intuitive. In fact, I would say many players would report it as a bug, even if some sort of indicator is added. Wouldn't it be a better idea to allow players to switch category and block independently?

Anuken:
Forgot to mention this:
> it should not conflict with the existing mechanism
Sure, but I don't think it's necessary anymore, and that control scheme is very confusing, in my opinion. Why have two different modes of block selection?

eldstal:
I would say that the two have different pros and cons.
The old system is less confusing (the grids correspond to the screen, which is nice) but a pain to use (modifier combinations contort your left hand).

This patch makes it more usable but less intuitive.

I suppose it would make sense to have this new system use separate keys for categories vs blocks, but then we're back to the problem of moving your hand around the keyboard looking for other keys.

Personally, I prefer a configurable hotbar as you would see in Factorio or Minecraft, which can then safely be bound to the number keys. This also does away with the confusion when you unlock new blocks and it moves around the old common ones in your menus.

If there is more interest in such a system, this key remapping is of little use.

Anuken:
A hotbar system may be the way to go, or perhaps an assignable radial menu with hotkeys could work?

Anuken:
Either way, I think adding these hotkeys will be redundant if a hotbar is implemented.

PR #778, opened 2019-10-05
Power Lasers Transparency

azureblue:
Sometimes power lasers look too bright and cover too much area (especially when they are clustered) and I've been missing a functionality that would allow to change that. So I implemented it.

It required some changes in block rendering (and scheduling blocks to render) code so that each visible laser is rendered **only once**. This particular change optimizes draws and increases FPS.

The opacity can by change in the Settings->Graphics->Power Laser Opacity.

I'v also slightly reduced lasers width.

How it looks like with different transparency values:
lasers-transparency


core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -281,9 +281,13 @@ public void drawLayer(Tile tile){
 
         for(int i = 0; i < entity.power.links.size; i++){
             Tile link = world.tile(entity.power.links.get(i));
-            if(linkValid(tile, link) && (link.pos() < tile.pos() || !(link.block() instanceof PowerNode) || !Core.camera.bounds(Tmp.r1).contains(link.drawx(), link.drawy()))){
-                drawLaser(tile, link);
-            }
+            if (!linkValid(tile, link))
```
Anuken: Unnecessary space after `if`, `continue` should be on the same line as the if-statement

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -281,9 +281,13 @@ public void drawLayer(Tile tile){
 
         for(int i = 0; i < entity.power.links.size; i++){
             Tile link = world.tile(entity.power.links.get(i));
-            if(linkValid(tile, link) && (link.pos() < tile.pos() || !(link.block() instanceof PowerNode) || !Core.camera.bounds(Tmp.r1).contains(link.drawx(), link.drawy()))){
```
Anuken: How does this handle cases where an extremely long-range node (e.g. surge tower) far off-screen is linking to a power generator that *is* on-screen, or vice versa? The whole reason power nodes are drawn twice is to fix potential issues of this sort.

core/src/io/anuke/mindustry/graphics/BlockRenderer.java
```diff
@@ -242,29 +242,39 @@ public void processBlocks(){
...
-            }else if(req.layer == block.layer){
-                block.drawLayer(req.tile);
-            }else if(req.layer == block.layer2){
-                block.drawLayer2(req.tile);
             }
+            drawBlock(request);
+        }
+    }
+
+    private void drawBlock(BlockRequest req) {
```
Anuken: I don't think there's a good reason to put this in a separate method. If it's *not* called multiple times or overly long, making new methods just creates more code and increases confusion.

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -281,9 +281,13 @@ public void drawLayer(Tile tile){
...
-                drawLaser(tile, link);
-            }
+            if (!linkValid(tile, link))
+                continue;
+
+            if (link.block() instanceof PowerNode && !(link.pos() < tile.pos()))
```
Anuken: unnecessary space, also see `continue` issue mentioned earlier

core/src/io/anuke/mindustry/ui/dialogs/SettingsMenuDialog.java
```diff
@@ -252,6 +252,7 @@ public void add(SettingsTable table){
         });
         graphics.sliderPref("fpscap", 240, 15, 245, 5, s -> (s > 240 ? Core.bundle.get("setting.fpscap.none") : Core.bundle.format("setting.fpscap.text", s)));
         graphics.sliderPref("chatopacity", 100, 0, 100, 5, s -> s + "%");
+        graphics.sliderPref("lasersopacity", 50, 0, 100, 5, s -> s + "%");
```
Anuken: If you're going to add a slider for laser opacity, then it would be a good idea to remove the setting for making power lasers visible, as it is no longer required. As an optimization, one could do `if(Core.settings.getInt("lasersopacity") == 0) return;` to prevent drawing anything when power lasers are hidden.

core/assets/bundles/bundle.properties
```diff
@@ -597,6 +597,7 @@ setting.crashreport.name = Send Anonymous Crash Reports
 setting.savecreate.name = Auto-Create Saves
 setting.publichost.name = Public Game Visibility
 setting.chatopacity.name = Chat Opacity
+setting.lasersopacity.name = Power Lasers Opacity
```
Anuken: I would call this "Power *Laser* Opacity".

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -332,8 +337,9 @@ protected void drawLaser(Tile tile, Tile target){
 
         float fract = 1f-tile.entity.power.graph.getSatisfaction();
 
-        Draw.color(Color.white, Pal.powerLight, fract*0.86f + Mathf.absin(3f, 0.1f));
-        Drawf.laser(laser, laserEnd, x1, y1, x2, y2, 0.3f);
+        Draw.color(Pal.powerLight, Pal.powerLightLowPower, fract*0.86f + Mathf.absin(3f, 0.1f));
```
Anuken: Why did you make the lasers even darker when at low power? It looks very muddy and ugly in your screenshot; I don't like it at all.

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
azureblue: Anything on a screen that has a power link will schedule blocks that it links to for rendering in BlockRenderer. So for every connection (a-b) on screen tile a and b will go to drawLayer(Tile tile) of PowerNode. It least that was I figured out and tested.

Btw. power connections seemed to be drawn more than twice duo rescheduling same tile in BlockRenderer.

core/src/io/anuke/mindustry/graphics/BlockRenderer.java
azureblue: Well, yeah I was going to make drawBlocks(Layer stopAt) cleaner. Will change that.

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
azureblue: Actually I agree...

core/src/io/anuke/mindustry/ui/dialogs/SettingsMenuDialog.java
azureblue: Yeah, why not. I was also considering caching the setting value between each laser render to prevent queries to settings map. But it's not easy if we want to see the change without restarting the game.

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
azureblue: Done. Though I think it makes code less readable.

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
azureblue: Reverted to original colors for now.

core/assets/bundles/bundle.properties
azureblue: Done.

core/src/io/anuke/mindustry/ui/dialogs/SettingsMenuDialog.java
Anuken: `graphics.checkPref("lasers", true);` is still in the settings, and `setting.lasers.name` needs to be removed from the main bundle as well.

azureblue: Forgot about that.

azureblue: Done

Anuken:
Seems good, thanks!

PR #779, opened 2019-10-05
Add download links to readme

1RedOne:
Lazy folks like me need links to do anything :p


Anuken:
Looks like you need an account to download the Oracle JDK now - not a fan of that. I would try and redirect people to a different JDK, like OpenJDK/Azul/AdoptOpenJDK.

1RedOne:
I will link them to the releases page from the Oracle project instead, one
minute.

1RedOne:
Updated, I linked them to the JDk instead.

Anuken:
That works.

PR #791, opened 2019-10-07
oh no bundle_oh_NO.properties; adding support for nonstandard locales

BasedUser:
oh no


Anuken:
...how does Java react to this locale?

AmateurPotion:
XD

core/src/io/anuke/mindustry/Vars.java
```diff
@@ -185,16 +185,16 @@ public static void init(){
...
+            for(int i = 0; i < locales.length-1; i++){
                 String code = stra[i];
                 if(code.contains("_")){
                     locales[i] = new Locale(code.split("_")[0], code.split("_")[1]);
                 }else{
                     locales[i] = new Locale(code);
                 }
             }
-
+            locales[locales.length-1] = new Locale("oh", "NO"); // ugly arc workaround
```
Anuken: 1. What does this have to do with Arc?
2. This is really hacky; I'd rather you change the locale resolution code to better support custom locales instead.
3. Does this even display properly?

BasedUser: 1. Arc loads locale codes, which I can't possibly add to (language code not defined, and even if i could I'd have to either append to Arc's data (scary) or hack it in
2. I'll try to rewrite that in next commits
3. It will display as a blank string, as this isn't an actual registered language and so doesn't have a display name (might need to implement in a format change for external bundles i.e. `languagename = displayNameHere`)

Anuken: 1. But that's not arc, that's just Mindustry code. Where do you see Arc doing things?

BasedUser: Mindustry requests Core, which is an Arc class, to load the locale codes from Java's database, just right before the for loop. Although I could add something to remedy this (codes not existing in any standard), probably coming with nonstandard bundles being loaded.

Anuken: Oh, that's what you're confused about. `Core.files.internal("locales")` is not loading a Java database, it's loading an autogenerated list of locales that the Mindustry build script creates in the assets folder. This list contains the names of all bundles in the "bundles" folder; even non-standard ones. You don't need to modify this.


The only thing that the Java database does is provide *display names* for those locales in the language select dialog (via `getDisplayName`, if I remember correctly).

Arkanic:
yes.

BasedUser:
continued in #879

PR #793, opened 2019-10-07
Adds Japanese menu items for Mods

1RedOne:
Adds Japanese menu content for Mods top level menu, and also mod menu items themselves


core/assets/bundles/bundle_ja.properties
```diff
@@ -21,6 +21,7 @@ load.map = マップ
 load.image = 画像
 load.content = コンテンツ
 load.system = システム
+load.mod = MOD
```
1RedOne: Some games would write this in Katakana, but English seemed to fit the menu more, and is commonly found in Japanese games.  This is why I chose English here but Katakana within sentences below.

PR #796, opened 2019-10-08
add Slag supply means

AmateurPotion:
In order to use slag more strategically, more means of dissemination are needed.


Anuken:
It's not *molten* rock. Pumps shouldn't be able to extract anything from them. If I add a slag resource, it'll be in the form of a real liquid block.

PR #801, opened 2019-10-09
inverse match sorter

azureblue:
I've added the ability for Sorter to split items to sides that do **not match** with the selected item. This feature was mentioned in the suggestions/ideas, and this is often desired to be able to get specific item from the conveyor.

It can be configured by selecting one of two buttons below the items table when the Sorter is clicked and whether it set up to inverse match it's also reflected as reddish (signalling that it blocks/prevent highlighted item to move forward) bar near the top.

I'm not sure if that interface is the best possible and I'm not very attached to the icons I've made :). The size of the icons is the same as in the Command Center. Anyway, It's just a draft.

Tell me what you think guys.

Peek 2019-10-09 19-222

And with icons centered:
Przechwycenie obrazu ekranu_2019-10-09_19-34-09


Anuken:
This should be a separate block that extends Sorter and splits off to the sides by default.

The reason for that is:
- Adding a new Boolean will break old saves unless steps are taken to make it backwards compatible.
- Block config must be stored in a single int for schematics and drone replacement to work. You don't seem to have synced this across the network..?

azureblue:
Yeah, I'm aware of the _configured(..)_ method callback and _int_ passing and saves :-). And yes, it's not yet synced. Of course adding a new block would also work, but I'm pretty sure that in the long run, the (more) configurable blocks are the way to go instead of multiplying them for only slightly different function (also that would be more cumbersome).

Perhaps some marshalling/unmarshalling based approach could be something to think about regarding transferring tile state over the network.

Anuken:
Configurable blocks are better in some circumstances, yes, but in this case, I would prefer a new block.

azureblue:
I think that the sorter instance is **the** perfect case for that. But I guess it's a matter of taste :wink:.

Anuken:
Added a separate inverted sorter block in 3d624f7eaf0a3fa6bb113f6c5a0e653300cc36fb.

PR #814, opened 2019-10-12
Add username labels to map/minimap

Maxattax97:

### DIFF
--- a/core/src/io/anuke/mindustry/graphics/MinimapRenderer.java
+++ b/core/src/io/anuke/mindustry/graphics/MinimapRenderer.java
@@ -84,8 +84,8 @@ public class MinimapRenderer implements Disposable{
         rect.set((dx - sz) * tilesize, (dy - sz) * tilesize, sz * 2 * tilesize, sz * 2 * tilesize);
 
         for(Unit unit : units){
-            float rx = (unit.x - rect.x) / rect.width * w, ry = (unit.y - rect.y) / rect.width * h;
-            Draw.color(unit.getTeam().color);
+            float rx = (unit.x - rect.x) / rect.width * w;
+            float ry = (unit.y - rect.y) / rect.width * h;
 
             if (withLabels && unit instanceof Player) {
                 Player pl = (Player) unit;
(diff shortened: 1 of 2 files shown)


Anuken:
How does this look on the small minimap, or in situations with more than a few players? I'm worried that in large servers, this would clog up the map and make things hard to see.

core/src/io/anuke/mindustry/graphics/MinimapRenderer.java
```diff
@@ -82,6 +86,15 @@ public void drawEntities(float x, float y, float w, float h){
         for(Unit unit : units){
             float rx = (unit.x - rect.x) / rect.width * w, ry = (unit.y - rect.y) / rect.width * h;
             Draw.color(unit.getTeam().color);
+
+            if (unit.getTypeID() == TypeIDs.player) {
```
Anuken: This is a strange way to compare types. Use `unit instanceof Player` instead.

Maxattax97:
I took out the local player's name from displaying originally because it didn't seem useful, but I kept the others' names on the minimap because I thought it might be okay. I've taken those out, but kept them on the large map dialog.

I was playing the other night with my friends and one would say "come here, look at this" but in order to figure out _where_ they were, geographical descriptions had to be given. I felt I'd add this feature to alleviate that.

Here are some new screenshots:

core/src/io/anuke/mindustry/graphics/MinimapRenderer.java
```diff
@@ -80,14 +84,28 @@ public void drawEntities(float x, float y, float w, float h){
         rect.set((dx - sz) * tilesize, (dy - sz) * tilesize, sz * 2 * tilesize, sz * 2 * tilesize);
 
         for(Unit unit : units){
-            float rx = (unit.x - rect.x) / rect.width * w, ry = (unit.y - rect.y) / rect.width * h;
+            float rx = (unit.x - rect.x) / rect.width * w;
+            float ry = (unit.y - rect.y) / rect.width * h;
+
+            if (withLabels && unit instanceof Player) {
```
Anuken: Use compact formatting (e.g. `if(){`). Happens a few times in this code.

Anuken:
Looks good, thanks.

PR #823, opened 2019-10-13
3 additional teams

ThatOneBepis:


Anuken:
But why? This will mess with a lot of UI layout and naming.

ThatOneBepis:
fair enough

Anuken:
In any case, I don't think the game is ready for this yet.

PR #824, opened 2019-10-13
Disallow dumping inside of enemy turf

Maxattax97:
See https://github.com/Anuken/Mindustry/issues/545#issuecomment-541382954


Anuken:
I'm worried that this may lead to more abuse. What if someone places a single block near the enemy base, thereby preventing them from building near the vicinity? Also, players may be confused about *why* they can't build in a location.

Maxattax97:
Your concern is warranted, we've been discussing that.

Briefly, the _why_ is a good point. A friend of mine suggested coloring floor tiles of the enemy team(s) when in build mode. I'll move forward with that if the rest of this is approved.

Personally, I feel building prevention is a non-issue. First of all, the radius can be greatly decreased to, say, 3 blocks. Second, if someone tried to wall another team inside of the Core protection zone, they can break them fairly easily by hand or with turrets. A bigger issue in my opinion is "claiming" plots of ore at the beginning of the game, but even then you can break it with relative ease.

After some discussion some possible solutions are:
 - Nerfing walls (more expensive and **longer build time**)
 - Redesigning turret AI to select targets in a more intelligent way
 - Buffing mechs or redesigning their AI to target buildings differently instead of only storming the core (this is could be a big change)

Ultimately you're the BDFL, and so far your design choices have led to a fantastic game, so I respect whatever decision you make and am happy to contribute code to help out.

P.S. One of us just made another PR for AI: https://github.com/Anuken/Mindustry/pull/827

Anuken:
I don't think this is needed anymore.

PR #825, opened 2019-10-13
Conservative generators

LionelBergen:
Makes it so generators do not consume resources When they aren't connect to anything, or the thing they are connected to is not running

Note: I left some TODO's. Will update these later


LionelBergen:
forgot to link issue: #825

azureblue:
What if they are inside a network which produce more power (as a whole) than is consumed?

LionelBergen:
The generates shouldn't consume more material than is needed for power (Unless there are batteries which need to be filled).
Just did some testing with 3 combusters & 1 consumer

core/src/io/anuke/mindustry/world/Block.java
```diff
@@ -554,6 +554,10 @@ public void handleBulletHit(TileEntity entity, Bullet bullet){
 
     public void update(Tile tile){
     }
+    
+    public void update(Tile tile, float amountOfPowerUsed){
```
Anuken: This shouldn't be a separate method. Amount of power used should be queried from the graph each update, not passed here.

core/src/io/anuke/mindustry/world/blocks/power/ItemLiquidGenerator.java
```diff
@@ -92,7 +92,12 @@ public boolean shouldConsume(Tile tile){
     }
 
     @Override
-    public void update(Tile tile){
+    public void update(Tile tile, float amountOfPowerUsed){
+    	// Should not update or do anything if we havn't consumed any power. This will leave the tower in a 'pause' state rather than consuming materials.
+    	if (amountOfPowerUsed <= 0f)
+    	{
```
Anuken: Don't use newline braces.

core/src/io/anuke/mindustry/world/blocks/power/ItemLiquidGenerator.java
```diff
@@ -92,7 +92,12 @@ public boolean shouldConsume(Tile tile){
     }
 
     @Override
-    public void update(Tile tile){
+    public void update(Tile tile, float amountOfPowerUsed){
+    	// Should not update or do anything if we havn't consumed any power. This will leave the tower in a 'pause' state rather than consuming materials.
+    	if (amountOfPowerUsed <= 0f)
```
Anuken: Use compact if statements. `if(){`

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -168,20 +169,24 @@ public void distributePower(float needed, float produced){
...
+                
+                amountOfPowerUsed += consumer.entity.power.satisfaction;
             }
         }
+        
+        return amountOfPowerUsed;
     }
 
-    public void update(){
+    public float update(){
```
Anuken: Why does this return a value? It doesn't seem appropriate that updating the graph returns an arbitrary float. The power consumed should be a field that is updated each `update()` instead and queried with a getter.

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -205,8 +208,17 @@ public void update(){
                 powerProduced -= chargeBatteries(powerProduced - powerNeeded);
             }
         }
-
-        distributePower(powerNeeded, powerProduced);
+        if (Mathf.isZero(powerNeeded))
```
Anuken: formatting

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -205,8 +208,17 @@ public void update(){
...
+        {
+        	return 0.0f;
+        }
+        else
+        {
+        	float powerUsed = distributePower(powerNeeded, powerProduced);
```
Anuken: formatting

tests/src/test/java/power/FakeGraphics.java
```diff
@@ -57,6 +57,11 @@ public int getBackBufferHeight(){
     public long getFrameId(){
         return frame++;
     }
+    
+    public void setFrameId(int frameId)
+    {
```
Anuken: This doesn't need to be a setter, just make it public. Formatting.

core/src/io/anuke/mindustry/entities/type/TileEntity.java
```diff
@@ -307,14 +307,17 @@ public void update(){
...
+        if(cons != null){
             cons.update();
         }
 
-        if(block == previous && power != null){
-            power.graph.update();
+        float amountOfPowerUsed = 0.0f;
+        
+        if(power != null){
+            amountOfPowerUsed = power.graph.update();
```
Anuken: This shouldn't be needed if power is queried in the block - don't change `TileEntity`.

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -205,8 +208,17 @@ public void update(){
                 powerProduced -= chargeBatteries(powerProduced - powerNeeded);
             }
         }
-
-        distributePower(powerNeeded, powerProduced);
+        if (Mathf.isZero(powerNeeded))
+        {
+        	return 0.0f;
```
Anuken: Why is this if-statement necessary? If no power is needed, the power balance should still be updated regardless.

core/src/io/anuke/mindustry/entities/type/TileEntity.java
```diff
@@ -307,14 +307,17 @@ public void update(){
...
         }
+        
+    	block.update(tile, amountOfPowerUsed);
```
Anuken: Hold on. If you have, say, 10 generators giving just enough power to supply a block with power, then one of the generators will have positive power used, and the other 9 will have 0, as the graph just updated, making them stop updating next frame. This will lead to all sorts of issues.

core/src/io/anuke/mindustry/world/blocks/power/ItemLiquidGenerator.java
```diff
@@ -92,7 +92,12 @@ public boolean shouldConsume(Tile tile){
...
+    	{
+    		return;
```
Anuken: What if a generator is sending out 50 power, but the grid requests 25? It will continue to burn fuel at the same rate as before, e.g. at 100% combustion speed. This will work only in situations where there is **no** power being requested from the grid, and those situations are very rare and don't present much of a problem in real games. The real problem arises when there's a large surplus of power that isn't going anywhere, but *part of it is still being consumed*.

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -168,20 +169,24 @@ public void distributePower(float needed, float produced){
                         }
                     }
                 }
+                
+                amountOfPowerUsed += consumer.entity.power.satisfaction;
             }
         }
+        
+        return amountOfPowerUsed;
```
Anuken: Why does this method return the sum of all power satisfactions?

Anuken:
Closing due to inactivity/a different approach being required.

PR #827, opened 2019-10-13
"Smarter" targeting for Turrets

Killian-LeClainche:
Some of my friends and I had been playing the PVP multiplayer and found that you can confuse the turrets by spamming copper walls on top of the opponents turrets and then turret push.

I felt like this reduces the competition of the PVP. Let me know your thoughts, thanks!

### DIFF
--- a/core/src/io/anuke/mindustry/world/blocks/defense/turrets/Turret.java
+++ b/core/src/io/anuke/mindustry/world/blocks/defense/turrets/Turret.java
@@ -19,6 +19,7 @@ import io.anuke.mindustry.entities.type.Bullet;
 import io.anuke.mindustry.entities.bullet.BulletType;
 import io.anuke.mindustry.entities.traits.TargetTrait;
 import io.anuke.mindustry.entities.type.TileEntity;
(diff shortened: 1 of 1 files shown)


indielm:
I think this is fine because it only works outside the base radius circle, which is adjustable with the map.

Anuken:
Maybe this functionality should be implemented for *all* units and blocks by putting the extra turret targeting step in `Units`? I feel like targeting walls when there's turrets available is a dumb idea in general.

Killian-LeClainche:
I like that, but ideally you'd weight the units/blocks around you and target accordingly. To explain further, air units would want to target anti-air turrets over ground turrets, etc. Just some food for thought for the future.

Also, some ramifications of doing this is that bombers, which (in my playthrough) ignore anti-air defense, would be less likely to behave the same. That means they won't attack the production lines if your air defense doesn't cut them down quickly.

Anuken:
Implemented configurable targeting priority in edfd402ccd1a9bc5d1c083977f90c0f9aa2407d2, so this should no longer be needed.

PR #837, opened 2019-10-18
Change selected block when changing category

clfer:
Hello

With this, when in building mode, the selected block will change when changing category.
Last selected block is stored by category

**Before:** 
You must remember to select the conveyor after changing category
mindustry-category-block__demo-old

**After:**
Block change automatically when you change category
mindustry-category-block__demo-new

it remembers previously selected block
category select demo

Have a good day

### DIFF
--- a/core/src/io/anuke/mindustry/type/Category.java
+++ b/core/src/io/anuke/mindustry/type/Category.java
@@ -1,7 +1,5 @@
 package io.anuke.mindustry.type;
 
-import io.anuke.mindustry.world.Block;
-
 public enum Category{
     /** Offensive turrets. */
     turret,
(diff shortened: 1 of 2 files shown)


core/src/io/anuke/mindustry/type/Category.java
```diff
@@ -23,4 +25,14 @@
     effect;
 
     public static final Category[] all = values();
+
+    Block selectedBlock = null;
```
Anuken: 1. This should not be stored in the category, as it will lead to very interesting bugs if a game reloads on Android or after a mod. Use an `ObjectMap` in input instead that gets cleared on every `ResetEvent`.
2. Don't use getters and setters.

clfer: 1. Done in https://github.com/Anuken/Mindustry/pull/837/commits/6ea949290e95005e50b652abc2ff7d23d25c7bca
2. For my personnal information, why shouldn't getters/setters be used in this case? I didn't do Java since university and back then we were taught that getters and setters are a must 😅

Anuken: Getters and setters bloat the code and aren't really necessary unless you're making a library. If you need to encapsulate the variable at a later point in time, IntelliJ can do that for you in a few clicks.

Anuken:
Seems to work well. Thanks.

PR #838, opened 2019-10-19
Added more rubble pngs

ado1928:
When you destroy anything larger than 4x4, the game crashes because there are no rubbles for that size, so I added them.


Anuken:
Note that it doesn't crash, there's just no rubble displayed. This can be tested by placing down a Nucleus core and seeing what happens when it gets destroyed.

PR #840, opened 2019-10-19
🚧 Add insulator wall

Quezler:
**problem**

(especially when) playing on servers you might want to keep certain power grids(/graphs) separate from one another for a multitude of reasons:

- a mass driver chain with its own power supply
- ditto for phase conveyors
- keeping the production & defense grids separate
- have self sufficient thorium & titanium mines for a power plant that powers the rest
- preventing vital systems going down by accidentally connecting them to other grids
- etc

on single player its still "somewhat easy" to oversee, but on multiplayer not everyone shares your grey matter so they wouldn't be fully aware of grid intentions unless e.g. message boards are placed everywhere.

**brainstorms**

on my road to a solution i stumbled upon several ideas:

- a "sorter" block that could be connected to a grid that colorized their wires for a visual difference, but then what if one or more graphs get forcibly merged via e.g. a solar panel placed between them? would the colors somehow merge? how would the gui look to set colors? etc? 

- a setting on the powerpoles somehow that prevented all automatic connections, but that shares the same issues as the above, moreover: how would players on a server find out where the power sorting blocks and/or configurable powerpoles are located? would be confusing if for some visually unknown reason a wire would not (want to) connect.

- a wall block that blocks "all" wires from going over, but then it would be weird to have awkwardly shaped gaps in the wall just for wires to pass through, and you'd also have to have logic in place that would check for all wires above a wall once placed, overall a technical nightmare.

- so eventually i've settled for a wall that simply blocks automatic connections from going over them, players can still connect them manually, this allows for a subsystem to be completely walled without gaps, and power is also easy to extract, and as a bonus it communicates to other players that the separate grid is intentional, instead of others mistaking it for a bug and "fixing" it by proudly connecting it to the grid and calling it a day.

**considerations**

- a 1x1 insulator block would be more than enough, the 2x2 insulator would probably not be used (a lot), but walls come in even numbers so i've added it for that reason.
- texture is the titanium wall saturated down to greyscale. (hence why its declaration in the files is right below titanium)
- i've tried to mimic the recent commits where new blocks got added, not sure if this is the exact method you would have used, but thats why its called a prototype :)
- as of writing this i noticed that it only applies to placed power poles, placing other things would still cause the powerpoles from "inside" the wall the connect over them to said machinery, will probably send a commit to solve that soon.
- for the costs i went with silicon and glass, since those are common things to use in (high voltage) insulators, this commit also includes a tech tree entry where i've placed them behind the kiln, since it felt more fitting there than in the wall section, but feel free to change that.

**visuals**

Screen Shot 2019-10-19 at 13 09 33
(static)

Untitled
(phase conveyors with their own power supply)

Untitled2
(just imagine the conveyor going to a very important power plant for very important stuff, you wouldn't want to accidentally drain the self sufficient grid for the miner)

Untitled3
(connects as normal if there is a gap in the wall)

Untitled4
(ugly setup illustrating that manual connections still work)

🍠


Quezler:
Screen Shot 2019-10-19 at 13 09 33

Quezler:
Untitled

Quezler:
Untitled2

(just imagine the conveyor going to a very important power plant for very important stuff, you wouldn't want to accidentally drain the self sufficient grid for the miner)

Quezler:
Untitled3

Quezler:
Untitled4

(pardon the horrible setup, but this is just to illustrate that you can manually connect wires)

Quezler:
thanks for reading, and as (not) promised: 🥔

Luxray5474:
+1

Very nice demonstration!

Ansraer:
Could you maybe rework the power node placement ui, so that it tells the player if a block (that it would usually connect to) is completely blocked by insulators and maybe even highlight them?

Quezler:
@Ansraer what about this? 📋 
Oct-25-2019 20-43-53

Anuken:
> stopping lightning before the wall looked to weird

I think this is the best option. There doesn't seem to be anything wrong with it besides the fact that it stops too early, yet continues to damage blocks in front. I'll look into implementing something like this.

> what about this?

That seems very subtle. I had to watch it twice to figure out what the change was. How about drawing that selection square over the block that's providing the insulation?

Quezler:
@Anuken lightning is stopped in front of the wall again, and the highlighter is moved to what is insulating, but pardon its messy/ugly code 😅 
Oct-25-2019 21-50-39

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -299,4 +317,27 @@ protected void drawLaser(Tile tile, Tile target){
         Draw.color();
     }
 
+    public static boolean insulated(Tile tile, Tile other){
+        return insulators(tile, other).size > 0;
+    }
+
+    public static Array<Tile> insulators(Tile tile, Tile other) {
+        Array<Tile> tiles = new Array<Tile>();
```
Anuken: Don't create any new objects in a method like this, that may be called many times per frame. If you absolutely must use an array, use a static temporary one and clear it every time.

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -299,4 +317,27 @@ protected void drawLaser(Tile tile, Tile target){
...
+        Array<Point2>points = PlaceUtils.normalizeDiagonal(tile.x, tile.y, other.x, other.y);
```
Anuken: This is a very... strange place to use this method. It gives unclean diagonals, which can be seen in your GIF.

How I would do this method:
```java
public static Array<Tile> insulators(int x, int y, int x2, int y2, Consumer<Tile> iterator){
    world.raycastEach(x, y, x2, y2, (wx, wy) -> {
      //do stuff to get the tile
     iterator.accept(world.ltile(wx, wy)); // if it fits the criteria
    });
}
```

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -299,4 +317,27 @@ protected void drawLaser(Tile tile, Tile target){
...
+
+        for(int i = 0; i < points.size; i++){
+            Point2 point = points.get(i);
+            Block block = world.tile(point.x, point.y).block();
```
Anuken: `world.tile(point.x, point.y)` can be null. This could crash.

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -299,4 +317,27 @@ protected void drawLaser(Tile tile, Tile target){
...
+
+            if (block instanceof BlockPart){
```
Anuken: Don't check for blockparts manually. Use `world.ltile(point.x, point.y)` to return the linked tile or null.

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -221,7 +226,20 @@ public void drawPlace(int x, int y, int rotation, boolean valid){
         Draw.color(Pal.placing);
         Drawf.circles(x * tilesize + offset(), y * tilesize + offset(), laserRange * tilesize);
 
-        getPotentialLinks(tile, other -> Drawf.square(other.drawx(), other.drawy(), other.block().size * tilesize / 2f + 2f, Pal.place));
+        getPotentialLinks(tile, other -> {
+            Drawf.square(other.drawx(), other.drawy(), other.block().size * tilesize / 2f + 2f, Pal.place);
+
+            insulators(tile, other).each(cause -> {
+
+                Block block = world.tileWorld(cause.drawx(), cause.drawy()).block();
```
Anuken: This can be null (so don't call `block()` without a check first), and use `ltileWorld` instead of checking whether it's a blockpart manually.

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -221,7 +226,20 @@ public void drawPlace(int x, int y, int rotation, boolean valid){
         Draw.color(Pal.placing);
         Drawf.circles(x * tilesize + offset(), y * tilesize + offset(), laserRange * tilesize);
 
-        getPotentialLinks(tile, other -> Drawf.square(other.drawx(), other.drawy(), other.block().size * tilesize / 2f + 2f, Pal.place));
+        getPotentialLinks(tile, other -> {
+            Drawf.square(other.drawx(), other.drawy(), other.block().size * tilesize / 2f + 2f, Pal.place);
+
+            insulators(tile, other).each(cause -> {
```
Anuken: Can be replaced with a direct consumer as mentioned previously.

core/src/io/anuke/mindustry/entities/effect/Lightning.java
```diff
@@ -131,4 +140,19 @@ public void draw(){
     public EntityGroup targetGroup(){
         return bulletGroup;
     }
+
+    protected static boolean hitsInsulator(Lightning l){
+        if (l.lines.size > 1){
+            Position from = l.lines.get(l.lines.size - 2);
+            Vector2 to = (Vector2) l.lines.get(l.lines.size - 1);
+
+            Tile a = world.tileWorld(from.getX(), from.getY());
```
Anuken: This can be null. Don't get the tile, just get its coordinates in world space.

core/src/io/anuke/mindustry/entities/effect/Lightning.java
```diff
@@ -131,4 +140,19 @@ public void draw(){
...
+            Tile b = world.tileWorld(to.getX(), to.getY());
```
Anuken: Same here.

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
Quezler: Eventually i had to go with an atomic boolean to get the result from the lambda 🙊

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
Quezler: I had no clue on how to get `Array<Tile>` out of your example 😅

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
Quezler: Should be gone now 😇

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
Quezler: Changed as per your code example above 😎

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
Quezler: Not entirely sure what you mean by this, but i may have accidentally solved it 😗

core/src/io/anuke/mindustry/entities/effect/Lightning.java
Quezler: Forwarded its coordinates 🔮

core/src/io/anuke/mindustry/entities/effect/Lightning.java
Quezler: Replaced it 👍

core/src/io/anuke/mindustry/entities/effect/Lightning.java
```diff
@@ -131,4 +138,16 @@ public void draw(){
     public EntityGroup targetGroup(){
         return bulletGroup;
     }
+
+    protected static boolean hitsInsulator(Lightning l){
+        if (l.lines.size > 1){
+            Vector2 from = (Vector2) l.lines.get(l.lines.size - 2);
```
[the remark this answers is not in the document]
Quezler: Because else the ide assumes its position object which didn't have a method/property that vector did have that i needed to use, i don't recall it anymore though, will remove the cast 🤔

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -299,4 +308,27 @@ protected void drawLaser(Tile tile, Tile target){
         Draw.color();
     }
 
+    public static boolean insulated(Tile tile, Tile other){
+        return insulated(tile.x, tile.y, other.x, other.y);
+    }
+
+    public static boolean insulated(int x, int y, int x2, int y2){
+        AtomicBoolean ref = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
Quezler: Thx, wasn't aware 😅

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -299,4 +307,27 @@ protected void drawLaser(Tile tile, Tile target){
...
+        });
+        return bool[0];
+    }
+
+    public static void insulators(int x, int y, int x2, int y2, Consumer<Tile> iterator){
+        world.raycastEach(x, y, x2, y2, (wx, wy) -> {
+
+            Tile tile = world.ltile(wx, wy);
+            if(tile != null && tile.block() != null && tile.block().insulated){
+                iterator.accept(tile);
```
Anuken: Once you hit an insulator, it may be a good idea to `return true` so the raycasting stops. `tile.block()` can't be null, so the condition here is redundant.

core/src/io/anuke/mindustry/entities/effect/Lightning.java
```diff
@@ -65,6 +66,11 @@ public static void createLighting(int seed, Team team, Color color, float damage
             Bullet.create(Bullets.damageLightning, l, team, x, y, 0f, 1f, 1f, dmg);
             l.lines.add(new Vector2(x + Mathf.range(3f), y + Mathf.range(3f)));
 
+            if(hitsInsulator(l)) {
+                l.lines.remove(l.lines.size -1);
```
Anuken: This seems to work, but:

image

The lightning never really hits the wall, it just stops early. A workaround may involve the following:

```java
//outside the for-loop:
boolean[] hit = {false};

//inside loop
if(l.lines.size > 1){
    hit[0] = false;
    Position from = l.lines.get(l.lines.size - 2);
    Position to   = l.lines.get(l.lines.size - 1);
    world.raycastEach(x, y, x2, y2, (wx, wy) -> {

        Tile tile = world.ltile(wx, wy);
            if(tile != null && tile.block().insulated){
                hit[0] = true;
                //snap it instead of removing
                l.lines.get(l.lines.size -1).set(wx * tilesize, y * tilesize);
                return true;
            }
            return false;
    };
    if(hit[0]) break;
}
```

This would effectively snap the last lightning line to the block, making it look like it connected to something instead of stopping just short of the wall.

Also, this is more of a problem with my code, but
`Array<Position> lines` should probably be changed to `Array<Vector2> lines`. The `Position` type had some other purpose that isn't needed here anymore.

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
Quezler: But then you cannot highlight all the walls that cause the insulation 😗

core/src/io/anuke/mindustry/entities/effect/Lightning.java
Quezler: > The lightning never really hits the wall, it just stops early. A workaround may involve the following:

+ you knew it stopped early right? 😅 

> I think this is the best option. There doesn't seem to be anything wrong with it besides the fact that it stops too early, yet continues to damage blocks in front. I'll look into implementing something like this.

Anuken: Well, yeah, I'm just saying it could look better with this method.

Quezler: Allrighty, i'll try to implement it later tonight 👍

Quezler: Had to rename hit (since it was taken), and replaced the xy's with variables i could find 😗

core/assets/bundles/bundle.properties
```diff
@@ -1076,6 +1078,8 @@ block.copper-wall.description = A cheap defensive block.\nUseful for protecting
 block.copper-wall-large.description = A cheap defensive block.\nUseful for protecting the core and turrets in the first few waves.\nSpans multiple tiles.
 block.titanium-wall.description = A moderately strong defensive block.\nProvides moderate protection from enemies.
 block.titanium-wall-large.description = A moderately strong defensive block.\nProvides moderate protection from enemies.\nSpans multiple tiles.
+block.plastanium-wall.description = Prevents automatic wire connections.\nmanual connections remain unimpeded.
```
Anuken: One last thing: how about
`A special type of wall that absorbs electric arcs.\nBlocks automatic power node connections. Manual connections remain unimpeded.`

...and likewise for the large wall.

Quezler: Looks a bit out of place though, are you sure? 😗 
Screen Shot 2019-10-27 at 21 27 33
Screen Shot 2019-10-27 at 21 27 40

Anuken: How about just `A special type of wall that absorbs electric arcs and blocks automatic power node connections`

Quezler: I tried to edit it in just now but github mobile doesn't want to cooperate, I'll try it tomorrow from my desktop, but feel free to merge & edit, or even edit the pull itself, i believe i had the checkbox on that allowed edits from maintainers 🤔

Quezler:
Anything else that needs to be done before this gets merged? 😇

Anuken:
No, I think that's it. Thanks for the contribution.

Quezler:
And thank you for all the code suggestions, i ended up learning a bunch of things :) 🍪

PR #848, opened 2019-10-21
Synchronize powergraph

Kieaer:
## The reason for doing this
When getting the value of PowerGraph, a #iterator error occurred if it did not match the tick on the server.

## How to reproduce the error
https://github.com/Kieaer/Essentials/blob/master/src/main/java/essentials/thread/Powerstat.java#L37

## Solution
Set synchronization to PowerGraph values

## Before
```
io.anuke.arc.util.ArcRuntimeException: #iterator() cannot be used nested.
        at io.anuke.arc.collection.ObjectSet$ObjectSetIterator.hasNext(ObjectSet.java:570)
        at io.anuke.mindustry.world.blocks.power.PowerGraph.distributePower(PowerGraph.java:149)
        at io.anuke.mindustry.world.blocks.power.PowerGraph.update(PowerGraph.java:209)
        at io.anuke.mindustry.world.blocks.power.PowerNode.update(PowerNode.java:158)
        at io.anuke.mindustry.entities.type.TileEntity.update(TileEntity.java:304)
        at io.anuke.mindustry.entities.EntityGroup.update(EntityGroup.java:49)
        at io.anuke.mindustry.core.Logic.update(Logic.java:204)
        at io.anuke.arc.backends.headless.HeadlessApplication.mainLoop(HeadlessApplication.java:99)
        at io.anuke.arc.backends.headless.HeadlessApplication$1.run(HeadlessApplication.java:64)
```

## After
ezgif com-optimize(2)
~~Please ignore block destroy~~

### DIFF
--- a/core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
+++ b/core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
@@ -57,21 +57,25 @@ public class PowerGraph{
 
     public float getPowerProduced(){
         float powerProduced = 0f;
-        for(Tile producer : producers){
-            if(producer.entity == null) continue;
-            powerProduced += producer.block().getPowerProduction(producer) * producer.entity.delta();
+        synchronized (producers){
+            for(Tile producer : producers){
+                if(producer.entity == null) continue;
+                powerProduced += producer.block().getPowerProduction(producer) * producer.entity.delta();
+            }
         }
         return powerProduced;
     }
 
     public float getPowerNeeded(){
         float powerNeeded = 0f;
-        for(Tile consumer : consumers){
-            Consumers consumes = consumer.block().consumes;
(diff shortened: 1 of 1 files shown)


Anuken:
Synchronization is not a cheap operation at all. This may noticably slow down the code and add unnecessary bloat just for one specific plugin. 

**Even if you synchronize the iteration**, you may still get crashes related to modifications of the graph during iteration. Why are you querying it in a new thread?

Kieaer:
In mod, can do ``update = true``, then ``public void update ()``, but the plugin can't follow game ticks.
The alternative is to use TimerTask or thread, runnable. (main class runs only once)

Another way is... this

(private to public)
This allows you to sync from a plug-in, not from a game source.

Kieaer:
I didn't know it :eyes:
Solved.

PR #849, opened 2019-10-21
⚔️ Fix stuttering conveyors between teams

Quezler:
**backstory**

A few days ago on an assault map i made an attempt to "steal" resources from an item source with a conveyor of my own, when that didn't work i repurposed that line for ammo, and not soon after noticed this bug:

**before**
Oct-21-2019 18-44-54

**after**
Oct-21-2019 18-45-00

**〃**
Oct-21-2019 19-39-02


Tazshelby2016:
You should allow enemy and friendly conveyers to mix, or simply cut any connections between them.

Anuken:
This actually occurred in a lot of different situations; I've applied a different fix.

Quezler:
If its not to much trouble, could you reference a commit hash? i'd like to see/learn how you solved it 🤔

PR #855, opened 2019-10-22
Server to server

Kieaer:
server to server
Usage
```
String serverip = "localhost";
String port = 6568;
Call.onConnect(player.con, serverip, port);
```

### DIFF
--- a/core/src/io/anuke/mindustry/core/NetClient.java
+++ b/core/src/io/anuke/mindustry/core/NetClient.java
@@ -206,7 +206,7 @@ public class NetClient implements ApplicationListener{
         logic.reset();
 
         Vars.netClient.beginConnecting();
-        net.connect(ip, port, Actions::hide);
+        net.connect(ip, port, Actions::new);
     }
     
     @Remote(targets = Loc.client)
--- a/core/src/io/anuke/mindustry/core/NetClient.java
+++ b/core/src/io/anuke/mindustry/core/NetClient.java
@@ -9,7 +9,6 @@ import io.anuke.arc.util.CommandHandler.*;
 import io.anuke.arc.util.*;
 import io.anuke.arc.util.io.*;
 import io.anuke.arc.util.serialization.*;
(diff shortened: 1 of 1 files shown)


core/src/io/anuke/mindustry/core/NetClient.java
```diff
@@ -198,6 +199,16 @@ public static String colorizeName(int id, String name){
...
     }
 
+    @Remote(called = Loc.client, variants = Variant.one)
+    public static void onConnect(String ip, int port){
+        netClient.disconnectQuietly();
+        state.set(State.menu);
+        logic.reset();
+
+        Vars.netClient.beginConnecting();
+        net.connect(ip, port, Actions::hide);
```
Anuken: What is `Actions::hide` for? There's nothing to hide here.

Kieaer: Can use ``net.connect(ip, port, null);``?

Anuken: No, but there's no reason to use an empty UI action.

core/src/io/anuke/mindustry/core/NetClient.java
```diff
@@ -198,6 +199,16 @@ public static String colorizeName(int id, String name){
...
     }
 
+    @Remote(called = Loc.client, variants = Variant.one)
+    public static void onConnect(String ip, int port){
+        netClient.disconnectQuietly();
+        state.set(State.menu);
+        logic.reset();
+
+        Vars.netClient.beginConnecting();
+        net.connect(ip, port, Actions::new);
```
Anuken: That's pretty much the same thing. `Actions` is a class for animated UI elements, it has nothing to do with this connections callback. Don't use it.

What you want is an empty callback, which would be `() -> {}`.

Kieaer:
I learned how to call an empty callback by Anuke!
~~knowledge 5 increase~~

PR #892, opened 2019-10-26
🥌 Do not count color codes toward maximum name length

Quezler:
The player limit is there for a reason (of course), but i was wondering which. 🤔 

- **purely visual length** then it would do no harm to subtract the length of formatting codes.
- **packet length something** yea then i can understand why color codes get counted along.

This pull is intended for those with names longer than 4 chars and wish to individually color them.

I kept this pull in draft mode since the regex that determines how many are actually colors is a bit crude and easy to exploit by just wrapping ones entire name within `[]`, let hear & share some thoughts. 😇

### DIFF
--- a/core/src/io/anuke/mindustry/core/NetServer.java
+++ b/core/src/io/anuke/mindustry/core/NetServer.java
@@ -689,7 +689,7 @@ public class NetServer implements ApplicationListener{
 
         StringBuilder result = new StringBuilder();
         int curChar = 0;
-        while(curChar < name.length() && result.toString().getBytes().length < maxNameLength){
+        while(curChar < name.length() && result.toString().getBytes().length < (maxNameLength + name.length() - name.replaceAll("(\\[.*?\\])", "").length())){
             result.append(name.charAt(curChar++));
         }
         return result.toString();


Anuken:
Even if this is changed, the text fields for inputting names on each platform would need to have some sort of way to prevent color codes from counting towards the limit. That won't be an easy task.

Quezler:
Hmm, yea, makes sence, i always kinda just hardcoded them (since copy paste in the name field does not work + its small + config is not in cleartext) 🙈 
Screen Shot 2019-10-27 at 16 57 11

Peter-Crawley:
Wouldn't it be possible to spam these char codes and potentially crash a low end server?

Quezler:
Colored player names should not be able to crash a server 🤔

Arkanic:
byte overload will be able to crash a server

Anuken:
You really can't crash a server just by sending it 65535 bytes.

Arkanic:
oh
i forgot about type sizes
:/

Anuken:
I don't think this approach is viable; an approach that covers all the platforms and edge cases would be very time-intensive and possibly exploitable. Shelving for now.

PR #900, opened 2019-10-26
Starting bundle_fi.properties

valtsu41:
This is far from ready, but because the original bundle I copied is so old now, I would like to get this file updated to avoid translating old stuff and get the new stuff to the bundle.


valtsu41:
So can this be merged..? Or does this need to be complete?

Anuken:
I passed over it previously, but if you want it in now I guess I can merge.

PR #904, opened 2019-10-27
Update tools:pack path on windows

thetrevdev:
Fix path for TexturePacker on windows.
Everywhere else in the build.gradle uses the same relative path.

### DIFF
--- a/tools/build.gradle
+++ b/tools/build.gradle
@@ -344,7 +344,7 @@ task pack(dependsOn: classes){
         }
 
         //pack normal sprites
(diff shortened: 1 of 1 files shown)


Anuken:
This breaks packing on Linux, and the current path works fine for me on Windows. I can also confirm that other people have been able to pack on Windows with the current script as well.

thetrevdev:
My mistake sorry about that.  I ran gradlew tools:pack instead of gradlew pack.
I have never used gradle before.  I will update the documentation

PR #905, opened 2019-10-27
Update bundle.properties

Ali-C-Ila:
removed description stating large power nodes have more connections.


Quezler:
Though its [technically true](https://github.com/Anuken/Mindustry/blob/master/core/src/io/anuke/mindustry/content/Blocks.java#L1057), it (indeed) pales in comparison to its range, and most players would not even notice since the small node rarely has so many blocks in range that would otherwise not be connected via shared graphs 🤔

PR #915, opened 2019-10-27
🗺 Locate existing mechpads

Quezler:
When playing on a server its pretty common for multiple mechpads to be scattered all over the place since players aren't aware where others might have placed theirs, if any at all 😵  

This change causes all mechpads on the map to be highlighted by (currently, subject to change*) its engine color as long as the build tab has that category selected, which should make it easier for other players (and yourself) to be aware of where they are 📌 
Oct-27-2019 21-20-48
(note that the low fps of the gif is due to filesize limitions, and not the constant checking of the entire grid 😗)


TheChosenEvilOne:
Checking the entire map like that is a horrible idea, it should just have a static list for mech pads where they get added when created and removed when destroyed.

Quezler:
Any suggestion(s) as to where in the code to put said list? 😗

Anuken:
Create a new `BlockFlag` (call it, say, `MechPad`), set that flag in the mech pad's constructor and call `indexer.getAllied(team, BlockFlag.mechPad)` to get a list of all pads in the map.

Quezler:
Thx, I'll try it tomorrow 😇

Anuken:
May need some graphical indicator changes later, but for now I don't have any issues with it. Thanks.

Quezler:
True, one of the colors looks to much like the allied player one, and another barely looks blue, but at least they are squares 🎛️

PR #925, opened 2019-10-28
Plastanium Wall

AmateurPotion:
This is Plastanium wall. Why do it uses metaglass?


Quezler:
Well it was intended to be called an insulator wall that used silicon & metaglass, but anuke suggested that it used plastanium instead of silicon in its recipe, so it might as well be called that since else the color might have been confusing 🤔 see #840 for more info 😇

Anuken:
Because metaglass in underused.

PR #936, opened 2019-10-29
🃏 Prevent inverted sorter chaining

Quezler:
A little to often on servers nowadays the inverted sorters get abused to make what could only be described as a `core 🐍`, of course there will be people hating me for making this pull thus taking away their toy, but i believe its for the best 🤔 

**abomination**
Screen Shot 2019-10-29 at 14 25 02
**before**
Oct-29-2019 14-18-30
**after**
Oct-29-2019 14-18-35
(code is not pretty, but it works, no idea if it has unintended side-effects, the above examples should be most if not all use cases, and those appear fine, hopefully this can assist in finding a solution 😗)

### DIFF
--- a/core/src/io/anuke/mindustry/world/blocks/distribution/Sorter.java
+++ b/core/src/io/anuke/mindustry/world/blocks/distribution/Sorter.java
@@ -82,6 +82,10 @@ public class Sorter extends Block{
         return other != null && other.block() instanceof Sorter && other.<SorterEntity>entity().sortItem == tile.<SorterEntity>entity().sortItem;
     }
 
+    boolean isInvertedSorter(Tile other){
+        return other != null && other.block() instanceof Sorter && ((Sorter) other.block()).invert;
+    }
+
     Tile getTileTarget(Item item, Tile dest, Tile source, boolean flip){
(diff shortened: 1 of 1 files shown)


Anuken:
Screenshot_2019-10-26-20-02-32

This may help, but there's still a lot of ways to abuse the system.

Quezler:
Oh dear lord 😱

matheod:
Does this fix prevent to use them as corner (which should be allowed in my opinion) ?

```
↑ MM
↑XMM
B↑
```

Where :
B : solid block
↑ : conveyor belt with coal
X : inverted sorter with lead as filter (or anything else than coal/copper)
M : mining machine with copper

Copper is not lead, so it is redirected to side but as no side is avaible it doesn't do anything
coal is not lead so it is redirected to side

Quezler:
@matheod 
Screen Shot 2019-11-05 at 20 15 13

iczero:
From what I understand, this simply prevents items from being transferred through 3 inverted sorters in a row. I would like to point out that this does prevent legitimate use cases, such as this:
image

Quezler:
Problem solved 🍪

PR #939, opened 2019-10-29
dispose event

J-VdS:
This event should be fired when the server closes.


Anuken:
Why? That event is only there for graphics.

J-VdS:
But how could you detect if the server is going to close? I thought you should launch an event for that.

PR #941, opened 2019-10-29
Fix #940

### LINKED ISSUE #940: Armored Conveyers visual glitch
**Platform**: *Windows*

**Build**: *99.2*

**Issue**: There is a problem with armored conveyers, they don't connect probably. I guess it's related to the last visual glitch I reported here @Quezler

image

**Steps to reproduce**: *Just rebuild the pyratit mixer als shown in the picture, it will output to the conveyer (which is correct) but the conveyer isn't connected to the building visually.*

Quezler:


Anuken:
Fixed it properly in the latest commit. There should be no conveyor class check.

Quezler:
Well it was a horrible fix anyways, still got no clue why it acted that was before i did my "fix", it just somehow worked 😅

PR #946, opened 2019-10-30
Update README.md

Shawak:


Anuken:
I'd rather put it at the bottom, considering how long it is now.

PR #949, opened 2019-10-30
🗜 Pressure conduit

Quezler:
Screen Shot 2019-10-30 at 13 09 27
When transporting liquids over long distances the `pressure/fullness/quantity/etc` tends to suffer, this is my attempt to add a conduit that can output at a slightly higher  `〃` than its input side.

While it technically may not change a lot (a high to low flow won't ruin or blow up stuff) it would be visually pleasing to be able to get a large quantity of liquids to a location and pressurize it somehow (that does not involve filling the entire path to 100% capacity) to get that sweet sweet fluid glow going.

I had originally considered that it would need power (trivia at bottom), i instead went for a high tech material, surge alloy in this case, to be the explanation for why it works without power, open to suggestions regarding this. 

The below example setup is a bit crude, but it illustrates how that over longer distances chained `pressure conduits` the "bonus" adds up, and it can "suck" lower `〃` conduits dry, one could view it as a `reverse overflow gate`, since it can pressurize specific parts of a pipe network first.

(though they don't need to be chained `back to back`, as long as they are somewhere in a flow path their bonuses add up, the pressure they output can be seen as a "checkpoint")

Oct-30-2019 13-07-56

(my initial attempts consisted of a `liquid tank / machine` hybrid with compression animations and what not, but that hit a few roadblocks along the way, so i tried a new conduit type instead 😗)


Anuken:
I think this should just replace the standard titanium conduit, which currently has little benefit over the standard one.

core/src/io/anuke/mindustry/world/blocks/distribution/Conduit.java
```diff
@@ -17,6 +17,8 @@
     protected TextureRegion[] topRegions = new TextureRegion[7];
     protected TextureRegion[] botRegions = new TextureRegion[7];
 
+    public float bar;
```
Anuken: This should be a property of `BlockStorage` to prevent an ugly cast, and should have a different name (like pressureScale or something).

Quezler:
Requested changes have been made, not sure if i should update the description of the `pulse-conduit`, most would simply neither notice nor care, and on top of that i have no clue what exactly to change it into 😅

core/src/io/anuke/mindustry/world/BlockStorage.java
```diff
@@ -143,7 +144,7 @@ public float tryMoveLiquid(Tile tile, Tile next, boolean leak, Liquid liquid){
 
             if(next.block().acceptLiquid(next, tile, liquid, 0f)){
                 float ofract = next.entity.liquids.get(liquid) / next.block().liquidCapacity;
-                float fract = tile.entity.liquids.get(liquid) / liquidCapacity;
+                float fract = tile.entity.liquids.get(liquid) / liquidCapacity * tile.block().liquidPressure;
```
Anuken: This can just be `liquidPressure`, as you are already calling this as the transportation block.

player20033:
Sorry, this idea was implemented on game already. add a liquid tank and liquid junction to profit it.

Quezler:
referencing #999

Anuken:
Should be safe to merge this now.

PR #950, opened 2019-10-30
📎 Auto junctions

Quezler:
A possible implementation for [this suggestion](https://feathub.com/Anuken/Mindustry/+19) by @constatinus

### DIFF
--- a/core/src/io/anuke/mindustry/input/InputHandler.java
+++ b/core/src/io/anuke/mindustry/input/InputHandler.java
@@ -433,7 +433,12 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
         if(over.block() != null){
             if(over.block() instanceof Conveyor){
                 if ((over.rotation() + rotation) % 2 != 0){
-                    block = Blocks.junction;
+
+                    Tile behind = over.getNearby((rotation + 2) % 4);
+
+                    if (behind.block() != null && behind.block() instanceof Conveyor){
+                        block = Blocks.junction;
+                    }
                 }
             }
         }
--- a/core/src/io/anuke/mindustry/input/InputHandler.java
+++ b/core/src/io/anuke/mindustry/input/InputHandler.java
@@ -32,6 +32,7 @@ import io.anuke.mindustry.ui.fragments.*;
 import io.anuke.mindustry.world.*;
 import io.anuke.mindustry.world.blocks.*;
 import io.anuke.mindustry.world.blocks.BuildBlock.*;
+import io.anuke.mindustry.world.blocks.distribution.Conveyor;
 
 import java.util.*;
 
@@ -418,9 +419,26 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
                 if(copy.hasConfig && copy.block.posConfig){
                     copy.config = Pos.get(Pos.x(copy.config) + copy.x - copy.originalX, Pos.y(copy.config) + copy.y - copy.originalY);
                 }
+
+                copy.block = didyoumean(req.x, req.y, req.block, req.rotation);
+
                 player.addBuildRequest(copy);
             }
         }
+    }
+
+    protected Block didyoumean(int x, int y, Block block, int rotation){
+        Tile over = world.ltile(x, y);
+
+        if(over.block() != null){
+            if(over.block() instanceof Conveyor){
+                if ((over.rotation() + rotation) % 2 != 0){
+                    block = Blocks.junction;
+                }
+            }
+        }
+
+        return block;
     }
 
     protected void drawRequest(BuildRequest request){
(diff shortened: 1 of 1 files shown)


core/src/io/anuke/mindustry/input/InputHandler.java
```diff
@@ -418,17 +419,50 @@ protected void flushRequests(Array<BuildRequest> requests){
                 if(copy.hasConfig && copy.block.posConfig){
                     copy.config = Pos.get(Pos.x(copy.config) + copy.x - copy.originalX, Pos.y(copy.config) + copy.y - copy.originalY);
                 }
+
+                copy.block = didyoumean(req.x, req.y, req.block, req.rotation);
```
Anuken: Blocks should be swapped only when you are placing a line, not when you are flushing them to the main build queue. The player should be able to see that their line request blocks are turning to junctions before they choose to place them, without using any special hacks to get it to display that way in the rendering code. Also, this won't work with mobile input. Instead of changing code here, you should change `updateLine` to replace the build request blocks.

core/src/io/anuke/mindustry/input/InputHandler.java
```diff
@@ -418,17 +419,50 @@ protected void flushRequests(Array<BuildRequest> requests){
...
+
                 player.addBuildRequest(copy);
             }
         }
     }
 
+    protected Block didyoumean(int x, int y, Block block, int rotation){
+        Tile over = world.ltile(x, y);
+
+        if(over.block() != null){
```
Anuken: Redundant condition, and again, `over.block()` **cannot be null**, only the tile itself can.

core/src/io/anuke/mindustry/input/InputHandler.java
```diff
@@ -418,17 +419,50 @@ protected void flushRequests(Array<BuildRequest> requests){
...
+                    for(int i = 0; i < lineRequests.size; i++){
+                        BuildRequest req = lineRequests.get(i);
+
+                        if(behind.x == req.x && behind.y == req.y){
+                            previous = req.block;
+                            break;
+                        }
+                    }
+
+                    if(previous != null && previous instanceof Conveyor){
```
Anuken: `previous` can't be null.

core/src/io/anuke/mindustry/input/InputHandler.java
```diff
@@ -418,17 +419,50 @@ protected void flushRequests(Array<BuildRequest> requests){
...
     }
 
+    protected Block didyoumean(int x, int y, Block block, int rotation){
+        Tile over = world.ltile(x, y);
+
+        if(over.block() != null){
+            if(over.block() instanceof Conveyor){
+                if((over.rotation() + rotation) % 2 != 0){
+
+                    Tile behind = over.getNearby((rotation + 2) % 4);
```
Anuken: `behind` can be null.

core/src/io/anuke/mindustry/input/InputHandler.java
```diff
@@ -418,17 +419,50 @@ protected void flushRequests(Array<BuildRequest> requests){
...
+                    Block previous = behind.block();
+
+                    for(int i = 0; i < lineRequests.size; i++){
```
Anuken: I don't think replacement should work this way. Instead of checking the existing previous block, you could check whether `lineRequests` contains two conveyor requests, one behind and one in front:
```java
lineRequests.contains(r -> r.x == x + Geometry.d4[rotation].x && r.y == y + Geometry.d4[rotation].y && r.block instanceof Conveyor && r.rotation == rotation) && lineRequests.contains(r -> r.x == x - Geometry.d4[rotation].x && r.y == y - Geometry.d4[rotation].y && r.block instanceof Conveyor && r.rotation == rotation)
```

core/src/io/anuke/mindustry/input/InputHandler.java
```diff
@@ -418,17 +419,50 @@ protected void flushRequests(Array<BuildRequest> requests){
...
                 }
+
+                copy.block = didyoumean(req.x, req.y, req.block, req.rotation);
+
                 player.addBuildRequest(copy);
             }
         }
     }
 
+    protected Block didyoumean(int x, int y, Block block, int rotation){
```
Anuken: I would call this `checkReplacement` or something similar. Additionally, it would be a good idea to make replacements get defined *in the blocks themselves*, so that each block could return a block it could be replaced with given a build queue:

```java
//in Block.java
public Block getReplacement(int x, int y, int rotation, Array<BuildRequest> requests /* or Eachable<BuildRequest>*/){
    return this; //override to return a junction, etc
}
```

This would allow blocks like conduits to override this behavior easily and implement replacements for liquid junctions.

Anuken:
Decided to implement it in 2c59288969cb27c9fb2ae9c4e61ca3a4a234f5b2.

PR #954, opened 2019-10-30
Move Force Projector visuals to (new) ProjectorTrait

l-Luna:
image
This moves the Force Projector's visuals to a new ProjectorTrait class, and adds similar visuals to the Overdrive and Mend projectors with it. This should make adding mod support for projectors simpler.


Anuken:
This could be useful for new types of shield generators in the future, but I really don't think it's necessary for mend/overdrive projectors. Considering how often they are used, this will add a lot of clutter to the screen.

Quezler:
But it would be neat to show the range of overdrive projectors (when placing new ones) so its more clear weather or not they overlap (to much) or leave a gap 😗

l-Luna:
Otherwise, is there anything that needs changing?
I agree that it's unnecessary on Mend Projectors, but since Overdrive Projectors are much less used, and don't (appear) to benefit from overlap, a visual indicator would be useful.

player20033:
It should be put in options so we could disable it.

l-Luna:
Added settings options, moved to Array and ArrayMap. Anything else that needs changing?

Anuken:
> a visual indicator would be useful

There are many different blocks with ranges (turrets, repair points...). I don't think they should be toggled individually in settings, and I also don't think that they should use the shield rendering code at all. Range indicators aren't shields, and the shields' wobblyness can make precise things harder to make out.

I may implement something like this later in development, but the structure will be very different from this PR.

PR #955, opened 2019-10-30
💼 Optionally show resources ingame

Quezler:
Screen Shot 2019-10-30 at 20 18 56
This is just a small modification i made to my client (just now) and will be using it in my custom builds, and i thought i would just share it (yay open source), not intended to be merged unless you think its useful (code is a bit crude), i personally find it neat that i now can view the resources of my team without having to take a peek at the core all the time or hover over a handful of buildings in the menu to get a glimpse, this also communicates nonverbaly that your team might be running out of a certain resource so you can act accordingly & ahead of time, instead of suddenly not being able to build anything anymore. (`false/off` by default)

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -621,6 +621,7 @@ setting.vsync.name = VSync
 setting.pixelate.name = Pixelate[lightgray] (disables animations)
 setting.minimap.name = Show Minimap
 setting.position.name = Show Player Position
+setting.resources.name = Show Available Resources
 setting.musicvol.name = Music Volume
 setting.ambientvol.name = Ambient Volume
 setting.mutemusic.name = Mute Music
--- a/core/src/io/anuke/mindustry/ui/ItemsDisplay.java
+++ b/core/src/io/anuke/mindustry/ui/ItemsDisplay.java
@@ -38,15 +38,24 @@ public class ItemsDisplay extends Table{
             int i = 0;
             for(Item item : content.items()){
                 if(item.type == ItemType.material && data.isUnlocked(item)){
-                    t.label(() -> format(item)).left().minWidth(50);
-                    t.addImage(item.icon(Cicon.small)).size(8 * 3).padLeft(4).padRight(4);
                     if(campaign){
+                        t.label(() -> format(item)).left().minWidth(50);
+                        t.addImage(item.icon(Cicon.small)).size(8 * 3).padLeft(4).padRight(4);
                         t.add(item.localizedName()).color(Color.lightGray).left();
                         t.row();
-                    } else {
+                    }else if(mobile){
                         if (++i % 2 == 0){
+                            t.label(() -> format(item)).left().minWidth(50);
+                            t.addImage(item.icon(Cicon.small)).size(8 * 3).padLeft(4).padRight(4);
                             t.row();
+                        }else{
+                            t.addImage(item.icon(Cicon.small)).size(8 * 3).padLeft(4).padRight(4);
+                            t.label(() -> format(item)).left().minWidth(50);
                         }
+                    }else{
+                        t.label(() -> format(item)).left().minWidth(50);
+                        t.addImage(item.icon(Cicon.small)).size(8 * 3).padLeft(4).padRight(4);
+                        t.row();
                     }
                 }
             }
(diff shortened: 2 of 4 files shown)


Anuken:
I would make a few modifications:
- The material names aren't important enough to take up space on the UI. If someone doesn't know what a material looks like, they can look at the building menu/database for a name. A tooltip could help if this turns out to be absolutely necessary.
- Materials that you have 0 of should not show up. There's a lot of wasted space from stuff you may not even have unlocked yet.
- Having 2 materials per row could help conserve space when in landscape mode (mobile) or on PC.

Quezler:
- makes sense, ill remove those
- personally i disagree on hiding resources you have 0 of, but only when you have researched them, seeing something actually be zero is an incentive to create/upgrade the production for it
- ill give it a go and see how it looks

Shawak:
Disagree on hiding resources which are zero also because if you are low on a resource and poeple building stuff especially on multiplayer it will jump from which will lead to constantly moving icons and numbers.

Quezler:
Exactly ^

Quezler:
Which one look better? 🤔 

Screen Shot 2019-10-31 at 08 14 47
Screen Shot 2019-10-31 at 08 15 17
Screen Shot 2019-10-31 at 08 16 04

Shawak:
I like this most I guess

image

Quezler:
@Shawak currently its in the [order of registration](https://github.com/Anuken/Mindustry/blob/master/core/src/io/anuke/mindustry/content/Items.java#L13), curious what @**Anuken** thinks regarding this 😗

Quezler:
look at this abomination i had accidentally created `(>.<)` (not part of pull)

Screen Shot 2019-10-31 at 08 39 01

jhitchco:
This will be a great addition

majora2007:
I actually like the horizontal one, esp if we think about mobile. Might be nice to be able to hide it if you don't want to see it. Imagine this:

image

image


Made in MS Paint.

Quezler:
A current issue is that it displays under the chat window 😗

majora2007:
Could we instead have the resource view top-left instead? 

The only reason I push for horizontal is vertical is a bit hard to read, but is better than nothing. Perhaps put the expand/collapse on that to think about mobile users.

Quezler:
But thats where the wave counter sits 🤔

majora2007:
Is the layout responsive in the sense that if there isn't enough x axis, it wraps down to another line? Because if so, we could have between wave counter and mini-map thing. It can be expanded. On mobile, we'd expect a few lines and they would tap it open/close when needed. Desktop users would see one row, but could collapse if they don't like it.

Jaeiya:
This is an amazingly QoL feature, any idea when this might be merged?

Anuken:
Most likely when a proper location for it is found and the details are finalized.

@Quezler

> if you are low on a resource and poeple building stuff especially on multiplayer it will jump from zero to one all the time

This is a valid issue, but it can be fixed by having resources never disappear from the menu once they appear as >0 once in any game. This would cleanly solve the issue of locked items as well, without needing a campaign specific check.

You also need to consider how this will look with mods. If you've ever tried using one of the bigger ones, you'll find that they have a massive amount of items. Displaying only items that have been previously obtained will save a lot of space without sacrificing any usability.

dlightle:
I just started playing and I was looking for exactly this myself.  If I get some time I might try to figure out the code and might try working up something myself, but until then, I wanted to say I also like the horizontal view (on PC, but seems reasonable for landscape on mobile too I think).

However, could you have it in the middle of the top, thinking something like older RTS titles like Warcraft (2/3) would do for resource display?  I was also thinking about power output, but I guess that's trickier if you have more than one power grid.

famo:
This is great!
It would also be very cool, while placing something the build cost would be displayed alongside this HUD element. So you can see on a quick glance, if you have enough resources or not.

And even nicer would be the highlighting of insufficient resources by displaying them in red numbers (by the build cost). :smiley: 

For the schematics cost I made this feature request.
https://feathub.com/Anuken/Mindustry/+542

Anuken:
Scheduling this for 6.0, as there are many things that may need to be taken into account when implementing.

MWestfall:
Would be veeeery useful.

MEEPofFaith:
People trying to build separators on nydus 3 seconds after stating:

Anuken:
Implemented for desktop as of the latest commit. Currrently displayed on the top of the screen.

Quezler:
`#dev-previews`?

Anuken:
I'm not very happy with where it's located in the interface yet. I'll post a preview when there's space for it on mobile screens.

PR #964, opened 2019-10-31
🔩 Add (plastanium) diode

Quezler:
A possible implementation for [this suggestion](https://feathub.com/Anuken/Mindustry/+20) by @constatinus (and many others)

**functionality**
This pull adds an insulated diode block, that (currently) forwards power (50% of the difference) from one battery to another if the power within `from` is greater than the `to` battery, a large battery can fully fill both small and large batteries, but small batteries will only fill large batteries to `4k`.

**example 1**
Below is an example of an `impact reactor` that will not shut down if the "outside network"'s power turns negative, since the battery inside the walls will only store/transmit "excess" energy, thus it will always have just enough power to stay running, but it will export all that it won't need.

Screen Shot 2019-10-31 at 12 34 50

**example 2**
Below is an example of high energy usage machines/parts of your base that you would only want to run if there is enough spare power for it.
Screen Shot 2019-10-31 at 13 16 54

**thoughts**
- i went with the `plastanium` border color to signify its insulated property, which fits nice with the existing power insulation walls, to signify & prevent accidental connections through a possibly one wide gap.
- base texture was a solar panel, which is overlayn by what most would consider a diode icon.
- recipe is a mix between solar panels and plastanium walls, since its both electric but also insulates.
- not sure if i used `entity.sleep()` correctly here, it would make some sense to not constantly check for batteries until a proximity update has been found.
- my intention is to add a functionality to "inject" into the other power graph somehow, thoughts on that are pending.
- it extends `Block` since that was the first place i found that had rotation support, though it should probably be moved to a power block type.
- went with interacting between batteries since i wasn't sure if/how a block could support being part of two graphs, and battery blocks happen to store power, so this simple thing just facilitates moving power from one battery to another, not caring which graph any is connected to.

🥔


AeronGreva:
This is a really good idea, and would definitely solve the downward spiral caused by relying on impact reactors.

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
```diff
@@ -28,24 +27,18 @@ public void update(Tile tile){
         Tile back = getNearby(tile, (tile.rotation() + 2) % 4);
         Tile front = getNearby(tile, tile.rotation());
 
-        if(back.block() != null && back.block() instanceof Battery &&
-          front.block() != null && front.block() instanceof Battery){
+        if(back.block() == null || front.block() == null) return;
```
Anuken: Blocks can't be null, only tiles can.

Quezler: Noted 👍

Quezler:
**current caveats** (yea working on it)
- receiving graph needs a battery to receive in
- if the amount sent is greater than the batteries on the receiving graph, its free

Dexapnow:
I thought surge is the one with electric based theme.

Anuken:
A proposed sprite change:

image

Yes, this doesn't blend in with the walls as much, but the original plastanium base really sticks out like a sore thumb in the build menu and in contexts where no plastanium is around. I think it should have the same gray background as all the other power blocks.

Also, the original diode sprite was too detailed for the style, so I replaced it with a simple arrow design.

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    public void draw(Tile tile){
+        Draw.rect(region, tile.drawx(), tile.drawy(), 0);
+        Draw.rect(arrow, tile.drawx(), tile.drawy(), rotate ? tile.rotation() * 90 : 0);
+    }
+
+    @Override
+    public TextureRegion[] generateIcons(){
+        return new TextureRegion[]{Core.atlas.find(name + "-arrow"), Core.atlas.find(name)};
```
Anuken: This can be removed (see my other comment).

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public void load(){
+        super.load();
+        arrow = Core.atlas.find(name + "-arrow");
+    }
+
+    @Override
+    public void draw(Tile tile){
+        Draw.rect(region, tile.drawx(), tile.drawy(), 0);
+        Draw.rect(arrow, tile.drawx(), tile.drawy(), rotate ? tile.rotation() * 90 : 0);
```
Anuken: You should also override the `drawRequestRegion(...)` method to also rotate it properly, since the block looks weird in schematics/placement (as the whole thing is rotated there).

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
```diff
@@ -0,0 +1,85 @@
...
+import io.anuke.arc.graphics.g2d.Draw;
+import io.anuke.arc.graphics.g2d.TextureRegion;
+import io.anuke.arc.math.Mathf;
+import io.anuke.mindustry.world.Block;
+import io.anuke.mindustry.world.Tile;
+
+import static io.anuke.mindustry.Vars.world;
+
+public class PowerDiode extends Block{
+
```
Anuken: This extra indent is unnecessary.

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
```diff
@@ -0,0 +1,85 @@
...
+        Draw.rect(region, tile.drawx(), tile.drawy(), 0);
+        Draw.rect(arrow, tile.drawx(), tile.drawy(), rotate ? tile.rotation() * 90 : 0);
+    }
+
+    @Override
+    public TextureRegion[] generateIcons(){
+        return new TextureRegion[]{Core.atlas.find(name + "-arrow"), Core.atlas.find(name)};
+    }
+
+    public Tile getNearby(Tile tile, int rotation){
```
Anuken: I would move this to Tile and call it `getNearbyLink`.

core/assets/bundles/bundle.properties
```diff
@@ -892,6 +892,7 @@ block.coal-centrifuge.name = Coal Centrifuge
 block.power-node.name = Power Node
 block.power-node-large.name = Large Power Node
 block.surge-tower.name = Surge Tower
+block.diode.name = Diode
```
Anuken: Maybe call it "Battery Diode" so people don't assume it works for graphs without batteries. Not sure about that, though.

Ansraer: I thought the fact that it only worked with batteries was a bug?

Anuken: No, it's a current limitation of how the diode works. Unless a different method of directional power transfer is used, it'll stay that way.

Quezler:
**psa** resuming progress on this likely sometime tomorrow 🌑

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
Quezler: i did try that as well originally, but my ide really didn't like that at all 🙄 

Screen Shot 2019-11-04 at 14 29 32
Screen Shot 2019-11-04 at 14 29 37
Screen Shot 2019-11-04 at 14 29 41
Screen Shot 2019-11-04 at 14 28 46
Screen Shot 2019-11-04 at 14 29 08

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
Quezler: I don't see which indentation is out of place 🤔

Quezler: i don't see which indentation is out of place 🤔

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
Anuken: Refresh your dependencies or remove/update your local version of Arc. That interface was moved about a week ago.

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
Anuken: Sorry, I meant blank line, not indent.

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
```diff
@@ -0,0 +1,94 @@
...
+            return backTile().entity.power.graph;
+        }
+        public PowerGraph frontGraph(){
+            if (!frontTile().block().hasPower) return null;
+            return frontTile().entity.power.graph;
+        }
+        public boolean connected(){
+            return backGraph() != null && frontGraph() != null;
+        }
+        //
```
Anuken: What's the empty comment for?

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public TileEntity newEntity(){
+        return new DiodeEntity();
+    }
+
+    public static class DiodeEntity extends TileEntity{
+
+        public Tile backTile(){
+            return tile.getNearbyLink(tile, (tile.rotation() + 2) % 4);
+        }
```
Anuken: Add a blank line between the methods.

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
```diff
@@ -0,0 +1,94 @@
...
+        insulated = true;
+    }
+
+    @Override
+    public void update(Tile tile){
+        super.update(tile);
+
+        DiodeEntity entity = (DiodeEntity) tile.entity;
+
+        if (!entity.connected()) return;
```
Anuken: extra spaces again

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
```diff
@@ -0,0 +1,94 @@
...
+        if (entity.input() > entity.output()) {
+            entity.transfer(entity.backGraph().getBatteryStored() * (entity.input() - entity.output()) / 2);
+        }
+    }
+
+    @Override
+    public void setBars() {
```
Anuken: space

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
Quezler: i did use the gradle refresh flag, without result.

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
Quezler: oh those just grouped methods of the same category.

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
Quezler: was part of a group ^

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
Quezler: tried --refresh-dependencies & removing the gradle cache folder, issues persist 🤔

Quezler: apparently also had to knock some sence into intelij 😞

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    public TileEntity newEntity(){
+        return new DiodeEntity();
+    }
+
+    public static class DiodeEntity extends TileEntity{
+
+        public Tile back(){
```
Anuken: Some notes about these methods:
1. The `front()` and `back()` methods seem like utilities that could be more useful in `Tile`, not this entity class. In fact, one of my branches has these methods in it.
2. I am not a fan of making turning everything into a mini-method like this. Trying to figure out what's going on in the main `update()` method is very difficult, considering I need to look through 2-3 different method calls in a different part of the file to figure out what it's actually doing. [This earlier version](https://github.com/Anuken/Mindustry/blob/2cd8fe0af9fcb1f3014e90f054ba52d69830265a/core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java#L25) seems a lot more readable: I can see what exactly it's doing, without having to untangle what each method's task is. If these methods were used extensively for many different classes, that would be another story, but in this case it seems like the only thing they will ever be used for is one `update()` method.
3. Even if you *needed* new utility methods, it would be a better idea to put them in the block class instead of creating a new `DiodeEntity` class. `TileEntity` classes are only for storing block specific data.

Quezler: 1. moved

2. reverted

3. noted

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
```diff
@@ -0,0 +1,87 @@
...
+        super.update(tile);
+
+        if(!tile.back().block().hasPower || !tile.front().block().hasPower) return;
+
+        PowerGraph backGraph = tile.back().entity.power.graph;
+        PowerGraph frontGraph = tile.front().entity.power.graph;
+        if(backGraph == frontGraph) return;
+
+        // 0f - 1f of battery capacity in use
+        Float backStored = backGraph.getBatteryStored() / backGraph.getTotalBatteryCapacity();
```
Anuken: Should be `float`, not Float.

Quezler: Weird how neither me nor my ide noticed that.

Quezler:
Hmm, i may or may not have messed up the (unused) left and right methods 🤔

core/src/io/anuke/mindustry/world/blocks/power/PowerDiode.java
```diff
@@ -0,0 +1,86 @@
...
+        // 0f - 1f of battery capacity in use
+        float backStored = backGraph.getBatteryStored() / backGraph.getTotalBatteryCapacity();
+        float frontStored = frontGraph.getBatteryStored() / frontGraph.getTotalBatteryCapacity();
+
+        // try to send if the back side has more % capacity stored than the front side
+        if(backStored > frontStored) {
+            // send half of the difference
+            float amount = backGraph.getBatteryStored() * (backStored - frontStored) / 2;
+
+            backGraph.useBatteries(amount);
```
Anuken: A possible issue that may arise:
Graph 1 is transferring power to graph 2. The power fractions are as follows:

1: 1.0 - 50000 / 50000
2: 0.5 - 5/10

This would transfer (1.0 - 0.5) / 2 = 0.25 of the first graph, or 12500 power. For at least one frame, that's going to transfer a lot more power than the second graph can handle. A solution would be to `clamp(amount, 0, frontGraph.getTotalBatteryCapacity() * (1 - frontStored))` so no extra power is transferred.

Shawak:
How about chaning the sprite to an "active" version, so it's visible when power get's transfered into another network?

Anuken:
It wouldn't be difficult to tint the diode sprite depending on flow rate - but I think that's outside the scope of this PR.

core/assets/bundles/bundle.properties
```diff
@@ -900,6 +902,7 @@ block.coal-centrifuge.name = Coal Centrifuge
 block.power-node.name = Power Node
 block.power-node-large.name = Large Power Node
 block.surge-tower.name = Surge Tower
+block.diode.name = Battery diode
```
Anuken: "diode" should be uppercase.

Quezler:
Probably because i edit textures upwards and all the assets are rotated clockwise 😅

willis936:
Tech tree wise: would this be available prior to thorium?  I’ve failed the salt flats scenario a few times now, mostly because of power failure.  The power failure comes from over demand starving the water pumps and coal drills that feed the combustion generators.  The plastanium diode looks like a solution to this.  I definitely think there should be a tutorial scenario (or some well written text) describing how to use it in game.  Not everyone will see this PR.

KenSharp:
The naming is insanely confusing as this is not how a real diode actually works. I had to come here just to work out why it behaves the way it does.

PR #967, opened 2019-11-01
Improve server host command

Shawak:
Start the server with a random map if none specified

### DIFF
--- a/server/src/io/anuke/mindustry/server/ServerControl.java
+++ b/server/src/io/anuke/mindustry/server/ServerControl.java
@@ -227,7 +227,7 @@ public class ServerControl implements ApplicationListener{
                     err("No map with name &y'{0}'&lr found.", arg[0]);
                     return;
                 }
-            } else {
+            }else{
                 Array<Map> maps = Vars.maps.customMaps().size == 0 ? Vars.maps.defaultMaps() : Vars.maps.customMaps();
                 result = maps.random();
             }
--- a/server/src/io/anuke/mindustry/server/ServerControl.java
+++ b/server/src/io/anuke/mindustry/server/ServerControl.java
@@ -211,19 +211,25 @@ public class ServerControl implements ApplicationListener{
             info("Stopped server.");
         });
 
-        handler.register("host", "<mapname> [mode]", "Open the server with a specific map.", arg -> {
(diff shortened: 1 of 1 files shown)


server/src/io/anuke/mindustry/server/ServerControl.java
```diff
@@ -211,19 +211,25 @@ private void registerCommands(){
...
-
-            if(result == null){
-                err("No map with name &y'{0}'&lr found.", arg[0]);
-                return;
+                if(result == null){
+                    err("No map with name &y'{0}'&lr found.", arg[0]);
+                    return;
+                }
+            } else {
+                Array<Map> maps = Array.with(Vars.maps.customMaps().size == 0 ? Vars.maps.defaultMaps() : Vars.maps.customMaps());
```
Anuken: `Array.with` should not be necessary here at all.

server/src/io/anuke/mindustry/server/ServerControl.java
```diff
@@ -211,19 +211,25 @@ private void registerCommands(){
...
+                result = maps.get(new Random().nextInt(maps.length));
```
Anuken: Use `maps.random()`, which just returns a random element.

server/src/io/anuke/mindustry/server/ServerControl.java
```diff
@@ -211,19 +211,25 @@ private void registerCommands(){
             info("Stopped server.");
         });
 
-        handler.register("host", "<mapname> [mode]", "Open the server with a specific map.", arg -> {
+        handler.register("host", "[mapname] [mode]", "Open the server with a random or specific map.", arg -> {
```
Anuken: This is not a very good description for this command. It should be:
`Open the server. Will default to survival and a random map if not specified.`

Anuken:
Also, some word of advice: don't try to write Java or submit pull requests without importing the project into IntelliJ first. Even with the simplest changes, it won't be easy to make do without an IDE. Run pull requests before submitting them.

Shawak:
Yeah I just didn't want to install Java on my machine. I got it running inside a docker image now, lemme rebase this.

Shawak:
@Anuken that should do it

PR #974, opened 2019-11-01
Clarrify where the option to view ping is located

Quezler:
I heard several people mentioning ping but i didn't know how or where to enable it, turns out it was hidden inside the `FPS` option, this pull simply lets it be known.

Screen Shot 2019-11-01 at 21 00 37


Quezler:
`#fastest merge ever`

PR #997, opened 2019-11-05
🐬 Factorio flavoured powerlines

Quezler:
**annoyance**
When trying to move & place powerpoles over long distances it often turns out roughly like this, inefficient and duplicate nodes:
Screen Shot 2019-11-05 at 08 29 58

**proposed**
This draft overrides drag-placing powernodes (since who needs those back to back anyways?) by not placing those middlemen.
By default it places them in a straight line, but they will use the convejor pathfinding when said key is held down.
It will not take existing powernodes on the map into account, thus the lines may interconnect in a strange way, though the beginning and end are guaranteed still connect one way or another.
Nov-05-2019 08-32-33

**caveats**
- doesn't check if a large powerpole can fit in a found 1x1 tile
- doesn't suggest a new location if the found one is entirely obstructed
- doesn't pathfind the shortest route over the terain
- probably more, but hey, its a draft :)

### DIFF
--- a/core/src/io/anuke/mindustry/input/InputHandler.java
+++ b/core/src/io/anuke/mindustry/input/InputHandler.java
@@ -535,28 +535,6 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
                 }
             });
         }
-
-        final int[] i = {0};
-        lineRequests.each(req -> {
-            if(!(req.block instanceof PowerNode)) return;
-
-            if(i[0]++ == 0 || i[0] == lineRequests.size){
-                // beginning & end should always be placed
-            }else{
-                // check with how many powernodes the *next* tile will overlap
-                int overlaps = 0;
-                for(int j = 0; j < i[0]; j++){
-                    if (((PowerNode) req.block).overlaps(lineRequests.get(i[0]).tile(), lineRequests.get(j).tile()) && lineRequests.get(j).block instanceof PowerNode){
-                        overlaps++;
-                    }
-                }
-
-                // if its more than two it can bridge the gap
-                if(overlaps > 1){
-                    req.block = Blocks.air;
-                }
-            }
-        });
     }
 
     protected void updateLine(int x1, int y1){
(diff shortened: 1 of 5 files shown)


Anuken:
> By default it places them in a straight line, but they will use the convevor pathfinding when said key is held down.

I think that making it just go diagonally would be a better idea in most cases. Conveyor pathfinding will definitely be sub-optimal when it comes to node placement, especially considering how it tries to avoid diagonal movement.

Quezler:
Diagonal is it's default, it wont do only straight lines, and when you hold the diagonal key down it changes from diagonal to conveyor mode 😗

Quezler:
edit: d57d177

Anuken:
> it changes from diagonal to conveyor mode

This is still a problem. The pathfinding isn't designed for blocks larger than 1x1, and gives very suboptimal results due to its axis locking. It's also short range, which makes it seldom useful for power node placement. 

How about defaulting to diagonal mode, and swapping to line mode when the key is held, if the user wants to place in a straight line for whatever reason? (Or you could just have the standard behavior)

Quezler:
Allright, i'll take the convejoy pathfinding away from powernodes 👍

core/src/io/anuke/mindustry/input/InputHandler.java
```diff
@@ -806,16 +807,50 @@ public void drawArrow(Block block, int x, int y, int rotation, boolean valid){
...
         if(diagonal){
             points = Placement.pathfindLine(block != null && block.conveyorPlacement, startX, startY, endX, endY);
         }else{
             points = Placement.normalizeLine(startX, startY, endX, endY);
         }
 
+        if(block instanceof PowerNode){
+            Array<Point2> skip = new Array<>();
+            final int[] i = {0};
+            points.each(req -> {
```
Anuken: Why not use a standard for-loop here? It's less lines, and avoids ugly array wrappers.

Quezler: guess it was just a remnant that stuck from something that worked 😅

core/src/io/anuke/mindustry/input/InputHandler.java
```diff
@@ -806,16 +807,50 @@ public void drawArrow(Block block, int x, int y, int rotation, boolean valid){
...
+                // check with how many powernodes the *next* tile will overlap
+                int overlaps = 0;
+                for(int j = 0; j < i; j++){
+                    // skip powernodes we have already crossed off as air
+                    if(skip.contains(points.get(j))) continue;
+
+                    Tile next = world.ltile(points.get(i).x, points.get(i).y);
+                    Tile loop = world.ltile(points.get(j).x, points.get(j).y);
+
+                    if (((PowerNode) block).overlaps(next, loop)){
```
Anuken: Extra space after `if` and after cast.

core/src/io/anuke/mindustry/input/InputHandler.java
```diff
@@ -806,16 +807,50 @@ public void drawArrow(Block block, int x, int y, int rotation, boolean valid){
...
             points = Placement.pathfindLine(block != null && block.conveyorPlacement, startX, startY, endX, endY);
         }else{
             points = Placement.normalizeLine(startX, startY, endX, endY);
         }
 
+        if(block instanceof PowerNode){
+            Array<Point2> skip = new Array<>();
+            
+            for(int i = 0; i < points.size; i++){
+                if(i == 0) continue; // beginning should always be present
```
Anuken: Instead of skipping 0, you could start at `i = 1`.

Anuken:
Looks good. Will make changes after merging if I find any issues.

Quezler:
By all means, all of my pulls should be considered a rough draft to convey an idea, there is always room for code improvements of them, i won't stop ya 😇

PR #998, opened 2019-11-05
Power productivity & Drone pad standby

Quezler:
This pull request introduces two chances:

- power productivity, consumers can now (unlike before) increase/decrease their power realtime.
- drone pad standby, consumes no power while not in use.

Since drones are autonomous once launched it would not make sense to lower their move/mine/attack speed based on the condition their pad is in, and the pad is literally & visually not doing anything while the drones are maxed, so it would make sense for them to use no power during.

Possible balance changes:

- increase base power usage of all drone pads, to compensate the long run.
- make standby drone pads use a fraction of the power while on standby.

This pull is mainly to prevent the constant (dis)connecting of (mainly miner) drones to save a few grains of power, which seems like a chore to have to babysit them until they grow up. (or die)

edit: found a related [feathub](https://feathub.com/Anuken/Mindustry/+117) entry.

### DIFF
--- a/core/src/io/anuke/mindustry/world/blocks/units/UnitFactory.java
+++ b/core/src/io/anuke/mindustry/world/blocks/units/UnitFactory.java
@@ -155,10 +155,8 @@ public class UnitFactory extends Block{
         UnitFactoryEntity entity = tile.entity();
 
         if(entity.spawned >= maxSpawn){
-            entity.power.productivity = 0f;
             return;
         }
-        entity.power.productivity = 1f;
 
         if(entity.cons.valid() || tile.isEnemyCheat()){
             entity.time += entity.delta() * entity.speedScl * Vars.state.rules.unitBuildSpeedMultiplier * entity.power.satisfaction;
--- a/core/src/io/anuke/mindustry/world/blocks/units/UnitFactory.java
+++ b/core/src/io/anuke/mindustry/world/blocks/units/UnitFactory.java
@@ -155,8 +155,10 @@ public class UnitFactory extends Block{
         UnitFactoryEntity entity = tile.entity();
 
         if(entity.spawned >= maxSpawn){
+            entity.power.productivity = 0f;
             return;
         }
+        entity.power.productivity = 1f;
 
         if(entity.cons.valid() || tile.isEnemyCheat()){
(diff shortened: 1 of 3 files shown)


12358abc:
👍

core/src/io/anuke/mindustry/world/consumers/ConsumePower.java
```diff
@@ -72,7 +72,7 @@ public float requestedPower(TileEntity entity){
         if(buffered){
             return (1f-entity.power.satisfaction)*capacity;
         }else{
-            return usage;
+            return usage * entity.power.productivity;
```
Anuken: This is actually redundant; many blocks already override a method that specifies whether they should consume things: https://github.com/Quezler/Mindustry/blob/30a6506f736e0fee4fefda38e2776f607728384d/core/src/io/anuke/mindustry/world/blocks/units/UnitFactory.java#L197. I just never implemented it for power.

This can be changed to 
`usage * Mathf.num(entity.block.shouldConsume(entity.tile))`.

(Mathf.num should just convert a Boolean to 1 or 0, so if shouldConsume is false usage would be 0).

Quezler:
Will change it later tonight 👍

core/src/io/anuke/mindustry/world/consumers/ConsumePower.java
Quezler: done.

core/src/io/anuke/mindustry/world/blocks/units/UnitFactory.java
```diff
@@ -155,8 +155,10 @@ public void update(Tile tile){
         UnitFactoryEntity entity = tile.entity();
 
         if(entity.spawned >= maxSpawn){
+            entity.power.productivity = 0f;
```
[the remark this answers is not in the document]
Quezler: done.

Quezler:
While we're on the topic, what about e.g. an idle airblast drill? should it/everything use no power when in standby mode? i'd like to hear your views & thoughts on it 🤔

Anuken:
Idle blocks will not use power after 3691ab7f8f208d1fa33a61d2140135b67cd3d6a6. I replaced the `canProduce` method since it was used for pretty much the same thing as `shouldConsume`.

Quezler:
That will free up so much power :)

`*happy generator noises*`

PR #999, opened 2019-11-05
🐉 Armored conduit

Quezler:
> The unholy offspring of [lightning](https://feathub.com/Anuken/Mindustry/+130) and [death](https://feathub.com/Anuken/Mindustry/+33) itself. Never engage this ~dragon~ conduit. Your only chance: Hide and pray it does not ~find you~ leak. (edit: and apparently [this](https://feathub.com/Anuken/Mindustry/+124) one as well)

This pull adds an conduit with the following properties:
- armored, duh.
- leaks significantly slower.
- same connection behavior as its conveyor counterpart.

Screen Shot 2019-11-05 at 15 00 19
named it `plated-conduit` instead of `armored-conduit` since the `phase-conduit` isn't named `titanium-conduit`, thus opposing the conveyor naming logic*


Arkanic:
Wraith fighter's demise.

core/src/io/anuke/mindustry/world/blocks/distribution/ArmoredConveyor.java
```diff
@@ -17,8 +17,12 @@ public boolean acceptItem(Item item, Tile tile, Tile source){
 
     @Override
     public boolean blends(Tile tile, int rotation, int otherx, int othery, int otherrot, Block otherblock){
-        return otherblock.outputsItems() && (Point2.equals(tile.x + Geometry.d4(rotation).x, tile.y + Geometry.d4(rotation).y, otherx, othery)
-        || ((!otherblock.rotate && Edges.getFacingEdge(otherblock, otherx, othery, tile) != null &&
-            Edges.getFacingEdge(otherblock, otherx, othery, tile).relativeTo(tile) == rotation) || (otherblock.rotate && Point2.equals(otherx + Geometry.d4(otherrot).x, othery + Geometry.d4(otherrot).y, tile.x, tile.y))));
+        return otherblock.outputsItems() && blender(tile, rotation, otherx, othery, otherrot, otherblock);
+    }
+
+    public static boolean blender(Tile tile, int rotation, int otherx, int othery, int otherrot, Block otherblock){
```
Anuken: Since this is used by multiple `Autotile`-ing blocks, I would put this in the `Autotiler` trait as a default method and call it `armoredBlend` or something.

core/src/io/anuke/mindustry/world/blocks/distribution/ArmoredConduit.java
```diff
@@ -0,0 +1,46 @@
...
+
+import io.anuke.arc.Core;
+import io.anuke.arc.graphics.g2d.Draw;
+import io.anuke.arc.graphics.g2d.TextureRegion;
+import io.anuke.mindustry.type.Liquid;
+import io.anuke.mindustry.world.Block;
+import io.anuke.mindustry.world.Edges;
+import io.anuke.mindustry.world.Tile;
+
+public class ArmoredConduit extends Conduit{
```
Anuken: This class needs to be re-formatted w/o spaces. Also has extra indent before `capRegion`.

core/src/io/anuke/mindustry/world/blocks/distribution/ArmoredConduit.java
```diff
@@ -0,0 +1,46 @@
...
+
+    protected TextureRegion capRegion;
```
Anuken: This region is so subtle that I don't think most players will notice it. Perhaps don't draw it at all, or make it a bit more solid.

Quezler: removing it also removes the justification as to why it leaks less (the end would literally be open), and if i made it any larger it would get in the way of the directional blends, personally i think its just the right size, most would likely notice it. 🤔

core/src/io/anuke/mindustry/world/blocks/distribution/ArmoredConduit.java
```diff
@@ -0,0 +1,45 @@
...
+        super.load();
+        capRegion = Core.atlas.find(name + "-cap");
+    }
+
+    @Override
+    public void draw(Tile tile){
+        super.draw(tile);
+
+        // draw the cap when a conduit would normally leak
+        Tile next = tile.getNearby(tile.rotation());
```
Anuken: Use your `front()` method so links are accounted for properly. Also make sure it's not null.

Quezler:
🎉

PR #1006, opened 2019-11-05
[graphic] configurable bridge transparency

Quezler:
Implementation of [this](https://feathub.com/Anuken/Mindustry/+31) based on #778 (shown = `50%`)
Screen Shot 2019-11-05 at 20 36 51
(i kept the default at `100%`, though it might make sense to lower it to `75%`, your call though)


Anuken:
Does this work for phase/liquid bridges too?

Quezler:
Ehmm, i don't know, didn't test those since they are already sort of transparent, if they extend this class & thus use the drawing method it's most likeky 🤔

Anuken:
Just checked, and no, they don't. Only the T1 extending bridges would be affected by this. I think it would be a good idea to make the other ones go more transparent with the setting as well.

Quezler:
Both at the same transparency, or will the phase one always be more transparent?

Anuken:
You would multiply its base transparency when setting colors with the configured bridge transparency. Phase bridges have transparent and non-transparent parts.

Quezler:
Done 👍

core/src/io/anuke/mindustry/world/blocks/distribution/ItemBridge.java
```diff
@@ -236,6 +236,10 @@ public void drawLayer(Tile tile){
         Draw.rect(endRegion, tile.drawx(), tile.drawy(), i * 90 + 90);
         Draw.rect(endRegion, other.drawx(), other.drawy(), i * 90 + 270);
 
+        int opacityPercentage = Core.settings.getInt("bridgeopacity");
+        if(opacityPercentage == 0) return;
+
+        Draw.alpha(opacityPercentage / 100f);
```
Anuken: This is incorrect. Alpha is being set a few lines above this call. Instead of setting the alpha here, you should put the opacity check up there and multiply the base opacity by the preferred settings opacity.

Quezler: Isn't that for the arrows?

Anuken: No, it's for the entire bridge.

Quezler: Hmm

Anuken:
Should this PR be considered stale?

Quezler:
I guess i kinda lost track since there are so many pulls from me 🙈

Anuken:
Implemented manually due to conflicts.

PR #1009, opened 2019-11-06
Conveyor upgrade assist

Quezler:
Haha, just noticed, title probably fooled you. 🃏 

Anyways, this (very rough draft pull with ugly code) adds upgrading capabilities to builder drones.

Here is a before and after:
Screen Shot 2019-11-06 at 12 34 29
Screen Shot 2019-11-06 at 12 35 09

Basically what this does is making idle builder drones go around and upgrading upgradable blocks to their next tier. (in steps)

`conduit > pulse conduit`
`conveyor > titanium > armored` (only upgrades to armored if it won't interfere with sides)
`copper wall > titanium > thorium > surge` (plastanium and phase intentionally kept out)

This (currently) only upgrades if the before & after blocks are of the same size, it does not go around upgrading 2x2 single walls to large ones nor 3x3 solar panels to 1 big one, ima let that up to the discussion of this pr.

Mildly related to [feathub](https://feathub.com/Anuken/Mindustry/+169)

### DIFF
--- a/core/src/io/anuke/mindustry/graphics/Pal.java
+++ b/core/src/io/anuke/mindustry/graphics/Pal.java
@@ -83,8 +83,5 @@ public class Pal{
     redDust = Color.valueOf("ffa480"),
     redderDust = Color.valueOf("ff7b69"),
 
-    plasticSmoke = Color.valueOf("f1e479"),
-
-    upgrade = Color.valueOf("#549de5"),
-    upgradeBack = Color.valueOf("#3e73a7");
+    plasticSmoke = Color.valueOf("f1e479");
 }
--- a/core/src/io/anuke/mindustry/graphics/Pal.java
+++ b/core/src/io/anuke/mindustry/graphics/Pal.java
@@ -83,5 +83,8 @@ public class Pal{
     redDust = Color.valueOf("ffa480"),
     redderDust = Color.valueOf("ff7b69"),
 
-    plasticSmoke = Color.valueOf("f1e479");
+    plasticSmoke = Color.valueOf("f1e479"),
+
+    upgrade = Color.valueOf("#549de5"),
+    upgradeBack = Color.valueOf("#3e73a7");
 }
--- a/core/src/io/anuke/mindustry/input/InputHandler.java
+++ b/core/src/io/anuke/mindustry/input/InputHandler.java
@@ -209,9 +209,7 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
     }
 
     public void drawBreaking(BuildRequest request){
-        if(request.upgrade){
-            drawUpgrade(request.x, request.y);
-        }else if(request.breaking){
+        if(request.breaking){
             drawBreaking(request.x, request.y);
         }else{
             drawSelected(request.x, request.y, request.block, Pal.remove);
@@ -230,17 +228,6 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
 
         drawSelected(x, y, block, Pal.remove);
     }
-
-    public void drawUpgrade(int x, int y){
-        Tile tile = world.ltile(x, y);
-        if(tile == null) return;
-        Block block = tile.block();
-
-        if (block.getUpgrade(tile) == null) return;
-
-        drawSelected(x, y, block, Pal.upgrade);
-    }
-
 
     public void useSchematic(Schematic schem){
         selectRequests.addAll(schematics.toRequests(schem, world.toTile(player.x), world.toTile(player.y)));
@@ -380,7 +367,7 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
         return null;
     }
 
-    protected void drawBreakSelection(int x1, int y1, int x2, int y2, boolean upgrade){
+    protected void drawBreakSelection(int x1, int y1, int x2, int y2){
         NormalizeDrawResult result = Placement.normalizeDrawArea(Blocks.air, x1, y1, x2, y2, false, maxLength, 1f);
         NormalizeResult dresult = Placement.normalizeArea(x1, y1, x2, y2, rotation, false, maxLength);
 
@@ -389,11 +376,7 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
                 Tile tile = world.ltile(x, y);
                 if(tile == null || !validBreak(tile.x, tile.y)) continue;
 
-                if (upgrade){
-                    drawUpgrade(tile.x, tile.y);
-                }else{
-                    drawBreaking(tile.x, tile.y);
-                }
+                drawBreaking(tile.x, tile.y);
             }
         }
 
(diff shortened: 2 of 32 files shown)


Quezler:
[discord discussion that changed the course of this pull](https://discordapp.com/channels/391020510269669376/396416151032299521/641610903779016706)

Quezler:
Nov-06-2019 14-17-08

KSean222:
Wouldn't the titanium rail upgrade to a armored rail break setups?

Quezler:
@KSean222 look at the original 2 pictures up top, it doesn't upgrade titanium to armored when there are "item output machines", it only upgrades where it won't break anything 😇

KSean222:
Also, potentially out of the scope of this pull request, but what about mod support for upgrading?

Quezler:
Ehmm, i haven't touched modding in any of my pulls so far & completely unfamiliar with its codebase, the score of this pr is currently just the base game 🤔

Quezler:
Snake prototype, follows the block you're building over until it ends. (turns permitted)
Nov-06-2019 14-51-36

Violet-Scarelli:
That snake is incredibly satisfying.

joshuaptfan:
I oppose the addition of this unit for game design reasons: the other existing drones mine/repair/speed up building/rebuild destroyed, which are almost always benign and desirable actions in any situation (except rebuilding can be undesirable if the destroyed blocks need a resource you're short of). Thus, almost no harm can be done by having those actions take place automatically.

However, upgrading is a thing that, until **very** late game, players only do to specific blocks as needed. There is a huge amount of time where upgrading every block in sight is not only undesirable, but basically griefing in a multiplayer game. Other blocks used poorly are merely useless, but a few drone factories built inappropriately can end a game.

Keep in mind that drones cannot be turned off temporarily after they are built, only by deconstructing their factory.

KSean222:
@joshuaptfan  It's been changed from an addition to the drones to an upgrade select.

Anuken:
I agree with @joshuaptfan - drones should only be doing basic tasks that cannot possibly cause harm or issues. Upgrading conveyors with a single click manually is definitely a desirable feature, but having drones do *any* placement automatically without a precedent is dangerous.

Quezler:
Upgrading via drones is long gone, read the first comment 🙄

Quezler:
As mentioned in my previous comment, this pull has since changed directions, and this is now the end result:

When holding the `shift/boost/fly` key it will pathfind along the chain of blocks under it up to where the cursor is, this allows for upgrading (for instance) conveyors with lots of twists and turns relatively easy.

Nov-09-2019 10-50-10

core/src/io/anuke/mindustry/input/Placement.java
```diff
@@ -56,6 +56,31 @@
...
+    public static Array<Point2> followLineUntil(int startX, int startY, int endX, int endY){
+        Pools.freeAll(points);
+        points.clear();
+
+        Tile cursor = world.ltile(startX, startY);
+        Block filter = cursor.block();
+
+        while(true){
+            if(cursor.block() != filter) break;                        // keep looking until encountering a different block
+            if(points.contains(new Point2(cursor.x, cursor.y))) break; // avoid biting its own tail
```
Anuken: Use a predicate, don't create new points for checks: `contains(p -> p.equals(cursor.x, cursor.y))`.

core/src/io/anuke/mindustry/input/Placement.java
```diff
@@ -56,6 +56,31 @@
...
+            points.add(new Point2(cursor.x, cursor.y));                // add cursor to points
```
Anuken: Use `Pools.obtain(Point2.class, Point2::new)` instead of a new point.

core/src/io/anuke/mindustry/input/Placement.java
```diff
@@ -56,6 +56,31 @@
...
+            if(cursor.x == endX && cursor.y == endY) break;            // do not look further than the cursor
+            cursor = cursor.getNearby(cursor.rotation());              // move cursor to next block to check
+        }
+
+        // if the end is nowhere on the line, don't return anything
+        if(!points.contains(new Point2(endX, endY))){
```
Anuken: Same isue with `contains()` and `new Point2(...)` here.

core/src/io/anuke/mindustry/input/InputHandler.java
```diff
@@ -792,13 +792,21 @@ public void drawArrow(Block block, int x, int y, int rotation){
 
     public void drawArrow(Block block, int x, int y, int rotation, boolean valid){
         Draw.color(!valid ? Pal.removeBack : Pal.accentBack);
+        if(Core.input.keyDown(Binding.dash) && valid){
```
Anuken: I would put this into the ternary statement, since it's part of the color assignment: `Core.input.keyDown(Binding.dash) && valid ? Pal.upgradeBack : !valid ? Pal.removeBack : Pal.accentBack`. Similarly for the back color.

core/src/io/anuke/mindustry/input/InputHandler.java
```diff
@@ -809,7 +817,7 @@ public void drawArrow(Block block, int x, int y, int rotation, boolean valid){
     void iterateLine(int startX, int startY, int endX, int endY, Cons<PlaceLine> cons){
         Array<Point2> points;
         boolean diagonal = Core.input.keyDown(Binding.diagonal_placement);
-
+        boolean follower = Core.input.keyDown(Binding.dash);
```
Anuken: Instead of using `dash` (which is a button frequently used for movement), a different keybind (say, U or Z) would be preferable. The exact button doesn't matter too much as long as it's not used.

core/src/io/anuke/mindustry/graphics/Pal.java
```diff
@@ -83,5 +83,8 @@
     redDust = Color.valueOf("ffa480"),
     redderDust = Color.valueOf("ff7b69"),
 
-    plasticSmoke = Color.valueOf("f1e479");
+    plasticSmoke = Color.valueOf("f1e479"),
+
+    upgrade = Color.valueOf("#549de5"),
```
Anuken: How about `6335f8` and `3b2dd1`?

core/src/io/anuke/mindustry/input/InputHandler.java
```diff
@@ -820,6 +828,8 @@ void iterateLine(int startX, int startY, int endX, int endY, Cons<PlaceLine> con
 
         if(diagonal){
             points = Placement.pathfindLine(block != null && block.conveyorPlacement, startX, startY, endX, endY);
+        }else if(follower){
+            points = Placement.followLineUntil(startX, startY, endX, endY);
```
Anuken: What happens if you try to upgrade walls or some other non-conveyor non-conduit like this?

Quezler: oh, then it just follows that line till the end, but it won't go around corners, so not that useful

CA6:
I think it should be editable so you can remove buildings that would be auto upgraded but eh I'm late to the party

Quezler:
closing in favor of #1722

PR #1013, opened 2019-11-07
Display bars for bridge contents

Quezler:
This is just my attempt of [visualizing bridge contents](https://feathub.com/Anuken/Mindustry/+21):

The typical black item slider probably won't work on this thing since its configurable and thus can't handle two gui's properly or something, so i coined this method.

Current issue: `lastDisplay` is cached, ~the icons/count don't update like the bars do.~ got that to work but now it displays in the bottom left corner, but overall that code is a hack :|

Screen Shot 2019-11-07 at 13 05 13
(note: this pull has gone through many iterations, see the `edited v` at the top to view them)

### DIFF
--- a/core/src/io/anuke/mindustry/input/InputHandler.java
+++ b/core/src/io/anuke/mindustry/input/InputHandler.java
@@ -576,7 +576,7 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
         //consume tap event if necessary
         if(tile.interactable(player.getTeam()) && tile.block().consumesTap){
             consumed = true;
-        }else if(tile.interactable(player.getTeam()) && tile.block().synthetic() && !consumed){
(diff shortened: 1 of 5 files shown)


Anuken:
As mentioned on Discord, another approach may be to make inventories editable for configurable blocks. This opens up the possibility of removing clogs, not just viewing them. Try removing the ["!consumed" condition from here](https://github.com/Anuken/Mindustry/blob/master/core/src/io/anuke/mindustry/input/InputHandler.java#L582) and see how well it works.

Quezler:
Hmm, that actually works surprisingly well, and doesn't interfere with the linking one bit 🤔

Anuken:
Closing due to conflicts; if this needs to be implemented, the changes consist of a single line anyway.

PR #1018, opened 2019-11-08
Add TapEvent/TapConfigEvent

Kieaer:
test
Q: Why is this necessary?
A: The destruction of griefers is evolving.
I can no longer catch griefers with existing Events, so i need to add events.
Todo: withdraw event

## Available methods for this event
* Blocking/detect change mass-driver change
* Blocking/detect change Bridge/Phase conveyor/conduit
* Blocking/detect change sorter

### DIFF
--- a/core/src/io/anuke/mindustry/game/EventType.java
+++ b/core/src/io/anuke/mindustry/game/EventType.java
@@ -141,6 +141,30 @@ public class EventType{
             this.player = player;
         }
     }
+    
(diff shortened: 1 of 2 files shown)


Kieaer:
smail problem: When I tap on the floor, I get a null error in tapevent.

Anuken:
@Kieaer If you have a `NullPointer`, figure out what's null and make sure you check for it *not* being null.

PR #1019, opened 2019-11-08
Add WithdrawEvent

Kieaer:
ezgif com-optimize

## Available methods for this event
* Blocking/detect take flammability items and suicide
Griefers can make a big fire and ruin the game.


core/src/io/anuke/mindustry/ui/fragments/BlockInventoryFragment.java
```diff
@@ -100,7 +101,7 @@ private void rebuild(boolean actions){
                         holding = false;
                         holdTime = 0f;
 
-                        Events.fire(new WithdrawEvent());
+                        Events.fire(new WithdrawEvent(tile, player, lastItem, amount));
```
Anuken: This will lead to duplicate withdraw events firing on the client when it is in singleplayer.

PR #1022, opened 2019-11-09
Allow for toggling the visibility of power line lasers

danmw3:
Configurable key-bind in controls and obeys the power line opacity setting.
The key-bind defaults to F7.

ezgif com-resize

### DIFF
--- a/core/src/io/anuke/mindustry/input/DesktopInput.java
+++ b/core/src/io/anuke/mindustry/input/DesktopInput.java
@@ -432,11 +432,9 @@ public class DesktopInput extends InputHandler{
 
         if(Core.input.keyTap(Binding.toggle_power_lines)){
             if(Core.settings.getInt("lasersopacity") == 0){
-                Core.settings.put("lasersopacity", Core.settings.getInt("preferredlaseropacity"));
+                Core.settings.put("lasersopacity", Core.settings.getInt("preferredlaseropacity", 100));
             }else{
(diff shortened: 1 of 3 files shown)


core/src/io/anuke/mindustry/input/DesktopInput.java
```diff
@@ -429,6 +429,17 @@ void pollInput(){
 
             mode = none;
         }
+
+        if(Core.input.keyTap(Binding.toggle_power_lines)){
+            if(Core.settings.getInt("lasersopacity") == 0){
+                Core.settings.put("lasersopacity", Core.settings.getInt("preferredlaseropacity"));
```
Anuken: I would change this to `Core.settings.getInt("preferredlaseropacity", 100)` to account for the slim chance that preferred laser opacity is missing.

core/src/io/anuke/mindustry/input/DesktopInput.java
```diff
@@ -429,6 +429,17 @@ void pollInput(){
...
+            }else{
+                if(!Core.settings.has("preferredlaseropacity")){
```
Anuken: Maybe don't have this condition here, so that the last power opacity gets saved every time you toggle lasers off.

danmw3:
Both changes are done :)

Anuken:
Thanks.

PR #1029, opened 2019-11-10
sand water outputs sand

ThatOneBepis:


YaFou:
Can we more information about this pull request @ThatOneBepis?

ThatOneBepis:
its self-explanatory

WaterRatje:
Do you mean that water can output sand?

If so, how?

YaFou:
But why this change? Why do you want output sand from sand water?

ThatOneBepis:
@WaterRatje yes

@YaFou its dumb to not drill sand in water NEAR SAND WITHOUT WATER (or in general)

WaterRatje:
oh, you mean just with a mining drill...

ThatOneBepis:
yes

Anuken:
Shelving this for now. Drilling sand that's underwater doesn't make a *whole* lot of sense, but the main issue is that this breaks the displayed resources for sectors and may affect balance in those areas. This may be reconsidered for 6.0.

ThatOneBepis:
Understood

Issue #1028, opened 2019-11-10
Stormtrooper aim implementation: Cannot hit immobile 1x1 target


AtomicStryker:
**Platform**: Windows

**Build**: 99.12

**Issue**: I suspect there is a problem around https://github.com/Anuken/Mindustry/blob/4c0607aba8fb94725b5a7a4216fa99ae57c0219f/core/src/io/anuke/mindustry/entities/type/base/GroundUnit.java#L169 Groundunit.behaviour which governs targeting. Without knowing much about your implementation, it looks like "firing when within a certain targeting angle" and "predict enemy movement, even if it is not moving" combined cause the problem.

**Steps to reproduce**: Have an enemy Dagger try to kill a 1x1 target that is mid-to-max range orthogonal to its movement direction. Its aim will always be off by about one tile. See screenshot:

![stormtrooper](https://user-images.githubusercontent.com/1858566/68546074-4f33d080-03d3-11ea-96d7-f6f7f704eebf.PNG)

Anuken:
![image](https://user-images.githubusercontent.com/10100000/68546275-93f34380-03a2-11ea-8d5c-5414926c5f39.png)

![image](https://user-images.githubusercontent.com/10100000/68546217-47a80380-03a2-11ea-8935-288d8a0e9c38.png)

No problems here, across varying distances and different unit movement formations.
I've "fixed" this problem so many times now that I don't really care about it anymore. 

What you're probably experiencing is some side effect of the bullets inheriting unit velocity and delta time calculations not taking that into exact consideration. It may be fixed if I decide to make bullets stop inheriting velocity, but until that time comes this won't be fixed for your edge case.

AtomicStryker:
Out of interest i debug stepped through that targeting algorithm and found immobile targets have a nonzero target velocity. That does not seem correct, unless it means something different in your codebase.

Forcing that to zero for immobile entities did not fix the issue, however.

Personally i wouldn't call something that happens in every map in which you send out drones an edge case but ... well if you fixed it so many times to be sick of it :) certainly know that feeling

AtomicStryker:
To clarify: I noticed the issue because on the Salt Flats level my entire Dagger army ran past Hail turrets taking pot shots multiple times, and all of them failed to hit them.

AtomicStryker:
Bwahahaa i found the crux of the issue: It's related to FPS. If i set the frame limit to a cinematic 30, the same issue happens ... IN THE OPPSITE direction.

[...]

PR #1030, opened 2019-11-10
fix https://github.com/Anuken/Mindustry/issues/1028

AtomicStryker:
aim of bullets being off depending on FPS.

With this change, bullets will no longer miss 1x1 targets when running at FPS other than 60.

### DIFF
--- a/core/src/io/anuke/mindustry/entities/Predict.java
+++ b/core/src/io/anuke/mindustry/entities/Predict.java
@@ -54,8 +54,12 @@ public class Predict{
     /**
      * See {@link #intercept(float, float, float, float, float, float, float)}.
      */
-    public static Vector2 intercept(TargetTrait src, TargetTrait dst, float v){
-        return intercept(src.getX(), src.getY(), dst.getX(), dst.getY(), dst.getTargetVelocityX() - src.getTargetVelocityX()/2f, dst.getTargetVelocityY() - src.getTargetVelocityY()/2f, v);
+    public static Vector2 intercept(TargetTrait src, TargetTrait dst, float v) {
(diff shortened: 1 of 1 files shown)


Anuken:
Have you tested this at varying frame rates? I've already tried several variants of delta time edits here, and none of them fixed it.

AtomicStryker:
Yes. Heres some measurements taken with a debugger, in mostly the same situation:

FPS 30

diffX = -0.8347168
diffY = -0.57470703
timeDelta = 2.036598

FPS 60

diffX = -0.32995605
diffY = -0.3305664
timeDelta = 0.95935196

FPS 120

diffX = -0.13586426
diffY = -0.1352539
timeDelta = 0.491676

There is a very clear correlation of "target velocity" to delta time. Which makes sense if you consider that target velocity is calculated by comparing this ticks position to last ticks, which is a value completely dependant on tickrate / FPS.

Arguably the more correct approach would be to fix velocity to be tick independent but i fear that would have unintended consequences.

AtomicStryker:
before PR, 30 fps
30fps
before PR, 120 fps
120fps

After PR, 120 fps
pr_120fps

Anuken:
How about if a turret tries to hit a moving enemy?

AtomicStryker:
I never noticed this bug regarding moving entities. Any idea why immobile entities have velocities in this algorithm?

A quick test with Salvos on 120fps looks like pinpoint center mass accuracy.

This makes a case for a comprehensive "aim verification map" however...

Anuken:
They don't. Bullets inherit unit firing velocity, which is why that needs to be accounted for.

Anuken:
Implemented slightly differently.

PR #1033, opened 2019-11-10
Big changes. "..." →…

Prosta4okua:


Anuken:
Why?

Prosta4okua:
I saw this in the recommendations for translators. Why not replace it in the original file?

Arkanic:
unicode ... rather than . . . ?

Anuken:
My concern is that it's harder to type and doesn't really solve any problems.

PR #810, opened 2019-10-10
[WIP] Desktop mouse control

clfer:
Hi Anuken,

I started to work on a new control style for desktop  more inspired by RTS.

Here is the first demo
_Click control:_
capture mouse demo - click

_Sidescreen scroll:_
capture mouse demo - mouvement

Are you interested in a feature like this?
Should I continue working on this?

Have a good day

PR #1037, opened 2019-11-11
Add mouse-based movement

joshuaptfan:
Turns the mouse into a pseudo-joystick, with the center of the game window as the axis, while a key is held down. In use, it feels like the ship is following the cursor. An alternative to #810, requiring less mouse movement.

Requires [Arc PR #5](https://github.com/Anuken/Arc/pull/5) for the default keybind, mouse back button, to function.

mouse movement demonstration

## Added
* 1 new keybind


core/assets/bundles/bundle.properties
```diff
@@ -660,6 +660,7 @@ keybind.press.axis = Press an axis or key...
 keybind.screenshot.name = Map Screenshot
 keybind.move_x.name = Move x
 keybind.move_y.name = Move y
+keybind.mouse_move.name = Mouse Move
```
Anuken: "Follow Mouse" would be more appropriate.

joshuaptfan: Done.

PR #1048, opened 2019-11-12
Visualize which mass drivers shoot at the selected one

Quezler:
https://feathub.com/Anuken/Mindustry/+238
Screen Shot 2019-11-12 at 10 35 57

### DIFF
--- a/core/src/io/anuke/mindustry/world/blocks/distribution/MassDriver.java
+++ b/core/src/io/anuke/mindustry/world/blocks/distribution/MassDriver.java
@@ -181,6 +181,11 @@ public class MassDriver extends Block{
 
         MassDriverEntity entity = tile.entity();
 
+        entity.waitingShooters.forEach(shooter -> {
+            Drawf.circles(shooter.drawx(), shooter.drawy(), (tile.block().size / 2f + 1) * tilesize + sin - 2f, Pal.place);
+            Drawf.arrow(shooter.drawx(), shooter.drawy(), tile.drawx(), tile.drawy(), size * tilesize + sin, 4f + sin);
+        });
+
         if(linkValid(tile)){
             Tile target = world.tile(entity.link);
             Drawf.circles(target.drawx(), target.drawy(), (target.block().size / 2f + 1) * tilesize + sin - 2f, Pal.place);


Violet-Scarelli:
Quez you mad genius. Builders against driver spaghetti, unite!

core/src/io/anuke/mindustry/world/blocks/distribution/MassDriver.java
```diff
@@ -181,6 +181,11 @@ public void drawConfigure(Tile tile){
 
         MassDriverEntity entity = tile.entity();
 
+        entity.waitingShooters.forEach(shooter -> {
```
Anuken: forEach isn't available on Android. Use a standard for-loop. In general, standard for-loops are preferable unless you are doing a one-liner (e.g. `array.each(e -> e.x = y)` would be a valid use case.)

core/src/io/anuke/mindustry/world/blocks/distribution/MassDriver.java
```diff
@@ -181,6 +181,11 @@ public void drawConfigure(Tile tile){
...
+            Drawf.circles(shooter.drawx(), shooter.drawy(), (tile.block().size / 2f + 1) * tilesize + sin - 2f, Pal.place);
+            Drawf.arrow(shooter.drawx(), shooter.drawy(), tile.drawx(), tile.drawy(), size * tilesize + sin, 4f + sin);
```
Anuken: This should be colored `Pal.place` to match the selection circle. This would require making `Drawf#arrow` have an extra color parameter, `color`, and making a new overload of `Drawf#arrow` that calls the new `Drawf#arrow` method with `Pal.accent` as the default color.

core/src/io/anuke/mindustry/world/blocks/distribution/MassDriver.java
Quezler: really? so much for java being cross platform, i'll get this fixed 😋

core/src/io/anuke/mindustry/world/blocks/distribution/MassDriver.java
Quezler: Noted, did cross my mind but i was unsure to the reasoning that the arrow color was hardcoded so hence the draft pull for comments 👍

core/src/io/anuke/mindustry/world/blocks/distribution/MassDriver.java
Anuken: "Write once, run anywhere" is basically a lie; once your program becomes more complex than a simple command line reader you have to start dealing with many platform-specific issues.

Quezler: Well personally i would expect to run into those kind of things when trying to make the device vibrate (sounds wrong, but you get the point), not with something as "simple" as a foreach :D

Arkanic: curious, isn't it? you'd think that android would come with the standard set of java se commands, seeing as the android api is built on the language.

Anuken: It might have been built on the same foundation, but many devices are old and don't support these ""new"" Java 8 features. It's not that forEach is completely unsupported; it just requires a newer API level than Mindustry is supposed to support.

PR #1049, opened 2019-11-12
Keep items on conveyor when upgrading

Quezler:
https://feathub.com/Anuken/Mindustry/+234
https://feathub.com/Anuken/Mindustry/+206

This is a just technical proof of concept, do not take this way to serious.

It carries over conveyor contents when the previous and next building block are both conveyors.

This is useful when lets say upgrading an incoming belt of titanium heading into your core without slowing your progress since you're killing the required titanium needed for upgrading while you're upgrading.

But it also has a flip side, often when you want to get rid of items on a conveyor (e.g. `sand/coal/scrap` mixed in) its common practice to just up/down grade the conveyor to get rid of those items, this pull would prevent that, hence the draft / technical proof of concept.

A different approach would be to send all the items (that are capable of entering the core) directly to the core when destroying a conveyor/block containing those items, similar to the "dropping of" of a player / mining drone inventory.

But since that might feel like teleportation somewhat it would maybe make sense to grab the items from the belt and put them into the inventory of the player, but that leaves edge cases for when multiple items are on the belt, since units can only carry one (afaik), and of course the carry limit comes into play.

I'd love to collect some thoughts on this topic 📥 

Nov-12-2019 11-24-11

### DIFF
--- a/core/src/io/anuke/mindustry/world/Build.java
+++ b/core/src/io/anuke/mindustry/world/Build.java
@@ -4,6 +4,7 @@ import io.anuke.annotations.Annotations.Loc;
 import io.anuke.annotations.Annotations.Remote;
 import io.anuke.arc.Core;
 import io.anuke.arc.Events;
+import io.anuke.arc.collection.LongArray;
 import io.anuke.arc.math.Mathf;
 import io.anuke.arc.math.geom.*;
 import io.anuke.mindustry.content.Blocks;
@@ -12,6 +13,7 @@ import io.anuke.mindustry.game.EventType.BlockBuildBeginEvent;
 import io.anuke.mindustry.game.Team;
 import io.anuke.mindustry.world.blocks.BuildBlock;
 import io.anuke.mindustry.world.blocks.BuildBlock.BuildEntity;
+import io.anuke.mindustry.world.blocks.distribution.Conveyor;
 
 import static io.anuke.mindustry.Vars.*;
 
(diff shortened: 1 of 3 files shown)


willis936:
I like this idea.  Handling rogue coal/scrap/sand is what the sorter/incinerator combo is for.  It’s satisfying when its implemented as a band aid because your conveyor is too long to reasonably cherry pick.  This change would make the incinerator more attractive during regular use.

Quezler:
You're absolutely right @willis936 🤔

Quezler:
Oh @**Anuken**, in regards to [this trello card](https://trello.com/c/KYQ3hFu8/248-optimize-conveyors), don't forget about sections of a conveyor snake being affected by overdrive projectors 😗 (not implying that you would have forgotten it or something, but seeing the discord gives the impression that you're diving deep into the optimization, so this is just a subtle heads up in case you overlooked this part)

Anuken:
I'm just not going to be implementing that, the system doesn't really support it.

Quezler:
Noted 👍

gm907:
"..send all the items (that are capable of entering the core) directly to the core when destroying a conveyor/block containing those items, similar to the "dropping of" of a player / mining drone inventory."
This.

Anuken:
I may consider implementing this later, but at the moment I don't see a pressing need for it. The code here isn't compatible with the new conveyors.

PR #1050, opened 2019-11-12
Skip to next building in queue when you don't have enough resources 

Quezler:
Proof of concept, button to skip to the next build in the queue when the progress does not increase (aka: out of building resources), instead of majestically trying to fly away out of range, so you can at least build the rest instead having to wait for the resources for that one block.

Nov-12-2019 12-16-47

(no idea why the button is red though)


Fayti1703:
Putting the input handling in `BuilderTrait` might cause some issues with Phantoms.

Quezler:
Thx for that heads up @Fayti1703 👍

joshuaptfan:
Why even have a keybind for it? Why not automatically skip unaffordable blocks?

iczero:
A few suggestions:
* Make keybind name `skip_unaffordable` instead of `afford_building` to clear up ambiguity
* Move input handling into DesktopInput so only the player responds to the skip, instead of every BuilderTrait entity
* Add a RPC call (Call.skipBuildingUnaffordable maybe?) to notify the server of the action so desyncs don't occur between the client and the server
  * I think `@Remote(target=Loc.both, called=Loc.server, forward=true)` on a static method in InputHandler should do it

iczero:
Actually correction: given that it just skips the current thing being built might as well rename it to "skip current build", the keybind to `skip_build` or something, and the RPC to `Call.onSkipCurrentBuild`

iczero:
On second thought it might be a better idea to put the remote call in Build

iczero:
Correction: It should probably be `@Remote(target = Loc.both, called = Loc.both, forward = true)` since the thing needs to be called on both sides to sync properly

iczero:
```diff
diff --git a/core/src/io/anuke/mindustry/input/Binding.java b/core/src/io/anuke/mindustry/input/Binding.java
index 8e5ca2bf6..83187098a 100644
--- a/core/src/io/anuke/mindustry/input/Binding.java
+++ b/core/src/io/anuke/mindustry/input/Binding.java
@@ -14,6 +14,7 @@ public enum Binding implements KeyBind{
     break_block(KeyCode.MOUSE_RIGHT),
     clear_building(KeyCode.Q),
     pause_building(KeyCode.E),
+    skip_building(KeyCode.R),
     rotate(new Axis(KeyCode.SCROLL)),
     rotateplaced(KeyCode.R),
     diagonal_placement(KeyCode.CONTROL_LEFT),
diff --git a/core/src/io/anuke/mindustry/input/DesktopInput.java b/core/src/io/anuke/mindustry/input/DesktopInput.java
index e3af574cf..8f467eb46 100644
--- a/core/src/io/anuke/mindustry/input/DesktopInput.java
+++ b/core/src/io/anuke/mindustry/input/DesktopInput.java
@@ -438,6 +438,10 @@ public class DesktopInput extends InputHandler{
                 Core.settings.put("lasersopacity", 0);
             }
         }
+
+        if(Core.input.keyTap(Binding.skip_building)){
+            Call.onSkipCurrentBuild(player);
+        }
     }
 
     @Override
diff --git a/core/src/io/anuke/mindustry/input/InputHandler.java b/core/src/io/anuke/mindustry/input/InputHandler.java
index 32e841bdf..99a1e9f87 100644
--- a/core/src/io/anuke/mindustry/input/InputHandler.java
+++ b/core/src/io/anuke/mindustry/input/InputHandler.java
@@ -3,6 +3,7 @@ package io.anuke.mindustry.input;
 import io.anuke.annotations.Annotations.*;
 import io.anuke.arc.*;
 import io.anuke.arc.collection.*;
+import io.anuke.arc.collection.Queue;
 import io.anuke.arc.func.*;
 import io.anuke.arc.graphics.*;
 import io.anuke.arc.graphics.g2d.*;
@@ -160,6 +161,13 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
         Core.app.post(() -> Events.fire(new TapConfigEvent(tile, player, value)));
     }
 
+    @Remote(targets = Loc.both, called = Loc.both, forward = true)
+    public static void onSkipCurrentBuild(Player player) {
+        Queue<BuildRequest> queue = player.buildQueue();
+        BuildRequest current = queue.removeFirst();
+        queue.addLast(current);
+    }
+
     public Eachable<BuildRequest> allRequests(){
         return cons -> {
             for(BuildRequest request : player.buildQueue()) cons.get(request);
```

Quezler:
@joshuaptfan well maybe someone is building a reactor or complicated setup where the build order is vital for some reason 🤔

Quezler:
@iczero actually it skips while you hold it down, not just once 😅

Quezler:
Implemented your diff, skipping to the next block in the queue makes more sense in the long run than skipping all of them in a big continuous loop 👍

iczero:
Oh. I think it's Core.input.keyDown?

Quezler:
Well correct me if im wrong but that will likely spam the connection, a tap would likely suffice in most cases, though i may be wrong.

iczero:
Oh yeah, true.

iczero:
@Quezler oh another thing, might want to check if queue is empty before firing off the rpc with Call.onSkipCurrentBuild

joshuaptfan:
@Quezler Schematics do not respect any predefined build order, they simply build bottom-to-top, left-to-right in the default orientation (rotating that if the schematic is rotated). The game only respects build order for manually-placed blocks, but even then, only sometimes (it does its best only for blocks in range).

Anuken:
recording1573619991
recording1573620006

Decided to implement something more like @joshuaptfan suggested myself in d8c997b355e3fefd58a3077e2fb89b48e90f15b6. Build requests that are "stuck" (e.g. no progress last frame + items are missing) will be skipped over, if possible. Request "ordering" code has also been cleaned up, and old requests out of range will no longer slow down anything you currently have queued.

Quezler:
> and old requests out of range will no longer slow down anything you currently have queued.

was about to suggest/pull it after this pull was finished, glad it got implemented! 🥇

matheod:
> Why even have a keybind for it? Why not automatically put unaffordable blocks at the end of the build queue?

If you are low on ressource and want something as soon as possible (for exemple a turret), while still being able to plan other building. You wouldn't want to other things to start building and using ressources you wanted to use for the thing you wanted to build as soon as possible.

PR #1052, opened 2019-11-13
Colorize powernodes based on network status instead of its powerlines

Quezler:
https://feathub.com/Anuken/Mindustry/+231

This pull is a draft for showing the status of the power graph visually on each node instead of their web of connections.

- idle/unconnected/texture/etc: `default/null`
- making more power than consuming: `green`
- using power from batteries: `orange`
- using more power than is being produced: `red`

Colors obviously need some work, way to bright, and the surge tower looks strange now.
(and needless to say `red` could be anywhere from no power available to almost all power available, it currently does not majestically fade between colors based on the exact satisfaction)

The above feathub link suggested leds, but instead of additional block(s) it would be more logical to visually modify the power nodes.

These changes should communicate the status of the power network more cleanly to the players:

- `green` all is fine
- `orange` check on power production
- `red` make more power available

Again, this is a technical proof of concept / draft / yadayada, its quality at this stage is nowhere near mergable, code quality is mediocre at best and the graphics look awful, this is just to get the ball rolling.

Nov-13-2019 11-44-24


core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -258,6 +260,15 @@ public void drawPlace(int x, int y, int rotation, boolean valid){
         Draw.reset();
     }
 
+    @Override
+    public void draw(Tile tile) {
+        super.draw(tile);
+        if(tile.entity.power.graph.getPowerNodeColor() == null || tile.block() instanceof PowerSource) return;
```
Anuken: `PowerSource` could override this method and draw the standard tile instead of having a special check here.

core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
```diff
@@ -26,6 +28,8 @@
     private final int graphID;
     private static int lastGraphID;
 
+    private Color led = null;
```
Anuken: Instead of just having a single color, how about:
- an enum, `PowerState`, which would be `insufficient` / `producing` / `consuming` / `none` (or something similar). Each enum would have a `Color` assigned in its constructor. The empty state would have `Color.clear` (**not** null).
- assign the state each frame instead of the color.
- keep the color, but name it `statusColor` or something and `statusColor.lerp(status.color, 0.1f * Time.delta())` every frame so it changes to the current status color.
- add a getter for the status for possible future applications

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -258,6 +260,15 @@ public void drawPlace(int x, int y, int rotation, boolean valid){
...
+        Draw.color(tile.entity.power.graph.getPowerNodeColor());
+        Draw.rect(led, tile.drawx(), tile.drawy(), (id%4)*90);
```
Anuken: What is `(id%4)*90` for?

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -328,9 +339,7 @@ protected void drawLaser(Tile tile, Tile target){
         x2 += t2.x;
         y2 += t2.y;
 
-        float fract = 1f - tile.entity.power.graph.getSatisfaction();
-
-        Draw.color(Color.white, Pal.powerLight, fract * 0.86f + Mathf.absin(3f, 0.1f));
```
Anuken: Displaying power status with lines could still be helpful. Is it a problem?

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
Quezler: i don't know, the code where i copied that from had it 😂 (cracks iirc)

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
Quezler: no, would actually make sence to keep it in 🤔

core/src/io/anuke/mindustry/world/blocks/power/PowerNode.java
Quezler: I'll try to move it there.

RedDragonMZ:
It can be an option on game settings

sk7725:
Why not change the laser color instead of its block color? The yellow palette is standard for power blocks, and by coloring lasers you can also point out which block is being greedy(thats not how nodes work rn, but...)

Anuken:
With the block status PR and the existing laser light coloration, I don't think this is really necessary anymore. The idea is simple to enough to implement later if I think it's worth revisiting.

PR #1053, opened 2019-11-13
Make chat show/hide setting vaiable

Kieaer:
I've thought of other ways, but nothing like that.

Usage
``NetClient.visual = true/false;``
If you want to show chat
```
Events.on(PlayerChatEvent.class, e -> {
    Call.sendMessage(e.message, colorizeName(e.player.id, e.player.name), e.player);
});
```

If you want bad words filtering
```
Events.on(PlayerChatEvent.class, e -> {
    if(!e.message.equals("where's your mom?"){
        Call.sendMessage(e.message, colorizeName(e.player.id, e.player.name), e.player);
    }
});
```

This can be a bad thing, but it's a must.

### DIFF
--- a/core/src/io/anuke/mindustry/core/NetClient.java
+++ b/core/src/io/anuke/mindustry/core/NetClient.java
@@ -49,6 +49,8 @@ public class NetClient implements ApplicationListener{
     private float timeoutTime = 0f;
     /** Last sent client snapshot ID. */
     private int lastSent;
+    /** Chat show/hide setting */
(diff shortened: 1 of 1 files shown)


Anuken:
Why not set chat visibility to 0% if you don't want to see it?

Kieaer:
No no. This's used by plugins and mods.

Kieaer:
ezgif com-video-to-gif

Kieaer:
I came up with a better way than this.

Issue #1036, opened 2019-11-11
Warn if any core is under attack


Quezler:
[`core is under attack!`](https://github.com/Anuken/Mindustry/blob/master/core/src/io/anuke/mindustry/ui/fragments/HudFragment.java#L301) only takes the first core into account, it would make sense if it would warn about all allied cores a map might have.

itskatt:
With an indicator pointing towards the attacked core.

Anuken:
Should be posted in the feature request page.

Quezler:
on*

ZkyLB:
nice

PR #1058, opened 2019-11-14
Warn if any core is under attack, the pull

Quezler:
I wasn't sure on how to best implement this, hence i originally made https://github.com/Anuken/Mindustry/issues/1036 as an issue without code, but now here we are: this pull takes all allied cores into account, its primary caveat is that technically if you repair one of your cores while another is under attack you won't be notified, but its arguably better than only checking the first core 🐡


Anuken:
This made me realize how hacky my core health checking code is.

Quezler:
Sorry 😅

Quezler:
Closing this pull since it appears @**Anuken** is on the path to commit something similar any moment from now, since the commits leading up are 1a26a1b & 230428f

Anuken:
Very slowly implementing something on my phone.

Quezler:
Well if you rather have me try to make a pull of a version that runs on your event system as shown in your recent commits, do let me know, at a desktop rn 😃

PR #1059, opened 2019-11-14
Core: seed

Quezler:
Screen Shot 2019-11-14 at 12 07 35
Screen Shot 2019-11-14 at 15 39 12

a buildable core block:

- cannot interact with tiles (`conveyors/unloaders/vaults/etc`)
- enemies do not go after this block (aka: not beeline to this core)
- will not count towards alive cores (implementation in pull inadequate)

intended usage:

- allows players to drop off or take from the core manually without being near an actual core (or just see the total available building resources)
- mining drones will drop off here as well, keeping the airspace around the real core(s) free
- can be used as an early warning system if a defence is breached, since it flashes the hud when damaged (depends on https://github.com/Anuken/Mindustry/pull/1058)

related:
- https://feathub.com/Anuken/Mindustry/+207
- https://feathub.com/Anuken/Mindustry/+30
- https://github.com/Anuken/Mindustry/pull/955

todo:

maybe refractor it away from `coreblock` so the enemy units / gameover logic / etc would not need an additional check, preventing player respawns, etc, though still keeping the properties mentioned above ^ (and probably turret target priority or something so increase the chance of it being hit)

trivia:

- originally planned to call it `core-fragment`, but since i used the bush as a base texture (round to signify that it should not connect to other blocks) `core-seed` sounded more appropriate  

some screenshots illustrating the above:

Screen Shot 2019-11-14 at 12 26 05
Screen Shot 2019-11-14 at 12 26 44
Screen Shot 2019-11-14 at 12 27 34


joshuaptfan:
Gameplay issues I can see with this:

* Griefers will build these into the front line defensive wall to give everyone false "Core under attack" warnings
* Checking core resources is currently inconvenient, but that should be addressed with a real solution that lets you check from anywhere, not just at more locations, like perhaps displaying it on pressing Tab (player list) key.

Quezler:
- true, good point, but people would likely catch the culprit sooner or later, and having a warning somehow when a defence is being breached should still be a welcome feature []
- yea well the original purpose of this block was just for the mentioned alert, but anuke isn't that fond of single purpose blocks (afaik) (e.g. original plastanium wall), so i threw a resource check in there since, well, its a core :)

Violet-Scarelli:
Oh I love this a lot. Now if only we could pair this with an improved mining drone and we could just put these near all the ore beds.

Quezler:
We should take into account that map makers might intentionally have certain ores far away from the core so drones/conveyors need to haul over long distances, a core seed would bypass all that.

Violet-Scarelli:
Also, if we're on the topic of cores/mining drone pathing, what, do you think, is the likelihood of either a 2x2 core (Core: Sprout, functional with conveyors and unloaders but has less than half the health of the core Shard, and is targeted automatically) or an improved mining drone?

Quezler:
Improved drones are already [planned](https://trello.com/c/zEP6l9Tv/556-t2-drones), and as for the core you are describing, sounds exactly like a normal one, just smaller?

Violet-Scarelli:
Oof yay me for not paying attention

Anuken:
I don't think cores should get any smaller than 3x3;

> allows players to drop off or take from the core manually without being near an actual core (or just see the total available building resources)

This should be implemented by #955 (or something similar)

> mining drones will drop off here as well, keeping the airspace around the real core(s) free

This should be resolved by dedicated drone dropoff points, or #1330.

PR #1064, opened 2019-11-15
Add info about Repair Point can heal mechs

XEN0PHIL:


Quezler:
Well _technically_ a mech falls under the unit umbrella, but i get your point 🤔

joshuaptfan:
Players can take the form of either mechs or ships, and they are not the same.

Anuken:
I think the term 'unit' should be enough to cover players and non-players.

PR #1073, opened 2019-11-16
Schematics size

ior68559:
I think it would be better to increase or make infinite size of schematics. First it would be easier to create big pixelarts. Second can create bigger industries like production line of anything.


Anuken:
1) This is the wrong commit. You just merged an unrelated branch.
2) Schematic size is limited for a good reason: Previews eat up a lot of memory, and the clipboard is unoptimized and causes significant lag when many blocks are in it. Even 32x32 is pushing the limit with a full selection.

PR #1026, opened 2019-11-10
Yes, of course! The Holy Hand Grenade of Antioch!

Quezler:
Suicide bombing 60 blast compound from a distance, with a thorium reactor explosion thrown in.

Screen Shot 2019-11-10 at 10 28 28
Nov-10-2019 10-32-30

(obviously this is a joke, gonna close the pull right after, leaving it here for those looking for a laugh)

PR #1080, opened 2019-11-17
[anti-air] tractor beam 🦈

Quezler:
> You know, I have one simple request. And that is to have sharks with frickin' laser beams attached to their heads! Now evidently my cycloptic colleague informs me that that cannot be done. Ah, would you remind me what I pay you people for, honestly? Throw me a bone here! What do we have?

Many players are occasionally requesting a early/mid game electric anti air turret (which admittedly would be mildly op)

This commit on the other hand adds an electric anti air tractor beam, which can be used to reel in enemy air units to your turrets instead of having to wall your base off completely. 

Having air units dragged in allows anti air turrets with less range to be more useful in late games, as well as not having to anti-air every nook and cranny of your base, but just a few _venus fly traps_ (tractor beam + turret(s) group) throughout your base.

The tractor beam uses a lot of power (when active) and in return adds velocity to the closest enemy air unit to drag them in, though its not high enough to be inescapable.

Effects stack.

Please observe the following example, with the laser set to friendly fire my glaive:
Nov-17-2019 21-00-47

known (potential) issues:
- when unit is right above the tractor beam the light flickers 
- not sure if boss/bigger units are dragged in proportionally to their size & mass
- probably more


Anuken:
Without looking at the code, I'll just mention that this falls into the same category as the silicon furnace (e.g. will be merged/reviewed later) and should probably be a larger block than it currently is (e.g. 2x2/3x3) to discourage spam.

Quezler:
I was fully aware of that fact, hence why it's in draft mode, and as for the latter, I'll take it into account :wink:

Quezler:
And i should note that I'm just throwing science at the wall to see what sticks, no rush whatsoever, just materializing ideas into existence, with the hope that it inspires others 😎

Arkanic:
pulling drones into scatter array

Luxray5474:
Quezler

# You are an absolute legend.

This is the greatest and most fun idea that's ever appeared on this repo in a looong time :)

To try to answer Anuken's statements, I have a few suggestions and questions:

* is the reason you got slower as you were tracked by the beam in the GIF because you stopped pressing movement keys?
* the building should be 3x3 to encourage building defense *around* the turret rather than building *the turret* around the defense. Your 'venus fly trap' idea is brilliant, and the 3x3 size would really encourage that. 
* the beam should be a little longer but track less faster. Probably enough to make it **only** drift a wraith, but still be enough to maybe pull a revenant closer a bit slowly. AI should then react to getting pulled by moving away but also in the direction it's going. That might introduce a new strategy of distracting enemies and slowing them down. 
* the beam should have an area-of-effect, like many tractor beams in movies and games, which pull not just one object, but multiple. That would counter the slow pull speed and make this very useful against swarms on the other hand, by pulling a cluster of enemies instead of just one. However the more enemies it's pulling, the more power it takes up.
* following with that 'pulling multiple things' mechanic, the beam's countenance should be in the likeness of, think the portal tractor beam: image ...and would get wider the more enemies it pulls in. That would probably be really hard to implement though... :/

that's it for now kthxbai

Quezler:
Are you insulting #1026 ? 😏

- yes, movement towards the tractor beam is currently linear, any other movement was me illustrating units can still escape somewhat.
- 3x3 feels obnoxiously large, but even for 2x2 I'll need to retexture it, it's currently also a surprise for me what my next step will be.
- interesting idea, like propper cat and mouse, I'll give it some thought.
- it currently pulls the closest (the more you have the stronger the pull), but it's single target since otherwise it might be rather overpowered to take out all air units in a wave with one shot.
- oh wow the aperture funnel is actually a pretty nice idea & graphic, the idea of a helix around it did cross my mind once (but to hard to implement at that time), but it refreshed my thoughts reading it.

Quezler:
Sort of demo gif based on the latest changes, not much of it is "visible" (due to the resolution, size & most changes being in the code), but it acts way smoother now than before 🙂 
Dec-24-2019 13-53-26
> oh, and laser strength now visually changes based on distance, notice the width & pull speed

Quezler:
size `2` version, not which base & category it should use though.

Screen Shot 2019-12-24 at 14 06 19

CA6:
this is interesting but a air nullification field would be awesome too as to would simply  "knockout" airborne  or in real talk just they would have a strong aversion to getting near it so kinda the reverse of combustion spam; however this seems more balanced and reasonable but if you introduce tractor beams ima gonna ask for aq ground equivalent of spreaders but thats a diffrent topic.

Quezler:
The code in here can easily target ground units as well if need be, what did you have in mind for spreader though?

CA6:
thats more of a fast intermadte of air and ground but thats its own idea

Quezler:
Well i'd love to hear it nontheless, chances are i'll even (try to) code it in :)

CA6:
well it is low heath i high speed can travel over water and oil however can't go though solid blocks also light weapons cheap and aiglie but bi no means a assault vehicle

Quezler:
Recently tested this in multiplayer, and it somehow desyncs when the target dies, client remains showing the lasers, cause: unknown 🤔

Quezler:
This one will likely be updated once the unit logic in `6.0` is deemed ready to be build upon 👾

Anuken:
Implemented in cecb5cf4b4896786895ff3a1ba9925ce3782d1c9

Quezler:
neat ❤️

PR #1084, opened 2019-11-18
Wololoo

Quezler:
leaving this here for those that are worthy ❤️ 

https://discordapp.com/channels/391020510269669376/396416151032299521/645975743196823562


Arkanic:
this is great, but sad how little atention it got

@Quezler

PR #1093, opened 2019-11-19
Educate players on tiles where a block is more efficient

Quezler:
As always this is just a draft to get the ball rolling, code quality isn't good enough to be merged.

Screen Shot 2019-11-19 at 10 47 49
Screen Shot 2019-11-19 at 10 48 11
Screen Shot 2019-11-19 at 10 48 03
Screen Shot 2019-11-19 at 10 47 56


core/src/io/anuke/mindustry/world/Block.java
```diff
@@ -537,6 +539,16 @@ public void setStats(){
         // Note: Power stats are added by the consumers.
         if(hasLiquids) stats.add(BlockStat.liquidCapacity, liquidCapacity, StatUnit.liquidUnits);
         if(hasItems) stats.add(BlockStat.itemCapacity, itemCapacity, StatUnit.items);
+
+        if(this.attribute != null) { // fixme: there is probably a better way for this v
+            for (Content block : Vars.content.getContentMap()[ContentType.block.ordinal()]) {
```
Anuken: Use `content.blocks()` for this and the Block type instead of Content

core/src/io/anuke/mindustry/world/Block.java
```diff
@@ -537,6 +539,16 @@ public void setStats(){
...
+                if (!(block instanceof Floor)) continue;
```
Anuken: isFloor() / asFloor() would be cleaner checks here.

core/src/io/anuke/mindustry/world/Block.java
```diff
@@ -537,6 +539,16 @@ public void setStats(){
...
+                Floor floor = (Floor) block;
+                float modifier = floor.attributes.get(this.attribute);
+                if (modifier == 0f) continue;
```
Anuken: I would use `Mathf.zero` here to prevent floating point direct comparisons

core/assets/bundles/bundle.properties
```diff
@@ -514,6 +514,7 @@ error.crashtitle = An error has occured
 blocks.input = Input
 blocks.output = Output
 blocks.booster = Booster
+blocks.attributes = Faster on
```
Anuken: "More efficient on"

Arkanic:
I can't find it in the commits list, so ill just put it here.
couldn't ```category.optional``` be reduced to "enhancements" because magama isn't optional for the thermal generator, which could lead to some confusion.

Quezler:
True, but i decided to add it in nontheless since it uses the same attribute block stat

Arkanic:
that makes sense
i'll think of a way to try and fix that

jinkside:
I'm new, but doesn't the thermal generator _only_ work on magma and such? It seems like there's some overlap here between "faster on..." and "placeable on...". The latter is listed for drills, but not thermal generators, which led to a lot of confusion for me. I'm a mobile player though, so I don't know how that relates.

Quezler:
well technically the magma tiles do increase the efficiency of the thermal generator, it just happens that its default efficiency is 0% and it can only be places on said tiles 😗

PR #1096, opened 2019-11-19
Added some Events

blackassasin1234:
adds PlayerBanEvent, PlayerUnbanEvent, PlayerIpBanEvent, and PlayerIpUnbanEvent..

### DIFF
--- a/core/src/io/anuke/mindustry/game/EventType.java
+++ b/core/src/io/anuke/mindustry/game/EventType.java
@@ -353,7 +353,7 @@ public class EventType{
         public final String ip;
 
 
-        public PlayerIpBanEvent(String ip) 
+        public PlayerIpBanEvent(String ip){
             this.ip = ip;
         }
     }
--- a/core/src/io/anuke/mindustry/game/EventType.java
+++ b/core/src/io/anuke/mindustry/game/EventType.java
@@ -366,5 +366,6 @@ public class EventType{
             this.ip = ip;
         }
     }
+    
 }
 
--- a/core/src/io/anuke/mindustry/game/EventType.java
+++ b/core/src/io/anuke/mindustry/game/EventType.java
@@ -332,33 +332,37 @@ public class EventType{
             this.player = player;
         }
     }
+    
     public static class PlayerBanEvent{
         public final Player player;
 
-        public PlayerBanEvent(Player player) {
+        public PlayerBanEvent(Player player){
             this.player = player;
         }
     }
+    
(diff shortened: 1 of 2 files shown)


core/src/io/anuke/mindustry/game/EventType.java
```diff
@@ -327,11 +327,40 @@ public PlayerConnect(Player player){
...
+
         public PlayerLeave(Player player){
             this.player = player;
         }
     }
-           
+    public static class PlayerBanEvent{
+        public final Player player;
+
+        public PlayerBanEvent(Player player) {
```
Anuken: Unnecessary space, use the formatting guidelines. Happens for every method here.

core/src/io/anuke/mindustry/game/EventType.java
```diff
@@ -327,11 +327,40 @@ public PlayerConnect(Player player){
...
+            this.player = player;
+        }
+    }
```
Anuken: Add a blank line after each class.

core/src/io/anuke/mindustry/net/Administration.java
```diff
@@ -126,7 +131,7 @@ public boolean unbanPlayerID(String id){
         info.banned = false;
         bannedIPs.removeAll(info.ips, false);
         save();
-
+        Events.fire(new PlayerUnbanEvent(Vars.playerGroup.find(p -> p.uuid.equals(id))));
```
Anuken: `player.uuid` can be null, causing a crash here. The condition should be inverted.

core/src/io/anuke/mindustry/net/Administration.java
```diff
@@ -88,7 +91,7 @@ public boolean banPlayerID(String id){
         getCreateInfo(id).banned = true;
 
         save();
-
+        Events.fire(new PlayerBanEvent(Vars.playerGroup.find(p -> p.uuid.equals(id))));
```
Anuken: Same issue with uuid.

core/src/io/anuke/mindustry/net/Administration.java
blackassasin1234: What do you mean by inverting the condition?

Anuken: `id.equals(p.uuid)` instead of `p.uuid.equals(id)`

blackassasin1234:
Formatted the code and used 
id.equals(p.uuid) instead of p.uuid.equals(id)

PR #1105, opened 2019-11-20
RockTheVote for server

Arkanic:
probably borked in about 50 ways

I need some help with the variables, they seem to be too global
i'll probably move them into the event

is < Player > necessary for non-plugin commands in this case? I think it's good to make sure things are working, but it might not be necessary in this case.

### DIFF
--- a/server/src/io/anuke/mindustry/server/ServerControl.java
+++ b/server/src/io/anuke/mindustry/server/ServerControl.java
@@ -33,7 +33,6 @@ import java.net.*;
 import java.time.*;
 import java.time.format.*;
 import java.util.*;
-import java.util.HashSet;
 
 import static io.anuke.arc.util.Log.*;
 import static io.anuke.mindustry.Vars.*;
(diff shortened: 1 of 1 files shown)


Arkanic:
is half a good amount or does it need a increase?

Anuken:
Read the [contributing guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md), you've violated a few of them here.

Arkanic:
i felt like i did

Arkanic:
i don't have a way of testing the commit though

Arkanic:
@Anuken what would i have to do to remove the boxing? i'm honestly not too sure what i'd do

Anuken:
You don't have any boxing there. I was mostly referring to how it doesn't compile (or, if it did, it wouldn't work)

Arkanic:
i'm assuming that was with HashSet rather than ObjectSet
i forgot to import it

Arkanic:
@Anuken does ```ObjectSet``` have a size or length function?

Anuken:
1. No, but it does have a variable.
2. Are you using IntelliJ? If not, get it. You won't need to ask questions like that with autocomplete around.

Arkanic:
yeah, i'm kinda giving up on eclipse with it's caches that don't auto update

Arkanic:
i'll fix this later, just becoming stale right now

PR #1112, opened 2019-11-22
Update bundle_de.properties

SosaPlays:
Added translations for many longer bits of text.


martin-mfg:
These translations aren't proper German. They very much look like automated translations, with all the errors that are to be expected from that approach.

DeltaNedas:
google translate bad
at least steal proper (?) translations like zac (citation needed)

PR #1124, opened 2019-11-24
Add storable indication in core database

matheod:
image

https://feathub.com/Anuken/Mindustry/+94

Tested on Windows

### DIFF
--- a/core/src/io/anuke/mindustry/ui/ContentDisplay.java
+++ b/core/src/io/anuke/mindustry/ui/ContentDisplay.java
@@ -85,6 +85,9 @@ public class ContentDisplay{
 
         table.left().defaults().fillX();
 
+        table.add(Core.bundle.format("item.corestorable", item.type == ItemType.material ? Core.bundle.format("yes") : Core.bundle.format("no")));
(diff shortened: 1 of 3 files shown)


core/src/io/anuke/mindustry/ui/ContentDisplay.java
```diff
@@ -85,6 +85,9 @@ public static void displayItem(Table table, Item item){
 
         table.left().defaults().fillX();
 
+        table.add(Core.bundle.format("item.corestorable", item.type == ItemType.material ? Core.bundle.format("yes") : Core.bundle.format("no")));
+        table.row();
+
```
joshuaptfan: Unnecessary line break.

matheod: I feel like "storable" should be separated from explosiveness/flammability/radioactivity because the last tree is "sort of grouped", but I can remove the line if required.

Anuken: I think it's okay with the indent.

PR #1126, opened 2019-11-24
Calculate powerbalance before batteries interfere

Quezler:
This allows the power node(s) to display the amount of power being used while there is still charge in the batteries.

Its useful to be able to see at what rate batteries are draining and not just by looking at the `K` bar for a while to try to notice a change.

And it is especially useful with diodes, when charging your main network its not uncommon to see `+0/s` basically everywhere, giving a poor indication of the actual usage of the network.

To my knowledge this chance is fully (backwards)compatible and is purely visual.

It should make the diodes slightly less horrible than they are in their current state. 🙂

### DIFF
--- a/core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
+++ b/core/src/io/anuke/mindustry/world/blocks/power/PowerGraph.java
@@ -192,6 +192,8 @@ public class PowerGraph{
         lastPowerNeeded = powerNeeded;
         lastPowerProduced = powerProduced;
 
+        powerBalance.addValue((lastPowerProduced - lastPowerNeeded) / Time.delta());
+
         if(!(consumers.size == 0 && producers.size == 0 && batteries.size == 0)){
 
             if(!Mathf.equal(powerNeeded, powerProduced)){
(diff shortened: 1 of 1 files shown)


Quezler:
Screen Shot 2019-11-24 at 13 16 58

Anuken:
This was intentionally changed during testing, as I had a situation where a graph with no power generation of its own was getting diode power from a different graph. In this destination graph, you would see something like "-100/s" power usage and 0 battery power stored, despite the fact that all machines connected to the graph actually had enough power to run via the diode.

Quezler:
0,o

Anuken:
Basically the issue is that, when you have a diode involved, "-100/s" really makes it seem like you're losing power or don't have enough generators, when the reality is that you have a diode refilling batteries every frame.

ph3rr3t:
Add a 4th bar labeled 'Diode' or something to signify how much power you're getting from them?

Quezler:
@Anuken well in that case the power wire color should signify that, but nullifying diodes makes it even more confusing to look upon a powernode than without imo 🤔 

@ph3rr3t the way it works is by equalizing battery levels to avoid the update loop of power graph, not entirely sure how to best keep track of it 😗

PR #1136, opened 2019-11-27
Adding the feature to switch doors using a logic controller

Remint32:
- [x] _Compiled and tested_

- 1) the **Player** and the **controller** can _equal_ access to the door switch.
- 2) the controller has the same _click delay_ as the player.
- 3) if controller reversed the signal _on positive_ - the door will is **open**, if reversed _on empty_ the door **closes**.
- 4) the Controller also switches the door provided that its signal has _changed_ (if the controller signal does not change and the player switches the door, then the **3** point is _ignored_ until the controller signal changes)
- 5) If there is a situation that the door did **not have time to switch** because of the delay and the signal **does not comply** with the rule of paragraph **3** **_AND_** the player does not switch the door _manually_, the door will change its state to the correct one after the delay.

[**> A short video showing the principle of operation <** (YouTube)](https://youtu.be/V7WosSEv1lw)

### DIFF
--- a/core/src/io/anuke/mindustry/world/blocks/defense/Door.java
+++ b/core/src/io/anuke/mindustry/world/blocks/defense/Door.java
@@ -97,10 +97,10 @@ public class Door extends Wall{
         DoorEntity entity = tile.entity();
 
         if((entity.lastSignal != tile.entity.enabled()) && tile.entity.timer.get(timerToggle, 30f)){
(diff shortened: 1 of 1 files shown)


Remint32:
_known issue:_
https://discordapp.com/channels/391020510269669376/396416151032299521/649146639415509013

Anuken:
Might as well merge this, as the logic branch isn't being worked on right now.

Issue #804, opened 2019-10-09
Conservative Generators


Anuken:
- [ ] Make generators stop consuming resources when no power is being requested. Somehow.

231ACR:
If possible create a ,boolean full, propose is that when the thing is full or clogged or over it will turn true which causes power consumption disabled but redundant power continue (Simply it will act as relay)(Silly approach I know)

azureblue:
Considering a connected graph, I think that the biggest problem is to choose the generator(s) to shut down, like what would be the criterion for that? In general If we consider only their output power then it looks like a _change-making problem_. But considering that some generators consume more resources per unit power it gets complicated (let alone considering different value or resources :)).

Tazshelby2016:
It will require changing how power works, from the source giving power to the users, to the users requesting power from the source.

azureblue:
Maybe.. but I think it doesn't matter. The current approach (having a power pool being created at every update and then distributed) seems just fine. 

Anyway, it won't significantly help to solve any of the challenges I've mention in the previous comment.

231ACR:
New Approach, (working on factories rather than powers system)
Make factories to detach to node on clog or full and automatically connect when something happens. Your own choice how to set this stimulation.

LionelBergen:
@Anuken 
I'm new to this game, what do you mean by 'no power is being requested'? 
Only things I can think are if their is no attached batteries, or the batteries attached are full

Anuken:
I'm referring to situations where you have a generator running on its own with no outputs, or a group of generators producing 100 power/sec while the grid only consumes 50 power/sec.

LionelBergen:
@Anuken 
Working on this now. Looking at code it looks like power is given fully to every consumer. Is that correct? 
For example if a tile is producing '50 power' and there are 2 units that need '50 power', will each one get 50? Or will they both get 25?

Anuken:
Each will receive 25.

LionelBergen:
@Anuken  PR #825 should do the trick. Let me know if theirs any issues or suggestions relating to it

runiArnsbjornson:
might be "naive" but why not act like the factories ? make a small reserve of power inside every generator (even a hidden one) then if power is not consumed (ie : the power grid is already in a positive state) the reserve will fill, at full capacity, stop production, as soon as reserve decrease, restart production ?

LionelBergen:
[...]

PR #1137, opened 2019-11-27
Conservative Generators

thomas-spriggs:
Implementation of https://github.com/Anuken/Mindustry/issues/804

When the power needed plus the power put into batteries is less than the
total generation capacity, conservative generators will conserve a
portion of their input items proportional to the percentage of optional
conservative power generation, which is unused. The conservation is
carried out based on extending the amount of time until the generators
will consume an additional item of fuel. So for example if the power
needed is 50% of the production capacity which is entirely conservative
then the generators will take 2x as long between consuming items of
fuel.


Anuken:
Decided to use a simpler implementation with https://github.com/Anuken/Mindustry/commit/01e3912827bbcd1285690a704ef5e4172b8333ea

PR #1139, opened 2019-11-28
Add Inverted Overflow Gate

l-Luna:
Implements [Inverted Overflow Gate](https://feathub.com/Anuken/Mindustry/+70).
Outputs to the sides; when the sides are full, tries to output forward.


martosss:
image
You can already achieve this in as little as 4 tiles

l-Luna:
You can do something similar for the sorter.

dandiez:
> 
> 
> image
> You can already achieve this in as little as 4 tiles

That does not help when you don't have enough space, for example if you want to "push" the load towards one of the sides when you have parallel tracks...
image

TsjipTsjip:
Would be a nice addition. We already have an inverted sorter, why not an inverted overflow gate?
While there is a workaround with the current overflow gate demonstrated above, being able to do this in a 1x1 area instead of a 2x2 area with offset belts would be a great enhancement.

Anuken:
Decided to implement this myself, due to formatting issues in this PR.

PR #1140, opened 2019-11-28
🎇 Observer mech

Quezler:
Screen Shot 2019-11-28 at 19 43 49

This is just a rough proof of concept, not intended to be merged straight away, yadayadayada, etc.

It adds a mech that (theoretically) can only observe, it can't: (well, shouldn't be able to)

- shoot
- build
- destroy
- take
- mine
- configure
- targeted
- etc

(it can however see block contents, block configurations, etc) (and can still use mech pads!)

The uses i had in mind:
- a mech to scout enemy bases on attack maps, since it won't be targeted, so you can have a peaceful look around.
- events on custom (modded/plugined) servers, where dead players or guests can still observe the game after they died or something, or perhaps a state to put new players in so they can be shown around without being able to grief anything.

(well now that i typed that out it sounds boring and unusable, but i'll continue typing nontheless to see where it can lead)

In a nutshell this is a highly situational mech, but the circumstances under which its useful do not yet exist, but its existence will catalyst the creation of the uses a mech like this provides.

Texture is based on parts of all flying ships, though i picked surge as a color to showcase its special status as compared to: `orange (attack) / blue (ambush) / green (support)`

Reused my `core-seed` texture for its (hidden) pad.

Nov-28-2019 19-45-31

(this pull has plenty of bugs & things overlooked, and just a bit unpolished in how it does things)

🥔


joshuaptfan:
Its function would be instantly intuitive if its sprite was just a translucent version of the dart, like a ghost.

Quezler:
Transparency is on the todo list, but i first wanted to see how well the idea was received.

DeltaNedas:
1. find someone busy working on defense before a wave
2. votekick
3. watch as they rage, unable to do anything because "semi-guilty until proven innocent/time has passed"

Quezler:
:ohno:

ZkyLB:
iloveiy ttr

Ali-C-Ila:
I believe this semi-guilty stuff is still plausible because depending on the number of players, usually losing one builder is far less dangerous than having one griefer. 

The one falsely started the vote can be kicked later.(also add a cooldown time for one player to start a vote.) 

However, this function should be disabled if there are below six players online.

CA6:
this is an interesting concept however until there are more attack maps it is kinda point less except in PvP and in multiplayer punishment.
if you could have your location also act as a unit rally point or drop unit "way points"(not included in multiplayer punishment). 
its your idea tho

joshuaptfan:
Thinking on discussion about this on Discord, votekick punishment should not apply all of these traits, because of how powerful players can be when transitioning from ghost back to normal behind enemy lines.

Votekick punishment should be a separate thing that applies only these traits:

* Cannot build or break blocks
* Cannot configure blocks
* Cannot take/mine items
* Cannot deposit items

They should still be targeted by enemies and take damage, so that they can't use this to bypass enemy defenses, and shooting is fine because it cannot be used to grief. So then, a "ghost" appearance would no longer be appropriate; the player ship should instead be tinted and/or outlined in red.

Abuse of the punishment by griefers can be prevented by not applying it until the votekick has been "seconded" (at least one other player has also voted to kick).

TsjipTsjip:
This would be a nice addition to the game for players to use on various maps, not just PvP maps.

On the subject of using this as a vote-kick mechanic, I don't see this as an appropriate solution. This for the reasons stated above but I also don't believe putting them in a state alike the observer mech but still target-able is a good idea. I'm looking specifically at multiplayer servers with a maximum player count. A punished player would still take up a slot, theoretically punishing the entire team for having one slot less. (This isn't an issue when the server isn't full, but on very intensive maps you need all the help you can get.) I believe vote-kicking as it is now is a good system.

This brings me to my second idea with this mech. It could be a possibility to add the punished state mentioned above separately. (As in vote-kicking is still possible.) This way, players would have the option to choose between two things of punishing griefers, ... This would be especially useful as this is applied as some sort of an in-game state meaning VPN users are more hindered by this punishment. (There are griefers who use a VPN to bypass vote-kicking.) In practice, I believe they would be able to leave and rejoin with a new connection "unknown" to the server, thus allowing them to build again. This along with the fact that it might over complicate things.

As for the base idea for an observer mech pad: I believe it is an awesome idea and I would use it if it was an in-game feature.

(Apologies for making this long.)

Anuken:
- If a fast scouting mech is needed, that could be the Dart, once the [mech rework](https://trello.com/c/aPHMbuNx/461-new-mech-roles) is completed. I don't think it should have *no* offensive or building capability. Very few (if any) people would use it in real games, as you can already get a good look at things just by flying in, and there's a large view range with a minimap anyway.
- If some sort of "spectator" mode is needed, adding a new mech seems like an un-intuitive and roundabout way to do so. I can see someone being very confused on why they've become this new mech and why they can't build or shoot properly. Instead of having a mech, the player should just be able to pan around the map while dead (or something similar). This would be relatively easy to implement - in fact, the mobile version already has such a feature.

joshuaptfan:
> A punished player would still take up a slot, theoretically punishing the entire team for having one slot less.

I'm talking about this punishment only as an immediate remedy for potential griefing while the votekick, which may or may not _eventually_ eject the player from the game, is in progress.

Arkanic:
> A punished player would still take up a slot, theoretically punishing the entire team for having one slot less.

a "null" or "empty" team which doesn't have a core

Anuken:
Closing this for the reasons I mentioned in my post above.

PR #1141, opened 2019-11-28
🎄 Festive mode

Quezler:
Screen Shot 2019-11-28 at 20 18 47

Just something i was working on as a surprise in secret, idea was to send a pull that added a christmas theme that would only be active on the 25th of december, with blinking light decorations / snow / red noses for mechs / snowball projectiles / etc / you name it, but i never came around to do more than adding red and green dots to buildings, and then i had to reallocate my time to work stuff, so it kinda just sat there all alone in my local repo, leaving it here so others might keep it company. 🎁


Anuken:
I was thinking about implementing a holiday seasonal update for a while... perhaps something like this could be part of it.

martosss:
a snowfall/snowflake overlay would be a nice addition to this :)

Anuken:
That's part of the plan

Arkanic:
thorium reactors display christmas theme colours

PR #1145, opened 2019-11-30
Present the file sharing on iPad in a popover - and prevent the crash.

emlynmac:
Fix crash on iPad when exporting game data; prevent the build from failing

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,3 @@
 org.gradle.daemon=true
-org.gradle.jvmargs=-Xms256m -Xmx1024m
+org.gradle.jvmargs=-Xms256m -Xmx4096m
 archash=c7f3d8125c8d13c0a31c5f14fdabdb3d0efbd676
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,3 @@
 org.gradle.daemon=true
-org.gradle.jvmargs=-Xms256m -Xmx4096m
+org.gradle.jvmargs=-Xms256m -Xmx1024m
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -1,3 +1,3 @@
 org.gradle.daemon=true
-org.gradle.jvmargs=-Xms256m -Xmx1024m
```
Anuken: This may cause builds to crash on some devices. Builds work fine for me, so I would recommend you only set this for your computer.

emlynmac:
My understanding was that increasing the heap max size will prevent crashes as we now don’t have the out of memory exception. 

On systems that don’t have that amount of physical RAM, virtual memory will be used which will be slower but should not crash. 

If you’re dead against this change I can remove it, but it should do more good than harm. 

Emlyn

Anuken:
In my experience, that's not the case. Several people with weaker PCs have reported "unable to allocate memory..." errors at Gradle startup, even when the max heap size is set to 1024. Reducing the number to 512 solved those problems for them. I would think that using a number as high as 4096 would definitely cause issues as well. You also have to consider the fact that Java doesn't try to GC until it needs to, and most people wouldn't like 4G of their ram being hogged without a good reason.

Considering I've **never** had issues with memory running out after more than two years of using Gradle, the problem most likely runs with your setup, not the repository.

emlynmac:
Ok, I’ll pull it out on Monday. 

AFK at the moment...

Anuken:
Decided to change it myself. Thanks.

PR #1146, opened 2019-11-30
🐝 Flight of the pootisbird 🐦

Quezler:
Just something silly ima leave here, movable cores (っ◕‿◕)っ

Nov-30-2019 15-39-44

### DIFF
--- a/core/src/io/anuke/mindustry/world/Tile.java
+++ b/core/src/io/anuke/mindustry/world/Tile.java
@@ -12,6 +12,8 @@ import io.anuke.mindustry.game.*;
 import io.anuke.mindustry.gen.*;
 import io.anuke.mindustry.type.*;
 import io.anuke.mindustry.world.blocks.*;
+import io.anuke.mindustry.world.blocks.storage.*;
+import io.anuke.mindustry.world.blocks.storage.CoreBlock.*;
 import io.anuke.mindustry.world.modules.*;
 
 import static io.anuke.mindustry.Vars.*;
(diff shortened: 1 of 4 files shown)


DeltaNedas:
stop

Dexapnow:
It's really good idea, but core shouldn't rotate.

Quezler:
@Dexapnow well my logic was that anything in the game faces the direction it flies in, it is in fact less work to have it not rotate at all, it would make the animation helluvalot smoother though 👍

Violet-Scarelli:
Shades of the Terrans. To top this off, all ya need to do is add different core behavior for each team color. Blue would do this. Red would have "creep" that transfers items and power. Green would have item/power pylons. Bwahahahaha!

Quezler:
latest

Violet-Scarelli:
Honestly the cutest depiction of a Zergling I have ever seen. Starcrafts is the best.

Arkanic:
why does this give me the same ironic feeling as hitchhiker's guide to the galaxy

Arkanic:
may i also ask why that core's shading is inverted

Quezler:
Ah, that's because textures are sprited facing the left and thus the rotation needs to be `+90`'d, didn't notice it, thx for pointing out out

sk7725:
This should return for 6.0

Anuken:
No.

Meltdown-Altair:
it's honestly *not* that bad of a choice for cores to be movable , but cant you just , set them to be pickable by paload units?
cores moving would cause all kinds of lag & stuff

USMP-lancer:
> it's honestly _not_ that bad of a choice for cores to be movable , but cant you just , set them to be pickable by payload units?
> cores moving would cause all kinds of lag & stuff

because of how gameover works, this would cause easy gameover on single core maps

PR #1150, opened 2019-12-01
Editor: protect blocks from being changed by players (e.g. item sources)

Quezler:
Dec-01-2019 19-19-00
[HD](https://discordapp.com/channels/391020510269669376/393111111853146112/650766369876475935)

This is just a suggestion flavored pull, it adds 3 new block restrictions:

- `unconfigurable`: config of the block can't be changed (e.g. lock output of item sources)
- `unremovable`: block cannot be deconstructed by its team (e.g. griefing item sources or having pregenerated parts of a factory in the map)
- `unkillable`: cannot take fatal damage (maybe for asthetics or something, unkillable scap walls)

(multiblock tiles can have several under them, but in the case of a single tile block i added a `immortal` one that inherits all 3)

original idea was just to make itemsources unconfigurable, but the next best thing griefers might do is remove it, and if they cannot do that they will just try to blow it up, so the bottom two are just out of necessity.

needless to say the code is messy and stuff happens where it should not happen, but its just enough to get this pull reasonably stable as a proof of concept.

caveats:
- if the unkillable is under a core, obviously that team cannot die
- during refractor lots of `set` got changed to `add`, to lazy to fix it for this prototype
- if an unconfigurable tile dies somehow, and something new gets build ontop of it, the new block is also unconfigurable
- probably more


joshuaptfan:
Variable naming suggestions:
* `indestructible` instead of `unkillable` (blocks are not living things).
* `invulnerable` instead of `immortal` (blocks are not living things).

Quezler:
That's just hurtful, blocks have feelings to 😭

Quezler:
Oh and its technically not invulnerable, it does take damage, it just won't go below `0.00001f` 🤔

Quezler:
And indestructible might be confused with the deconstruction planner 😗

joshuaptfan:
It may be a better hint to players that something cannot be destroyed if it doesn't take damage to begin with, rather than wasting their time depleting its health only for them to think the game is bugged (prepare for bug reports) when the block doesn't get destroyed.

joshuaptfan:
Checking the thesaurus... `indelible`? no... `inexterminable`? no... `irrefrangible`? no... ah, `unbreakable`.

Quezler:
Well there would be a handful of reasons to want to show cracks thus attracting menders/drones/attention, straight up immortal doesn't fit the style imo

Oh and its a draft, no perfect thoughts were poored into naming & code, and i did in fact use thesaurus for naming something in here

: )

joshuaptfan:
If a block on a tile with only the `unconfigurable` property set is deconstructed, players can then build their own blocks on it which will become unconfigurable. `unconfigurable` should only be available in conjunction with the other properties.

Quezler:
hence why i added a bunch

CA6:
Would intouduce specific item covering from some maps

Anuken:
Interesting concept, but I think this should be implemented as a set of configurable int flags on a tile entity, not as an overlay. Until the editor gets a major overhaul, this will have to wait.

PR #1158, opened 2019-12-04
💰Upgrade bannedblocks to allowance

Quezler:
Whereas banned blocks can either be set to completely allow or block a block, allowance will allow a limit/cap to be set.

This change is backwards compatible for map files, since it automatically imports banned blocks. (though its not backwards compatible with older clients, they will need to upgrade to use it)

The name for it is subject to debate, i've tried a few but none sounded fitting enough:
- `limit` sounds to broad
- `max` of what?
- `quota` sounds like a goal to reach
- `manymore` lorem ipsum

Current caveats:
- players can 'cheat the system' by placing many blocks in their building state, since the flag only gets applied once the building is fully constructed, i'll look into fixing that if this pull gets enough traction.

Oh and the placement of where the allowance shows is up for debate as well, it might be a bit weird for it to show up as a building material for it, but its the least out of place looking location i could have come up with.

Screen Shot 2019-12-04 at 11 39 16
Screen Shot 2019-12-04 at 11 39 26
Screen Shot 2019-12-04 at 11 39 39
Screen Shot 2019-12-04 at 11 40 10
Screen Shot 2019-12-04 at 11 43 14
Screen Shot 2019-12-04 at 11 43 24
Screen Shot 2019-12-04 at 11 43 35
Dec-04-2019 11-44-18

🥔


Quezler:
`#itworksonmymachine ¯\_(ツ)_/¯`

Screen Shot 2019-12-04 at 12 01 34

joshuaptfan:
This could create a negative teamwork dynamic where people feel bad about others "wasting" the allowance, leading to fighting in chat and vote kicks. There's no such tension when blocks are simply either banned or not.

Anuken:
What is the use case for this? Challenge maps?

Ali-C-Ila:
This cannot be used in multiplayer. But yeah, this may add some challenge and fun to some solo maps.

Quezler:
> What is the use case for this? Challenge maps?

hmm, well the general idea was to encourage/discourage certain behavior:

- limiting small solar panels, reducing their spam so people opt for 3x3 ones (or other sources of power, instead of paving entire sections of the map with em)
- limiting steam generators, mainly against those free power schematics, while still allowing it to be used as a stepping stone
- limiting the tier 1 graphite presses, pushing players to scale to the tier 2 ones which are generally more efficient
- limiting types of generators, forcing players to use a mix of everything-ish
- limiting defensive stuff (turrets/walls/etc) requiring strategic placements
- perhaps limiting (lower) tier conveyor to encourge uses of mass drivers over long distances
- etc, basically any stuff a map maker or server owner might want to experiment with

> "sniping"

yea it would be an issue that players will/could steal blocks from builds of others 🙈

Legomonster33:
You can fix the issue of people "stealing" each other's blocks by having it be an allowance per person instead of just in general.

DeltaNedas:
+1 :rocket:

itcannotbe:
>You can fix the issue of people "stealing" each other's blocks by having it be an allowance per person instead of just in general.

That would not be an issue if all the players are COOPERATING.

Yes the emphasis is needed as it is almost forgotten and doesn't matter on PvE maps a.k.a. Survival.

Quezler:
exactly ^

joshuaptfan:
That's naive. Just because players are on the same team, does not mean they will always work together. It only takes 1 bad apple to ruin everything.

itcannotbe:
Challenge maps aren't for everyone.

Also that bad apple could be subject to a vote kick.

Anuken:
I don't think this should be added to the game.

- It's tricky to implement correctly, and may have a lot of edge cases. Bugs are practically guaranteed.
- It adds more bloat and maintenance burden to the already bloated rule/build code.
- It adds more complexity to the rule selection system.
- It's not difficult to implement as a server-side plugin with an action filter.
- It's a heavy handed approach to solve a problem that's related to balance. Instead of preventing players from placing too many of a single type of block, the game should have homogeneous block spam be less effective as a tactic. I realize that some servers may want to experiment with their own changes, but... again, use a plugin instead, it's much more flexible.

PR #1164, opened 2019-12-05
Correct description falsely stating attack mode has no waves

Ali-C-Ila:


YaFou:
I think your suggestion doesn't provide anything. Information about the different game modes are in the help when you create a new game.
**In my point of view, I don't like so much this idea.**

PR #1165, opened 2019-12-05
Update bundle_zh_TW.properties

crackycookie:
Updated unit.thousands = k, unit.millions = mil in Chinese


Ali-C-Ila:
I think this update shouldn't be merged. 
I'm one of the zh TW translators and I thinks helps are always appreciated. However, for this case, using "mil" and "k" is totally understandable to the gaming community, but something like "254 千" or "13百萬" are very weird and does not exist in Chinese. 
If you want to discuss, a great place is the official Mindustry discord: https://discord.gg/mindustry.

crackycookie:
True....so in order to make that statement available in Chinese, it requires changes of how to display the numbers in the game?

Ali-C-Ila:
Yeah, that going to be very complicated. You will need to adapt the code to function differently when different language is choosed. I don't believe anything similar exists currently.

crackycookie:
I see. I have some background of coding but not proficient, I guess I will just put this in my mind and maybe help with other things. Thanks for your time, I will close this pull request :)

PR #1167, opened 2019-12-05
Create bundle_zh_HK.properties

crackycookie:
Mainly referenced from bundle_zh_TW.properties, but modified words and phrases to suit players from Hong Kong, including myself haha.


Anuken:
I'm not sure if `bundle_zh_HK` is a valid name for a locale, or if Java will recognize it as such.

crackycookie:
I've referred to this link: https://www.science.co.il/language/Locale-codes.php
where the LCID string is "zh-HK". I've tried to put the bundle.properties in my Windows 10 Mindustry folder and it works.

Anuken:
The problem with this bundle is that it's largely the same as another bundle, with some minor changes. It would be nice if the bundle only contained the keys of what has changed, so that any updates in the "parent" bundle would show up in this one unless explicitly overridden - otherwise, without a dedicated person keeping it up to date, the bundle would become outdated quickly.

PR #1175, opened 2019-12-06
add liquid-distributor block

Arkanic:
basically just a larger liquid router, this allows for easier liquid chaining for things like defense and recursive factory layouts.

~~(the sprite is pretty bland)~~

Screenshot (6)
I could probably add some kind of detail to the top of the block


Anuken:
Why not just use 4 routers?

Arkanic:
quezler told me the same thing

Arkanic:
tbh the liquid system just means that it evenly distributes

Arkanic:
it has more capacity now, giving it an actual reason to be built

Anuken:
If you want capacity, there are tanks, and if you want distribution, there's routers. I don't think this block is really needed.

Arkanic:
that's fine by me, i'll just find something else to make

geoffrey1900:
hey this is a good idea if so why is the the distrbutor in the first place

geoffrey1900:
examples bad but i actualy really want this lol

geoffrey1900:
so we dont have to deal with the powers problem as much of using router chains that each input is half the last and it doesent work well of say (still makes me sad f) i had som thor reactors and i put dow some cyro flow to 5 of them and the 5th blew up from getting like no cyro and deleted front line and i died

geoffrey1900:
so even fluid flow would be nice lol

geoffrey1900:
*add it*

Anuken:
Use better piping instead. I won't be adding this.

Arkanic:
reviewing this, i can agree with anuke, this idea wasn't too well thought of.

PR #1189, opened 2019-12-10
Add Ability to choose what resource is mined

eun4786:
https://feathub.com/Anuken/Mindustry/+243
ore selection
When you click on a drill, the resource table is displayed and mine the selected resources from the table.

### DIFF
--- a/core/src/io/anuke/mindustry/world/blocks/ItemSelection.java
+++ b/core/src/io/anuke/mindustry/world/blocks/ItemSelection.java
@@ -40,4 +40,29 @@ public class ItemSelection{
 
         table.add(cont);
     }
+
+    public static void buildDrillItemTable(Table table, Array<Item> items, Prov<Item> holder, Cons<Item> consumer){
+
+        ButtonGroup<ImageButton> group = new ButtonGroup<>();
+        group.setMinCheckCount(0);
+        Table cont = new Table();
+        cont.defaults().size(38);
+
+        int i = 0;
+
+        for(Item item : items){
+            if(!data.isUnlocked(item) && world.isZone()) continue;
+
+            ImageButton button = cont.addImageButton(Tex.whiteui, Styles.clearToggleTransi, 24, () -> control.input.frag.config.hideConfig()).group(group).get();
+            button.changed(() -> consumer.get(button.isChecked() ? item : null));
+            button.getStyle().imageUp = new TextureRegionDrawable(item.icon(Cicon.small));
+            button.update(() -> button.setChecked(holder.get() == item));
+
+            if(i++ % 4 == 3){
+                cont.row();
+            }
+        }
+
+        table.add(cont);
+    }
 }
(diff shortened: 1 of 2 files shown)


Quezler:
+1

Kieaer:
+1

Anuken:
I don't agree with making this configurable. Even if it's hidden behind a menu, this adds unnecessary UI and complexity that isn't needed in most cases.

DeltaNedas:
why not a drill setting tool that sets target resource for any drill you click instantly, no bloat ui for drills just one thing

core/src/io/anuke/mindustry/world/blocks/ItemSelection.java
```diff
@@ -40,4 +40,29 @@ public static void buildItemTable(Table table, Prov<Item> holder, Cons<Item> con
 
         table.add(cont);
     }
+
+    public static void buildDrillItemTable(Table table, Array<Item> items, Prov<Item> holder, Cons<Item> consumer){
+
```
YaFou: Why there is a line break? There is no need.

Arkanic: try importing the mindustry code format
~~i've done the same too many times~~

YaFou: Mindustry IntelliJ code format: https://github.com/Anuken/Mindustry/blob/master/.github/Mindustry-CodeStyle-IJ.xml

Jaeiya:
> I don't agree with making this configurable. Even if it's hidden behind a menu, this adds unnecessary UI and complexity that isn't needed in most cases.

I've played several games where I wanted to max out production efficiency of a specific resource but it was too close to another (sand, scrap, coal, etc...). I agree it's a QoL feature that may not be necessary, but there is definitely a precedence for it.

XEN0PHIL:
What about making it drill everything by default?
For example there's 3 lead and 1 copper in a 2x2 square, then mechanical drill will output 75% lead, 25% copper.

Jaeiya:
@XEN0PHIL But then you'd have to use a sorter and that adds unnecessary complexity to your lines. I find in most cases, if there's a resource bleeding into another, I only want one of them; I very rarely if ever have found myself needing both resources equally.

YaFou:
I manage easily my drills with multiples ores below. And often, the ore is far from others, so we don't need to select the resource which we want to mine.
**Also, this isn't necessary but it's can be a good extra feature as an option in settings.**

matkoniecz:
"extra feature as an option in settings" are making testing and avoiding bugs significantly harder.

CA6:
eh this is QoL stuff that would be nice but I'd like like how delta said it a singular menu in the build menu to save ui computing

Anuken:
This PR can still be used for discussion, but I will not be merging it in its current form.

PR #1191, opened 2019-12-11
Surge Capacitor (higher tier battery)

Arkanic:
Screenshot (16)

effectively just a giant battery.
uses high amounts (100) surge to construct, as to make sure that only higher level players/teams construct them & to discourage spamming them. can hold 300k :eyes: power.

I was discussing an idea about making the battery take consumables in return for extra storage space, or make it require cooling, but in the end i felt just making it overly expensive would deter people from creating vast swathes of them.

Ironically, the main purpose of the surge conductor is to hopefully reduce the size of the "battery fields" that people create in order to keep their power stored (like below).
Screenshot (18)

might need a nerf though

Edit: i revamped sprite
image


core/assets/bundles/bundle.properties
```diff
@@ -1175,6 +1176,7 @@ block.surge-tower.description = An extremely long-range power node with fewer av
 block.diode.description = Battery power can flow through this block in only one direction, but only if the other side has less power stored.
 block.battery.description = Stores power as a buffer in times of surplus energy. Outputs power in times of deficit.
 block.battery-large.description = Stores much more power than a regular battery.
+block.surge-capacitor.description = A massive and expensive battery, the surge capacitor is the ultimate solution to power storage.
```
Quezler: not sure about the usage of `ultimate solution` here, kinda prevents future storage solutions and/or mods 🤔

joshuaptfan:
The justification for this feature may need analysis: do players build battery fields because they actually need that much reserve power, or just to fill unused space?

core/assets/bundles/bundle.properties
Arkanic: i can think of something else, tbh i didn't think that was a good solution in the first place

Arkanic:
@joshuaptfan people do it for both, I've been on maps where people are complaining about the fields of batteries where they want to build a factory/defence, which block would be useful for (storing large amounts of power in a small space), and there are also times when people intentionally did this, and the capacitor allows people with more advanced cores & factories to have more power storage:space, effectively allowing people with higher level stuff to have a lot more power storage (currently large batteries fill up rather quickly when you have a couple of impacts running).

core/assets/bundles/bundle.properties
```diff
@@ -944,6 +944,7 @@ block.surge-tower.name = Surge Tower
 block.diode.name = Battery Diode
 block.battery.name = Battery
 block.battery-large.name = Large Battery
+block.surge-capacitor.name = Surge Capacitor
```
YaFou: For the French translation: `block.surge-capacitor.name = Condensateur de Surtension`.

core/assets/bundles/bundle.properties
```diff
@@ -1175,6 +1176,7 @@ block.surge-tower.description = An extremely long-range power node with fewer av
 block.diode.description = Battery power can flow through this block in only one direction, but only if the other side has less power stored.
 block.battery.description = Stores power as a buffer in times of surplus energy. Outputs power in times of deficit.
 block.battery-large.description = Stores much more power than a regular battery.
+block.surge-capacitor.description = A massive and expensive battery, the surge capacitor uses surge plating for extremely high power storage.
```
YaFou: For the French translation: `block.surge-capacitor.description = Une batterie massive et coûteuse, le condensateur de surtension utilise un placage de sutension pour un stockage d'énergie extrêmement élevé.`.

Arkanic: thank you

blackassasin1234:
I think it should be renamed to surge battery instead of capacitor, it would make it fit better with the other batteries' name

Arkanic:
@blackassasin1234 I discussed that with some people on #development, in the end people agreed on a high capacity battery rather than a kind of meta requiring items to increase the storage space because of some loopholes such as where does all the excess power go, and how the material should be processed.

Arkanic:
The entire thing is stale, once anuke takes interest/places suggestions, criticism ill update it again

Arkanic:
I'm also going to improve the sprite because the current one looks really sad

Arkanic:
bye bye

PR #1199, opened 2019-12-11
Fixed ore texture names

DeltaNedas:
ore-myore -> ore-myore-1.png not mydrop1.png

**DO NOT MERGE UNTIL ALL MODDERS ARE GIVEN A GOOD AMOUNT OF TIME TO UPDATE MODS**
(btw i made pr on examplemod too, it works nicely)
((btw works on my mod too))


Arkanic:
interesting

DeltaNedas:
ok arkanic didnt say its bad so modders pls fix your mods or theyll break its 99% official 100% official if/when our lord says its ok

Anuken:
> DO NOT MERGE UNTIL ALL MODDERS ARE GIVEN A GOOD AMOUNT OF TIME TO UPDATE MODS

This isn't going to happen. Even if I publically announce this and personally ask everyone to do it, some mods will remain broken and users will complain. A single hyphen just isn't worth all this effort, and in my opinion defining new blocks is easier without it.

DeltaNedas:
it also uses drop
which makes no sense, why would an ore use its drop name for texture?
maybe it would be possible to look for both old and new?

Anuken:
> why would an ore use its drop name for texture?

Because it's simple and immediately recognizable. 'Ore' is unnecessary and inaccurate, as not all resources on the ground may necessarily be ores (coal ore? sand ore?)

DeltaNedas:
ore prefix is for OreBlock which does `super("ore-" + name)`, for other stuff yea you could just have name-*variant*.png no `ore-`

PR #1201, opened 2019-12-12
Handle missing locale keys for link descriptions

TsjipTsjip:
This change prevents the game from crashing when a locale key is missing for a link description listed in the "About" tab. Previously this used `I18NBundle.getNotNull()`, which would kill the process when it finds a given key isn't present. It now uses `I18NBundle.getOrNull()` which will return null values if a key is missing.
Returned null values are handled by listing the keys in a new ArrayList<String> , I may have violated the contribution guidelines here but I am unsure. When the "About" tab is clicked for the first time, an in-game error message is displayed with all the keys that are missing. After dismissing this message, the game just continues as is.

These changes add the following new locale keys: `missinglocale`, `missinglocale.single`. These can be used for displaying missing locale keys elsewhere in the project, however they only serve a purpose in the about tab in this pull request. If these keys are not present, the used text will default to `Missing locale key(s):`.

This is a whole lot for such a minor thing, but this is one of my first contributions to the project and I didn't really know how the project works. I now know much more of this and am still eager to learn.


TsjipTsjip:
Resolved conflicts with `bundle_nl_BE.properties` and `bundle_it.properties`.

Anuken:
Looking over the changes, I think this is over-complicating things. You have additional bundle keys and significant extra code for something that will most likely never happen to an end user - why not replace `getNonNull` with `get` instead?

TsjipTsjip:
> Looking over the changes, I think this is over-complicating things. You have additional bundle keys and significant extra code for something that will most likely never happen to an end user - why not replace `getNonNull` with `get` instead?

I used `I18NBundle.getOrNull()` to know more easily when a key is missing, would need to compare a substring otherwise. If the error message popup is unnecessary I can remove this, doing so would simplify the entire system to an extent where the Links class looks very alike the one currently in the release builds (only difference being `I18NBundle.get()` instead of `I18NBundle.getNotNull()`). I can commit this to this PR if you want me to, I do see I was a little ambitious.

core/src/io/anuke/mindustry/ui/Links.java
```diff
@@ -27,6 +34,19 @@ private static void createLinks(){
...
+            if(missingLocales.size()==1){
+                key = "missinglocale.single";
+            }
+            String m = Core.bundle.getOrNull(key);
+            if(m==null)m="Locale key(s) missing:\n";
+            for(String s : missingLocales){
+                m = m+s;
+            }
+            String finalM = m;
+            Time.run(1f, ()->{ui.showErrorMessage(finalM);});
```
YaFou:
```suggestion
            Time.run(1f, () -> { ui.showErrorMessage(finalM); });
```

YaFou: `finalM` is used for nothing. `m` variable can be used for `Time.run`.
```java
            Time.run(1f, ()->{ui.showErrorMessage(m);});
```
*With my suggestion:*
```java
            Time.run(1f, () -> { ui.showErrorMessage(m); });
```

core/src/io/anuke/mindustry/ui/Links.java
```diff
@@ -38,7 +58,13 @@ private static void createLinks(){
...
             this.name = name;
             this.color = color;
-            this.description = Core.bundle.getNotNull("link." + name + ".description");
+            String key = "link." + name + ".description";
+            String desc = Core.bundle.getOrNull(key);
+            if(desc == null){
+                missingLocales.add(key+"\n"); // <- LINE 62
+                desc = key;
+            }
+            this.description = desc;
```
YaFou: You can add line breaks before and after the block to clarify the code like this:
```java
            this.color = color;

            String key = "link." + name + ".description";
            String desc = Core.bundle.getOrNull(key);
            if(desc == null){
                missingLocales.add(key+"\n"); // <- LINE 62
                desc = key;
            }
            this.description = desc;

            this.link = link;
```
You can also add more line break into the new block.
``` java
            String key = "link." + name + ".description";
            String desc = Core.bundle.getOrNull(key);

            if(desc == null){
                missingLocales.add(key + "\n"); // <- LINE 62
                desc = key;
            }

            this.description = desc;
```

core/src/io/anuke/mindustry/ui/Links.java
```diff
@@ -38,7 +58,13 @@ private static void createLinks(){
         public LinkEntry(String name, String link, Color color){
             this.name = name;
             this.color = color;
-            this.description = Core.bundle.getNotNull("link." + name + ".description");
+            String key = "link." + name + ".description";
+            String desc = Core.bundle.getOrNull(key);
+            if(desc == null){
+                missingLocales.add(key+"\n"); // <- LINE 62
```
YaFou: Why `// <- LINE 62`?

core/src/io/anuke/mindustry/ui/dialogs/AboutDialog.java
```diff
@@ -31,6 +32,7 @@ public AboutDialog(){
     }
 
     void setup(){
+        if(linkEntries==null)linkEntries = Links.getLinks();
```
YaFou:
```suggestion
        if(linkEntries == null) linkEntries = Links.getLinks();
```

core/src/io/anuke/mindustry/ui/Links.java
TsjipTsjip: My bad! This was a note I added when I was troubleshooting a NullPointerException.

core/src/io/anuke/mindustry/ui/Links.java
TsjipTsjip: m cannot be used in this case due to it being used in a lambda expression. (These need to be final or effectively final. I am unsure how else to achieve this.)

Anuken:
Resolved in a different way - see commit linked above.

PR #1206, opened 2019-12-12
Make scroll zoom by default, remove zoom hold keybind

joshuaptfan:
This PR makes scroll zoom without needing to hold another key.

Scroll will not zoom when building rotateable blocks and schematics, or when holding "R" to rotate rotateable blocks.

## Removed
* "Zoom Hold" keybind in all localizations

## Modified
* Zoom and rotation conditions in DesktopInput.java
* Tutorial text concerning zoom

## Needs localization
* Modified tutorial text concerning zoom

## Needs testing
* Mobile zoom/rotate behavior
  * MobileInput.java checked for the "Zoom Hold" keybind for some reason, was it just a mistaken copy/paste?


DeltaNedas:
yes

Anuken:
I don't see any code here that would handle the rotation keybind conflict. What happens if you try to rotate a block you're about to place with the scrollwheel?

> MobileInput.java checked for the "Zoom Hold" keybind for some reason, was it just a mistaken copy/paste?

No, it just used the same controls as desktop. Why would it be a mistake?

joshuaptfan:
> What happens if you try to rotate a block you're about to place with the scrollwheel?
It gets rotated, if it can be rotated, and zoom doesn't happen. If it can't be rotated, zoom happens.

Anuken:
Testing it out, I feel like Ctrl should be kept as an override for zooming *even when you have a rotating block selected* - possibly as a hidden keybind. There are times when I want to zoom while placing conveyors.

joshuaptfan:
Alright, I've made the "Diagonal Placement" keybind (which is also Ctrl by default) now modify scroll to zoom regardless of rotateable things. This has no effect on the "Diagonal Placement" keybind's original use, as scroll is usually not involved in that, though they do work together (placing diagonal conveyors and scrolling at the same time).

Anuken:
Alright, that works, thanks.

PR #1210, opened 2019-12-13
Liquid void

TsjipTsjip:
We have an item source, an item void, a power source, a power void and a liquid source but no liquid void.

Adds a liquid void as a sandbox utility tool for quickly clearing out liquids from any system.


joshuaptfan:
Isn't that already accomplished by building an open-ended conduit that spills liquid on the ground? In what cases would that solution not suffice?

TsjipTsjip:
An open-ended conduit does not suffice in specific situations. When you are testing designs which only take up a small amount of space and you mess up on conduits, an easy space for an open-ended conduit isn’t that evident.
One of the other reasons I made this is due to me feeling like this was missing. As stated in the initial post, we have every sandbox utility but this one.
(Side reference: I tested incinerators and they do practically the same, however they also take items and heat up. This is not ideal for certain situations.)

core/assets/bundles/bundle.properties
```diff
@@ -965,6 +965,7 @@ block.mechanical-pump.name = Mechanical Pump
 block.item-source.name = Item Source
 block.item-void.name = Item Void
 block.liquid-source.name = Liquid Source
+block.liquid-void.name = Liquid Void
```
YaFou: For French translation: `block.liquid-void.name = Destructeur de Liquide`.

core/assets/bundles/bundle.properties
```diff
@@ -1128,6 +1129,7 @@ block.power-source.description = Infinitely outputs power. Sandbox only.
 block.item-source.description = Infinitely outputs items. Sandbox only.
 block.item-void.description = Destroys any items. Sandbox only.
 block.liquid-source.description = Infinitely outputs liquids. Sandbox only.
+block.liquid-void.description = Removes any liquids. Sandbox only.
```
YaFou: For French translation: `block.liquid-void.description = Détruit n'importe quel liquide. Bac à sable seulement.`.

YaFou:
@itskatt, have you a better translation for `Destructeur de Liquide`? I didn't find a better one. It's not too good with `Destructeur` with liquids.

itskatt:
@YaFou `Vaporisateur de liquide` probably

YaFou:
Yes good! Thank you for your help @itskatt!

core/assets/bundles/bundle_fr.properties
```diff
@@ -958,6 +958,7 @@ block.mechanical-pump.name = Pompe Mécanique
 block.item-source.name = Source de Ressources
 block.item-void.name = Destructeur de Ressources
 block.liquid-source.name = Source de Liquide
+block.liquid-void.name = Destructeur de Liquide
```
YaFou:
```suggestion
block.liquid-void.name = Vaporisateur de Liquide
```

RogueRedHawk:
@joshuaptfan I know it's a bit late to be responding now but also I think it's worth mentioning that it's not always ideal to just dump all the contents of a liquid tank onto the floor, such as when said liquid is flammable.

joshuaptfan:
Flammable liquids are no more dangerous than water until ignited by enemies. Usually when testing in sandbox, no enemies are around.

itcannotbe:
I would like to point out that a. You can't set your setup on fire on accident and second the pulse conduit is more than capable to clear a system with a huge liquid capacity of liquids in no time.

TsjipTsjip:
Slag can put your setup on fire, but all this talk is quite irrelevant. I simply don’t see why this block shouldn’t be in the game while we have all the other utilities to use.

Anuken:
I don't have any issues with this block being added, but I'll wait until right before the next build to get merge it.

core/src/io/anuke/mindustry/world/blocks/sandbox/LiquidVoid.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public LiquidVoid(String name){
+        super(name);
+        hasLiquids = true;
+        solid = true;
+        update = true;
+    }
+
+    @Override
+    public boolean acceptLiquid(Tile tile, Tile source, Liquid liquid, float amount){ return true; }
```
Anuken: `return true` should be on a new line and formatted like a standard method.

TsjipTsjip: Done!

Quezler:
[_sprite submission_](https://github.com/TsjipTsjip/Mindustry/pull/1)
liquid-void

PR #1217, opened 2019-12-15
Second-tier illuminator

TsjipTsjip:
Spamming illuminators all over the place can be quite annoying. This 2x2 version of the illuminator provides more light at more cost. (Double the material cost, double the energy cost per second.)

This is still a draft request as I am currently lacking artwork for this; I will be working on this later on.

This request is low-priority as it serves to resolve a nuisance that arises when playing on lighting-enabled maps which is an experimental feature.


core/src/io/anuke/mindustry/content/Blocks.java
```diff
@@ -1832,6 +1832,14 @@ public void draw(Bullet b){
             consumes.power(0.05f);
         }};
 
+        largeIlluminator = new LightBlock("large-illuminator"){{
+            requirements(Category.effect, BuildVisibility.lightingOnly, ItemStack.with(Items.graphite, 8, Items.silicon, 4));
+            brightness = 1.34f;
```
Anuken: Brightness is the light opacity; it can't and shouldn't be greater than 1.

TsjipTsjip: I'll play around with this a bit, thanks for mentioning this. If a lack of better option arises, I'll leave it at the value from the standard illuminator.

CA6:
good idea

Arkanic:
maybe thin the walls down a bit, but a great idea

WaterRatje:
Shouldn't the light and build costs be 4x higher? 2x2 is 4x larger than 1x1. And otherwise to stimulate but also balance it out, you can make it 3x higher instead of 2

TsjipTsjip:
> Shouldn't the light and build costs be 4x higher? 2x2 is 4x larger than 1x1. And otherwise to stimulate but also balance it out, you can make it 3x higher instead of 2
> 
> Edit: after taking a look at the file changes, I'd say 10 and 6

The block's size isn't scaled accordingly as it is impossible to do so for Mindustry in the current moment. Amplifying the projected light by 4x the original amount is a bad idea which causes numerous graphical issues. (Light isn't rendered if it is outside a certain offscreen boundary, this issue has actually already surfaced with 2x amplification.) I've based the required power and materials on the effects it currently gives, which is 2x.

WaterRatje:
> 
> 
> > Shouldn't the light and build costs be 4x higher? 2x2 is 4x larger than 1x1. And otherwise to stimulate but also balance it out, you can make it 3x higher instead of 2
> > Edit: after taking a look at the file changes, I'd say 10 and 6
> 
> The block's size isn't scaled accordingly as it is impossible to do so for Mindustry in the current moment. Amplifying the projected light by 4x the original amount is a bad idea which causes numerous graphical issues. (Light isn't rendered if it is outside a certain offscreen boundary, this issue has actually already surfaced with 2x amplification.) I've based the required power and materials on the effects it currently gives, which is 2x.

Okay, I understand. Everything looks fine for me

core/assets/bundles/bundle.properties
```diff
@@ -927,6 +927,8 @@ block.inverted-sorter.name = Inverted Sorter
 block.message.name = Message
 block.illuminator.name = Illuminator
 block.illuminator.description = A small, compact, configurable light source. Requires power to function.
+block.large-illuminator.name = Large illuminator
+block.large-illuminator.description = A larger version of the illuminator. Provides configurable light for a larger area. Requires power to function.
```
YaFou: For the French translation: `block.large-illuminator.description = Un plus grande version de l'illuminateur. Il fournit une lumière configurable sur une surface large. Il requiert de l'énergie pour fonctionner.`

core/assets/bundles/bundle_nl_BE.properties
```diff
@@ -866,6 +866,10 @@ block.distributor.name = Distributor
 block.sorter.name = Sorter
 block.inverted-sorter.name = Inverted Sorter
 block.message.name = Message
+block.illuminator.name = Licht
+block.illuminator.description = Een compacte, instelbare lichtbron. Heeft energie nodig om te werken.
+block.large-illuminator.name = Groot licht
```
YaFou:
```suggestion
block.large-illuminator.name = Groot Licht
```

core/assets/bundles/bundle_fr.properties
```diff
@@ -924,6 +924,8 @@ block.inverted-sorter.name = Trieur Inversé
 block.message.name = Message
 block.illuminator.name = Illuminateur
 block.illuminator.description = Une petite source lumineuse compacte et configurable. Nécessite de l'énergie pour fonctionner.
+block.large-illuminator.description = Grand Illuminateur
+block.large-illuminator.description = Un plus grande version de l'illuminateur. Il fournit une lumière configurable sur une surface large. Il requiert de l'énergie pour fonctionner.
```
YaFou: Sorry for the mistake 😅
```suggestion
block.large-illuminator.description = Une plus grande version de l'illuminateur. Il fournit une lumière configurable sur une surface large. Il requiert de l'énergie pour fonctionner.
```

kapzduke:
are flood lights from thorge expansion what you're looking for?

Anuken:
Closing in favor of overdrivable illuminators with a slightly higher base radius.

PR #1239, opened 2019-12-20
[server] uuid security level

Quezler:
> technical draft (not backwards compatible with 101.*, merge when 102 is around the corner)

This pull adds a security level to go along with the uuid. (inspired by teamspeak)

To increase your security level you need to sacrifice some computing power to compute the next level, which duration grows pretty much exponentially.
(1-4 are relatively fast but 5+ will take a while)

Main purpose is to obstruct those using uuid changers, since you will need to regenerate the security level for each, which can take a considerable amount of time depending on the security level of the server you want to join.

Of course mobile users will have a harder time generating their uuid than powerful desktop users, but there's nothing i can really change about that.

Also malicious users could leave security level generators running 24/7 or rent large computing power/etc, so it might just get in the way a bit of normal users. (though normal users would only have to generate their desired security level once, which can be done with the device left afk or over multiple attempts, the progress is saved)

The technical implementation is currently one big mess:
- uses outside packages
- hijacks the mod list since i could not get the serializer to add a new field
- probably horrible from a cryptographic standpoint, but i'm not a cryptographer
- not sure how backwards compatible propper packet implementations for it will be
- currently settable for steam/local servers
- etc
- (oh and technically a server you're connecting to can steal your uuid & security level, but that "changes nothing", though it should probably use public/private key pairs alltogether)

Screen Shot 2019-12-20 at 13 59 21
Screen Shot 2019-12-20 at 13 59 29
Screen Shot 2019-12-20 at 13 59 39

🥔

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -122,6 +122,9 @@ mod.scripts.unsupported = Your device does not support mod scripts. Some mods wi
 
 about.button = About
 name = Name:
+securitylevel = Security Level:
+securitylevel.improved = Security level improved.
+securitylevel.changed = [accent]{0}[] -> [accent]{1}[]
 noname = Pick a[accent] player name[] first.
 filename = File Name:
 unlocked = New content unlocked!
--- a/core/src/io/anuke/mindustry/core/NetClient.java
+++ b/core/src/io/anuke/mindustry/core/NetClient.java
@@ -75,10 +75,15 @@ public class NetClient implements ApplicationListener{
                 net.disconnect();
             });
 
+            Metadata meta = new Metadata();
+            meta.categories.put("mod", mods.getMetaStrings());
+            meta.categories.put("securitylevel", new StringMap(){{
+                put("checkpoint", Integer.toString(Core.settings.getInt("securitylevel-checkpoint", 0)));
+            }});
+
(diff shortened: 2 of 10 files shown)


Arkanic:
do you think you could make the system automatic in a way? most not-so tech savvy users might just see it as an annoying obstruction to them playing online...

Quezler:
@Arkanic automatic in what way? silently nomming away processing power might prompt players to get rid of the game since "its slow on their device without a clear reason as to why", and the default level for servers is 0 anyways, so tech-unsavy users don't get hindered by default, only when a server owner explicitly opts for a security level will it matter, but normal players would likely figure out that the kick message is the same as whats next to that new weird upgrade arrow and manage to click it 🤔

Arkanic:
yeah, i was just thinking some people might find it annoying an unecessary, as good of an idea as it is

Quezler:
annoying, somewhat, normal players would only have to do it once, since their uuid is static.
unecessary, perhaps, depends on the amount of people that change their uuid a lot to grief with.

Arkanic:
from experience, i've only ever seen it once

Arkanic:
but yeah, this is a good idea

Anuken:
This should not be necessary anymore. If the problem resurfaces, I may take another look.

PR #1241, opened 2019-12-21
[power] small armored node

Quezler:
> prototype

**backstory**
Most players would find it rather annoying if an air unit breaks through the defences of a large base which in turn dissect all small power nodes, while the large power node has quite a lot more health (which i only found out halfway through this commit 😳), this powernode is aimed to securely deliver power over long distances without to much worry in regards to in defence, and as a slight bonus it somewhat acts insulated since it can only connect in straight lines, which may or may not prevent some accidental connections.

**caveats**
- currently uses duplicate code
- two power connections in the same direction are possible, resulting in loss of a potential side
- should probably use a node/block property instead of a specific block check
- not sure if the white/glass makes the texture to detailed or not
- uncertain as to how useful this block will be in practice
- probably more 

Screen Shot 2019-12-21 at 08 04 19
Screen Shot 2019-12-21 at 08 04 01


CA6:
A) It should have a graphite cost, due to graphite being described as insulting.
B) It should have a difrent connect sprite.
C) It should work on a contact way and be able to connect normal power nodes.
D) I like the idea of it and this would make life simpler.

Quezler:
`A)` would make sense, by the time thorium & metaglass are both reached most lines are already in place, though it should not make it _to_ early game that most could skip over it with little effort
`B)` what did you have in mind? would love to hear it :)
`C)` contact mode still works as usual, any lines coming out from it must simply be stright
`D)` is D really a point?

CA6:
No D) wasn't 
And the reason for C) was because I was thinking of more of a bridge like sprite

Quezler:
hmm, not sure how it would look bridge-y

Quezler:
i had something in mind just now but that would make it look & function way to similar to a surge tower

Violet-Scarelli:
I'd like to suggest a few possible tweaks:
The power laser this thing generates should be thicker than other lasers and tinted more blue/purple, signifying that it's an armored connection.
It should have a small portion of battery capacity, as it's designed to be a front-line block, and if one of its fellows get destroyed, the network it's supplying shouldn't be completely helpless.

Quezler:
coloring lasers can sometimes be a bit tricky due to the base laser color, but it would make sense.

not sure how well the code will handle a powernode that is also a buffer, but its a neat idea though.

Arkanic:
@CA6 when i made the graphite description, i kinda intended for heat/electrical insulation, not kinetic/attack "shielding" insulation

CA6:
Yeah I was unsure about my sprite idea but I like violets idea much more

Arkanic:
i think the dark-shiny lines need a slight colour increase, but other than that, this is something i could find myself using

Quezler:
This pull however is a bit stale, i don't like my texture and the code, but if it stays open i'll look into revising it more for `6.0`, it currently doesn't feel as though it has a place in the game as-is 🤔

Anuken:
Yeah, I don't think this is too useful in its current state. Perhaps the game could use an ultra-long-range armored beam transport block instead, or something of the sort.

Dexapnow:
It would be better if it was 2x2 and with range close to surge tower because I don't see myself using something of this scale in endgame. Also maybe witch more of pint-to-point mechanic (just like 3.5 lasers used to be)

USMP-lancer:
ayo i found a pr for v3 power lasers, sick dude

PR #1242, opened 2019-12-21
[code] remove duplicate condition from if in launchpad

Quezler:


Anuken:
Fixed manually to avoid conflicts.

PR #1244, opened 2019-12-21
[multiplayer] techtree

Quezler:
> prototype (no negative reactions plox, its not meant as a serious merge candidate in this stage)

**random list**
- a rough draft that makes the techtree available in custom games & multiplayer
- unlocks some blocks by default so you can actually launch stuff to research with
- due to me not being smart enough only the server host can research (not player/admins)
- launched items are essentially lost and can never be called down again
- probably a bit unstable and messy, but it works somewhat reliably, feel free to pull & test :)
- custom rule dialog should probably use an interactive techtree instead of a bannedblocks
- should probably have an alert or something if another player researches something
- maybe requiring a vote on what to spend the riches on when playing multiplayer?
- probably needs some balancing since all the work does not get carried over, yet
- other stuff probably
- oh and here are some pics from the custom game (aka: not campaign) V

Screen Shot 2019-12-21 at 10 39 32
Screen Shot 2019-12-21 at 10 39 42
Screen Shot 2019-12-21 at 10 39 50

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -767,6 +767,7 @@ rules.title.unit = Units
 rules.title.experimental = Experimental
 rules.lighting = Lighting
 rules.ambientlight = Ambient Light
+rules.techtree = Use tech tree
 
 content.item.name = Items
 content.liquid.name = Liquids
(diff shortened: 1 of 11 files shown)


Arkanic:
oh no

Quezler:
:ohno:

CA6:
:oh ok:

Anuken:
I'll take a look at this if a server-based campaign is ever implemented. Until then, I don't think this will be merged.

PR #1245, opened 2019-12-21
[drone] firefighter

Quezler:
> prototype (not backwards compatible, liquid module got added to a block)

This pull allows spirit repair drones to optionally firefight by supplying their factories with the appropriate liquid.

caveats:
- laser/refueling looks ugly
- unit carrying a liquid uses the droplet
- spray does not look that nice
- empty factory liquids locks the drones in place
- will also try to extinguish fires in enemy bases
- improper liquid handling 
- hacky weapon logic
- other stuff i'm forgetting

Dec-21-2019 13-35-58

https://feathub.com/Anuken/Mindustry/+87


CA6:
Would introduce some offensive capabiltys like them seeking out enemys if you say put oil or slag in them... tho separate it from the spirit drone as it can have other issues

Anuken:
This is a good proof of concept, but the code here won't be usable since I'll be implementing the mechanics differently.

PR #1246, opened 2019-12-21
Let version string be translatable.

DeltaNedas:
Version string can now be half translated (version type ("release") cant be translated right now, will add if you want it).

See below for example (not in pr for obvious reasons :])
screenshot


Arkanic:
nice

DeltaNedas:
anuke says no

PR #1249, opened 2019-12-21
[chat] dynamic team color

Quezler:
> prototype, this can be done way better and less hacky, probably, this gets the ball rolling

All this does is replace `[teamcolor]` in the player's name with the color of the team they are in at the time of sending a message, though dynamic variables such as those should probably use something shorter and be usable without typing it yourself, also it shows the tab menu wrongly now, obviously 😗 

Screen Shot 2019-12-21 at 17 07 29
Screen Shot 2019-12-21 at 17 11 44

https://feathub.com/Anuken/Mindustry/+487

### DIFF
--- a/core/src/io/anuke/mindustry/core/NetClient.java
+++ b/core/src/io/anuke/mindustry/core/NetClient.java
@@ -196,7 +196,7 @@ public class NetClient implements ApplicationListener{
     public static String colorizeName(int id, String name){
         Player player = playerGroup.getByID(id);
         if(name == null || player == null) return null;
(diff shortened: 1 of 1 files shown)


DeltaNedas:
why not just [team]

Quezler:
probably to generic, and i just copied from the feature suggestion 😛

CA6:
i like it and would make it easier to identify team mates by chat but then again /t

Anuken:
This won't truly be viable unless I rewrite the font renderer to allow for more complex tags. Shelving for now.

PR #1256, opened 2019-12-24
Add "Reload Mods" button to Mods dialog.

DeltaNedas:
Currently, to reload mods you need to either:
* Restart Mindustry
* Disable a mod, scroll down to find it and enable it

This pr adds a Reload Mods button that does this all in 1 click.
(Ideally it would just "overlap" the Reload Required text when its not visible but we can't have nice things as it's beyond my knowledge)
See video for use [here](https://cdn.discordapp.com/attachments/396416151032299521/659121850630012949/Capture_195342.mp4).


Arkanic:
i was just about to do that :)

DeltaNedas:
currently closed because anukes rewriting mod ui stuff soon

PR #1260, opened 2019-12-27
[conveyor] plastanium

Quezler:
> prototype, posting in this state since people seem genuinely interested, so open sourcing it :)

Adds a `plastanium` conveyor with a high(er) throughput than its counterparts, items on it get transported by roomba-like `crater`s. 
(for more info check the discord's `#development` around/before the time of posting this pull)

- code: unstable, messy, plenty of bugs, hacky & edge cases
- sprite: cute as hell
- balanced: barely
- foo: bar

(i have tried higher throughput conveyors before but i ran into the conveyor code wall, and by random luck the train thingy i made recently sparked the idea of moving items on conveyors via entities, bypassing said code entirely, though this is arguably quite heavy because of the increased entity count, while it lacks tons of features, its a nice draft to showcase & gather feedback)

Some graphics of how it looks: (note that it has no power/item cost for each crater)

Screen Shot 2019-12-27 at 20 47 02 1
Screen Shot 2019-12-27 at 20 47 09
Dec-27-2019 20-53-09

> _comming soon to ~~stores~~ conveyors near you!_ (well, if anuke likes it that is 🙃)


ZkyLB:
i want to compile this but i cant

Quezler:
thats because you need a higher arc version, i had this in my working tree which i did not commit on purpose since the master branch was already above this one at the time of making the pull:

Screen Shot 2019-12-28 at 09 35 51

> and i didn't want to deal with a merge conflict

ZkyLB:
well yes but i actually cant compile anything because of my sucky computer

Quezler:
branch should be fixed now though ^

ph3rr3t:
I notice in the example gif, the 1st Crater loaded with Plastanium gets pushed off the belt and vanishes.  Has this been fixed / taken into account?

Quezler:
> may the odds be ever in your favor

well sadly he got hungergamed, they die if they leave the conveyor, deleting their cargo
(they aren't yet patient enough to avoid collisions)

CA6:
I love it but I feel as in the vain of mindustry
They should have a power or item cost and or have to do a return to origin thing on a separate belt but I agree they are cute as frik

joshuaptfan:
Mechanically, this is just a higher throughput conveyor with extra steps.

Quezler:
> state of the ~~union~~ roomba (oh and happy newyear 🎉)

- changed from being entities to being drawn onto conveyors
(so not all of them have to get synced over the net and such)


- has more health than titanium conveyors, but less than armored conveyors
(since its more complicated, and armored conveyors should still be an option)

- optimal (unload) throughput (based on if craters have to wait for one another when unloading)
```
normal:      8, one titanium conveyor
overdrived: 12, two titanium conveyors
phased:     18, three titanium conveyors
```

- situational
(for best unboosted throughput you should just use titanium conveyors, and armored if you need health, only if those conditions aren't met this comes out up top)

- mergability 
(code quality should be good enough in a few days to merge, though thats entirely up to anuke)

- optimization
(not exactly sure how to best measure this vs the current conveyor code, but it feels as though its more optimized, but i might be completely wrong)

- screenshot
(the only added stat menu right now, disregard the ones up top)

Screen Shot 2020-01-03 at 13 01 43

> 🥔

TsjipTsjip:
+1 and following this.

Balancing looks good, haven't looked into the code changes just yet.

core/assets/bundles/bundle.properties
```diff
@@ -931,16 +931,15 @@ block.hail.name = Hail
...
 block.armored-conveyor.name = Armored Conveyor
-block.armored-conveyor.description = Moves items at the same speed as titanium conveyors, but possesses more armor. Does not accept inputs from the sides from anything but other conveyor belts.
 block.junction.name = Junction
 block.router.name = Router
 block.distributor.name = Distributor
 block.sorter.name = Sorter
 block.inverted-sorter.name = Inverted Sorter
 block.message.name = Message
 block.illuminator.name = Illuminator
-block.illuminator.description = A small, compact, configurable light source. Requires power to function.
```
Quezler: moved a few block descriptions to where the others are

core/src/mindustry/world/blocks/distribution/BaseConveyor.java
```diff
@@ -0,0 +1,94 @@
...
+import mindustry.world.*;
+import arc.graphics.g2d.*;
+import mindustry.graphics.*;
+import mindustry.world.meta.*;
+import mindustry.world.blocks.*;
+import mindustry.entities.type.*;
+import mindustry.entities.traits.BuilderTrait.*;
+
+import static mindustry.Vars.tilesize;
+
```
Quezler: extracted some conveyor code into a base conveyor class so i could use it in this fork more cleanly compared to a sub class of conveyor and/or straight up duplicate code

core/src/mindustry/world/blocks/distribution/BaseConveyor.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public boolean isAccessible(){
+        return true;
+    }
+
+    static abstract class BaseConveyorEntity extends TileEntity{
+        int blendbits;
+        int blendsclx, blendscly;
+
+        float clogHeat = 0f;
```
Quezler: not exactly sure if clogheat belongs in here, but its being used in draw and idlesound stuff

core/src/mindustry/world/blocks/distribution/CraterConveyor.java
```diff
@@ -0,0 +1,271 @@
...
+    // has no crater conveyors facing into it
+    private boolean isStart(Tile tile){
+        Tile[] inputs = new Tile[]{tile.back(), tile.left(), tile.right()};
+        for(Tile input : inputs){
+            if(input != null && input.getTeam() == tile.getTeam() && input.block() instanceof CraterConveyor && input.front() == tile) return false;
+        }
+
+        return true;
+    }
+
```
Quezler: these checks could/should probably interact with super's (undocumented) blendbits

core/src/mindustry/world/blocks/distribution/ItemConveyor.java
```diff
@@ -59,7 +43,9 @@ private static int compareItems(long a, long b){
     @Override
     public void setStats(){
         super.setStats();
+
         stats.add(BlockStat.itemsMoved, speed * 60 / itemSpace, StatUnit.itemsSecond);
+        stats.add(BlockStat.boostEffect, "$blocks.itemsmoved");
```
Quezler: made it more clear what a boost effect does to conveyor blocks
(really confused me when i was a starting player that conveyors were affected as well)
and it makes the boost entry for the pull's conveyor less out of place

Anuken:
I really don't think these should be called "craters". I've seen several people get confused by the word's meaning already, and it doesn't make a lot of sense since crates aren't used directly.

Quezler:
@Anuken  Ah, i thought it nicely fit the theme of how ground units were called, `crater` is not to far of from `eruptor` (both named after parts of a volcano), and it contains items so its a nice reference to a crate as well, but i understand the confusion, especially since there is a map called craters, but since its not actually an unit anymore it would make sense to remove its reference in the bundle. (since tons of units was probably not the way to go)

Are you okay with `crater` being referenced in the code, or would you prefer if that got changed to `roomba/batch/etc` as well?

(also, do you like this idea in general? just curious 🙂)

Quezler:
Some small remarks found while playing on the test server: (will probably work on those)
- they need to be in view in order for the ~~craters~~ items to move apparently
- if the line is just one tile it can input from the front
- speed probably needs tweaking (& maybe in sync with all of them)
- capacity/throughput when overdrive may need a slight upgrade

CA6:
Couriers would work as a good name the block could be called something like
 courier roads 
A conveyer like building but it has striped crawler frames running items down it

That s my two cents and I love there adorable nature ether way

Quezler:
@CA6 well anuke basically said he didn't want them to have names since they are not actually units, but since the block is basically a conveyor its probably best to keep that named after `plastanium`

CA6:
@Quezler that was just a suggestion but what anuke not wanting them to have a name makes sense as you'd rarely be able to refer to  them with out having to refer to the plastium compression conveyer thingy anyhow

Quezler:
~~well personally i'll probably keep calling them craters until the end of time~~

but essentially `the plastanium conveyor(s)` is currently what users could/should call them 🤔

CA6:
__I'll most defanatly keep calling them item roombas__


That makes sense

Violet-Scarelli:
Alright, I tried it, and it is incredible. However, given that it's a completely new type of system, I would suggest new types of item flow control, seeing as the transport paradigm for this has been moved from "items on belt" to "robots carrying items". 

Noted bugs:
two streams of courier robots cannot merge properly, different from the weaving mechanic of regular conveyors. 
A robot can get stuck at the start of a plastanium conveyor chain, which makes the conveyor unusable until the robot has been filled so it can leave. 
Possibly more? Haven't been ultra thorough as of now.

Suggestions are as follows:
Proper routing and sorting. Right now, the routing/sorting system for the robots relies on the item conveyor systems, which can get a little janky and tends to clog the system a little. If the robots could have their own sorters and routers, it would keep things moving and possibly allow greater logistic control. These components would most likely look like regular routers, but with plastanium edges or something similar.
Faster loading/unloading with a proper block. This is mostly self-explanatory, something like a distributor but with plastanium highlights that can hold a set amount of robots and dumps all their items like a distributor would to all attached normal conveyors or containers faster than the robots unload (or even use it as a hub to split robot payloads up quickly).
Robots have a timer for getting filled. If they receive no new items within a certain amount of time, they automatically depart. 
Smarter pathfinding, so robots can wait for one another at intersections. This would solve the merging problems.

Quezler:
> visual amount representation instead of numbers (shown: 18, 12, 8 & 1)

Screen Shot 2020-01-07 at 11 29 42

CA6:
so this seems like a a to b type of transport not a split or add in style. that said it shouldn't be less so at least able to handle merging. how you could do this is a if then statement so the merge point is synced be the origins. so if connection to another origin is made then one side (preferably the less full side) becomes the "waiter" the other sends more. but if they are equal in input then they send with a synced pause. that pause is about the rate of the curator drones send rate for them to merge properly.

CA6:
bruh get pushed back

jakyote:
Probably not helpful but what if the "crater" was a block that just packed items (using power and maybe sacrificing 1 item/package?) into a "package" item that could be output to any conveyor/block/etc? perhaps also requiring the unloader (or a new uncrater) to unpackage them?

Quezler:
That would probably add to much blocks that only have a single purpose, and a lot of overhead 🤔

CA6:
hmmm yeah that is not as clean cut and less satisfying as what is already in the pull

Violet-Scarelli:
This got the 6.0 stamp so I'm guessing that's good

CA6:
Yeah it is.... saw on the discord that it is basicly a eventually thing it will happen just when

Quezler:
not exactly, it'll be reviewed when 6.0 rolls around & merged if it works well, according to anuke 🙂

Quezler:
i will have to rework this pull due to the recent conveyor optimizations, postponing temporarily 🤔 
(since more are probably on their way since it looked glitchy during playtesting, so waiting 🙂)

Quezler:
scratch that, i tried making a hjson mod for it, but that was infinitely more complicated than dealing with this merge conflict by hand 😗

Quezler:
my changes/use of the `ItemModule` is probably incompatible with #1495

Quezler:
https://github.com/Anuken/Mindustry/issues/1637 https://github.com/Anuken/Mindustry/pull/1260/commits/ecb09c0f046fb4c3be7f12aa2ae338dfc046b110 <-- bodged implementation that i really wanted in here 😗

Quezler:
note: in the commits today i tried to blend it with normal conveyors, but failed to find a proper way.

Quezler:
and i will definitely update this one to how `6.0` structures tile/entity/update logic 👑

Quezler:
> <initial port to 6.0 (contains some master commits that haven't been merged into the branch)>

Quezler:
Took this pull out of draft mode in case @Anuken wants to merge & adapt it into the official 6.0 🤔

Anuken:

DeltaNedas:
the meek shall inherit the earth

Quezler:
i'll repack the conflicting sprites once i research the layer changes blocking it:

Screen Shot 2020-04-23 at 20 50 15

bvcbvc:
i like the orginal animation for it, has a bit of realism to is i guess

Quezler:
> i like the orginal animation for it, has a bit of realism to is i guess

the original animation actually used units, moved away from that one since it would have never been merged as-is 😅 
(net syncing / performance heavy / collisions / desyncs / etc)

bvcbvc:
> > i like the orginal animation for it, has a bit of realism to is i guess
> 
> the original animation actually used units, moved away from that one since it would have never been merged as-is 😅
> (net syncing / performance heavy / collisions / desyncs / etc)

no wonder it looks so real, and probably would have been that buggy lol, still i think it would have been a nice cosmetic feature to have it have a little it of offset, the rng could be client side, since it would not effect gameplay

LogicalMelon151:
can i get one as a pet?

deltanedas:
mindustry indonesia

LogicalMelon151:
> mindustry indonesia

🇮🇩

PR #1261, opened 2019-12-28
check if Votekick's voteRequired has been reached every vote

blackassasin1234:
As the title implies, this PR makes it so that it doesn't check if the votes are more than required only when the timer has ran out, but also when someone votes.


XEN0PHIL:
Votekick should NOT kick as soon as it reaches 3/3 because there might be people who want to vote against the kick.

DeltaNedas:
this
no, closed permanently

Anuken:
I'm going to use my own implementation.

Quezler:
what do you roughly have in mind?

Anuken:
basically the same thing, but with some more testing

PR #1265, opened 2019-12-28
web move dialog

AmateurPotion:
I don't want to be suddenly moved.


Anuken:
I'll just change the icon to a link.

DeltaNedas:
what even is that pr

PR #1270, opened 2019-12-29
PlayerChatEvent change

Kieaer:
This allows the plugin to take care of when the player speaks abusively.

example

and

and can this (prefix)
69009041-c26db180-0994-11ea-808f-5625b5d0d8f1


AmateurPotion:
앗아 직접 만드셨구나..

Kieaer:
omg please accept :sob:
This can be used not only for abusive language but also for special events.

Anuken:
Implemented in a different way with latest commit. Having an `original` field didn't make much sense.

PR #1271, opened 2019-12-29
Add ServerLoaded event

Kieaer:
This event fires after the server loads.
example
It work.


Kieaer:
duplicate ServerLoadEvent

PR #1279, opened 2019-12-30
Overall improvement of Czech translation

PetrGasparik:
Translation with line to line comparison with english counterpart. Up to line 297 now.


PetrGasparik:
yes, there is more recent. I started to make PR from my fork

PR #1286, opened 2019-12-30
Display cumulative production speed of multiple blocks

mode89:
Hi,

I would like to propose a new feature. I think, it might come in handy to have a display showing the current production speed of multiple blocks. For example, one might use it for estimating the exact number of the upstream consuming blocks that need to be built.

Here is how it works:
* Press CTRL
* Drag the mouse to select the blocks of interest with the bounding box
* A window pops up in the bottom left corner, showing the cumulative production speed per resource.
* To discard selection, press CTRL again

video

I've made this pull request as a draft implementation. I'm not sure, whether this feature fits your vision of the game. I would like to hear your comments.


Quezler:
it currently only works for drills?

mode89:
Yeah. As for the initial iteration, I decided to make it simple and implement drills only. It shouldn't be hard to scale to the other kinds of blocks.

I think, if the code is fine, the current functionality can be used right away, and the other blocks can be included later. But, if it's necessary, I would be happy to do the implementation for other blocks here as well.

mode89:
Also, I am not sure, whether I selected right UI and control scheme.

famo:
I like the idea of seeing the production sum of several buildings.
However I think it should only display the theoretical maximum, taking into account boost state (water and/or overdrive) and number of ore fields below a drill.

It is not showing in your screencast, but since the production is non-continuous I suspect that in a non-paused game the value would have a lot variance. And hence wouldn't be really readable / usable.

For measuring the actual economy I would favor a dedicated counting transport block (for conveyors and conduits), which can be placed by the player at any point in his production complex. This counting gate would display the values from an average of like 5 seconds in the HUD.
(For statistic nerds it could even show the total items counted and time elapsed since placed on the map.)


See also:
https://feathub.com/Anuken/Mindustry/+575

mode89:
Thanks @famo. Yeah, now I see it. You are right: the theoretical maximum would work better.

Anuken:
I don't think this feature should be in the base game.

famo:
@Anuken would you be open to a **_block_** that counts the items and calculates the throughput on a conveyor/conduit?

See also my comment above:
https://github.com/Anuken/Mindustry/pull/1286#issuecomment-572775380

Anuken:
No. All transport blocks in 6.0 are already capable of calculating resource flow rate.

famo:
Ah, interesting! Is there a preview of how this looks like?

Anuken:
It's not done, but I posted a preview on Discord some time ago.

PR #1290, opened 2019-12-31
UX improvements for showTextInput

davidmfritz:
- Added keyDown support (Enter, Escape, Back)
- Made Keyboard cursor auto-focus on TextField end

### DIFF
--- a/core/src/mindustry/core/UI.java
+++ b/core/src/mindustry/core/UI.java
@@ -307,15 +307,15 @@ public class UI implements ApplicationListener, Loadable{
                     hide();
                 }).disabled(b -> field.getText().isEmpty());
                 buttons.addButton("$cancel", this::hide);
-                this.keyDown(KeyCode.ENTER, () -> {
+                keyDown(KeyCode.ENTER, () -> {
                     String text = field.getText();
                     if(!text.isEmpty()){
                         confirmed.get(text);
                         hide();
                     }
                 });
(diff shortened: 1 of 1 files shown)


core/src/mindustry/core/UI.java
```diff
@@ -307,6 +307,15 @@ public void showTextInput(String titleText, String dtext, int textLength, String
                     hide();
                 }).disabled(b -> field.getText().isEmpty());
                 buttons.addButton("$cancel", this::hide);
+                this.keyDown(KeyCode.ENTER, () -> {
```
Anuken: Don't use `this` unless necessary. You use it here 3  times without a purpose.

davidmfritz: Thank you for the notice. Didn't know this wasn't necessary in Java.
I pushed the changes for further investigation.

core/src/mindustry/core/UI.java
```diff
@@ -316,7 +316,10 @@ public void showTextInput(String titleText, String dtext, int textLength, String
                 });
                 keyDown(KeyCode.ESCAPE, this::hide);
                 keyDown(KeyCode.BACK, this::hide);
-            }}.show();
+                show();
```
davidmfritz: show() call had to be moved inside instead of daisychaining, because the TextField would not be available at this point to set the focus on.

PR #1295, opened 2019-12-31
Updated some german translations

Aaron2550:
block.tar.name = Teer is right, but the only thing you get from tar is oil so why not name it oil?


MetaMuffin:
Right... That should be changed

PR #1296, opened 2019-12-31
Server Autosave Feature

Killili:
With this the server can make autosaves on each wave. 

- Adds Autosave Feature
- Adds configuration Option

This really helps finding out why your unattended base exploded ;)

### DIFF
--- a/server/src/mindustry/server/ServerControl.java
+++ b/server/src/mindustry/server/ServerControl.java
@@ -173,13 +173,17 @@ public class ServerControl implements ApplicationListener{
         });
 
         Events.on(WaveEvent.class, e -> {
-            if(state.is(State.playing) && Config.valueOf("enableAutosave").bool() ){
+            if(state.is(State.playing) && Config.enableAutosave.bool()){
                 Fi file = saveDirectory.child("autosave-" + Strings.capitalize(world.getMap().name()) + "-" + state.wave + "." + saveExtension);
-
+                
                 Core.app.post(() -> {
-                    SaveIO.save(file);
-                    Call.sendMessage("[scarlet][Server]:[] - Autosave Done");
-                    info("&lyServer: &lb{0}", "Autosave Done");
+                    try{
+                        SaveIO.save(file);
+                        info("&lbAutosave completed.");
+                    }catch(Exception ex){
+                        String msg = "Autosave failed with: " + ex.getMessage();
+                        Log.err(msg);
+                    }
                 });
             }
         });
--- a/server/src/mindustry/server/ServerControl.java
+++ b/server/src/mindustry/server/ServerControl.java
@@ -174,12 +174,17 @@ public class ServerControl implements ApplicationListener{
 
         Events.on(WaveEvent.class, e -> {
             if(state.is(State.playing) && Config.enableAutosave.bool()){
-                Fi file = saveDirectory.child("autosave-" + Strings.capitalize(world.getMap().name()) + "-" + state.wave + "." + saveExtension);
+                Fi dir = saveDirectory.child("autosaves");
+                Fi file = dir.child(Strings.capitalize(world.getMap().name()) + "-" + state.wave + "." + saveExtension);
+                Fi oldSave = dir.child(Strings.capitalize(world.getMap().name()) + "-" + (state.wave-5) + "." + saveExtension);
                 
                 Core.app.post(() -> {
                     try{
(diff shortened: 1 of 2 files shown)


Quezler:
when playing a map with the same name, older saves get overridden?

server/src/mindustry/server/ServerControl.java
```diff
@@ -172,6 +172,18 @@ public ServerControl(String[] args){
             }
         });
 
+        Events.on(WaveEvent.class, e -> {
+            if(state.is(State.playing) && Config.valueOf("enableAutosave").bool() ){
```
Anuken: Why do `Config.valueOf("enableAutosave")` when you can do `Config.enableAutosave`? You also have an unnecessary spaces after `bool()`.

server/src/mindustry/server/ServerControl.java
```diff
@@ -172,6 +172,18 @@ public ServerControl(String[] args){
...
+                Fi file = saveDirectory.child("autosave-" + Strings.capitalize(world.getMap().name()) + "-" + state.wave + "." + saveExtension);
+
+                Core.app.post(() -> {
+                    SaveIO.save(file);
```
Anuken: This should be `try-catch`-ed and `Log.err` the exception, if any occurs. An autosave failure due to lack of disk space or an IO error shouldn't crash the server.

server/src/mindustry/server/ServerControl.java
```diff
@@ -172,6 +172,18 @@ public ServerControl(String[] args){
...
+                    Call.sendMessage("[scarlet][Server]:[] - Autosave Done");
```
Anuken: I don't think players should know or care about the autosave. Don't send a message about it.

server/src/mindustry/server/ServerControl.java
```diff
@@ -172,6 +172,18 @@ public ServerControl(String[] args){
...
+                    info("&lyServer: &lb{0}", "Autosave Done");
```
Anuken: Why prefix this with `Server:`? The console log isn't a chat, and this formatting seems unnecessary since the argument is always the same. Just print `info("&lbAutosave completed.")`.

server/src/mindustry/server/ServerControl.java
```diff
@@ -172,6 +172,22 @@ public ServerControl(String[] args){
...
+        Events.on(WaveEvent.class, e -> {
+            if(state.is(State.playing) && Config.enableAutosave.bool()){
+                Fi file = saveDirectory.child("autosave-" + Strings.capitalize(world.getMap().name()) + "-" + state.wave + "." + saveExtension);
+                
+                Core.app.post(() -> {
+                    try{
+                        SaveIO.save(file);
+                        info("&lbAutosave completed.");
+                    }catch(Exception ex){
+                        String msg = "Autosave failed with: " + ex.getMessage();
```
Anuken: `Log.err("Autosaving failed.", ex)` would be preferable. Exception messages can be `null`, and they usually don't tell you what went wrong.

server/src/mindustry/server/ServerControl.java
Killili: Copy pasted the wrong code apparently.

Killili:
Any suggestion regarding the file name issue? I noticed "save" makes backups so i thought that would suffice?

server/src/mindustry/server/ServerControl.java
Killili: You are right

Killili:
@Anuken updated to current head.
Am i missing something or why does it still say Changes Requested?

Anuken:
However, I'm not sure if saving every wave is a good idea, since many servers don't run survival, and I'm also not sure if saving in the original save folder while without deleting older map backups will work well.

DeltaNedas:
maybe have saves > autosaves > mapname_wave_time.msav

Killili:
It is off by default, so everbody that wants to use it will have to activate it.
So its not a problem for everyone else.

I could add a save rotation like, save the last 10 waves and delte old ones?

Killili:
Update feature to:
-Use a sub directory for autosaves (saves/autosaves)
-Keep only the last 5 waves (delete autosave-5 if it exists)
-Sort savegames (autosaves and save listing)

Anuken:
Implemented slightly differently.

PR #1299, opened 2020-01-01
move download links higher in README.MD

Arkanic:
currently people have to scroll right to the bottom to see the download buttons


Anuken:
That's a bit too high. I moved it above the suggestions page instead.

Arkanic:
ok

PR #1300, opened 2020-01-01
let boolean and double be used by mods

DeltaNedas:
why do i even have to do this jesus christ
see my meltdown when scripting a multiple weapon mech for more context
before: crash because REEE YOU CANT USE java.lang.Boolean/Double
after: :) it work


DeltaNedas:
see #1302

PR #1301, opened 2020-01-01
make turnCursor: false mechs not cross eyed

DeltaNedas:
before it was hardcoded for a mech to target 1 unit in front of it if turnCursor was false
now, it can be configured (still 1 by default)
targetDistance variable lets you set how far a mech will try and shoot!


DeltaNedas:
@Anuken please like 2 people crashed because they didnt patch this in

DeltaNedas:
actually no not crash but still its unplayable with 1 unit target forced

DeltaNedas:
thank you mr cat

PR #1302, opened 2020-01-01
Set primitive wrapping to false for scripts

KSean222:
Rhino usually converts Java primitives to their JS equivalents. However, Rhino actually converts their boxed equivalents (`Boolean` for `boolean`, `Integer` for `int`, etc.) into `JavaObject`s. These `JavaObject`s aren't allowed in `ClassAccess.java`, which is another thing. For now, this PR sets primitive wrapping to false, which converts their boxed values into the primitive equivalents, so that `Boolean` gets converted to a JS `boolean` and is usable.


DeltaNedas:
outdoes #1300

DeltaNedas:
anuke please

Anuken:
...ok

PR #1305, opened 2020-01-01
Estimate and display conveyor resource flow

mode89:
Motivation

Knowing an approximate amount of resources, traveled through a conveyor block, may give an insight on how many upstream consumer blocks need to be built to exploit the whole flow capacity. Also it may tell, whether the conveyor is going to be clogged, if not enough consumers were built.

The flow per each item is being shown in the PlacementFragment, when the mouse hovers above a conveyor block:
video

Should it be optional for each block? For example, by having a block configuration UI, that turns on/off the function. This may save some CPU time and memory.

### DIFF
--- a/core/src/mindustry/world/blocks/distribution/Conveyor.java
+++ b/core/src/mindustry/world/blocks/distribution/Conveyor.java
@@ -254,22 +254,27 @@ public class Conveyor extends Block implements Autotiler{
         // TODO It seems, the better way is using the real framerate, rather
         // then simply 60
         float flowEstimationBegin = currentTime - flowEstimationWindow * 60.0f;
-        for(Item item: entity.itemHistory.keys()){
-            FloatArray itemHistory = entity.itemHistory.get(item);
-            int removeUpUntilIndex = -1;
-            for(int index = 0; index < itemHistory.size; index ++){
-                if(itemHistory.get(index) < flowEstimationBegin){
-                    removeUpUntilIndex = index;
-                }else{
-                    break;
+        for(ObjectMap.Entry<Item, FloatArray> entry: entity.itemHistory){
+            Item item = entry.key;
+            FloatArray itemHistory = entry.value;
+            if(itemHistory.size > 0){
+                int removeUpUntilIndex = -1;
+                for(int index = 0; index < itemHistory.size; index ++){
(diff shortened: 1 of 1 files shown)


DeltaNedas:
thank you mode89, very cool!

CA6:
I would use it and it being an optional overlay is great. The question is how it acts with mobile as a good portion of players in aware of play on moblie.

mode89:
Hi, @CA6. Sorry, I haven't tested the mobile version. Where do you foresee possible issues would be?

DeltaNedas:
i say you tap a button to enable resource mode, tap a block and then it acts like you just left your mouse on it
when youre done tap it again to disable resource mode and build stuff or whatever

mode89:
Thanks, @DeltaNedas. I will try it out.

mode89:
@DeltaNedas, originally I was thinking of an implementation, that is similar to the UI, that is used for selection of an item type in the sorter block on the desktop version. How about that? Does the mobile version has the same or a similar UI? Sorry, I have never played it on mobile.

DeltaNedas:
you cant add a configure menu to every block
on mobile you tap block then tap button like on desktop

player20033:
I wish you could do this also on bridge and mass drivers.

mode89:
@DeltaNedas, what if we add a button (or a checkbox) to the PlacementFragment, either when building a new conveyor, or when tapped on an existing conveyor block?

mode89:
@player20033, that's a good idea. Seems, it shouldn't be complicated. I will look into the source code. I think, it would be better, if we made it after this pull request got merged. Or do you think, it is necessary to implement it as a part of this PR?

CA6:
@mode89 sorry for the long get back time.
mainly with just the amount of things already on screen so it being toggleable makes it less of an issue 
cheers

mode89:
OK. Thanks @CA6.

Anuken:
This is useful, but will have a major performance impact if left enabled for all conveyors. I'm not a fan of adding a persistent toggle, either. A different solution would be needed.

I'll be looking into implementing this for all item/liquid transport blocks, not just conveyors.

ejmixmaster:
how do I fix my mindustry?

-----Original Message-----

PR #1306, opened 2020-01-03
Update bundle_ru.properties

whert-dev:
Fixed some minor flaws, changed "Копьё" -> "Глефа" and "Пластаний" -> "Пластан" to more correspond original version.


BasedUser:
we do not approve of this change

XEN0PHIL:
1. Direct damage (прямой урон) There's nothing about "direct" damage in original version. If the damage is not direct, it has a special description. For example: area dmg
2. Plastanium (пластаний/пластан). In russian we have short names, for example titanium is titan (титан), but vanadium is vanadiy (ванадий). Plastanium can be both plastan or plastaniy, so we'll keep the one people got used to.
3. Glaive (копьё/глефа). Word glaive can mean both spear and a throwable bladed weapon. In this case we decided to use the more common meaning of a spear, but it may change if Anuke tells us that he meant the throwable melee.
4. Everything else (mostly) is acceptable. We have to wait for Anuke's word.

XEN0PHIL:
UPDATE: Anuke told us to use your new Пластан and keep our previous Копьё.

XEN0PHIL:
Approved changes were merged with "RU updates"
https://github.com/Prosta4okua/Mindustry/commit/2e114d09ea418398368f2ce416ae4da2905cb478#diff-84cd9dbd131353c048e3892ba097c098

Anuken:
Merged the other PR.

PR #1307, opened 2020-01-03
Made power nodes retain connections in schematics

KSean222:
https://feathub.com/Anuken/Mindustry/+42
Still working out all the kinks in the code, leaving this here to see what y'all think.
Demonstration of power node link memory


KSean222:
Should probably rework the code to avoid copious use of `Call` when connecting nodes

KSean222:
Helps with https://feathub.com/Anuken/Mindustry/+42

ado1928:
I see this being abused using many connections to overload players on a server with connection packets

Ajedi32:
IIRC, each node can only have 20 connections each (30 for large nodes). Is that really _that_ much more data than the schematic itself would be?

CA6:
Eh due to size of schmatics it can't be abused to gard

Quezler:
a variation on this (without looking at the code prior) can be found [here](https://github.com/Anuken/Mindustry/compare/master...Quezler:scidustry) 🙂

Anuken:
Implemented in 6.0 branch.

PR #1309, opened 2020-01-03
let mechs override drawStats and do custom power cell stuff

DeltaNedas:
(it doesnt break vanilla mechs dont worry i thought of that)
see it in action https://streamable.com/n5ph5
the custom rotation value is used which means power cell is drawn properly
see https://github.com/DeltaNedas/vbucks/blob/master/scripts/mother-hen.js

### DIFF
--- a/core/src/mindustry/entities/type/Player.java
+++ b/core/src/mindustry/entities/type/Player.java
@@ -273,10 +273,6 @@ public class Player extends Unit implements BuilderMinerTrait, ShooterTrait{
         this.team = team;
     }
 
-    public float getItemtime(){
-        return itemtime;
-    }
-
     //endregion
 
     //region draw methods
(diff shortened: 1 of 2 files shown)


DeltaNedas:
you write 5 line of actual code
Yes.

DeltaNedas:
@Anuken please look at it

DeltaNedas:
anyone have feedback?

core/src/mindustry/entities/type/Player.java
```diff
@@ -273,6 +273,10 @@ public void setTeam(Team team){
         this.team = team;
     }
 
+    public float getItemtime(){
```
Anuken: Instead of making a getter for this, you should add a method to `Player`:
```java
public float drawBackItems(){
    drawBackItems(itemtime, isLocal);
}
```

...and call this method in the mech.

DeltaNedas: gotcha, will update it

DeltaNedas: implemented void drawBackItems in player and mech drawStats

DeltaNedas:
yay

PR #1312, opened 2020-01-03
Make peoples are possible to alliance. Hex war related.

player20033:
It passed to compile, however I don't know how to test or debug does it works.

### DIFF
--- a/core/src/mindustry/game/Team.java
+++ b/core/src/mindustry/game/Team.java
@@ -81,6 +81,18 @@ public class Team implements Comparable<Team>{
     public boolean isEnemy(Team other){
         return state.teams.areEnemies(this, other);
     }
+    
+    public boolean isAllies(Team other){
+        return state.teams.areAllies(this, other);
+    }
+
+    public void addAllies(Team other){
+        state.teams.addAllies(this, other);
+    }
+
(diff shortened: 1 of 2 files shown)


player20033:
I don't understand what it does means.
https://github.com/Anuken/Mindustry/blob/d6d6dc29dc5cca2d702a67e09c78b36b3b84996d/core/src/mindustry/ai/BlockIndexer.java#L158

player20033:
Travis check is failed was false alarm.
So, I need people to help me testing the alliance function.
Related: https://github.com/Anuken/HexedPlugin/pull/2

Arkanic:
this doesn't just affect hexWars, this affects all gamemodes that involve multiple teams, which is a real issue

example:
2 team PvP, if they make an alliance, the game will never end
it will literally never end

player20033:
> this doesn't just affect hexWars, this affects all gamemodes that involve multiple teams, which is a real issue
> 
> example:
> 2 team PvP, if they make an alliance, the game will never end
> it will literally never end

I see, however I don't see a way to help me debug alliance function. and alliance function should be integrated on main game code.

player20033:
I am give up for this, sorry.

PR #1313, opened 2020-01-03
use findAll to iterate through mod content

SimonWoodburyForget:
Recursive search for files in content directory, such that they may be further organized with sub-directories.


DeltaNedas:
anuke please :pray:

PR #1314, opened 2020-01-04
Add preview outline for Force Projectors on hovering

KSean222:
([Demo here](https://gyazo.com/d3531e0eb40ba4af32574988d80ae6d8))
This adds hover preview outlines to force projectors.


Quezler:
hmm, what about showing both ranges (& during placement) so you also know its size if it were phase boosted from that position, without needing to have phase in it to toggle it?

core/src/mindustry/world/blocks/defense/ForceProjector.java
```diff
@@ -32,6 +32,7 @@
     public float cooldownLiquid = 1.5f;
     public float cooldownBrokenBase = 0.35f;
     public float basePowerDraw = 0.2f;
+    private boolean phaseValid;
```
Anuken: This shouldn't be a variable. It's shared for every single block. You can't rely on this being the correct value. Calculate it every time it's needed instead.

KSean222: Is this good now?

Anuken: On second glance, why not use `realRadius(ForceEntity)` instead or calculating it yourself?

KSean222: `realRadius(entity)` would be equivalent to the size of the shield bubble. Since this is already displayed, I opted to show the "max" size, not accounting for overheating or not enough power.

KSean222:
Will break with #1323

Anuken:
Unlike other projectors, force projectors display their radius visually when active. Even when they're broken, you can visualize their radius by middle clicking, so I don't think this is necessary.

PR #1323, opened 2020-01-05
Rotatable Force Projectors

KSean222:
https://feathub.com/Anuken/Mindustry/+543
A rotated force projector
[Demo](https://gyazo.com/114288dfc3a4669de9da62d607adfa94)
Breaks #1314


famo:
Cool! :smiley: 

Additional checks:
Is the set rotation also saved in schematics and does the force field also rotate with the schematic while placing it?

E.g. common use case you have a wall and a force projector in a schematic.

KSean222:
@famo Yes, the force field rotates and it's saved in schematics. The rotation is saved just like all other rotatable blocks.

KSean222:
Current state of PR:
Preview demo
Rotation demo
Schematic demo
Collision demo
Collision demo 2
Simple mode demo
Backwards compatibility demo
(Backwards compatibility demo)
I think we're ready 😛

ThatOneBepis:
A cool QoL if you ask me

Anuken:
I'm just not a fan of this idea. The whole concept of 'rotating' a force projector's shield simply *feels* like a strange thing to do, as it undermines the reason for making the shield a hexagon in the first place. 

If I wanted shields to be easy to use, they could have just been squares or circles; with hexagons, you have to put a little thought into how you're going to space out your shields/defenses for the optimal hex grid to cover your base. A rotated grid throws that out the window, as you can now have flat rectangular edges everywhere.

ZkyLB:
i agree with the ~~naughty~~ cat

famo:
@Anuken 
Sorry, but this seem kind of inconsistent to me. So we can have flat edges on one axis (horizontal), but not on the vertical axis?

This gets especially awkward with 4 player PvP maps like the one below, where two players have a horizontal line of defense and the other two a vertical one:
a

If you don't want flat lines on force projectors, I think using circles or a slightly rotated octagons would be the best and most fair solution.

PR #1324, opened 2020-01-05
Fix Steam Achievement italian translation error

alexmaurizio:
As reported on the discord channel, here's the fix for the Steam Achievement italian translation of "Lone Guardian" which was translated in "Guardiano dell'amore" (which is LoVe Guardian) that did not make sense.


alexmaurizio:
Added another translation error fix relative to Revenant Factory

PR #1326, opened 2020-01-05
[sandbox] ability to switch team

Quezler:
> proof of concept, ui is dirty, code could probably need some work, etc, etc

While testing out the plastanium conveyors on my test server the other day (well, actually today) the thought occurred to me that i could merge my other concepts in there as well (e.g. the tractor beam), but that would only be fun to play with if players could switch between teams.

This is currently only possible if the mode is set to editor, but it would make sense to allow this for sandbox mode (though not by default) as well.

But to prevent abuse its possible to allow only admins to switch teams (in this initial draft)

Not all quadrillion team colors are supported, just the basis 6.

Screen Shot 2020-01-05 at 15 09 49
Screen Shot 2020-01-05 at 15 11 44

> the player option greys out if admin below it is selected* 🥔

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -777,6 +777,8 @@ rules.title.unit = Units
 rules.title.experimental = Experimental
 rules.lighting = Lighting
 rules.ambientlight = Ambient Light
+rules.bipartisan.all = Players can switch teams
+rules.bipartisan.admin = Admins can switch teams
 
 content.item.name = Items
 content.liquid.name = Liquids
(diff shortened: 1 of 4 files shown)


Arkanic:
yes

CA6:
Hmmmm

Anuken:
I would like to add this someday, but it'll have to wait for the UI rework.

Quezler:
ok ❤️

Anuken:
Not much left to discuss here, so I'll put this on the Trello and close this PR.

Quezler:
as you wish 🙂

PR #1329, opened 2020-01-06
[6.0] conduit pump connection

Quezler:
> prototype, it should probably use blendbits, and this example only connects from front & back.

Screen Shot 2020-01-06 at 10 58 56
Screen Shot 2020-01-06 at 11 00 51

- [ ] use blendbits somehow
- [x] figure out what should happen if multiple lawns would draw over the same corner
- [ ] cleanup the code
- [ ] research what impact it has on rendering
- [x] render connections from the sides as well
- [ ] find out if blocks have a flag to easily check if the texture has transparent bits
- [x] figure out how to draw half a texture, e.g. for the small pump, it would cover all 4 corners
- [x] wait for reactions on this post before doing more


joshuaptfan:
So this will work with modded blocks of any shape?

Quezler:
99% chance ^

Violet-Scarelli:
This is 700% more pleasing than what's in the game right now.

joshuaptfan:
What happens if conduits/conveyors are entering a block's corner from 2 sides?

Quezler:
> [ ] figure out what should happen if multiple lawns would draw over the same corner

currently one would completely draw over the other based on rendering order

Quezler:
possibly relevant commit: https://github.com/Anuken/Mindustry/commit/f46be924b9628c35cc0c805ebd10249647c19a01#diff-f7607809ff1961e63c269e2955c84a8cR10

Quezler:
> started revising it for 6.0, now with better knowledge of the code, currently only works for the front & back of conduits, conveyors & sides are pending, as well as a cleanup of region code:

Screen Shot 2020-04-17 at 14 07 30

> uses some form of magic to cut the existing textures in two, so the small pump looks good ^

Quezler:
Tagging #1884 since they are somewhat related, and this one is buried in the date sorted pulls 🗝️

Quezler:
Screen Shot 2020-04-17 at 14 24 09

Quezler:
> it now works for conduits, but the code is a mess, and i am not seeing the obvious solution 😶 

Screen Shot 2020-04-17 at 16 47 11
Screen Shot 2020-04-17 at 16 47 16

Quezler:
i cannot think straight right now, so i'll just await a review of @Anuken, specifically for:
- is this idea worth looking into more at all, or would you rather close it?
- should i actually make the split textures in the generator or loader?
- any tips on clearing up that ghastly block of rotation code?

Quezler:
shower thought: alternatively it could be turned into a draw chance on the pump block (or all non-square liquid outputting blocks), which then would somehow always be called despite blocks overriding most stuff, so I'm not entirely sure where to move it to if any 🤔

Anuken:

Quezler:
guess i'll have to look into the recent rendering changes/conflicts ^

Quezler:
front texture apparently is slightly longer when in a corner, will investigate:
Screen Shot 2020-04-24 at 11 52 48

BlueWolfGaming:
i literally just read everything in this pull to see if i could implement it into a dark theme mod... guess ill just resprite the pumps for now xD

Quezler:
thats probably the easiest approach for now ^

Anuken:
This took a lot of fiddling around, but I managed to implement it without making the code too terrible:
image

image

PR #1330, opened 2020-01-06
[drones] miner drone dropoff

Quezler:
> proof of concept, code is not all that serious, this is to kickstart the topic :)

instead of placing the miner drones in the far edges of the map to get free minerals until the end of time, changes them so you actually have to take the items out of them one way or another.

its been discussed a few times to have a dedicated drop off point, but it would probably make most sense to make them drop off at their spawnpoint or cointainers right next to it.

this probably won't go well for drones that require items to spawn since it will unload.

but this would allow for some more precise mining drone usage, since you can get the right amount of copper/lead from anywhere in the map, instead of finding a patch or unloading it from the core 10 miles away.

Screen Shot 2020-01-06 at 12 33 12

🥔

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1627,7 +1627,7 @@ public class Blocks implements ContentList{
         //endregion
         //region units
 
-        draugFactory = new UnitFactory("draug-factory"){{
+        draugFactory = new MinerFactory("draug-factory"){{
             requirements(Category.units, ItemStack.with(Items.copper, 30, Items.lead, 70));
             unitType = UnitTypes.draug;
             produceTime = 2500;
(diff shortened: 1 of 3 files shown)


joshuaptfan:
I'd say for copper and lead mining, the current implementation is ideal because:
* Dropping off at the core means no space needed for conveyors
* Unlimited throughput due to not using conveyors

Which is fine for copper and lead because current game balance expects copper and lead to be abundant.

Dropping off at spawn seems best for advanced mining drones that can mine titanium, etc. because the inconvenience of having to run a conveyor from the drone spawn to the core makes it more balanced compared to drills:
* Need to run conveyors just as with drills
* Throughput limited by conveyors (this is why we need to be careful about adding very high throughput conveyors to the vanilla game! #1260)

It is better than drills because:
* Infinite drones can mine from a single tile of ore

So, dropping off at drone spawns is more balanced and suitable for advanced resources.

Anuken:
This can't reasonably be implemented with 6.0 factories, so I'll have to close it.

Quezler:
(i didn't even realize it was still open)

PR #1331, opened 2020-01-06
[quality of life] mass driver place range assist

Quezler:
> line size/color/divisions/etc will probably need tweaking

Just a small quality of life thing, the ability to see if the driver you're trying to place is actually in range of the one you have selected, especially when you try to place one right on the edge 🤔 

Jan-06-2020 17-07-49


DeltaNedas:
very cool

PR #1333, opened 2020-01-07
Authentication implementation for servers

iczero:
Allow mindustry servers to require authentication before allowing users to join.

Authentication api implementation: https://github.com/hellomouse/mindustry-authentication-server

Not implemented yet:
* Account-based whitelists/bans/admins
* Logging out from client
* Web interface for the authentication service (registration, etc)


TheMrAlbert:
Uhh... I think there is conflict by the time I send this comment.

Shawak:
Are you really sending the password in plaintext? :thinking:

HaphLife:
> Are you really sending the password as plaintext? 🤔
Ah yes, very secure. The hackers will try to decrypt it first and be confused. /s

They really should encrypt the data using an up to date encryption method, and use a hashing algorithm with salt and only send the hash over the network, and only store the hash, so they can't leak the clients password (which might be re-used), and don't have to worry about that liability.. Plaintext is absolutely wrong.

Bohaska:
@iczero Can you update this pull request? Some problems have already been pointed out.

iczero:
Passwords are not sent in plaintext as https is used as an encrypted transport between the client and the auth server. The auth server does a salted hash of the password before storing it in the database.

Encrypting the password on the client is pointless. Hashing the password on the client would only *possibly* help in cases of password reuse, and even then, it is not possible to do a salted hash on the client without an extra round trip, which would allow rainbow tables to be used.

As there does not appear to be much interest, I am closing this pull request. I am currently quite busy with university. I may update this PR if I have time in the future.

PR #1339, opened 2020-01-08
Rename bundle_pt.properties to bundle_pt_PT.properties

artur309:
It´s better since theres is a bundle_pt_BR, and not being confuse for some


PetrGasparik:
I agree, on some specific languages, like Portuguese, English, Chinese, it is better to use lang_COUNTRY syntax. for rest, lang is enough

PR #1350, opened 2020-01-10
[quality of life] visualize solar panel efficiency

Quezler:
> makes it (more) clear that on darker maps solar panels produce less power.

Screen Shot 2020-01-10 at 18 16 17

Screen Shot 2020-01-10 at 18 35 06

Screen Shot 2020-01-10 at 18 35 12

> not sure if this "should" be merged, since lighting is still listed as experimental.

### DIFF
--- a/core/src/mindustry/world/blocks/power/SolarGenerator.java
+++ b/core/src/mindustry/world/blocks/power/SolarGenerator.java
@@ -1,7 +1,10 @@
 package mindustry.world.blocks.power;
 
+import arc.*;
 import arc.struct.*;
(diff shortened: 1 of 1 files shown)


Quezler:
~~hmm, just sorta realized the power bar is already a sort of indicator for it~~

Anuken:
I didn't implement this because it seemed unnecessary; why display efficiency when the power output bar already does so? Yes, you don't see the exact percentage, but that's not very important.

Quezler:
ima go ahead and close it then i suppose, unless efficiency during placement is still worth doing?

PR #1351, opened 2020-01-10
Update bundle_pt_PT.properties

artur309:
A new update


artur309:
It´s being updated with new translations

PR #1356, opened 2020-01-11
Remove false description on Omega's armor

Ali-C-Ila:
Omega does not actually block 90% damage. Made an correction on that.


Quezler:
it "can" block damage, but only whilst moving/firing 😗

ZkyLB:
hey que

Anuken:
It does block damage, as mentioned above.

PR #1365, opened 2020-01-13
batteries have glow based on power stored

Arkanic:
unfinished

the orangey things on the batteries should glow based on their buffered power
alpha formula is calculated by `power/maxPower`


Arkanic:
i borked gradle, i'll make a new commit to match

DeltaNedas:
arkanen

Arkanic:
uh oh

PR #1375, opened 2020-01-14
There was a missing f

CA6:
F


KSean222:
The `f` does nothing of significance in this case, since it's an integer value.

Anuken:
It's better without an F.

CA6:
okay

Arkanic:
Big f

ado1928:
F for our fallen soldier

PR #1376, opened 2020-01-14
Orange on batteries change colour based on power contents

Arkanic:
done

batteries now have two parts
`battery.png` is now the base
`battery-lights.png` is now the orange things on the battery

now it cycles between mindustry standard dark gray to battery orange, the formula is `power/maxPower`

i playtested it but forgot to take screenshots

it seems to go orange to quickly sometimes, i might divide it by `1.5f` or `2f`


PetrGasparik:
please make screenshorts:) or even better, short movie clips with gyazo

Arkanic:
Ok i think i know what's going on here, git thinks that the new file is out dates, ill close the pr and re open it to see if that works

PR #1380, opened 2020-01-14
edit fastlane

AmateurPotion:


AmateurPotion:
So many update...

Anuken:
I do not recommend translating changelogs, as they cannot be seen until the next update and are guaranteed to have duplicates.

PR #1384, opened 2020-01-14
Orange on batteries shade depending on power content

Arkanic:
Redo because git can't comprehend sprite packing


nichrosia:
Sorry about that, accidentally mentioned this pull instead of an issue

PR #1385, opened 2020-01-15
orange colour variance depending on power redo (done)

Arkanic:
the orange "cells" on batteries now change from black to orange depending on their buffered power (power stored).
I was originally using `power/maxPower` but anuke told me to use `power.status` which is a float that contains the current amount of power given compared to the required.

now lights have two sprites, `battery.png`, which is the base, where the cells are the standard mindustry dark dark grey, and `battery-lights.png` which is a white outline of the cells, for the alpha shader.

this took so long because of git and conflicting files problems :/

edit: pics :)
Screenshot (59)

in the preview the batteries seem pretty dark, i don't know whether or not it is github preview shading weirdly or the sprite packer compressing it, but here are the sprites when i inserted them
  
Screenshot (61)


core/src/mindustry/world/blocks/power/Battery.java
```diff
@@ -1,10 +1,39 @@
...
+    public void load(){
+        super.load();
+        lightsRegion = Core.atlas.find(name + "-lights");
+    }
+
+    @Override
+    public void draw(Tile tile){
+        super.draw(tile);
+        TileEntity ent = tile.ent();
+        PowerModule entity = ent.power;
```
Anuken: Don't create a variable if you're only going to use it once; use this directly instead.

Arkanic: ok, will do
be doesn't seem to compile

Arkanic:
Thanks, ill pack the sprites tomorrow
 ~~how and why am I taking so long~~

CA6:
ah yes cool lag..... I like it but it should be like shield animations toggleable

Arkanic:
@CA6 literally every generator does this

Anuken:
Some thoughts:
image

Why are the empty spots black? Mindustry doesn't black anywhere in its block palette; use the dark gray present in the battery sprites instead.

image

The displayed icons for the batteries should have them at 100% charge (e.g. with the overlay). See how drills override `generateIcons` for an example of how to add a sprite onto the icon.

CA6:
@Arkanic oof, true but I do think that those animations should be also able to be disabled as well tbh

Arkanic:
@Anuken the sprites i intend to use used the impact reactor background gray, i think i might have picked the wrong sprites

I've been on a hiatus for six days, ill finish this sometime soon

Arkanic:
I'll also try the drill override, that'll been interesting

Arkanic:
aha
```
@Override

public TextureRegion[] generateIcons(){

    return new TextureRegion[]{Core.atlas.find(name), Core.atlas.find(name + "-rotator"), Core.atlas.find(name + "-top")};

}
```

Arkanic:
aaaaa more conflict

PR #1398, opened 2020-01-19
[quality of life] mechpad checkpointing

Quezler:
Currently if you want to spawn from a new mech pad you have to:
`build it > power it > use it`

And if your mech is already the one the pad you placed spawns:
`build it > power it > use it > use it`

This pull changes it so you can "tap" a (powered) mech pad while outside of its "use" range to set your spawn there for your next death(s):
`build it > power it > tap it`

(might not be that mobile friendly since they tend to move onto the pad a little more eagerly)

**video**
in [discord](https://discordapp.com/channels/391020510269669376/663853133587546138/668536324965531648) since it supports a higher resolution than a gif #filesize 🙂 

**spawning from**
Screen Shot 2020-01-19 at 20 21 47

**changable to**
Screen Shot 2020-01-19 at 20 22 12

**unpowered/unavailable**
Screen Shot 2020-01-19 at 20 22 35

> _spawning from_ has a higher visual priority than _unpowered/unavailable_

### DIFF
--- a/core/src/mindustry/world/blocks/units/MechPad.java
+++ b/core/src/mindustry/world/blocks/units/MechPad.java
@@ -47,11 +47,12 @@ public class MechPad extends Block{
 
     @Remote(targets = Loc.both, called = Loc.server)
     public static void onMechFactoryTap(Player player, Tile tile){
-        if(player == null || tile == null || !(tile.block() instanceof MechPad) || !checkValidTap(tile, player)) return;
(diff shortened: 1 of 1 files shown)


Anuken:
If you're wondering why I added the 6.0 tag, it's because mech pads will have different mechanics there, and I'm not sure if merging this now is a good idea.

Quezler:
~~but 6.0 isn't around the corner yet, and this is a simple change that can easily be undone?~~ 🤔

Anuken:
No longer relevant, due to the removal of mech pads.

PR #1399, opened 2020-01-19
If construction was simplified.

TodorBalabanov:
This simplification helps for easier understanding of the if checking done in this method.


DeltaNedas:
a few newlines would do that and probably create faster bytecode

Quezler:
why not save a few lines?

```
if(values[dx][dy] >= value) continue;

if(current != null && values[dx][dy] >= tl) continue;

if(other.solid() == true) continue;

if(other.floor().drownTime > 0) continue;
```

Anuken:
Since there is only ever one `continue` needed here, this should be separated into newlines with `&&` and optional comments instead. I'm not a fan of redundant additional if-statements.

PR #1407, opened 2020-01-21
Fixed BlockBuildEndEvent sometimes not firing in multiplayer

ado1928:
BlockBuildEndEvent really seemed unpredictabe when building in multiplayer, sometimes firing, sometimes not. After i moved it to onConstructFinish() method, it seems like it works every time. Not sure exactly how, because onConstructFinish() was only ever called in constructed(), but it fixed it


Kieaer:
This can cause tile NPE to be faster.

Anuken:
This makes the block you get after finishing building a `BuildBlock`, always.

Quezler:
the block in a4ac318c9e804b08839ae5c1deb16f47b93fb21c tends to be a buildblock a lot of the time as well 👀

Anuken:
It should be. You can't deconstruct a block without having it be a `BuildBlock`.

PR #1421, opened 2020-01-23
[server] pausable

Quezler:
> prototype "it just sorta works" (code not deemed worthy to be merged)

- adds the ability for servers to be paused
- [ ] add server console (un)pause command?

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -50,6 +50,8 @@ public class NetClient implements ApplicationListener{
     /** Last sent client snapshot ID. */
     private int lastSent;
 
+    public boolean paused = false;
+
     /** List of entities that were removed, and need not be added while syncing. */
     private IntSet removed = new IntSet();
     /** Byte stream for reading in snapshots. */
@@ -61,6 +63,7 @@ public class NetClient implements ApplicationListener{
         net.handleClient(Connect.class, packet -> {
             Log.info("Connecting to server: {0}", packet.addressTCP);
 
+            paused = false;
             player.isAdmin = false;
 
             reset();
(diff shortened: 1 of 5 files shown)


ZkyLB:
you scare me

Quezler:
[this](https://github.com/Quezler/mindustry__plugin--weepingangel/blob/master/src/main/java/weepingangel/WeepingangelMod.java#L14) scares me more 👀

ZkyLB:
> [this](https://github.com/Quezler/mindustry__plugin--weepingangel/blob/master/src/main/java/weepingangel/WeepingangelMod.java#L14) scares me more 👀

no you scare me more, you cute roomba

Quezler:
krater

CA6:
oh no........ tho quezler you're making things that are pretty cool keep it up

Quezler:
thx ❤️

core/src/mindustry/core/NetClient.java
```diff
@@ -50,6 +50,8 @@
     /** Last sent client snapshot ID. */
     private int lastSent;
 
+    public boolean paused = false;
```
Anuken: I'm concerned about the fact that there are now *4* different states to check for the pausing (some of which may be accurate, some may not): `netClient.paused`, `netServer.paused`, `state.isPaused()` and `state.is(State.paused)`. This is rather confusing, to say the least.

Quezler: well i wasn't sure where to put them and how to properly sync them all 🙈

Anuken:
Since entities are now updated with a single call and there's no update delay, this will be easier to implement in 6.0. Will consider it there.

PR #1425, opened 2020-01-24
PRE 103 RU_updates

XEN0PHIL:


Anuken:
The dot cannot be part of the translation. Adding it would lead to text overlap in block inventories:

image

PR #1437, opened 2020-01-25
yet another batteries glow pr

Arkanic:
the orange "cells" on batteries now change from black to orange depending on their buffered power (power stored).
I was originally using `power/maxPower` but anuke told me to use `power.status` which is a float that contains the current amount of power given compared to the required.

now lights have two sprites, `battery.png`, which is the base, where the cells are the standard mindustry dark dark grey, and `battery-lights.png` which is a white outline of the cells, for the alpha shader.

this took so long because of git and conflicting files problems :/

edit: pics :)
Screenshot (59)

in the preview the batteries seem pretty dark, i don't know whether or not it is github preview shading weirdly or the sprite packer compressing it, but here are the sprites when i inserted them
  
Screenshot (61)


Arkanic:
e.e.e.e.e.ee.e

PR #1444, opened 2020-01-25
add time played to end of game score screen

ahouts:
https://feathub.com/Anuken/Mindustry/+126
image
I was not able to test the pvp end screen.
Also added it to multiplayer end of game info
image


DeltaNedas:
uptime epeen, now on mindustry

ahouts:
Just noticed that `0f` is used elsewhere, not `0.f`. Fixing

core/src/mindustry/io/SaveVersion.java
```diff
@@ -67,6 +67,7 @@ public void writeMeta(DataOutput stream, StringMap tags) throws IOException{
             "build", Version.build,
             "mapname", world.getMap() == null ? "unknown" : world.getMap().name(),
             "wave", state.wave,
+            "gametime", state.gametime,
```
Anuken: A few lines above this, an equivalent variable (`playtime`) is already written here. You shouldn't need to add another variable to `GameState` for it, as saves already store time played. I implemented the client side of this PR using the existing variables in 2721aa550d85b5146187a5a93bc5fb4518446a14.

Yes, if you're playing on a server without a save, the time information isn't stored, but I think server-side match duration should be calculated with a simple 'last match' timestamp and elapsed check instead.

ahouts: Figured it was somewhere but I was looking in the wrong places. Didn't expect the save object to be reachable from the main loop of the game.

PR #1453, opened 2020-01-26
Adding Steam Rich Presence support.

danieljennings:


danieljennings:
Sorry, trying to figure out how to get this PR built was a pain in the ass, typing up a real description now because it dropped my description.

danieljennings:
Hi friends,

Submitted for your consideration is a patch to DesktopLauncher that adds support for Steam's Rich Presence support, as documented here: https://partner.steamgames.com/doc/features/enhancedrichpresence

I've piggybacked off of the existing updateRPC method used for DiscordRPC's similar rich presence support, by just expanding that method to do the Steam parts as well (if the Steam bool is set). I wasn't sure if it made more sense to live inside of SNet, since that's where most of the Steam-specific code is, or here because this is where the existing Rich Presence code was written and it shares a lot of the work figuring out what strings we want to show.

Because of (what I consider to be) a deficiency in the Steam Rich Presence API here, I did have to create and upload the stub of a localization file to the Steam Partner site in order to get this to work at all (apparently we expect the root-level "steam_display" key to always be a Steam localization token, and don't allow explicitly setting a rendered string there, but I'm going to dig into that requirement more.)

Here's the contents of what I uploaded, which can be downloaded/uploaded via this page on the partner site: https://partner.steamgames.com/apps/richpresence/1127400
`"lang"
{
	"english"
	{
		"tokens"
		{
			"#steam_status_raw"	"%steam_status%"
		}
	}
}`

Here's what it looks like as an example:
image

Should be a straightforward review, I matched your coding style as closely as possible. Let me know if you have any questions about this or Steam's Rich Presence in general and I should be able to answer them; if you have questions about other parts of the Steam API I can dig into the code to help you answer them, and failing that I can find the responsible party and ask.

The CI tests are failing on what I believe is an unrelated change? I got the same thing locally after syncing back up with HEAD when trying to make the PR.

Also, I need to fix the PR because I left a debugging line in there that you don't want, so as soon as I figure out how to update the PR here I'll do that.

danieljennings:
Also, if you want to reach out about any of this via email, you can reach me at danielj (AT) valvesoftware.com, but I'll be quicker here outside of work hours.

Anuken:
Looks good, thanks!

danieljennings:
Thanks!!

PR #1464, opened 2020-01-28
[quality of life] impact reactor startup sequence

Quezler:
https://feathub.com/Anuken/Mindustry/+626

changes it so the impact reactor needs to be full of blast compound in order to begin starting up.
(once its startup animation has started it will continue spinning with any amount of blast*)

this should help greatly since impact reactors tend to omnomnom away if there's only 1 in them.
and it might also make it a bit easier to spot (potential) throughput issues, among other things.

block-impact-reactor-full

> picture is just here to make the pull request look a little less boring (^-^)

### DIFF
--- a/core/src/mindustry/world/blocks/power/ImpactReactor.java
+++ b/core/src/mindustry/world/blocks/power/ImpactReactor.java
@@ -71,7 +71,7 @@ public class ImpactReactor extends PowerGenerator{
     public void update(Tile tile){
         FusionReactorEntity entity = tile.ent();
 
-        if(entity.cons.valid() && entity.power.status >= 0.99f){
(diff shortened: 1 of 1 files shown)


ZkyLB:
amazing

Quezler:
impressive, a conflict on a pull that only changes part of one line 🥇

Quezler:
feel free to either close this or ask me to update it ^

Anuken:
> this should help greatly since impact reactors tend to omnomnom away if there's only 1 in them.

Making reactors only consume when full doesn't fix that problem, it only delays it until the reactor gets full. The root cause is lack of input; when the reactor fills up, it will consume the last item, and immediately stop warming up since its inventory isn't full anymore and the blast compound input (still) can't keep up. Your end result is the same, and the reactor won't work either way.

PR #1480, opened 2020-01-30
Gitignore update: mods/ and .bat

PetrGasparik:
Hi,
please accept this update, which adds to .gitignore following
`mods/` - obvious reasons, there will be no mods in Mindustry and github 
`*.bat` - that's what I use for effective contributing (remote pull etc), I believe there will be no *.bat file in future

This change should make no harm to anyone and it will help me a lot :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,3 +153,5 @@ gradle-app.setting
 .DS_Store
 Thumbs.db
 android/libs/
+mods/
+*.bat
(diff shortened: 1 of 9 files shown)


PetrGasparik:
unfortunately it is not how I wanted it to be

DeltaNedas:
gradlew.bat...
ok so some questions
1. why is your mods folder in your source folder
2. why are you using batch scripts at all, just `git pull remote branch`, or at least just dont add them to your pr

PR #1488, opened 2020-02-01
added my server

Micro526:


Quezler:
you might want to solve the merge conflict 👀

Anuken:
I'll have to pass on this one. I generally only verify servers from hosts I know or have seen other people mention, and this is the first time I've seen you or this server. The fact that it is "still in the development stage", as the description states, doesn't help much either.

PR #1494, opened 2020-02-02
[quality of life] server list categories

Quezler:
> graphical prototype, contains duplicate code and probably some un-caught edge cases

splits the server list in 3 distinct categories:
- local servers
- remote servers
- global servers

global servers and local servers are no longer part of the same internal "list" so you can still press/see/find the refresh local games button if there is a global server in the list.

local servers have been put way up top, because if there is a local game available on your network its pretty likely that its your own testing server or a server you would probably want to join for a plethora of reasons, justifying them being above your own remote server list.

global servers go at the bottom, since you'd want your own server list first, so you don't have to scroll for the servers you actually care about.

Feb-02-2020 15-48-01

🥔 
> and yes, at the time of typing the potato i remembered i didn't use the `bundle.properties`


DeltaNedas:
ok now delete the global servers or add a little folding arrow thing to hide it

Quezler:
i don't think delete or hiding the global list is the best possible solution though

Prosta4okua:
Add to bundle.

Quezler:
> Add to bundle.

obviously. read the two quotes in the post, both mention reason(s) for it not being the case, yet. 🤐

Quezler:
i'll probably wait for a reaction from @Anuken on this before proceeding with bettering/closing it 🙄

Anuken:
This seems useful. A button to collapse the sections would be nice to add, even if it doesn't completely solve the problem. I may look into implementing this myself, as collapsing layout can be a hassle to get right.

Quezler:
This pull doesn't feel "quite right" yet, but i'll put it in normal mode so anuke can tinker with it 🔧 
> (if he so pleases, as he mentioned above ^)

Anuken:
Implemented with collapser buttons per category. The server list still needs some work, but that'll come later.

PR #1495, opened 2020-02-03
Improved ItemModule::take() behavior

InvalidError404:
The original take() behavior spams items in whatever order they appear in the items list until each index is depleted, which is problematic when non-specific unloaders are competing against dedicated unloaders for a low-index resources.

My modification makes the take() loop persistent so take() will do complete laps around the item list starting from wherever the previous call returned from, never repeating the same item twice in a row unless there is nothing else to return. A significant improvement IMO.

How is this an improvement? With the original behavior, if you converge a bunch of belts on a storage block or launcher, want to unload a few belts of specific resources (ex.: copper, lead, silicon and titanium for surge alloy) and pass everything else including overflows along using non-specific unloaders, you are out of luck when the resources you want are at the top of the list since most of those will get swept away by non-specific unloaders. With the rotating take(), non-specific unloaders are equal-opportunity across all available resources, which gives single-resource unloaders that many more chances to unload more of their resources before non-specific unloaders get to them. It also reduces the rate at which items further down the list that may hardly ever get touched by the existing implementation will race toward the storage block's cap. The even drain across all items will help prevent things like mass driver stalls due to receivers filling up with an excess resource that isn't getting cleared.

It would be even nicer if dedicated unloaders had priority over non-specific ones (non-specific unloaders don't touch resources with dedicated unloaders unless the dedicated unloaders are overflowing), though that would require a rework beyond my current "getting the stupid IDE and tool chains to work" level of familiarity with Java development.

To sum it up: trivial change, big impact for people who like mixing belts and sorting them out with unloaders.

### DIFF
--- a/core/src/mindustry/world/modules/ItemModule.java
+++ b/core/src/mindustry/world/modules/ItemModule.java
@@ -86,7 +86,7 @@ public class ItemModule extends BlockModule{
             if(items[i] > 0){
                 items[i]--;
                 total--;
-    			// save the next position so the next call to take() can resume from there.
+    		// save the next position so the next call to take() can resume from there.
                 takeRotation = (i + 1) % items.length;
                 return content.item(i);
             }
--- a/core/src/mindustry/world/modules/ItemModule.java
+++ b/core/src/mindustry/world/modules/ItemModule.java
@@ -12,6 +12,9 @@ public class ItemModule extends BlockModule{
     private int[] items = new int[content.items().size];
     private int total;
 
+    // Make the take() loop persistent so it does not return the same item twice in a row unless there is nothing else to return.
+    protected int takeRotation;
+
     public void forEach(ItemConsumer cons){
         for(int i = 0; i < items.length; i++){
             if(items[i] > 0){
@@ -65,9 +68,6 @@ public class ItemModule extends BlockModule{
     public int total(){
         return total;
     }
-
-    // Making the take() loop persistent so it does not return the same item twice in a row unless there is nothing else to return.
-    protected int takeRotation;
 
     public Item take(){
     	// 0-to-length loop broken in two parts. First resume the loop where it previously left off.
--- a/core/src/mindustry/world/modules/ItemModule.java
+++ b/core/src/mindustry/world/modules/ItemModule.java
@@ -75,7 +75,7 @@ public class ItemModule extends BlockModule{
             if(items[i] > 0){
                 items[i]--;
                 total--;
-    			// save the next position so the next call to take() can resume from there.
+    		// save the next position so the next call to take() can resume from there.
                 takeRotation = (i + 1) % items.length;
                 return content.item(i);
             }
--- a/core/src/mindustry/world/modules/ItemModule.java
+++ b/core/src/mindustry/world/modules/ItemModule.java
@@ -70,25 +70,14 @@ public class ItemModule extends BlockModule{
     }
 
     public Item take(){
-    	// 0-to-length loop broken in two parts. First resume the loop where it previously left off.
-        for(int i = takeRotation; i < items.length; i++){
-            if(items[i] > 0){
-                items[i]--;
-                total--;
-    		// save the next position so the next call to take() can resume from there.
-                takeRotation = (i + 1) % items.length;
-                return content.item(i);
-            }
-        }
-
-        // Then start a new lap which ends where the call started from if empty.
-        for(int i = 0; i < takeRotation; i++){
-            if(items[i] > 0){
-                items[i]--;
-                total--;
-    		// save the next position so the next call to take() can resume from there.
-                takeRotation = (i + 1) % items.length;
-                return content.item(i);
+        for(int i = 0; i < items.length; i++){
+            int index = (i + takeRotation);
+            if(index >= items.length) index -= items.length; //conditional instead of mod
+            if(items[index] > 0){
+                items[index] --;
+                total --;
+                takeRotation = index + 1;
+                return content.item(index % items.length);
             }
         }
         return null;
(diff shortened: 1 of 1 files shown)


Quezler:
it should probably be written to the stream as well, for new joins & syncs 🤔

core/src/mindustry/world/modules/ItemModule.java
```diff
@@ -66,11 +66,28 @@ public int total(){
...
+    protected int takeRotation;
+
     public Item take(){
-        for(int i = 0; i < items.length; i++){
+    	// 0-to-length loop broken in two parts. First resume the loop where it previously left off.
+        for(int i = takeRotation; i < items.length; i++){
+            if(items[i] > 0){
+                items[i]--;
+                total--;
+    			// save the next position so the next call to take() can resume from there.
```
Quezler:
```suggestion
    		// save the next position so the next call to take() can resume from there.
```

core/src/mindustry/world/modules/ItemModule.java
```diff
@@ -66,11 +66,28 @@ public int total(){
...
+                return content.item(i);
+            }
+        }
+
+        // Then start a new lap which ends where the call started from if empty.
+        for(int i = 0; i < takeRotation; i++){
             if(items[i] > 0){
                 items[i]--;
                 total--;
+    			// save the next position so the next call to take() can resume from there.
```
Quezler:
```suggestion
    		// save the next position so the next call to take() can resume from there.
```

InvalidError404:
> it should probably be written to the stream as well, for new joins & syncs 🤔

At the moment, my java-specific dev skills are: cursing at the IDE. If what you mean by adding it to the stream is adding it to the read/write functions at the end of the file, that messes up the bytestream formatting and makes it incompatible with saves. 'Slightly' problematic unless there is a mechanism elsewhere for handling extensions.

core/src/mindustry/world/modules/ItemModule.java
```diff
@@ -66,11 +66,28 @@ public int total(){
...
+    	// 0-to-length loop broken in two parts. First resume the loop where it previously left off.
+        for(int i = takeRotation; i < items.length; i++){
+            if(items[i] > 0){
+                items[i]--;
+                total--;
+    		// save the next position so the next call to take() can resume from there.
+                takeRotation = (i + 1) % items.length;
+                return content.item(i);
+            }
+        }
```
Anuken: This method can be written as a single loop.

```java
for(int i = 0; i < items.length; i++){
    int index = (i + takeRotation) % items.length;
    if(items[index] > 0){
        items[index] --;
        total --;
        takeRotation = (index + 1) % items.length;
        return content.item(i);
    }
}
```

core/src/mindustry/world/modules/ItemModule.java
```diff
@@ -66,11 +66,28 @@ public int total(){
         return total;
     }
 
+    // Making the take() loop persistent so it does not return the same item twice in a row unless there is nothing else to return.
+    protected int takeRotation;
```
Anuken: Instance variables should always be declared at the top of a class.

core/src/mindustry/world/modules/ItemModule.java
InvalidError404: Sure, it can, though modulo remains an extremely expensive instruction on modern hardware (from 15-25 cycles for IDIV/r8 to 50-70 cycles for /r64 vs less than one cycle effective cost for most other 1-5 cycles ALU instructions due to pipelined superscalar execution) and depending on how the java compiler re-factors this, you can conceivably spend more CPU time waiting for IDIV to produce the remainder the rest of the loop depends on than doing useful work in tight loops like that.

As a mostly low-level/bare-metal ASM/C/C++/VHDL programmer, putting an exceptionally expensive operation in an otherwise trivial loop's control path when it can so easily be avoided bothers me.

Hope you found this bit of low-level hardware perspective enlightening.

core/src/mindustry/world/modules/ItemModule.java
InvalidError404: Of course, sorry about that.

Just wanted to put the tweak out for people to try and hopefully see the modified behavior or something resembling it make it into a future release.

core/src/mindustry/world/modules/ItemModule.java
Anuken: I am aware that modulo is generally considered expensive, but in this case there isn't much of a benefit to splitting things up into two loops. Adding an extra loop adds more code and makes this method harder to read and modify.

You also have to consider the fact that this is extremely premature optimization. `take()` is a method that is called only once per frame for every input bridge and unloader. Considering that most active games have few amounts of these blocks being used, it really doesn't matter which implementation is slower, because the difference is only going to be a few nanoseconds. I would rather have more readable code than something that's ugly and imperceptibly faster.

That being said, I tried benchmarking both versions to see how significant the speed difference was, and the performance difference (with random items) for these two is about ~12 ns, with the non-modulo one taking 30 and one taking 42. Rewriting the mod implementation to mostly use subtraction instead of modulo actually made it run faster than yours, with 21 ns taken per invocation. This benchmark probably isn't too accurate, but when dealing with nanoseconds, it really doesn't matter.

Code:
```java
            for(int i = 0; i < items.length; i++){
                int index = (i + takeRotation);
                if(index >= items.length) index -= items.length; //conditional instead of mod
                if(items[index] > 0){
                    items[index] --;
                    total --;
                    takeRotation = index + 1;
                    return content.item(index % items.length); //called rarely so it doesn't really matter here, according to benchmarks
                }
            }

            return null;
```

InvalidError404: Guessing this is speculative execution at work: branch prediction may not like my front loop having a pseudo-random starting point so your more predictable loop ends up cheaper overall despite having an extra test.

As for being imperceptibly faster, imperceptibly faster bits of code add up over time.

PR #1499, opened 2020-02-03
Update bundle_nl.properties

LauweJoster:
Improved some grammar and sentencing that were very off. Also added a few translations.


Quezler:
good job ❤️

PR #1506, opened 2020-02-04
Seperate flame and topRegion rendering, for better code readability.

TechFish3000:
if you would like me to stop suggesting this, just say so.
The only reason i want this is so i can have a `topRegion` but not a flame.

### DIFF
--- a/core/src/mindustry/world/blocks/production/GenericSmelter.java
+++ b/core/src/mindustry/world/blocks/production/GenericSmelter.java
@@ -30,7 +30,15 @@ public class GenericSmelter extends GenericCrafter{
 
         GenericCrafterEntity entity = tile.ent();
 
-        //draw glowing center
+        //draw topRegion
+        if(topRegion != NULL && entity.warmup > 0f){
+            Draw.color(1f, 1f, 1f, entity.warmup);
+            Draw.rect(topRegion, tile.drawx(), tile.drawy());
+            Draw.color();
+        };
+        
+        
+        //draw "flame" in center
         if(entity.warmup > 0f && flameColor.a > 0.001f){
             float g = 0.3f;
             float r = 0.06f;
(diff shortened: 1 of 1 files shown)


DeltaNedas:
set flame color to be empty
it already exists

TechFish3000:
yes, but that gets rid of the `topRegion`

TechFish3000:
Would i use `Null` or `None` to check if `topRegion` exists?

core/src/mindustry/world/blocks/production/GenericSmelter.java
```diff
@@ -41,7 +49,7 @@ public void draw(Tile tile){
             Draw.tint(flameColor);
             Fill.circle(tile.drawx(), tile.drawy(), 3f + Mathf.absin(Time.time(), 5f, 2f) + cr);
             Draw.color(1f, 1f, 1f, entity.warmup);
```
Quezler: this line is now duplicate, you might want to move it above the first if, and call reset when both are done 👀

core/src/mindustry/world/blocks/production/GenericSmelter.java
```diff
@@ -30,7 +30,15 @@ public void draw(Tile tile){
...
+        //draw topRegion
+        if(entity.warmup > 0f){
+            Draw.color(1f, 1f, 1f, entity.warmup);
+            Draw.rect(topRegion, tile.drawx(), tile.drawy());
+            Draw.color();
+        };
+        
+        
+        //draw "flame" in center
         if(entity.warmup > 0f && flameColor.a > 0.001f){
```
PetrGasparik: you may enclose this `if` into the upper one and get rid of "entity.warmup > 0f &&" here, plus one less check

TechFish3000:
There, it is efficient now (hopefully).

PetrGasparik:
> There, it is efficient now (hopefully).

I approve! :)

Anuken:
This has some incorrect indentation and isn't compatible with 6.0's entity rewrite.

PR #1514, opened 2020-02-05
make batteries glow based on power contents (a redo that works)

Arkanic:
i went all the way as to add seperate icon sprites

as a fact, this is the fifth redesign to fix the anti aliasing

 now the holes in the battery are transparent, with the colour appearing behind them
the icon is an entirely seperate file of the old sprite


core/src/mindustry/world/blocks/power/Battery.java
```diff
@@ -1,10 +1,41 @@
 package mindustry.world.blocks.power;
 
+import arc.*;
+import arc.graphics.Color;
+import arc.graphics.g2d.*;
+import mindustry.world.*;
+
 public class Battery extends PowerDistributor{
 
+    public TextureRegion baseRegion;
```
Anuken: The base region is just a white square, so it can be drawn with `Fill.square()`.

Anuken:
I'm going to merge this and implement some fixes on top myself.

Arkanic:
nice
my first actual commit

core/src/mindustry/world/blocks/power/Battery.java
Arkanic: ok

PR #1517, opened 2020-02-06
For consistency

MEEPofFaith:
Swap "liquids" with "liquid." Also proposing because the category for liquid blocks is "liquid" and not "liquids" when modding.


Arkanic:
nice

MEEPofFaith:
image

PR #1518, opened 2020-02-06
New storage block: Max Unloader

InvalidError404:
This update implements a new unloader that dumps most abundant item first out of the container it is attached to which should help players manage container overflows.

I could have added rotation in case multiple items are at max capacity, though the primary purpose of this block is to avoid getting there in the first place since you are kind of screwed no matter what when ingress bandwidth > egress bandwidth anyway.

Needs a sprites atlas rebuild, new icon included - added to mine by hand since I could not get the image packer tool to work properly.

Here's a screen crop of Max Unloader in action:
MaxUnloader Action

Max unloader is only touching the three resources that are at ~320.


ZkyLB:
i cant even find a use for this

InvalidError404:
> i cant even find a use for this
I have two common use cases for myself:
1- on the output vault of a mass receiver to prevent/delay any item from reaching the item cap and causing the mass driver/receiver to stop, starving whatever lies downstream of other resources
2- unloading whatever there is the most of gives sorted unloaders the best chance of filling their belts, could be made even better in that regard by adding a threshold below which max unloader becomes inactive to leave the remainder to regular unloaders, in which case I'd change the name to something like Threshold Unloader - "Unloads whichever item there is the most of when the block is over 50% full." or something of that sort.

core/src/mindustry/content/TechTree.java
```diff
@@ -38,9 +38,10 @@ public void load(){
                             });
                         });
                         node(container, () -> {
-                            node(unloader);
+                            node(unloader, () ->{
```
Quezler:
```suggestion
                            node(unloader, () -> {
```

Anuken:
Even if there are a few uses for a block, I generally don't add new transportation features unless there's very high demand for them (as was the case for inverted sorters/gates). This seems like a niche block most people won't understand or use.

InvalidError404:
I'd say it is one of those "you don't know you want it until you've used it" kind of thing.
Typical MaxUnload
This is how I use it on pretty much any map I plan to stay on for a while: use a mass receiver to feed a vault, use sorted unloaders to pick up stuff - surge alloy ingredients in this case going out toward the bottom-right - and max unloaders dumping excess stuff on the left, one belt disconnected here to let the vault fill up so the dedicated copper/silicon/lead/titanium belts can have full output. Max Unload keeps all stacks even so inventory can flow smoothly through the mass receiver and vault.

In this instance, I was testing a modified version that does not unload unless there are 50+ items in the source block, a much lower effort alternative to actually giving dedicated unloaders priority. Looks promising so far, though limiting the buffer to 50 to accommodate the regular launchpad (set the threshold to start draining low enough that mass drivers won't systematically trigger a launch) is a bit sad, will have to look into checking the underlying block type to tweak thresholds accordingly. If I go down that route, I'll be changing the name to "Threshold Unloader" and look into making the threshold configurable between 1/4, 1/2 and 3/4.

This is a continuation of my first patch's theme of making non-specific unloaders steal fewer items from starving dedicated unloaders. Giving dedicated unloaders some sort of priority would require considerable re-work, creating some sort of low-priority unloader on the other hand is relatively trivial.

Arkanic:
@InvalidError404 i could pack the sprites for you, then make a pull

InvalidError404:
> 
> 
> @InvalidError404 i could pack the sprites for you, then make a pull
> just out of curiosity, what would you use this for?

See my previous reply in this pull request, I think it describes fairly well one situation where I would use it all the time - want to prime a vault to feed a high volume resource sink (ex.: surge alloy smelter) with full belts but don't want mass drivers to stall due to full receiver vault either.

As for the sprites, I may take you up on the offer once I have decided how far I want to go with the threshold unloader idea, will probably need to create config icons/overlays. Currently looking for a way to distinguish launchers from vaults/containers so I can adapt behavior to the two different item counting styles.

InvalidError404:
Worked on the "Threshold Unloader" upgrade to my maxunload some more...
Config ThresUnloader

As you may be able to infer from the four different amounts of green filling the unloader's red center and the four setup bar icons, the Threshold Unloader has four threshold levels. The unloaders are inactive until their inventory threshold is met, at which point they start unloading whatever there is the most of until inventory dips back down below their respective tresholds.

This is a much more useful block than my plain max idea and can behave like my first idea too on its 0% threshold since it reuses the same max search function, albeit with max initialized to the threshold count so only items with higher quantities get considered,

InvalidError404:
Closing this old pull request, will open another for my more useful upgraded version later.

InvalidError404:
Threshold Unloader pull request is up: https://github.com/Anuken/Mindustry/pull/1539

PR #1532, opened 2020-02-08
[byte logic] 103 master merge

Quezler:
> sat down for about an hour meticulously merging stuff by hand, this version compiles

some stuff is commented out (or even removed) to assist with compilation:
- `buildTable`
- `buildConfiguration`
- some @ annotation tags
- probably some more stuff
- pretty sure i forgot about some code
- some of it is marked with the `fixme` tag, intelij will know how to find them

as mentioned above, this version compiles, that's all it does, expect a fair amount of logic within it to be broken to any possible degree, i am fairly unfamilair with the byte logic branch (heck, i first ran it this morning before merging in the master branch), but the existence of this pull will likely assist those who can squash the last bugs in it, since the worst part (merging) is behind them 🙂 

Screen Shot 2020-02-08 at 06 56 37

> note: i currently have no intention to do (much) more on this, i'm basically giving this pull away to those that have it in them to actually iron out the bugs caused by the master merge 😗


KK6DOC:
Quezler you didn't have to do this but thank you sooo much :-)

Quezler:
i was up early and i needed some challenge to occupy myself with 😋 

(don't celebrate to early tho, this pull is fairly broken 😗)

Legomonster33:
een if its still broken its a start to keeping bytelogic alive (until anoooooooooke decides how to do the new system)

Quezler:
as @iczero pointed out somewhere, a clean rewrite would probably be better than a catch up merge, closing this since its stale & searchable 😗

Quezler:
note to self: https://github.com/Anuken/Mindustry/compare/byte-logic

Quezler:
for those still receiving notifications for this pull, see its successor linked above ^

PR #1534, opened 2020-02-08
return map from WorldLoadEvent

Kieaer:
Probably for my plugin :eyes:


Anuken:
This is misleading; the world can load without a map. Can't you call `world.getMap()`?

Kieaer:
Always return "Editor playtesting"

Kieaer:
I found why world.getMap() isn't working

Quezler:
i'm currently on the trail of maps having `unknown` or `Editor playtesting` as a name when loading from a save instead of a map, any pointers to help in my research? 🤔 
wisdom_of_the_ancients

PR #1535, opened 2020-02-08
Additional Category(mod)

AmateurPotion:
Mod blocks and resources are overflowing!

* Who can tell me how can i upload image to github?


DeltaNedas:
aaa

Anuken:
This can completely break layout and is too specific to be added to the base game. While new categories would be nice, this is not the right way to go around implementing them.

PR #1539, opened 2020-02-08
Threshold Unloader

InvalidError404:
An evolved version of my Max Unloader. This variant starts unloading the most plentiful item only when the storage it is attached to is filled above the selected threshold of 0%, 25%, 50% or 75%. The threshold is applied on a per-item basis for vaults (won't unload any item that is below threshold) and total basis for launchers since launch is triggered on total. Behaves like the Max Unloader when set to 0% since it reuses the same max search function, albeit with the initial max value set to capacity*threshold for vaults so items with insufficient quantity get ignored.

Basically, this is an overflow gate for containers that lets you pick how empty (or full) you would like to keep the vault it is attached to.

ThreshUnloader
In the Tar Field run I started to test the threshold unloader one last time before opening my push, my core spawned in the SW corner, so I had to setup this little relay/sorting station in the corridor leading to large open area in the south.

The copper, lead, titanium and silicon unloaders are feeding surge alloy smelters as usual, a pair of regular non-sorted unloaders trickle-feeds my core a healthy mix of whatever passes by, one threshold unloader set at 50% lets the vault fill up to at least 50% of any one item before kicking in so it does not interfere with any items I am low on (it is unloading a mix of plastanium, thorium and lead in this case, the three resources above 50% of 1000) and a backup threshold unloader kicks in at 75% to help ensure mass receivers never run out of room to dump their loads.


Anuken:
I don't think this block is necessary.
- Your use case is too specific. Mixing several types of resources in a mass driver isn't something I've seen much of, and many people (including me) were confused about its intended use in the initial PR discussion. It's not intuitive.
- There are already far too many block variants. For a new transport block to be added, there must be a very good reason for it, or significant community support.
- It could be implemented with a logic system toggling an unloader, which may be added in 6.0.
- It's much more complex than the other transport blocks. Complexity should be avoided unless absolutely necessary.

PR #1549, opened 2020-02-11
Allow admins to bypass the playerlimit [server]

fuzzbuck:
admins often have a hard time joining their own servers because of a set playerlimit

### DIFF
--- a/core/src/mindustry/core/NetServer.java
+++ b/core/src/mindustry/core/NetServer.java
@@ -131,7 +131,7 @@ public class NetServer implements ApplicationListener{
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


KSean222:
`admins.getPlayerLimit() > 0 && playerGroup.size() >= admins.getPlayerLimit() && !netServer.admins.isAdmin(uuid, packet.usid)`
if the player limit is larger than zero AND the player group is larger than the limit AND (the player is NOT an admin)

fuzzbuck:
> 
> 
> hang on, doesn't this prevent all non-admin players from joining? its 3 `&&`'s 🤔

I tested it a few times with multiple players, should work fine

PR #1550, opened 2020-02-11
[byte logic] 104

Quezler:
> rebased by hand, instead of merged by hand: #1532

The pull mentioned above was messy, so here is my 2nd attempt.

Its probably more stable than the other one, since the [diff](https://github.com/Anuken/Mindustry/compare/byte-logic) was easier to read.

Includes all (well, i hope so) of the `byte-logic` branch, plus some extras:
- controllable `item/liquid` `source/void/incinerator`, it kinda made sense
- (more have been added since, find them in the commit messages bellow*)

Screen Shot 2020-02-11 at 18 57 18
> since i cannot suggest a new branch via a pull request i instead aimed this version at being mergable, via the implementation of an experimental rule 😗


Arkanic:
nice

Quezler:
`// todo, figure out exactly what part of the conveyor freezing logic breaks the travis`

ZkyLB:
> `// todo, figure out exactly what part of the conveyor freezing logic breaks the travis`

is it fixed :<

Anuken:
As mentioned in `#development`, I'm not planning to use byte-logic in any capacity. Those who wish to play with the mechanic on their own can still use the proposed branch.

PR #1562, opened 2020-02-12
translations; missing tags added

heckelson:
I added more translation after going through the file once more, and I added all tags that were missing from the German file (`bundle_de.properties`) compared to the English one (`bundle.properties`).


martin-mfg:
looks good 👍

heckelson:
Did not get pulled.

PR #1563, opened 2020-02-12
[quality of life] illuminator color on the minimap

Quezler:
> this is both beautiful and disastrous at the same time

This'll be a nice place for its debate, with some working code attached to it 🙂 

- takes image > sorter converters to the extreme
- though lighting is off by default
- should the pixels darken if they are without (full) power?

Screen Shot 2020-02-12 at 17 09 59
Screen Shot 2020-02-12 at 17 09 56

 > feel free to close it any time, i have mixed feelings about weather this should exist at all 🤐


Quezler:
actually, i'll just save everyone the trouble and just close it, in its current state it should not exist 💣

PR #1566, opened 2020-02-12
[quality of life] better visiualize sorters with no filter

Quezler:
> a possible solution, there are more, feel free to suggest the latter.

Currently some filter types (cough: silicon) look way to much like an unfiltered sorter.
This pull steals the red cross from the byte-logic branch to more clearly mark them:
Screen Shot 2020-02-12 at 19 36 09
> yes, the 2 center ones are set to silicon, same color, just a different size, pretty hard right?

### DIFF
--- a/core/src/mindustry/world/blocks/distribution/Sorter.java
+++ b/core/src/mindustry/world/blocks/distribution/Sorter.java
@@ -66,12 +66,10 @@ public class Sorter extends Block{
         if(entity.sortItem == null){
             Draw.color(Pal.remove);
             Draw.rect(Icon.cancelSmall.getRegion(), tile.drawx(), tile.drawy());
-            Draw.color();
-            return;
+        }else{ // ▲ unfiltered | filtered ▼
+            Draw.color(entity.sortItem.color);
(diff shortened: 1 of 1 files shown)


geoffrey1900:
make it not 3d bc it makes it look bumping out from the actual sorter

Quezler:
`Icon.cancelSmall.getRegion()` just happens to look 3d apparently, but its a placeholder 🤔

Quezler:
this seems the culprit: (no idea currently how to add an exception for it)
```
    static FreeTypeFontParameter fontParameter(){
        return new FreeTypeFontParameter(){{
            size = 18;
            shadowColor = Color.darkGray;
            shadowOffsetY = 2;
            incremental = true;
        }};
    }
```

Anuken:
Drawing icons in-game causes a texture swap. I think this should be implemented with a gray X drawn specifically for the sorter instead; red stands out too much.

joshuaptfan:
Or could just add a subtle inner shadow effect to make it look "empty", like there's a hole there that can be filled, that would be instantly intuitive.

PR #1570, opened 2020-02-12
Add "pardon" command for votekicks

fuzzbuck:
Allows admins to "pardon" a votekick ban, useful for when a player gets unfairly votekicked and has to wait **60 minutes**, but he can instead be pardoned and join right away.

### DIFF
--- a/server/src/mindustry/server/ServerControl.java
+++ b/server/src/mindustry/server/ServerControl.java
@@ -661,7 +661,8 @@ public class ServerControl implements ApplicationListener{
         });
         
         handler.register("pardon", "<ID>", "Pardons a votekicked player by ID and allows them to join again.", arg -> {
-            PlayerInfo info = netServer.admins.getInfo(arg[0]);
+            PlayerInfo info = netServer.admins.getInfoOptional(arg[0]);
+            
             if(info != null){
                 info.lastKicked = 0;
                 info("Pardoned player: {0}", info.lastName);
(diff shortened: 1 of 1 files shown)


server/src/mindustry/server/ServerControl.java
```diff
@@ -659,6 +659,16 @@ private void registerCommands(){
                 err("That IP/ID is not banned!");
             }
         });
+        
+        handler.register("pardon", "<ID>", "Pardons a votekicked player by ID and allows them to join again.", arg -> {
+            PlayerInfo info = netServer.admins.getInfo(arg[0]);
```
Anuken: Use `getInfoOptional`; `getInfo` will always create an info if it does not exist, and may never return null.

server/src/mindustry/server/ServerControl.java
```diff
@@ -659,6 +659,16 @@ private void registerCommands(){
...
+            if(info != null){
+                info.lastKicked = 0;
+                info("Pardoned player: {0}", info.lastName);
+            } else{
```
Anuken: extra space

Anuken: I meant the space in `} else{` before the `else`.

fuzzbuck: oh

server/src/mindustry/server/ServerControl.java
```diff
@@ -666,7 +666,7 @@ private void registerCommands(){
             if(info != null){
                 info.lastKicked = 0;
                 info("Pardoned player: {0}", info.lastName);
-            } else{
+            }else{
                 err("That ID hasn't been votekicked!");
```
Anuken: Might be more accurate to say "That ID can't be found" instead.

fuzzbuck: oke

PR #1571, opened 2020-02-12
Added new lines and some small translation.s

artur309:
I added new lines and some small transaltions that was missing.

In future how can i know what is the new lines to translate? I had to search in the lines.


stuffyAI:
To see what lines were added and thus need to be translated simply follow commits to the english language bundle, there you can see which line was added where and then go to your language bundle on the same line to translate.

PR #1593, opened 2020-02-15
Separate ambient light and solar panel efficiency

DeltaNedas:
Before, solar panel multiplier was 1 - ambient light opacity
Now, solar panel multiplier can be set in game rules.
Requested by at least two people that have no affiliation with me!


Anuken:
Solar panel efficiency should default to the ambient light rule if nothing is set for this rule. Perhaps have it default to -1, and make solar panels use ambient efficiency if it's < 0.

DeltaNedas:
ok did

PR #1596, opened 2020-02-16
ItemSource and Liquid source always show all items

ExterminatorX99:


Anuken:
Why? Those blocks aren't (and shouldn't) used in the campaign. I can only see this being used for cheating.

ExterminatorX99:
That's exactly my intention. I've been trying to add an ItemSource block into a mod and have been failing to make it work when trying to overload the buildConfiguration method.

Anuken:
Why do you need to cheat in the campaign?

ExterminatorX99:
I've already completed the campaign and am now getting into mods. To save myself the grind and to test some of the features of a specific mod(Real life mindustry mod) I want to just put down an ItemSource and let it run for some time. My actual issue is that the ItemSource, as is, doesn't show all unlocked items. For instance raw copper ore and the ingots added by the mod even though I have them researched.

ExterminatorX99:
Further testing proved me wrong. The problem was that I was trying to use the ItemSource to create items I haven't yet revealed in the core database and that was the reason they weren't showing up in the list.

PR #1607, opened 2020-02-19
Reset local player name on disconnect

iczero:
Some servers may change the player name in game. This is incorrectly persisted after disconnecting from the server, causing it to carry over to other servers.

This is especially problematic when Mindustry is launched from steam, as player names cannot be easily changed there.


Kieaer:
I think this's my server lol

PR #1608, opened 2020-02-19
Prevent construction/deconstruction of blocks on other teams

iczero:
If blocks on other teams are manually added to the build queue, it is possible to construct/deconstruct them. This is abusable to steal resources from the enemy team during construction of blocks.


Anuken:
I fixed this yesterday, [here](https://github.com/Anuken/Mindustry/blob/master/core/src/mindustry/entities/traits/BuilderTrait.java#L78).

iczero:
Oh whoops. Sorry!

Quezler:
:ohno:

PR #1615, opened 2020-02-21
[votekick] neutral ballot 

Quezler:
just something i had in my own custom server jar for a while in case someone votes against 🤔


mRebornM:
Can i ask why?

Quezler:
> Can i ask why?

`foo has voted to kick bar` makes sense if you vote yes,
but when you vote no, the `kick` word still points to kicking said person,
so `kicking` just means the player has voted either yes or no.

mRebornM:
So its normally the same message for both yes and no, you fixed it?

Quezler:
yes, now the same message applies to both positive and negative votes

PR #1625, opened 2020-02-23
[serverlist] nydus

Quezler:
> nydus server submission //\\(oo)/\\\\

Screen Shot 2020-02-23 at 10 05 38

using a non default port to allow rewiring which server it point to on my end:
Screen Shot 2020-02-23 at 10 09 15

the reasoning behind this is to be able to switch the port away from the main server in the event that i'm testing a (minigame) plugin that needs a bunch of players, so i can leach of the nydus entry in the global server list.

i intend to keep a close eye on the stream of (new) players that join through this new listing, i already have a small fleet of admins (which i intend to grow more over time), as well as developing new(er) antigrief measures as reaction(s) to what happens on this server substrate.

furthermore i was curious about the "legality" of being allowed to turn off the port under certain circumstances (huge influx of griefers / filled to capacity / etc), since the `SERVERLIST.md` does not mention requiring a 24/7 uptime sla, so i am curious on your standpoint on it.

also, the player limit (as can be seen above) is currently 11, which is not "entirely" true: 
Screen Shot 2020-02-23 at 10 16 51
(there's a "secret" double cap in place, so players who aren't scared of the player cap being full can still get in if they want to, this change grandfathers the change that admins can always join)
of course i plan to adjust the player limit over time based on the influx of new players if this pull gets accepted.

that's about it, thanks for reading, let me know if you (or anyone) has questions :) 🥔 

> oh, and with `modified jar` i mean that i have some server sided enhancements in place to slightly modify the experience, while staying within the constraints what the server can pull off without players requiring custom clients and/or mods.

### DIFF
--- a/servers.json
+++ b/servers.json
@@ -9,6 +9,9 @@
     "address": "mindustry.indielm.com"
   },
   {
+    "address": "mindustry.nydus.app:1337"
+  },
+  {
     "address": "mindustry.ecansol.com:6597"
   },
   {


Anuken:
> the SERVERLIST.md does not mention requiring a 24/7 uptime sla, so i am curious on your standpoint on it.

Offline servers aren't displayed on players' lists, so this shouldn't be a problem unless the server gets abandoned completely and starts cluttering up the JSON file.

Arkanic:
that could be solved with some travis scripting
ping servers in global servers file on every build
after a certain amount of time, it removes the server in question from the JSON file

Anuken:
It's not enough of a (potential) problem to spend time writing a script to count offline time and auto-commit a file.

Quezler:
> Offline servers aren't displayed on players' lists, so this shouldn't be a problem unless the server gets abandoned completely and starts cluttering up the JSON file.

i noticed that part, i was mostly worried about clients needlessly pinging server ip's each time they open the list 🙂

PR #1630, opened 2020-02-24
[question] adminaction

Quezler:
Is there a specific reason the block is not set to null? (e.g. to educate reading out the type first) 😗


Anuken:
I guess I forgot to do it.

PR #1648, opened 2020-02-28
add shebang to convert_sounds.sh

DeltaNedas:
title

### DIFF
--- a/core/convert_sounds.sh
+++ b/core/convert_sounds.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 #convert from stereo to mono
 cd assets/sounds/
 for i in *.ogg; do


Anuken:
Does this fix anything specific? I've never had a problem without it.

mRebornM:
Im not sure if it fixes something but it defitely makes a difference

Anuken:
Right, but can you provide a specific example of the script failing to execute correctly due to absence of that line? I'd like to know if this was caused by a real problem that someone had, or if it's just 'standard practice'.

Quezler:
mostly depends on if the file is executable or not, so the os knows which to run when launched by hand iirc 🤔

mRebornM:
Why would you launch it by hand?

Anuken:
You wouldn't. This script is only supposed to be run through Gradle during iOS compilation.

PR #1655, opened 2020-03-02
bug_report.md table uppercase

bitt0008:


Anuken:
This is a file used by Github; the keys don't need to be uppercase, and in fact everything may break if they are.

PR #1662, opened 2020-03-04
Reverted RU translation of the Tower Defense genre

XEN0PHIL:
This genre name should not be translated (and the translation was wrong anyways)


Prosta4okua:
We should translate the genre. 
Screenshot_20200304-153053

PR #1674, opened 2020-03-05
[projector] visualization

Quezler:
> prototype implementation of https://github.com/Anuken/Mindustry-Suggestions/issues/14

Mar-05-2020 11-34-13

shows "to be affected" tiles when you hover a projector, crude implementation (to be improved)

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -716,6 +716,17 @@ public class Block extends BlockStorage{
         Draw.reset();
     }
 
+    public void drawTinted(Tile tile, Color tint){
+        if(tile.entity == null) return;
+
+        Draw.reset();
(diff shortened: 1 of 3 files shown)


Quezler:
ehmm, i wanted to set it to draft mode instead of a pull, yay not being able to toggle between them.

MEEPofFaith:
Wait if you close a pull request are you not able to open it again?

And why did it take so long for me to decide to ask that?

MEEPofFaith:
One thing I can think of is that blocks on the edge of the ring don't partially highlight like with the menders because the whole circle has the same healing.

Also, maybe if you have a block selected to be built, if you hover over a location where the block would be destroyed by the wave spawn, it would flash red so that you don't have to place it down and wait for the next wave to see if it implodes or not.

PR #1678, opened 2020-03-06
Touched the player and var classes to add ability to pull getPing() f…

Nicknakin:
…unction, along with added a call to the getPing() function in the playerlist ui element


Nicknakin:
Intended to address https://github.com/Anuken/Mindustry-Suggestions/issues/10

Anuken:
The client doesn't know the ping of other users, and your implementation will display every single user with the client's local ping.

PR #1684, opened 2020-03-08
set current mod to required mod

DeltaNedas:
or you will need to prefix mod name in requires of the required mod
instead you now need to prefix current mods name after youve required from another mod


Anuken:
I still don't want to make/test a proper mod yet, so I'll assume you know what you're doing here.

PR #1686, opened 2020-03-08
nearby water extractors decrease efficiency

DeltaNedas:
https://github.com/Anuken/Mindustry-Suggestions/issues/28
for every nearby extractor, it should decrease efficiency by 3%, compound
currently its dodgy and im seeing that newly placed ones have better efficiency or even over 100%!
todo: fix


mRebornM:
h for water extractor designs

PR #1691, opened 2020-03-09
[6.0] affinity antonym

Quezler:
> pardon the switch statement, but that was the most compact thing on my mind right now 😗
 
Screen Shot 2020-03-09 at 13 34 55

> some space between the lines might be necessary, they nearly touch one another's edges 🤔

### DIFF
--- a/core/src/mindustry/world/meta/BlockStats.java
+++ b/core/src/mindustry/world/meta/BlockStats.java
@@ -1,5 +1,6 @@
 package mindustry.world.meta;
 
+import arc.math.*;
 import arc.struct.*;
 import arc.struct.ObjectMap.*;
 import mindustry.*;
@@ -43,11 +44,7 @@ public class BlockStats{
 
     public void add(BlockStat stat, Attribute attr, int sign){
         for(Block block : Vars.content.blocks()){
-            switch(sign){
-                case-1: if(!block.isFloor() || block.asFloor().attributes.get(attr) >= 0) continue; break;
-                case 0: if(!block.isFloor() || block.asFloor().attributes.get(attr) == 0) continue; break;
-                case+1: if(!block.isFloor() || block.asFloor().attributes.get(attr) <= 0) continue; break;
-            }
+            if(!block.isFloor() || !Mathf.equal(Math.signum(block.asFloor().attributes.get(attr)), sign)) continue;
             add(stat, new FloorValue(block.asFloor()));
         }
     }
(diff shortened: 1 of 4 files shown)


fuzzbuck:
perhaps rename them to "Efficient on" & "Inefficient on"? Affinities & Aversions might be hard to understand for players who are not fluent in english

fuzzbuck:
Otherwise nice pr, makes sense that it wont work at full speed on literal lava lol

Quezler:
I'll leave the naming of them up to anuke, i just went with this word since it mirrors the existing one, and the java does look appealing but it could indeed use some improvements, any ideas? 🤔

Quezler:
as discussed in the #1093 pull "more/less efficient on" would take up space, here's a concept:
Screen Shot 2020-03-10 at 12 25 14
> obviously this will require some invasive code edits to get right, bundles/stats and such 😗

core/src/mindustry/world/meta/BlockStats.java
```diff
@@ -38,8 +38,16 @@ public void add(BlockStat stat, Liquid liquid, float amount, boolean perSecond){
     }
 
     public void add(BlockStat stat, Attribute attr){
+        add(stat, attr, +1);
+    }
+
+    public void add(BlockStat stat, Attribute attr, int sign){
         for(Block block : Vars.content.blocks()){
-            if(!block.isFloor() || block.asFloor().attributes.get(attr) <= 0) continue;
```
Anuken: The attribute sign check can be written (without the switch) as `if(!block.isFloor() || !Mathf.equal(Math.signum(block.asFloor().attributes.get(attr)), sign)) continue;`

Anuken:
Another option for naming is to ditch the concept of affinities/aversions completely and display a `Efficiencies:` list with each tile having a % number on it, possibly colored in red/green to help indicate whether it's better or worse.

Quezler:
> Another option for naming is to ditch the concept of affinities/aversions completely and display a `Efficiencies:` list with each tile having a % number on it, possibly colored in red/green to help indicate whether it's better or worse.

how would that translate to the amount of tiles the (multi)block is on? could get confusing 🤔

Anuken:
> could get confusing

Not at all. +20% could simply mean "20% more efficient for every block this producer is on", and -20% would mean the reverse. Being on two 20% blocks and having everything else be normal would lead to 140% efficiency.

PR #1696, opened 2020-03-10
Create short-description.txt

unakiritan:
Japanese Steam metadata.


Anuken:
I can merge this, but the Steam page won't work properly unless all the data is present (including the long description).

PR #1700, opened 2020-03-12

laohuaji233:


linglan512572354:
New items are ok and there a new rules recently.

However, Some Blocks name like "Phantom Builder", this commit changed it to the earlier version(maybe v99). I think change Block names frequently are not wise for a game, although the older version is more familiar to me.

Then the power Infinite is "power-source" not "power-source-node" please remove "节点” character. Furthermore, we call Maltdown "融毁” in daily communication but remain "炮” character can tell green hand player this block is a defense tower because "融毁” means nothing in Chinese. (others like arc, wave, etc.)

Finally, I make a typo in the previous commit, the alpha-mech-pad name is wrong, please amend it in your update. I think if there are many pull requests for one language will make Anuke confuse.

laohuaji233:
感谢你对我的翻译提出的建议。
我想你是一位中国人，所以我就直接用中文与你交流好了。
有关新版本的无人机的翻译，我会想出一个好名字来代替的。
至于炮塔后面加炮，我觉得有些炮塔没有必要加，就犹如我们说AK47而不说AK47枪一样，像幽灵改名叫幽灵炮我觉得不妥。其他一些炮塔同理。
还有一些翻译我也会改正过来的。

PR #1722, opened 2020-03-17
Conveyor/Conduit upgrade Assistance

ma44:
Crossed out no longer applies, assistance only works with rotatable blocks
~~Not really working well with walls due to some neat behaviors, see discord link for example
https://discordapp.com/channels/391020510269669376/396416151032299521/689548505311739977~~

This is a rehashing of #1009 but incorporates change requests by Anuken including a color change, dedicated keybinding and some code replacement


core/src/mindustry/input/Placement.java
```diff
@@ -56,6 +56,30 @@
...
+    public static Array<Point2> followLineUntil(int startX, int startY, int endX, int endY){
+        Pools.freeAll(points);
+        points.clear();
+
+        Tile cursor = world.ltile(startX, startY);
+        Block filter = cursor.block();
+
+        while(true){
+            if(cursor.block() != filter) break;                        // keep looking until encountering a different block
+            if(points.contains(Pools.obtain(Point2.class, Point2::new).set(cursor.x, cursor.y))) break; // avoid biting its own tail
```
ma44: While the original PR did have a change request to make this into a lambda statement, this utilizes a changing `cursor.x, cursor.y` making lambda expression usage error out, I don't have enough knowledge to figure out an alternative way of doing this so any suggestions would be helpful.

Anuken:
```suggestion
            if(points.contains(p -> p.equals(cursor.x, cursor.y))) break; // avoid biting its own tail
```

core/src/mindustry/input/Placement.java
```diff
@@ -56,6 +56,30 @@
...
+            points.add(Pools.obtain(Point2.class, Point2::new).set(cursor.x, cursor.y));
+            if(cursor.x == endX && cursor.y == endY) break;            // do not look further than the cursor
+            cursor = cursor.getNearby(cursor.rotation());              // move cursor to next block to check
+        }
+        // if the end is nowhere on the line, don't return anything
+
+        if(!points.contains(p -> p.equals((Pools.obtain(Point2.class, Point2::new).set(endX, endY))))){
```
Anuken:
```suggestion
        if(!points.contains(p -> p.equals(endX, endY))){
```

core/src/mindustry/input/Placement.java
```diff
@@ -56,6 +56,30 @@
...
+
+        Tile cursor = world.ltile(startX, startY);
+        Block filter = cursor.block();
+
+        while(true){
+            if(cursor.block() != filter) break;                        // keep looking until encountering a different block
+            if(points.contains(Pools.obtain(Point2.class, Point2::new).set(cursor.x, cursor.y))) break; // avoid biting its own tail
+            points.add(Pools.obtain(Point2.class, Point2::new).set(cursor.x, cursor.y));
+            if(cursor.x == endX && cursor.y == endY) break;            // do not look further than the cursor
+            cursor = cursor.getNearby(cursor.rotation());              // move cursor to next block to check
```
Anuken: I don't think blocks without rotation should be part of the line. Even if this works with walls, it's confusing and unintuitive.

core/src/mindustry/input/Placement.java
```diff
@@ -56,6 +56,30 @@
...
+        }
+        // if the end is nowhere on the line, don't return anything
+
+        if(!points.contains(p -> p.equals((Pools.obtain(Point2.class, Point2::new).set(endX, endY))))){
+            Pools.freeAll(points);
+            points.clear();
+            points.add(new Point2(startX, startY));
```
Anuken:
```suggestion
            points.add(Pools.obtain(Point2.class, Point2::new).set(startX, startY));
```

core/src/mindustry/input/Placement.java
ma44: bloop
Fails to compile

core/src/mindustry/input/Placement.java
ma44: Agree with this, added a check to see if the block can rotate or not so only conveyors/conduits get changed

core/src/mindustry/input/Placement.java
ma44: Addressed

Anuken:
I won't be merging this after all - the keybinds are already very bloated, and it's usually not too hard to just upgrade conveyors manually. I may revisit this if someone comes up with a way to do it intuitively without new keybinds.

PR #1733, opened 2020-03-21
Add Category Key Bindings

cyrusduong:
Related https://github.com/Anuken/Mindustry-Suggestions/issues/41

Creates the ability to re-assign at least the category hotkeys. 

This is a passive change. The user will have to manually go re-assign the category keys from their default (will continue to act acts like block selection in master). 

If they reassign the keys they can then use a different key combo to select blocks.

For example: 

1. A user rebinds category 3 to the key 'B' (conveyors category assuming the full tech tree is unlocked).
2. When a user is in-game. They can select the 3rd block in that category by pressing the combo `B-3`. 

Tested on Windows using `gradlew.bat desktop:run`.


cyrusduong:
I did notice that the keybinding hints 
image
Don't seem to be updating when the keybinding is changed... I'll see if I can figure that one out. Hopefully, it's not a property/resource string....

zethnest:
Screenshot_20200409_033847
The keybinding hints do update when the keybinding is changed. This is release 104.10

Anuken:
I would rather implement a new hotkey/bar system than keep changing this one.

PR #1738, opened 2020-03-23
Updated tutorial text and added emojis

XEN0PHIL:
Changed lines from .intro to .launch


XEN0PHIL:
1. Is there issues with an updated text?
2. Should I put brackets around all emojis?
3. Should I remove all brackets around emojis?
4. Should I add another space between the word and and emoji?

Quezler:
question for anuke: which way of typing them is preferable?
-  "chinese style"
- \uf822 "unicode escape"
- https://github.com/Anuken/Mindustry-Suggestions/issues/18 "color tag-ish"

Anuken:
I would use "chinese-style" for now. :emoji: style would be an improvement in readability, but requires some further consideration to determine how exactly it will be implemented.

Anuken:
> Is there issues with an updated text?

There are some improvements in clarity, but I see numerous grammatical errors in your changes. I'll need to look over it.

PR #1745, opened 2020-03-24
Update and rename core/assets/bundles/bundle.properties to  ядро/ акт…

NickZhYT:
…ивы/ связки/bundle.properties

Some description edits


Anuken:
See other comment

PR #1759, opened 2020-03-26
Make buildTime, time, and speedScl public

z0mbiesrock:
Also added variables for allowing a custom spawn effect and/or sound.

This should make scripting a custom UnitFactory animation easier.


Anuken:
I can't accept this pull request, as unit factories are rewritten in 6.0 and this would only cause merge conflicts.

PR #1775, opened 2020-03-28
Update bundle.properties

NickZhYT:
Some description changes


Anuken:
I don't think it's necessary to list all the materials that the generator accepts. If someone needs to, they can just check the stats.

You've also misspelled "pyratite".

PR #1785, opened 2020-03-30
Fix use 'name' from bandle

burunduk:
To improve localization, please split the use of this variable.


Anuken:
Unless you can specify a problem that this PR solves,  it doesn't seem necessary to create more duplicate bundle entries. If anything, `editor.name` should be removed and replaced with `name`.

burunduk:
"name" for player and "name" for map in my language is diferent words.
"Імʼя" - plyer name, sensate
"Назва" - map name, insensate

Anuken:
Couldn't you just change the map dialog code to use `editor.mapname` then, and avoid adding any new entries?

Issue #776, opened 2019-10-04
Overflow conveyors sometimes divert resources on a fully compressed but moving belt


Prezombie:
**Platform**: Windows

**Build**: Steam v96

**Issue**: I had a fully filled line of copper going to the core, and set up an overflow belt running to a launch pad. about every 5th piece got diverted to the side belt, leaving gaps in the main output line. I think there's some error in how the overflow block is checking for space in the line ahead, when it should only ever be diverting resources if the line ahead is blocked _and stationary._

Anuken:
I'm not sure if this is really a bug. If the conveyor belt is full, then the overflow gate can't pass any items forward, regardless of whether the conveyor is stationary or not. What if it's trying to pass an item to a conveyor from the side? It may be unable to do so, but the conveyor is technically *not* stationary, as more items can be accepted from the back, and it is still moving. 

The real issue here is that items don't move at perfectly consistent speeds across conveyors, which leads to occasional 'overflow'. Unfortunately, there isn't a viable fix for that.

JacksonKearl:
@Anuken, do the overflow gates have any sort of “debounce”? Such that if the output is momentarily blocked it will wait a bit for output to open before switching to overflow mode?

serdarakkilic:
> @Anuken, do the overflow gates have any sort of “debounce”? Such that if the output is momentarily blocked it will wait a bit for output to open before switching to overflow mode?

This will slow down item transportation dramatically. Each overflow gate you add to the setup will even slow down the transportation/distribution more.

albert-92:
**Platform**: `Windows 10 x64`

**Build**: `steam build 121.4`

The problem depicted by https://github.com/Anuken/Mindustry/issues/776#issuecomment-576293101 is still existent in the newest steam build.

Anuken:
It won't ever be fixed.

PR #1791, opened 2020-03-31
Nerf overflows.

ritzip:
set overflow speed to be the same as titanium conveyors

### DIFF
--- a/core/src/mindustry/world/blocks/distribution/OverflowGate.java
+++ b/core/src/mindustry/world/blocks/distribution/OverflowGate.java
@@ -12,7 +12,7 @@ import java.io.*;
 import static mindustry.Vars.world;
 
 public class OverflowGate extends Block{
-    public float speed = 1f;
+    public float speed = 6f;
     public boolean invert = false;
 
     public OverflowGate(String name){


ZkyLB:
hamburger

bitt0008:
why

ritzip:
This aims to stop unhealthy "sneks" that have op throughput.
unknown

ado1928:
oh come on ritzip

ph3rr3t:
There was no problem before with Overflows & Sorters/Inverted.  What changed?  I only noticed this after Underflows were added.  If Over/Unders be nerfed, have it be a max of 41 items per second.  Why 41 and not 40/30/60?  Close to 4 full Titanium Belts worth of items is all I was ever able to pass through Overflow Sorter chains before we got Inverted or Underflows.  Any designs that used this mechanic before it got out of hand would still function.  I knew it was inevitable, but completely removing the mechanic altogether kills a lot of design compaction.  Just my thoughts on the matter.

ritzip:
Screenshot_20200331-205000~2

ph3rr3t:
The problem currently is that you can pass an infinite amount of items through a single chain right?  Input any material you want at any point, and output it at any point?
Well to keep if from being a true snek, just limit to about what it was before, that is to say, close to 50-51 items per second.

ado1928:
No. Items are passed every frame, allowing for 60 items/s. IMO that is good enough, and can take 5 titanium conveyors

ado1928:
Also, no the throughtput didnt change after the addition of underflows, as long as i know, it has always been 60/s

ph3rr3t:
Serdar informed me that it wasn't until this fix that it changed:
https://github.com/Anuken/Mindustry/issues/776

Anuken:
There is a [line in the source code](https://github.com/Anuken/Mindustry/blob/master/core/src/mindustry/world/blocks/distribution/Sorter.java#L92) that can prevent chaining if uncommented. It hasn't been implemented due to controversy.

This PR has negative side effects, so I'd rather not merge it.

ZkyLB:
cute cat

Anuken:
stop

PR #1794, opened 2020-03-31
Create bundle_by.properties

KeviTV:
Full translation of the game for the Belarusian language.


Anuken:
The Belarusian [language locale](https://www.science.co.il/language/Locale-codes.php) code is `be`, so you should name the file `bundle_be.properties`.

PR #1804, opened 2020-04-01
fix for js print() function printing null

Blockzilla101:
js print() function prints null, if zero or false is passed to it as an argument.


Anuken:
Fixed in a simpler way, see commit.

PR #1809, opened 2020-04-03
Updated Phantom description

XEN0PHIL:
Added info "Rebuilds destroyed blocks"


not-purple:
yea, it will rebuild. but it travels to the next block, can it be program to fix near block first

PR #1810, opened 2020-04-03
[quality of life] resuming construction in bulk

Quezler:
Currently when you run out of building resources (or when you die) half way through construction you're left with one or more build blocks with barely any progress, and the only way to complete what was intended is to either click all of them by hand, or delete them in order to replace them.

This draft allows people to resume the build by allowing it to be added to the build queue if the block being constructed matches the current type.

Some things that are up for debate:
- it only allows the original block to be constructed, no wall/conveyor/drill hotswap upgrading.
- it keeps the rotation the original block was in, could either be annoying or fairly useful.
- and of course the code, no idea if i thought of everything, haven't tested it a lot yet. 

Apr-03-2020 16-03-46
> gif shows me running out, then scavenging copper to resume via dragging a new line 🥔

### DIFF
--- a/core/src/mindustry/world/Build.java
+++ b/core/src/mindustry/world/Build.java
@@ -87,7 +87,7 @@ public class Build{
         if(tile == null) return false;
 
         if(type.isMultiblock()){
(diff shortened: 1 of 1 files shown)


not-purple:
what if there is a hotkey to resume in bulk, instead of placing each type

PR #1812, opened 2020-04-03
[backport] sorter cross

Quezler:
first off, it looks absolutely amazing 😊 
secondly, i cannot wait to get my hands on it, so i hereby submit this backport request 📚 
finally, how do you feel about backport(s) in general? (since i might try some more) 🤔 

unknown-2


Anuken:
I don't think backports should be necessary; 5.0 seems stable right now, and now that April 1st is over, most platforms are very unlikely to get an update. This will only end up causing some merge conflicts with the sprite files.

Quezler:
@Anuken could this either be merged or closed please? 🙂

PR #1835, opened 2020-04-10
I18n es tiny fix

mdtrooper:
There was a tiny error: 

> Alcanzar la oleada{0} en {1}

The fix is:

> Alcanzar la oleada {0} en {1}


mdtrooper:
Thanks :beers:

PR #1843, opened 2020-04-12
Extended indicator options

zethnest:
When playing survival or attack mode in multiplayer, the indicators are both useful and annoying because it shows both allies and enemies at the same time. So I separated them individually so they can be toggled in the settings. 

Also I didn't know there was a mech indicator so I put that in the settings menu as well.


Quezler:
the mech indicator changes will cause problems upstream in 6.0 (where they are removed) 😗

zethnest:
I see. I'll remove the mech part if there is a need to merge.

Anuken:
Implemented in 6.0.

PR #1845, opened 2020-04-12
[playerlist] change votekick hammer to ballot

Quezler:
Currently the `hammer` icon votekicks for normal players, while the same icon allows for instant kicking by an admin, imo it would make sense of the icon for normal player looked more like a ballot / voting box of some sort, but since such an icon wasn't readily at hand i picked the ticket icon while we sort the details out in this pull discussion 🤔 (its not a problem, just a qol change 🙂)
> Screen Shot 2020-04-12 at 12 22 49


Quezler:
correction, for admins its a ban, not a kick 😗 (which only increases the need for clarity)

Quezler:
suggested icon from a community member:
SPOILER_unknown

Quezler:
https://www.flaticon.com/free-icon/kick_1928 courtesy of @Blockzilla101

Anuken:
I don't think an icon of a ticket/ballot is intuitive at all; if I saw that in a game, I wouldn't be able to tell what it does. A boot icon could work, but that particular one doesn't fit Mindustry's icon style very well.

Quezler:
I could try to design one, which parts of it conflict with the design style from your point of view? 🤔

Anuken:
It's too detailed and has a lot of curved rotated lines, especially with the spiky bits at the right.

Quezler:
noted, i'll get to work 🎨

Quezler:
well this inkscape pixel > vector conversion was a nice disaster:
Screen Shot 2020-04-13 at 09 26 35
Screen Shot 2020-04-13 at 09 26 24

Anuken:
Honestly, I don't think a boot icon would clear anything up here. Yes, the same icon is used for admin kicks and for votekicks, but as long as people know that both actions involve kicking, that's not a problem.

PR #1867, opened 2020-04-14
Better Bridge Indicator

zethnest:
Currently, when you overlap a lot of bridges, it becomes harder to find which goes to which. So I came up with a visual that shows where the input comes from and where the output goes to if there are any connected bridge designated for output.

Screenshot_20200414_202005
This works with all ItemBridge types.


Anuken:
At a glance, I can't tell what's going on in the screenshot, and I don't think many players will be able to, either. It's not obvious which bridge is connected to what, or which way the items are flowing.

zethnest:
I see. That'd make sense. I guess some arrows might do.

zethnest:
Peek 2020-04-16 22-40
It now looks like this. I used arrows to show where it goes and now uses the proper colours matching the indicator of drawConfig. Yellow shows where it comes from. Blue shows where it goes.

zethnest:
I'm making a new pull.

PR #1868, opened 2020-04-15
Client whisper command

Arkanic:
simple player-to-player messaging
works like whispering in minecraft


Anuken:
It's hard to type out user's names correctly in Mindustry, as many of them could have color tags or icons. Unless you can find a way to circumvent that, this won't be very useful.

zethnest:
@Arkanic a way to circumvent that, you can add a new button at the player list that will take you to the chat pre typed with the player name. Also add '/r' to reply to the recent incoming whispered player.

My idea of making it easier.

Arkanic:
I failed horribly last time i messed with gui
i'll try my best though

Arkanic:
i shortened command names down to "w" for whisper and "r" for reply

Anuken:
I still think that this won't be particularly useful without UI, considering how complex player names can get. I'll reopen this if there's demand for a whisper command.

Arkanic:
Fair enough, seeing as ID's are kind of clunky and names aren't always typeable with exotic non-dvorak characters and the like

PR #1869, opened 2020-04-15
Server locale support and filtering

DeltaNedas:
This pr adds a `locale` field to the server list which specifies what language it uses, if not English.
Clients check this against their current language to filter out servers.
Currently there is no way to disable it, but I wouldn't mind adding a setting for it.


ActualWina:
idk how much useful this is, most servers already have a language clarification in title. I suggest adding this as a requeriment for uploading a server to public server list, and just that.

DeltaNedas:
travis pls

Anuken:
It would be a better idea to just let players hide community servers.

PR #1870, opened 2020-04-15
Added instructions for a edge case

iDenis124:
For when the font is missing a language's character, and general instructions on how to fix it.


Anuken:
This isn't very helpful information, and it's not the course of action I would suggest. It's usually best that I add the characters myself, since I know how the tools work and I need to be sure the font you're adding characters from has the right license.

iDenis124:
So it would better to just translate, and wait for you do add the characters, or translate and also specify which characters are missing?

ActualWina:
Exactly, I would add the clarification to warn when there are missing characters, but adding it by itself sounds somewhat more complicated and can give unnecessary errors.

Anuken:
You should let me know what's missing in the PR.

PR #1875, opened 2020-04-16
Update bundle.properties (color code correction)

Ali-C-Ila:


Anuken:
This isn't correct.

`"{0}[white]"?` makes sure that any color tags in the user's name don't affect the question mark and quote. Since the message is already white, writing `[white]` *before* the user's name does nothing. Placing `[]` after the name isn't a good idea, as it would reset the message to the last color code, and player names may contain multiple color codes.

PR #1883, opened 2020-04-17
Fix the Reaper Sprite

MEEPofFaith:
Edges used to be weird and blurry and it was way bigger than the original sprite.


Anuken:
The reaper isn't even a real unit in 6.0. The art asset used was only a test, there's no need to change it. I won't be using either of these sprites in the release.

MEEPofFaith:
:ohno:

Wait so you're yeeting the current units out of existence?

kapzduke:
Wait, so the 6.0 reaper will be the same as the legacy reaper??

PR #1884, opened 2020-04-17
[6.0] conduit/conveyor cornerpieces

Quezler:
> fills in the "triangle" gaps in the corners when there are multiple snakes next to eachother

Just something that has been bothering me since forever, during normal gameplay & whilst working on my plastanium conveyor draft: #1260

Instead of trying to tack it onto that pull request, i have it here as a standalone:
- conveyors get corners when the two other edges are of the **same color** and do not face out
- conduits get corners when the two other edges are **also conduits** and do not face out

technical issues:
- not sure if the way of detecting it is up to standard
- when zooming in there are slight pixel gaps visible

(slightly fixed that last part by adding inward pixels, but its still slightly noticable)

Screen Shot 2020-04-17 at 12 02 27
Screen Shot 2020-04-17 at 12 05 24
Screen Shot 2020-04-17 at 12 12 54

> i'll leave it here as a draft for now, where it can gather dust and reactions over time 🧹


Quezler:
weird, the discord bot didn't warn about this one being closed ^

HaphLife:
Not sure what I think about armored conveyors not connecting to other conveyors and other conveyors to armored conveyors. In all other cases armored conveyors connect to conveyors and vice versa. However given how armored conveyors work, I understand the idea.

Anuken:
This doesn't really bother me, and I don't think the additional code/sprites (and more requirements for conveyor modders) is worth it.

PR #1889, opened 2020-04-18
[6.0] Better bridge indicator from #1867

zethnest:
Port from pull request #1867 compatible with 6.0


zethnest:
image
Pardon for the FPS drop. It now looks like this. I used arrows to show where it goes and now uses the proper colours matching the indicator of drawConfig. Yellow shows where it comes from. Blue shows where it goes. Ported to v6

Anuken:

PR #1915, opened 2020-04-25
New power cells for everything

MEEPofFaith:
I know you mentioned that the old units probably won't exist anymore, but I was bored and decided to this anyways.


Anuken:
I'll probably change/delete many of these, but they're a good starting point.

PR #1919, opened 2020-04-26
Add Mo Town servers to community servers

Momiro:
I have two dedicated servers running and I would like them to be visible in community servers. One is for maps that I make, another is a hexed pvp with no unit factories.

### DIFF
--- a/servers.json
+++ b/servers.json
@@ -41,4 +41,10 @@
   {
     "address": "aamindustry.play.ai"
   }
+  {
+    "address": "motownmindustry.duckdns.org"
+  }
+  {
+    "address": "motownmindustry.duckdns.org:6577"
+  }
 ]


Anuken:
> hexed pvp with no unit factories

How would that work?

Momiro:
I have simply banned all of the unit factories with server rules.
If you mean how would the competitive gamemode work, players can use long ranged turrets to damage other players structures, capturing hexes still works. I would rather have a limit on how many unit factories can be placed at once, but I don't know of a way to do that.

The reason I made the no unit hexed pvp server is because I've seen people complaining that if they join in the middle of a hexed pvp game, they immediately get steamrolled by other players existing units.

Anuken:
> they immediately get steamrolled by other players existing units.

Artillery sounds like even more of a disaster. It was banned because of constant spawnkilling and sniping.

Momiro:
Hmm, is there a way to limit tge number of certain blocks spawning? Or would it be better to limit drones to a certain type?

Anuken:
Neither are possible without custom plugins.

Momiro:
The default server wrapper lets you ban certain unit factories, If I unban some of the less "OP" units, would that be more viable?

Momiro:
Perhaps just add the server with my survival maps on it for now, until I come up with a better solution for a more beginner friendly Hex PvP?

ActualWina:
If u can prevent players building next to each other, that hex pvp should work

Momiro:
I would do that if I could. I could also give players a largsr starting supply of copper so they can build some basic defences first

sk7725:
Making a newly spawned player 'immune'(Force projector-style force field that persists for 120 seconds) is a good way to discourage spawnkilling...Heck i should make a plugin for that later

PR #1921, opened 2020-04-26
[doodle] terrain schematics

Quezler:
> this is just me playing around with the code to get closer to the below explained goal

Screen Shot 2020-04-26 at 12 38 12

_basically, allowing any block, floor & overlay in a schematic:_

Currently we have `Loadouts.java`, which contains as you know the starting core layouts, but that only consists of normal blocks, and its hardcoded to place ores under drills.

I propose embedding support into schematics for floor & overlay layers as well.

The reason i'm diving into this now is since i'm planning to explore the 6.0 `tile.setFloor` calls, and it would be nice to have some predefined "ruins" or "decorated ore patches" that are placable on the map (by the server) in realtime.

And to do that i'll either have to modify the schematic format / come up with some sort of editor generation setting for predefined embeds / create like a secondary map to be loaded in that functions like a spritemap of some sort that i'll have to map out / etc, basically the schematic approach sounded most logical.

I don't really intend for players to be able to use it in single player or multiplayer, but while developing this it does have some promising usages for the ingame editor.

The reason i'm leaving this draft here is for discussion on what would be a nice approach, since i'd rather not bodge in my own hacky solution that lack official support, i'm willing to bet that more people find this idea interesting, and thus a more official implementation would be required.

Thanks for reading, please leave a reaction (either emoji or text)

**tldr** easily defining structures that use all 3 layers that the server can paste into the world 🥔 

> needless to say this code currently only sorta does the graphical part, not much else (yet)

### DIFF
--- a/core/src/mindustry/Vars.java
+++ b/core/src/mindustry/Vars.java
@@ -125,6 +125,8 @@ public class Vars implements Loadable{
     public static boolean steam;
     /** whether typing into the console is enabled - developers only */
     public static boolean enableConsole = false;
+    /** whether schematics can be used as blueprints - developers only */
+    public static boolean enableBlueprint = true;
     /** application data directory, equivalent to {@link Settings#getDataDirectory()} */
     public static Fi dataDirectory;
     /** data subdirectory used for screenshots */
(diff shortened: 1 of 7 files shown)


Anuken:
This would be fitting as part of a 7.0+ editor rework, when I improve the in-game editor and add a selection copy-pasting tool.

I don't think there's much to discuss here - the technical details depend on the use case. I'm not entirely sure that the schematic format (e.g. storing each tile separately as an x/y coordinate) would be a good fit for terrain at all, but we'll see.

PR #1927, opened 2020-04-27
[QOL] Better Visual of Bridge Indicator

zethnest:
This is how the bridge indicator looks like.

### DIFF
--- a/core/src/mindustry/world/blocks/distribution/ItemBridge.java
+++ b/core/src/mindustry/world/blocks/distribution/ItemBridge.java
@@ -175,23 +175,25 @@ public class ItemBridge extends Block{
                     if(!linkValid(tile,other,false)) continue;
                     boolean linked = other.pos() == link;
                     if(!linked && !(other.<ItemBridgeEntity>ent().link == tile.pos())) continue;
-                    float tx = tile.drawx(), ty = tile.drawy(), ox = other.drawx(), oy = other.drawy();
-                    float alpha = Math.abs((linked ? 100 : 0)-(Time.time() * 2) % 100f) / 100;
+
+                    float tx = tile.drawx(), ty = tile.drawy();
+					float ox = other.drawx(), oy = other.drawy();
+                    float alpha = Math.abs((linked ? 100 : 0)-(Time.time() * 2f) % 100f) / 100f;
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/blocks/distribution/ItemBridge.java
```diff
@@ -167,6 +167,37 @@ public void drawConfigure(){
...
+                    if(!linked && !(other.<ItemBridgeEntity>ent().link == tile.pos())) continue;
+                    float tx = tile.drawx(), ty = tile.drawy(), ox = other.drawx(), oy = other.drawy();
+                    float alpha = Math.abs((linked ? 100 : 0)-(Time.time() * 2) % 100f) / 100;
+                    float x = Interpolation.linear.apply(ox, tx, alpha);
+                    float y = Interpolation.linear.apply(oy, ty, alpha);
+
+                    Tile link = linked ? tile : other;
+                    Tile otherLink = linked ? other : tile;
+
+                    Draw.color(Color.valueOf("444444"));
```
Anuken, line 186: `Color.valueOf` creates garbage. Cache this as a final static variable.

core/src/mindustry/world/blocks/distribution/ItemBridge.java
```diff
@@ -167,6 +167,37 @@ public void drawConfigure(){
...
+                    Lines.stroke(2.5f);
+                    Lines.circle(ox, oy, 2f);
```
Anuken, line 188: Circles don't really fit here, and may reduce performance. Draw a 45° rotated square instead (`Lines.poly()`)

core/src/mindustry/world/blocks/distribution/ItemBridge.java
zethnest: I found that `Pal.gray` is very close to `Color.valueOf("444444")` so I used that.

zethnest:
gifine
This is what it currently looks like. I've tried using it in 5.0, and it doesn't impede or even disrupt what I'm doing at the moment at all. It sometimes helps me in identifying which goes to which.

Anuken:
Merged in 2280ed6

PR #1932, opened 2020-04-27
Water extractor now requires 25 metaglass rather than 25 graphite to construct

ma44:
Despite being a building for extracting liquids, it required graphite rather than metaglass; this change is mainly to just make it require metaglass rather than graphite to construct. 

With the recent PR of nerfing water extractors, this comes with the added bonus of already needing metaglass income beforehand so you can actually do piping rather than wonder why you spent 25 graphite on some building

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1223,7 +1223,7 @@ public class Blocks implements ContentList{
         }};
 
         waterExtractor = new SolidPump("water-extractor"){{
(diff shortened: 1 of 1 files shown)


ActualWina:
maps without sand will be almost dead with this. The nerf is enough, why more.

ThatOneBepis:
Scrap is a thing you know

BlueWolfGaming:
@ActualWina :bruh: if a map dont have sand then that map shouldnt be in existence.

Anuken:
I can change this easily myself, so there's no need for a PR. The fate of water extractors will be decided later.

PR #1933, opened 2020-04-27
Update bundle-fr.propeties : adds regions

feu-follet:
Adds regions with comments to make it easier finding a content to update
Not finished at all, but I'll continue if get validated


Anuken:
This won't work! Bundle comments don't start with `//`, and the updater script automatically removes all extra space.

PR #1935, opened 2020-04-28
[projector] visualization v6 port

zethnest:
Port of #1674 to v6 API


Anuken:
Implemented differently:

recording1588094127

PR #1944, opened 2020-04-30
Overdrive projector visualization on plastanium conveyors

Quezler:
> only shows the overdrive indicators for the loading dock

Screen Shot 2020-04-30 at 16 01 11

> personally not so happy with the way this now does a double check though

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -277,6 +277,10 @@ public class Block extends UnlockableContent{
         return rotate;
     }
 
+    public boolean shouldOverdrive(Tile tile){
+        return canOverdrive;
+    }
+
     /** Adds a region by name to be loaded, with the final name "{name}-suffix". Returns an ID to looks this region up by in {@link #reg(int)}. */
     protected int reg(String suffix){
         cacheRegionStrings.add(name + suffix);
--- a/core/src/mindustry/world/blocks/defense/OverdriveProjector.java
+++ b/core/src/mindustry/world/blocks/defense/OverdriveProjector.java
@@ -99,7 +99,7 @@ public class OverdriveProjector extends Block{
         public void drawSelect(){
             float realRange = range + phaseHeat * phaseRangeBoost;
 
-            indexer.eachBlock(this, realRange, other -> other.block().canOverdrive, other -> Drawf.selected(other, Tmp.c1.set(baseColor).a(Mathf.absin(4f, 1f))));
+            indexer.eachBlock(this, realRange, other -> other.block().canOverdrive && other.block().shouldOverdrive(other.tile()), other -> Drawf.selected(other, Tmp.c1.set(baseColor).a(Mathf.absin(4f, 1f))));
 
             Drawf.dashCircle(x, y, realRange, baseColor);
         }
--- a/core/src/mindustry/world/blocks/distribution/StackConveyor.java
+++ b/core/src/mindustry/world/blocks/distribution/StackConveyor.java
@@ -102,6 +102,11 @@ public class StackConveyor extends Block implements Autotiler{
         return super.rotatedOutput(x, y);
     }
 
+    @Override
+    public boolean shouldOverdrive(Tile tile){
+        return tile.<StackConveyorEntity>ent().state == stateLoad;
+    }
+
     public class StackConveyorEntity extends TileEntity{
         public int state, blendprox;
 


Anuken:
Why? Looking at the code, any segment has its capacity increased by the projector.

Quezler:
@Anuken the code that checks for the overdrive effect only exists on the loading dock, once its filled above its capacity that way it can move over track pieces that aren't overdriven, this pull will help players realize that (unlike normal conveyors) they do not need to boost the entire line ^

Anuken:
Perhaps that shouldn't be the case? It's not very intuitive behavior.

Quezler:
its one of the key features of the plastanium conveyor that only the loading dock needs to be overdriven, since that is where the overdrive projector tends to be, which in turn can be used to compactly load items onto the roomba, and send that off to its destination without having to worry about 1 or more conveyor tiles that are just outside its range and thus voiding the entire boosted status

Quezler:
vaguely similar to those japanese train pushers, they help the ~~people~~ cargo get in, and then it can go it its destination unhindered, which incentivizes running the line entirely to its destination without breaking it with junctions or other item related blocks (which basically rewards good layout/designs with them)

Anuken:
I don't think that's a good enough reason to change mechanics in such a way; I can see players being confused about why their conveyors aren't moving any faster. It's very easy to make them speed up instead:

recording1588270000

Quezler:
it lists the effect in the info dialog at the bottom though:
Screen Shot 2020-04-30 at 20 09 33
Screen Shot 2020-04-30 at 20 09 40

Anuken:
Not many people would see that, and I still don't see why item capacity should be preferred over speed.

> which incentivizes running the line entirely to its destination without breaking it with junctions or other item related blocks 

Wouldn't a speed increase do the same?

Quezler:
the thought behind them was to reward players for (phased) overdrive projector usage, without it the throughput is comparable (iirc even slightly below) the titanium conveyor, and since this conveyor is meant to be used industrially its rewarding to them together, and once having them packed to capacity it would be a priority to get them (in)to their destination without decompressing them, so bridging over or around them would be preferable than cutting through it with junctions/sorters/othermeans, it was a design choice to promote clean lines

Quezler:
a speed increase instead of item capacity would of course still make them fairly great, but it in turn doesn't give the player a reason to not interupt the line since then the only thing holding them back is the `recharge` speed (amount of loading docks to fully saturate a line)

Quezler:
(and from a personal standpoint, boosting a very long line all the way to the core with overdrive projectors (or even with phase in them) is fairly wasteful, not just on power but also on space, and phase fabric is often fairly rare, and best spent on the factories that you need, and if you use plastanium conveyors as the output lines of those you get rewarded appropriately)

Quezler:
on a slightly (un)related note, the inner darker circle on the roomba is bothering me, it looks great as a texture file, but once there is a small stack of anything on it it kinda interferes with the shadows in (from what i believe) an ugly way:

Screen Shot 2020-04-30 at 20 22 47

Anuken:
> it in turn doesn't give the player a reason to not interupt the line

Theoretically, shouldn't both methods require the same throughput? A line with 2x the items per stack and 1x the speed is processing the same amount of items, on average, as a line with 1x the items at 2x the speed. Both methods will require any intermediate blocks like junctions or conveyors to have the same throughput for uninterrupted movement.

> is fairly wasteful

Sure, but it's supposed to be. Making every conveyor in the line be able to stack more items just because the loading dock is overdriven feels like magic. If there's a problem with overdrive balance or cost, that should be handled in the overdrive projector itself, not in this particular conveyor.

Quezler:
> magic
well normally the overdrive projector boosts the activities of the block itself, and only the loading dock actually does something, all the other tiles in the line are just tracks that don't visibly do anything other than letting the roombas drive on them

> throughput
well they would still need to wait (aka: recharge, which also avoids collisions*) before they can launch, so boosting their speed would be a nerf to them, but either will still decompress at junctions/sorters/etc causing cooldowns and such

Quezler:
**tldr** i would prefer if that mechanic stays, but in case you really want it gone, this is my recommendation:

- remove the item capacity boost entirely (since that would be replaced with movement speed)
- set the item capacity to 10 (to slightly make up for the change, and rounded is nice, the only reason it was 8 is since the overdrive projector increments in steps of 25%, so it had to be a multiple of 4 in order to come to a number without a float)
- ensure they cannot collide into eachother in where they slow down again

> tbh it kinda makes sense since the increased capacity doesn't even visibly show anymore

Anuken:
> all the other tiles in the line are just tracks that don't visibly do anything other than letting the roombas drive on them

Sure, but the roombas themselves are in range of the projector, which is what makes them faster.


> ensure they cannot collide into eachother in where they slow down again

Why? Since the conveyors at the end can no longer sustain the same throughput, collisions are bound to happen. I tested this out, and it looks fine.

Quezler:
> boosted roombas

very well then

> social distancing

well if it works it works

Anuken:
2dbb099d79099a62eb42e4329f580c44546aa847

Quezler:
also, did you change something major regarding `tools:pack`?
i've noticed a while ago that in the sprites the inner bits are at an angle while ingame they were square, and when i just now tried to remove the inner bit of the roomba it didn't appear to update:

Screen Shot 2020-04-30 at 20 46 48

@Anuken

Quezler:
disregard the above message, i figured out what caused it ^

BasedUser:
why not increase the frequency of conveyor loads?

Anuken:
Increasing speed already does that.

Issue #1952, opened 2020-05-01
It scares me


BlueWolfGaming:
**Platform**: *Android*

**Build**: *104.10*

**Issue**: *I went to edit the "plated-conduit" sprites in Novix and the file seems to be corrupted, another android user has had the same issue after testing. Picture provided below.*

**Steps to reproduce**: *Download any of the "plated-conduit" top sprites and run on Novix.*

**Link to mod(s) used, if applicable**: *None, strictly vanilla.*

---

*Place an X (no spaces) between the brackets to confirm that you have read the line below.*  
- [X] **I have searched the closed and open issues to make sure that this problem has not already been reported.**
![Screenshot_2020-04-30-22-44-10](https://user-images.githubusercontent.com/61173114/80783477-2d700680-8b37-11ea-80d0-7c4469edd144.png)

MEEPofFaith:
#1953 

This might be the fix.

DeltaNedas:
PNG image data, 32 x 32, 4-bit colormap, non-interlaced

Anuken:
This isn't a bug; the image renders fine in-game, and Novix is a completely different application.

PR #1953, opened 2020-05-01
Unjank the Plated Conduit sprites.

MEEPofFaith:
For some reason, the transparrent part was, like, [5 or so shades of colors](https://gyazo.com/dc59600fad032ec0450cbeb9ac40929e). Of course, you couldn't see them, but they werere there.


MEEPofFaith:
I don't know why the custom cells from earlier are still here.

BlueWolfGaming:
I'm afraid that the transparency shouldn't affect how the sprite is saved, though re-uploading should fix it. Possibly a viable solution to #1952.

MEEPofFaith:
Well, it's fewer colors to load.  And maybe the things I did in Paint.net might fix the weird corruptedness.

BlueWolfGaming:
It may be fewer, yet the transparent colours exist in all sprites with transparency. Check the other sprites.

DeltaNedas:
re encode as 8 bit rgba to fix jank

PR #1955, opened 2020-05-01
Add throughput module to sandbox blocks

Quezler:
> when playing around with #1954 i needed more-accurate-then-eyesight measurements.

so i started working on idea's for a measurement block sprite, and how it should function (sorters are instant / needs to support liquids / maybe like a junction / etc), but then it just dawned on me i could just add it to the `item` & `liquid` `sources` & `voids`:

basically those `4` blocks now measure their throughput over a period of `10 seconds`.
(during the **initial 10 seconds** the bar will slowly fill up, indicating it is measuring/loading)

> cursor is over the liquid void next to the cryofluid mixer (shown below)
Screen Shot 2020-05-01 at 11 18 52

during my limited amount of testing it does appear to flicker a bit.
(like between `4.2` and `4.3` for normal conveyors, but that [somewhat makes sense](https://github.com/Anuken/Mindustry/blob/master/core/src/mindustry/world/blocks/distribution/Conveyor.java#L53))

Screen Shot 2020-05-01 at 11 10 35

i am not sure how large of a performance impact this would have, nor if the measurement frame is of the right size, it takes a while to load but it seams nice to measure over a longer period of time, but its entirely up for debate.

> i'll probably post a video of this in action in `#pull-requests` later today 🥔

### DIFF
--- a/core/src/mindustry/entities/def/TileComp.java
+++ b/core/src/mindustry/entities/def/TileComp.java
@@ -55,6 +55,7 @@ abstract class TileComp implements Posc, Teamc, Healthc, Tilec, Timerc, QuadTree
     PowerModule power;
     ItemModule items;
     LiquidModule liquids;
+    ThroughputModule throughput;
     ConsumeModule cons;
 
     private transient float timeScale = 1f, timeScaleDuration;
@@ -1054,6 +1055,10 @@ abstract class TileComp implements Posc, Teamc, Healthc, Tilec, Timerc, QuadTree
 
         if(power != null){
             power.graph.update();
+        }
+
+        if(throughput != null){
+            throughput.update();
         }
 
         updateFlow = false;
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -598,6 +598,7 @@ bar.progress = Build Progress
 bar.spawned = Units: {0}/{1}
 bar.input = Input
 bar.output = Output
+bar.throughput = Throughput: {0}/s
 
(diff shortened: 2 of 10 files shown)


core/src/mindustry/entities/def/TileComp.java
```diff
@@ -55,6 +55,7 @@
     PowerModule power;
     ItemModule items;
     LiquidModule liquids;
+    ThroughputModule throughput;
```
Anuken, line 58: This is a very specific use case; it shouldn't be a module. Better to have it in each relevant void class instead.

Quezler: that was my first thought, but having 4 identical bar codes scattered around as duplicate seemed bad imo?

Anuken: The throughput class could still exist with some utility methods (like adding bars), but without extending `BlockModule` and only in the void entities.

Quezler: where would you like the throughput class to be? an interface like autotiler?

Anuken: Keep it the way it is, but remove the `BlockModule` superclass, add some utility methods and add fields to the void entities containing an instance of it.

Quezler: how would the utility methods know that the entities have those two variables?

Quezler: and how will the addbars on the block access the module inside the entity to add them?

PR #1957, opened 2020-05-01
add a sorter item filter timeout setting

DeltaNedas:
Adds slider, that when not zero, will make sorters forget their last filtered item after some time.

Prevents setting a large amount of sorters to an unwanted item when you forget about it.
Timeout can be between a second and a day.


Quezler:
might as well do it for the two sandbox sources as well while you're at it?

Anuken:
Sorry, no. This is a very specific feature and it'll just end up being more bloat that has to be maintained.

PR #1961, opened 2020-05-01
Unjank Plated Conduit Background Transparency

MEEPofFaith:
The background transparent part was for some reason 3-6 different colors. Invisible, but still there.

Also fixed my pull request to not have the previous stuff in it.


MEEPofFaith:
#1952

PR #1971, opened 2020-05-03
Add filter reset keybind [J]

DeltaNedas:
A simple way to reset filters on sorters, unloaders and item/liquid sources.
Adds Trigger.resetFilters event that is fired when J (no thought it's just unused) is pressed.
Mods can use this in custom filterable blocks if they want.

TODO:
- Mobile button?

Replaces #1957

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -757,6 +757,7 @@ keybind.chat_history_next.name = Chat History Next
 keybind.chat_scroll.name = Chat Scroll
 keybind.drop_unit.name = Drop Unit
 keybind.zoom_minimap.name = Zoom Minimap
+keybind.reset_filters.name = Reset Filters
 mode.help.title = Description of modes
 mode.survival.name = Survival
 mode.survival.description = The normal mode. Limited resources and automatic incoming waves.\n[gray]Requires enemy spawns in the map to play.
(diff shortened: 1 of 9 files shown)


Quezler:
or maybe just add it to `Q` since that clears the current build plan?

Anuken:
Why not just double click?

Quezler:
resetting sorter (and alike) blocks with a key seems a little more practical than placing one down and interacting with that in order to clear it

Anuken:
I don't think pressing yet another random key is a good way of changing the next filter, either. There's also quite a lot of duplicated code involved here.

Quezler:
> yet another random key

hence why i'm counter-suggesting just adding it onto **Q**

Quezler:
if something like this is gonna be implemented at all, i'm requesting bridges to be added to it as well, placing down bridges is a bit of a pain if one of the axis somehow aligns with the last bridge placed, down here is a poor example but it illustrates the point, pressing _Q_ or something with my pinky while doing so would be lovely:

May-03-2020 17-12-50

DeltaNedas:
as i said, i put 0 thought into what key is default as long as it wasn't taken already

DeltaNedas:
hmm would it be possible to have an action unbound by default?

DeltaNedas:
if i add bridge reset, should the feature be renamed to "block reset" or what?

PR #1973, opened 2020-05-03
Ctrl Click Bridge Reverse Connect

zethnest:
Peek 2020-05-04 00-42

In the GIF above, I `CTRL + Click` the source of the two bridge link and the bridge link will be reversed where the source of the two bridge will become the destination.


ActualWina:
looks nice

Anuken:
Not a fan of hidden controls like this. An alternative:

recording1589132892

PR #1984, opened 2020-05-04
Bullets don't target what they can't hit.

MEEPofFaith:
Honestly, I don't fully know what I'm doing, but if it works it works.


MEEPofFaith:
Seems like someone did this already. Closing pull.

PR #1993, opened 2020-05-06
[QOL] Show Overdrive Range when Building Overdrive

zethnest:
Screenshot_20200506_124151
Currently you need to either guess or memorize the size of the overdrive range when you want to place down an overdrive with minimal overlap. This pull will show all of the overdrive range when OverdriveProjector is selected.

Update: Maybe won't be updated for a month or two. Feel free to do a PR on this

TODO:
- [ ] Show overdrive range around the mouse only
- [ ] Toggleable shaders to show area coverage like shield projectors


Anuken:
Making a separate block flag and range display *just* for overdrive projectors (and no other blocks) seems too hacky; I would rather implement a generalized solution for all range-based blocks later.

PR #2003, opened 2020-05-07
Custom Drone Building Color

MEEPofFaith:
I'm making it my goal to find small details like this and make them modding affectable.


Anuken:
The building shader will be removed/replaced later, so this won't be necessary.

MEEPofFaith:
error.png

PR #2006, opened 2020-05-07
Customizable Fragging Angle

MEEPofFaith:
If you wanted something that doesn't frag 360, you can make something that doesn't frag 360.


Anuken:
Please test your changes before proposing them. Not only will the cone here always be offset to one side, but the rotation and velocity components will all have different rotations, leading to a mess.

MEEPofFaith:
I'll go figure out how to compile java later.

MEEPofFaith:
took a few tries

PR #2014, opened 2020-05-09
small fix

KeviTV:


Anuken:
What does this fix?

KeviTV:
The incinerator will not work after this change if there is no charge on the battery

Anuken:
That's a problem with the incinerator itself.

PR #2019, opened 2020-05-10
Shift Click to Same Tile Configure

zethnest:
Peek 2020-05-10 18-08
In the GIF above, I am Shift-Clicking the power nodes to configure the power nodes itself. It's the same as clicking the power node twice or double-click. This too works with all other double-click configurable blocks such as bridges and sorters.

### DIFF
--- a/core/src/mindustry/input/InputHandler.java
+++ b/core/src/mindustry/input/InputHandler.java
@@ -610,6 +610,11 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
         }
         boolean consumed = false, showedInventory = false;
 
+        if(tile.block().configurable && tile.interactable(player.team())){
+            consumed = true;
+            tile.onConfigureTileTapped(tile);
+        }
+
         //check if tapped block is configurable
         if(tile.block().configurable && tile.interactable(player.team())){
             consumed = true;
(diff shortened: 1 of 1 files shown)


Anuken:
What if you happen to be boosting, and you want to configure a tile? Is this really significantly more convenient than just double clicking?

zethnest:
Sometimes you want to just reset the connections on power nodes. They're also sometimes close to each other and you need to click on an empty tile first before clicking on the power node beside it.

zethnest:
I just thought using the Dash binding would be good since it's actually a really low probability for you to be dashing whilst configuring something.

Anuken:
> Sometimes you want to just reset the connections on power nodes.

So why not just double click? It's nearly as fast.

> it's actually a really low probability for you to be dashing whilst configuring something.

I don't think so.

zethnest:
> So why not just double click? It's nearly as fast.

It is. Unless the power nodes are close to each other. You need to click away before double clicking the next.

> I don't think so.

I see, What do you think if this doesn't trigger if the player is moving?

Anuken:
> You need to click away before double clicking the next.

That specific problem can be solved by making power nodes deselect themselves when double clicked.

recording1589133423

PR #2024, opened 2020-05-10
Small russian translation fix

Summet-V:


Anuken:
Is this really correct?

"башенная защита" is more like "defense *with* towers", and "защита башен" is "defense *of* towers".

Summet-V:
ok, leave it as it is

PR #2034, opened 2020-05-12
Slider Togglable Unit Opacity

zethnest:
Peek 2020-05-12 18-07
It is as the title and GIF shows. This solves the problem where there are flying units nearby of where you want to build and you can't see what's underneath the unit.

The default key to toggle unit opacity is F7, right beside power line opacity and bridge opacity which are F5 and F6.

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -690,6 +690,7 @@ setting.playerlimit.name = Player Limit
 setting.chatopacity.name = Chat Opacity
 setting.lasersopacity.name = Power Laser Opacity
 setting.bridgeopacity.name = Bridge Opacity
+setting.unitopacity.name = Unit Opacity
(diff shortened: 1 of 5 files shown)


zethnest:
Implements https://github.com/Anuken/Mindustry-Suggestions/issues/108

Anuken:
Sorry, no.

- This significantly complicates unit rendering. Every single time I add a new function or custom drawing for a unit, I'll have to test make sure it renders correctly. Mods are guaranteed to break this if they use `Draw.color()` somewhere, and maintenance will be a general mess.
- It doesn't look very good.
- You should be able to move the units somewhere else instead of making them ghosts.

zethnest:
> You should be able to move the units somewhere else instead of making them ghosts.
This works for land units but not flying units

Anuken:
It *would* work for flying units, if you could control them. That's not possible right now, but it will be.

PR #2041, opened 2020-05-12
6.0/multiForge 

SnakkiZXZ:
how about getting a few factories in one. I propose to combine a large silicon smelter with large meltglass forge. 

miniGif_20200512205627
of course my code is not perfect and requires refinement, but I just wanted to show how it should work

P.S. I don't understand how to create requests correctly, so it happened.


Anuken:
No.

PR #2043, opened 2020-05-12
Update server's

MINDUSTRYRU:
Add
mindustry.ru (Survival)
mindustry.ru:9990 (Survival 2)
mindustry.ru:9991 (PvP)
mindustry.ru:9992 (Attack)
mindustry.ru:9993 (HardEarth)

### DIFF
--- a/servers.json
+++ b/servers.json
@@ -24,7 +24,16 @@
     "address": "mindustry.ru"
   },
   {
-    "address": "mindustry.ru:7000"
+    "address": "mindustry.ru:9990"
+  },
+  {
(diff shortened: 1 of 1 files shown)


DeltaNedas:
why not leave it as is with the hub and not 5 servers

MINDUSTRYRU:
To maintain the lobby, system resources are required, as well as if the lobby of the server is turned off, access to other servers is limited

KeviTV:
I still don't see the point in adding all 5 servers ^

Kieaer:
Use a Lobby server

PR #2056, opened 2020-05-14
 Energy-dependent incinerator speed limit

KeviTV:
GIF 14 05 2020 20-33-49


Summet-V:
Translation: Energy-dependence incinerator speed limit

Anuken:
Due to the use of random numbers, this solution ends up being both FPS-dependent and desynced on clients. I think it's better if it's kept to being only functional at 90% or higher.

PR #2057, opened 2020-05-14
Add a mindustry server to global server list

bot196:
Please add this server to the Mindustry server list:

IP: diskfloppy.ddns.net


Quezler:
i would suggest making a new pull request that does not reuse existing branches ^

Anuken:
I wonder how many times people have tried to merge byte-logic now.

PR #2065, opened 2020-05-15
Clear Last Config Key

zethnest:
Peek 2020-05-16 01-51

Adds feature where when you want to place a block that have previous configuration, such as bridges and sorters, you can press the `Binding.clear_building` key to clear the configuration. If there are no configuration then pressing `Binding.clear_building` will clear building queue.


DeltaNedas:
1. CONTRIBUTING.md, code style
2. mods cant extend it and its messy to add more blocks
3. you havent added support for item/liquid sources
I suggest using an event that takes a block as argument, fire it if block isnt null.
Add handlers in each configurable block.

zethnest:
@DeltaNedas 
1. Fixed
2. What do you mean? Please enlighten me
3. I have already shown that ItemSource is supported. LiquidSource is supported as well, I forgot to show it in the gif.

> event that takes a block as argument
What do you mean by this? It doesn't have to be an event.

DeltaNedas:
2: having to do lots of checks in one messy place, that mods cant add to
For events, doing it in an extendible manner needs a callback so events are the easiest option.

zethnest:
Can you point me to a few mods where this will be a problem?

zethnest:
Oh wait nevermind. There aren't any mods because it's 6.0.

Anyways, I don't there will be a problem with mods. There's only so much you can add to logistic blocks. Sorter, Unloader, and Item/LiquidSource have the same API on clearing the config. ItemBridges are an exception as they don't have `saveConfig` flag as of 603cb4295. Other configurable blocks such as the UnitFactory and MassDriver don't save config.

Am I missing something else?

zethnest:
Ew. Ugly commits. opening a new one.

PR #2069, opened 2020-05-16
Schematic Drill Indicator

zethnest:
This implements https://github.com/Anuken/Mindustry-Suggestions/issues/165

screenshot200517011457749952295
screenshot200517011503916301425

Two pictures above shows the mined ore + the mining speed

screenshot200517004303805256764

Picture above only shows the mined ore.

### DIFF
--- a/core/src/mindustry/world/blocks/production/Drill.java
+++ b/core/src/mindustry/world/blocks/production/Drill.java
@@ -76,7 +76,7 @@ public class Drill extends Block{
         if(returnItem == null) return;
 
         Draw.color(returnItem.color);
-        Draw.rect("drill-top", req.drawx(), req.drawy(), 1f);
+        Draw.rect(topRegion, req.drawx(), req.drawy(), 1f);
         Draw.color();
     }
 
--- a/core/src/mindustry/world/blocks/production/Drill.java
+++ b/core/src/mindustry/world/blocks/production/Drill.java
@@ -73,24 +73,19 @@ public class Drill extends Block{
         if(tile == null) return;
         countOre(req.tile());
         if(returnItem == null) return;
-        /*
+        /* This commented code shows only the icon
         float dx = req.drawx() - size * tilesize/2f, dy = req.drawy() + size * tilesize/2f;
         Draw.mixcol(Color.darkGray, 1f);
         Draw.rect(returnItem.icon(Cicon.small), dx, dy - 1);
         Draw.reset();
         Draw.rect(returnItem.icon(Cicon.small), dx, dy);*/
+        /* This below code shows the icon + the drill speed */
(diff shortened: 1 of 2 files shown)


zethnest:
Current implementation as of e454199d9 looks like this

Anuken:
The item colors are useful, but I don't think numbers should be displayed. They create a lot of visual noise, especially when drills are placed in clusters.

core/src/mindustry/world/blocks/production/Drill.java
```diff
@@ -73,57 +71,13 @@ public Drill(String name){
...
-        font.setUseIntegerPositions(ints);
-        font.setColor(Color.white);
-        font.getData().setScale(1f);
-        Draw.reset();
-        Pools.free(layout);
-        /* end of code section */
-
-        /* This commented code colors their ore region */
         Draw.color(returnItem.color);
         Draw.rect("drill-top", req.drawx(), req.drawy(), 1f);
```
Anuken, line 79:
```suggestion
        Draw.rect(topRegion, req.drawx(), req.drawy(), 1f);
```

core/src/mindustry/world/blocks/production/Drill.java
```diff
@@ -66,6 +67,19 @@ public Drill(String name){
...
+    @Override
+    public void drawRequestConfigTop(BuildRequest req, Eachable<BuildRequest> list){
+        Tile tile = req.tile();
+        if(tile == null) return;
+
+        countOre(req.tile());
+        if(returnItem == null) return;
+
+        Draw.color(returnItem.color);
+        Draw.rect(topRegion, req.drawx(), req.drawy(), 1f);
```
Anuken, line 79: Aren't you drawing it with 1 degree of rotation here?

core/src/mindustry/world/blocks/production/Drill.java
```diff
@@ -66,6 +67,19 @@ public Drill(String name){
         idleSoundVolume = 0.003f;
     }
 
+    @Override
+    public void drawRequestConfigTop(BuildRequest req, Eachable<BuildRequest> list){
+        Tile tile = req.tile();
+        if(tile == null) return;
```
Anuken, line 73: Technical fix: You should also skip drawing if `!req.worldContext` (so schematics don't grab world state).

core/src/mindustry/world/blocks/production/Drill.java
zethnest: I didn't check before. but `topRegion` makes them look like they have a mole.
screenshot200517121852617205364

core/src/mindustry/world/blocks/production/Drill.java
zethnest: I didn't notice before as I copied it from `draw()`. Removing `1f` doesn't seem to change anything so I'll remove that and from `draw()`

PR #2075, opened 2020-05-17
Added server Omega Sandbox

RebornTrack970:
What do i need to do to make my server added? Or get known? Because i dont know nearly any one who ones the servers on community servers and i dont think %95 of people does knkw eather so can you add community.mindustry.io:9004 to community servers?

### DIFF
--- a/servers.json
+++ b/servers.json
@@ -55,5 +55,8 @@
   },
   {
     "address": "twsmindustry.24x7.hk"
+  },
+  {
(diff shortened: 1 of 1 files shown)


RebornTrack970:
If not tell me why and how i can add it.

itcannotbe:
You don't
That address is well know to be the container server for the io group of servers that server people who request to use that particular server. Also being said that you have no actual right over it.

PR #2078, opened 2020-05-18
fix pick_block on schem error

zethnest:


Anuken:
Fixed in a slightly different way.

PR #2079, opened 2020-05-18
[QOL] Search Ignore Symbols

zethnest:
Peek
**Opening schematic menu** using keybind will **automatically focus** the search field
**Pressing enter** when focused on the search field will **select the first match**.
You can now **use a simple fuzzy search** for searching schematics.
The fuzzy search is **togglable** and is **Default** to **false**.

EDIT: Oops, forgot to add description
Sorry for laggy gif

Alternative solution to https://github.com/Anuken/Mindustry-Suggestions/issues/94

TODO:
- [X] Use different ~~fuzzy~~ algorithm


zethnest:
I feel like I need to open another PR for this. And I should drop the fuzzy thing.

PR #2085, opened 2020-05-19
liquid-generator

SnakkiZXZ:
How about adding a hydroelectric power plant to the game? I propose to add a special generator, which will produce a small amount of energy if water is brought to it, and generates a little more energy if put on water. 
I believe that this new type of energy generation will help in getting it on the maps with a lot of water and a small amount of resources

code is not going, and requires some improvements
sprite apply here, but i think it does not fit
liquidGenerator

> I use JavaScript code, don't hit sneakers

### DIFF
--- a/liquid-generator.js
+++ b/liquid-generator.js
@@ -1,0 +1,21 @@
+var xx = "water"
+var yy = "sand-water"
+var zz = "darksand-water"
+var cc = "darksand-tainted-water"
+const liquidGenerator = extendContent(SolarGenerator,"liquidGenerator",{
+getPowerProduction(tile){
+    if(tile.floor() == xx || tile.floor() == yy || tile.floor() == zz || tile.floor() == cc){
+        return 1.4
+    } else {
+        if(tile.entity.liquids.get(Liquids.water) > 0.1){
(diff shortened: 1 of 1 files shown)


zethnest:
This is a mod

liquid-generator.js
```diff
@@ -0,0 +1,21 @@
+var xx = "water"
+var yy = "sand-water"
+var zz = "darksand-water"
+var cc = "darksand-tainted-water"
+const liquidGenerator = extendContent(SolarGenerator,"liquidGenerator",{
+getPowerProduction(tile){
+    if(tile.floor() == xx || tile.floor() == yy || tile.floor() == zz || tile.floor() == cc){
+        return 1.4
```
Quezler, line 8: instead of using the alphabet you should probably use liquid drop / affinities ^ (though way more is wrong here*)

SnakkiZXZ:
this code was written only to show how the principle works, I'm not sure that it will be used

itcannotbe:
If that was the case I don't thing a in this repo would the best thing to do.

Suggestions Repo: https://github.com/Anuken/Mindustry-Suggestions/issues

PR #2087, opened 2020-05-19
Add clear() function.

Summet-V:
Add a clear() function that clears the console log.


Anuken:
Why not just invoke `Vars.ui.scriptfrag.clearMessages()` manually when you need it? This doesn't seem like a very commonly used utility, and `clear()` by itself is not a very descriptive method name.

Quezler:
isn't that a bit to ambiguous?

Summet-V:
This would be a shorthand for cleaning the log, many people do not know how to do this.

PR #2089, opened 2020-05-20
[QOL] Search Schematic

zethnest:
Seperating this QOL from https://github.com/Anuken/Mindustry/pull/2079.

**Opening schematic menu** using keybind will **automatically focus** the search field
**Pressing enter** when focused on the search field will **select the first match**.

### DIFF
--- a/core/src/mindustry/ui/dialogs/SchematicsDialog.java
+++ b/core/src/mindustry/ui/dialogs/SchematicsDialog.java
@@ -58,6 +58,14 @@ public class SchematicsDialog extends FloatingDialog{
         cont.pane(t -> {
             t.top();
             t.margin(20f);
+
+            t.update(() -> {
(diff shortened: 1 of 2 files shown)


core/src/mindustry/ui/dialogs/SchematicsDialog.java
```diff
@@ -46,17 +47,25 @@ void setup(){
...
         }).fillX().padBottom(4);
 
         cont.row();
 
         cont.pane(t -> {
             t.top();
             t.margin(20f);
+
+            t.update(() -> {
+                if(Core.input.keyTap(Binding.chat) && Core.scene.getKeyboardFocus() == searchField && firstSchematic != null){
```
Anuken, line 63: Where is `firstSchematic` defined? This doesn't look like it will compile.

zethnest: Whoops. Sorry. Forgot to check.

PR #2093, opened 2020-05-20
[QOL] Center Map to Player Implementation

zethnest:
Opening the map will center the map to you. So you have a better time navigating yourself instead of searching for yourself in a cluster of blocks with the same colour or searching for yourself in a very huge map.

TODO:
- [X] Make it not ignore zoom


zethnest:
Currently independent of zoom

PR #2097, opened 2020-05-22
[QOL] Quick Copy Implementation

zethnest:
Peek 2020-05-22 14-00
You can copy **and** break blocks at the same time. This works when you want to break blocks and press the schematic select, or copy schematic key at the same time.

TODO:
- [ ] Support for mobile.
- [ ] Different color(?) for the lines
- [ ] Update GIF (too laggy)


zethnest:
Maybe this one should be called `mode = cutting`

PR #2100, opened 2020-05-23
Ability To Disable Auto Connect of Power Nodes.

Summet-V:
Now you can disable auto node connection in settings.

### DIFF
--- a/core/src/mindustry/world/blocks/power/PowerNode.java
+++ b/core/src/mindustry/world/blocks/power/PowerNode.java
@@ -120,13 +120,15 @@ public class PowerNode extends PowerBlock{
         Draw.color(Pal.placing);
         Drawf.circles(x * tilesize + offset(), y * tilesize + offset(), laserRange * tilesize);
 
-        getPotentialLinks(tile, other -> {
-            Drawf.square(other.x(), other.y(), other.block().size * tilesize / 2f + 2f, Pal.place);
+        if(Core.settings.getBool("autonodeconnection")) {
+            getPotentialLinks(tile, other -> {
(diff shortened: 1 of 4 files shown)


core/src/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -266,32 +268,34 @@ public static void insulators(int x, int y, int x2, int y2, Cons<Tilec> iterator
...
-                (other.block().outputsPower && !other.block().consumesPower) || other.block() instanceof PowerNode) && linkValid(this, other)
-                && !other.proximity().contains(this) && other.power().graph != power.graph;
-
-            tempTileEnts.clear();
-            Geometry.circle(tile.x, tile.y, (int)(laserRange + 2), (x, y) -> {
-                Tilec other = world.ent(x, y);
-                if(valid.get(other)){
-                    if(!insulated(this, other)){
-                        tempTileEnts.add(other);
+            if(Core.settings.getBool("autonodeconnection")) {
```
Anuken, line 271: This won't work on servers. Automatic node connection is done server-side to prevent spam of packets from the client.

Summet-V: So i should edit server-side?

Anuken: You *did* edit server-side; what I'm saying is, when a client disables node connections on their end, it won't change how the server behaves.

Summet-V: Then how to implement it?

ifritdiezel:
oh yes let it happen
at least in singleplayer

DeltaNedas:
Maybe it's fine as a server setting?

PR #2102, opened 2020-05-24
Call that

MEEPofFaith:
I tried using the damage all in a script and it gave me an error, it was trying to use the team one, so if the call is changed it can then be called properly.


Anuken:
Are you sure your parameter list was correct?

Anuken:
image

Yeah, I don't see what the problem is. Make sure your script is correct.

PR #2103, opened 2020-05-24
[Mod] Deep Water Placeable Implementation

zethnest:
Peek 2020-05-24 18-49

Implements https://github.com/Anuken/Mindustry-Suggestions/issues/196

GIF above shows a mod that contains a block that uses the variable `requiresDeepWater` and is only placeable on deep water


MEEPofFaith:
What about being able to place in normal and deep water?
Land and water?

zethnest:
Denied

MEEPofFaith:
oh no

garyType1:
If the block could be placed on deep water though, they should have a "sink" look

zethnest:
@garyType1 That would mean the blocks are underwater

Anuken:
I don't see much use for requirements of *deep* water specifically. This can be implemented by making `requiresWater` allow deep water placement instead.

Issue #2107, opened 2020-05-25
Setting liquidBoostIntensity to 1 doesn't stop showing "Optional Enhancement" for drill


asl97:
**Platform**: tested on *Android+Linux*

**Build**: *104.10* & git tag 104.6

**Issue**: I am trying to hide the `"Optional Enhancement"` in the stat page by setting LiquidBoostIntensity to 0 but that stop the drill from working.

Setting it to 1 doesn't hide it ether.

A way to fix this is require the boost to be above 1 before showing the stat here  (unless negative boost is a thing):
https://github.com/Anuken/Mindustry/blob/0e59a22cbecf544205c5f9918e214086800c7f15/core/src/mindustry/world/blocks/production/Drill.java#L161-L163

Another way is to ensure the speed isn't 0 here:
https://github.com/Anuken/Mindustry/blob/0e59a22cbecf544205c5f9918e214086800c7f15/core/src/mindustry/world/blocks/production/Drill.java#L265-L267

**Steps to reproduce**:  example drill that require water to run:

`"./content/blocks/drills/mechanical-drill.json"`:

```
consumes:{
  liquid:{
    liquid:water
    amount: 0.05
    optional:false
    booster:false
  }
}
liquidBoostIntensity:0
```

*Place an X (no spaces) between the brackets to confirm that you have read the line below.*  
- [x] **I have searched the closed and open issues to make sure that this problem has not already been reported.**

itcannotbe:
I wonder why...…
What's the product of some number and zero again?

asl97:
@itcannotbe but there isn't any item boosting so there shouldn't be a multiplication boost of 0.

a proper fix to be to do something in the stats.add to check if there is both a boost and boost item.

itcannotbe:
I think that not the case. The only reason why setting it to zero and the drill not working in your test is

https://github.com/Anuken/Mindustry/blob/7dff023928413b503be1e19501db0f30253bc7be/core/src/mindustry/world/blocks/production/Drill.java#L260

Is returned as true as you didn't have anything that's it can consume set to optional

asl97:
@itcannotbe I get that part, I could set it as 1 and it would work as expected.

With this part modify to greater than 1.
https://github.com/Anuken/Mindustry/blob/0e59a22cbecf544205c5f9918e214086800c7f15/core/src/mindustry/world/blocks/production/Drill.java#L161-L163

[...]

PR #2108, opened 2020-05-25
Hide "Boost Effect" when liquidBoostIntensity == 1

asl97:
https://github.com/Anuken/Mindustry/issues/2107


asl97:
I ain't quite sure if it should be hidden if it's below 1,

Somebody could possibly implement some kind of gunk ore where if there too much in the miner, the miner would eat up the item and slow down, that is if it is currently supported.

As for negative amount, I not sure how that would work.

PR #2112, opened 2020-05-26
6x6 CoreBlock, "Core: Nexus"

z0mbiesrock:
An even bigger core block for even larger still bases

Name and description need to be translated to the other bundles, though.


Kieaer:
I'm not sure why this is necessary.
image

Anuken:
Sorry, I don't really see the need for a new core block.

PR #2114, opened 2020-05-26
Shots should spawn with recoil

MEEPofFaith:
Btw, why is the second tr.trns have y as the second instead of the third value?


Anuken:
Why should they spawn with recoil? The recoil is happening *after* the bullet exists the turret.

MEEPofFaith:
hm... oki

PR #2119, opened 2020-05-27
Allow modders to make longer range Mass Drivers

MEEPofFaith:
Increased lifetime and made the bullet just oof if it passes over an oofed target.

Compiling is down, so no prior testing has been made.

### DIFF
--- a/core/src/mindustry/entities/bullet/MassDriverBolt.java
+++ b/core/src/mindustry/entities/bullet/MassDriverBolt.java
@@ -51,7 +51,7 @@ public class MassDriverBolt extends BulletType{
         float dst1 = b.dst(data.from);
         float dst2 = b.dst(data.to);
         
-        //if the target is dead, just keep flying until the bullet explodes
+        //if the target is dead, oof upon passing the destination point
         if(data.to.dead()){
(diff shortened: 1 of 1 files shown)


Anuken:
Even with longer range, mass driver bullets still slow down in transit.

Syncinus:
I'm pretty sure the ``b.remove`` method calls despawned anyway, so you don't have to called despawned and then remove.

core/src/mindustry/entities/bullet/MassDriverBolt.java
```diff
@@ -47,14 +47,17 @@ public void update(Bulletc b){
...
-        //if the target is dead, just keep flying until the bullet explodes
-        if(data.to.dead()){
-            return;
-        }
-
         float baseDst = data.from.dst(data.to);
         float dst1 = b.dst(data.from);
         float dst2 = b.dst(data.to);
+        
+        //if the target is dead, oof upon passing the destination point
```
Anuken, line 54: Why should the bullet just disappear when it flies past a mass driver that isn't there anymore? There's nothing to stop it.

MEEPofFaith:
Ima fix my PR

PR #2132, opened 2020-06-02
Health color change

Slava0135:
I've been always wondering why health is red while all healing effects are lime... This will make more sense I believe.


zethnest:
I beg to differ. Red as health is stapled to if not all, most of the games. Green/Lime as healing effects is also widely used in most games. 

In effects' kind of language. Green means something good. Red means something bad.
In UI terms, Red means something important. Green means something has passed.

Warning signs and interfaces has always been red. Red is a very strong color, which means it needs to be noticed.

Green in UIs should be used for something like efficiency.

Anuken:
To add to the comment above, the lime color used for healing simply looks awful when used as a bar; this isn't a good change.

PR #2133, opened 2020-06-02
Add Atanner Gaming Survival Server

AydenRennaker:
Seems like there is a lot of interest in the survival game mode so we want to add a new survival exclusive server.

### DIFF
--- a/servers.json
+++ b/servers.json
@@ -48,6 +48,9 @@
     "address": "mindustry.atannergaming.com"
   },
   {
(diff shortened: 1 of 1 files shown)


Anuken:
image

Seems to be on attack mode, not survival.

AydenRennaker:
Right shuffle mode was not on built-in. Its been corrected now.

AydenRennaker:
The server has been updated with our new moderation tools.

PR #2138, opened 2020-06-03
[Fix] Mark Block.solid False By Default.

Summet-V:
Why are other fields marked by default but `Block.solid` not?


Anuken:
Probably some leftover legacy code; no initializer is needed here at all. I'll remove it.

PR #2143, opened 2020-06-06
Updated Segment and Lich sprite

EyeOfDarkness:
changed Segment's decal,
changed Lich entirely.


Syncinus:
The Lich sprite looks a lot better, and still seems to fix within the Mindustry style.

Syncinus:
The Lich sprite looks a lot better, and still seems to fix within the Mindustry style.

PR #2146, opened 2020-06-06
[Server] Alias Command.

Summet-V:


DeltaNedas:
Then a script that registers aliases for a through zzzzzzzzzz comes along and aliases can't be used anymore.

Anuken:
This doesn't save. What is the use case?

Summet-V:
> This doesn't save. 
Maybe save it in config dir?

PR #2147, opened 2020-06-06
Improve zoom controls at high and low zooms

blahblahbloopster:
At zoomed-in scales it wasn't very responsive and at zoomed-out scales it was
too responsive


Anuken:
Dividing by `5` doesn't feel responsive enough; I'll merge this manually.

PR #2149, opened 2020-06-07
Reduce voteCooldown

Kieaer:
### Reason
If the vote fails to expel the griefer, during the cool time, the griefers can destroy everything.
It's really enough time.

The best solution is a remove voteCooldown or 5~10 second cooldown.


DeltaNedas:
or, remove global cooldown and have a cooldown only if a player tries to repeatedly votekick

Anuken:
Until the single-vote limit is worked around, this will only cause problems. Note than in 6.0, the votekick cooldown is *per-player*, so even if the vote fails, someone else can try starting a vote.

Kieaer:
Does anybody know that :eyes:
Need command description update

Anuken:
I want to try refactoring it into a UI box instead.

PR #2150, opened 2020-06-07
#Refactoring & Design Pattern

huiung:
#Refactoring
1.Used refactoring operations
 Extract Method

2.Classes which have been changed
  Effects.java, WeaphonsComp.java

3.Why were they refactored
-Effects.java has duplicated codes -> extract method
-WeaphonsComp.java's update method is so complex -> extract method each of role

#Desing Pattern
1. Used Design Pattern
    Singleton

2. Classes which have been changed
   MusicControl.java, Control.java

3. Why used this Design Pattern
   I think that MusciControl class need only one for running
   but, it can be created more than one because it isn't singleton.
   applying singleton is effective regarding to memory and this hazards


Anuken:
This PR is invalid, you've pulled in many unrelated commits. I can't see the changes you mention in the massive changelog; open a new PR.

PR #2166, opened 2020-06-15
Make coreblocks have their team represented on sprite

ThePythonGuy3:
I made this on my custom build, and it turned out very well. As I can't upload files, I can't upload the sprites required, but look in my build's sprites: [block sprites](https://github.com/ThePythonGuy3/Mindblowstry/tree/master/core/assets-raw/sprites/blocks/storage)

If you don't like mine, you'll need: name (sharded team block), name-top (The part that changes depending on team) and name-base (The base of the core).

Here's how it turns out:
image

PR #2177, opened 2020-06-19
Make coreblocks have their team represented on sprite 2: Electric Boogaloo

Quezler:
This pull is just a complete & cleaned up version of #2166 <-- credits obviously go there ❤️ 

(doesn't contain repacked sprites since their diff looks a bit to weird compared to previous)

todo:
- render it correctly in the editor
- somehow avoid having sharded hardcoded
- etc

Screen Shot 2020-06-19 at 20 14 09


core/src/mindustry/world/blocks/storage/CoreBlock.java
```diff
@@ -164,6 +168,18 @@ public void onProximityUpdate(){
...
+            if(team == Team.sharded){
+                Draw.rect(block.region, x, y, block.rotate ? rotdeg() : 0);
+            }else{
+                Draw.rect(baseRegion, x, y, block.rotate ? rotdeg() : 0);
+                Draw.color(team.color);
+                Draw.rect(topRegion, x, y, block.rotate ? rotdeg() : 0);
+                Draw.color();
+            }
+        }
+
```
Quezler, line 182: not sure if you want to have `, block.rotate ? rotdeg() : 0` there or not ^ (modding support?)

core/src/mindustry/world/blocks/storage/CoreBlock.java
```diff
@@ -181,6 +197,10 @@ public void drawSelect(){
             Draw.reset();
         }
 
+        @Override
+        public void drawTeam(){
+            //
```
Quezler, line 202:
```suggestion
            //no
```

ThePythonGuy3:
Thanks for making my PR better!

Quezler:
You're welcome @ThePythonGuy3 ❤️

Quezler:
for @Anuken's reference, this is how the sharded core would currently look if not excluded:

Screen Shot 2020-06-19 at 20 38 20

Quezler:
What i'm noticing however is that just like some other pulls i worked on (like #1884 for example) there tend to be white/transparent/alignment/etc issues where the texture overlays meet, any idea how to best remedy those whenever they occur?

ThePythonGuy3:
> for @Anuken's reference, this is how the sharded core would currently look if not excluded:
> 
> Screen Shot 2020-06-19 at 20 39 52
> 
> 

That's why I added a shared team sprite also

ThePythonGuy3:
> What i'm noticing however is that just like some other pulls i worked on (like #1884 for example) there tend to be white/transparent/alignment/etc issues where the texture overlays meet, any idea how to best remedy those whenever they occur?

I ran inte the exact same problem, but as my PR didnt compile in the newer versions of v6, I did not add the solution to mine. But the solution is simple: you need to make the base be "fluent" with the colors. Like this:
 image

Anuken:
> you need to make the base be "fluent" with the colors.

The reason colors "bleed" through is due to the antialiasing/filtering of the block underneath; the base edges under the "top" region blend into the nearby colors and make them lighter. That's why coloring the base sprite properly fixes it.

---

A true solution to the problem with shading could be to automatically generate sprites at pack-time with the correct palette and shading, then have the cores use a generated sprite based on team. Of course, only the important teams (crux, sharded, maybe derelict) would get sprites - other teams would use the fallback tinted sprite. 

This has several advantages:

- You'll only need the top sprite and the bottom sprite, no other copies or manual resprites need to be made
- Any block can be made to use a team color pattern - all you need is the top sprite
- Once you define a team palette, it can be used for every tinted team block - no need to remember the exact colors

Quezler:
well i suppose rewriting the pack engine to do the above falls above my scope so I'll archive this 😗

Anuken:
Final result:

image

Quezler:
you've kept the team stripes in the bottom left corner? (graphics look nice btw) ❤️

Anuken:
I *could* remove them, but they aren't much of a problem right now.

Issue #805, opened 2019-10-09
Better Team Indication


Anuken:
Brainstorm ways to make the team indicator on blocks/units look better **without** incurring expensive performance costs on color shaders or outlines.

231ACR:
use thickened markers with black borders.

ThatOneBepis:
For arcs, change default color of blue bits to gray, which could be used for Gray team. In other teams, change the color of said bits to the color of the team (if you understand what i tried to say)

Tazshelby2016:
It seems that the art for the blocks/units needs to be redesigned. As @ThatOneBepis said, this can be done by recoloring the sprites to the team's color, or by using an overlay sprite colored in the team colors.

231ACR:
How this is can be more effective?
changing colors only lessens it memory wise, which will not lessen any burdens on GPU's processor.

Tazshelby2016:
@231ACR  It will add nothing to the GPU, as it is simply making the colors different.

LionelBergen:
What is currently done to indicate team?
We could check how expensive shaders/outliners are, if we ensure they are done once per sprite.
I agree with above comments about simply changing the colors to the teams color.

ThatOneBepis:
Current indication on the team is quite unoriginal considering every building has it (stripe of color in the bottom right)

Arkanic:
either a simple kind of "hovering" icon over the block/area in question, or have the "default" (background) block colour change to that specific team
e.g. the core's background "grey" colour changes to red

ThatOneBepis:
Concerning custom blocks from modding, how would they behave? background color, while it may seem something straight outta b63, is something what may solve this issue

Luxray5474:
Personally really loved the nice ***t h i c c*** outlines back when the game had pixelated graphics. I wish those would come back, but I recognize the performance implications.

* Are shadows a class that is applied to blocks or chunks of blocks individually? If they are then tint the shadow of a team member with their color. 
* The good old method, which is editing the sprites to match the color of the player's team, is probably the easiest. That includes their bullets too! However that just doubles the amount of player sprites in the spritesheet.
* Place a circle around the player with their team color. 
* For blocks I agree with Arkanic. However it would look really ugly with the current art style. Any direct modification to the blocks' sprites might result in an almost doubling of the size of the spritesheet. I think the best way would to make a realtime outline, the same way that blocks have an outline when you are choosing what to drop your items into.

Arkanic:
[...]

PR #2192, opened 2020-06-26
Implementation of team colored force fields

Remint32:
# Implementation of color force fields

The current realization exploits existing crutches with the rendering of colored shield frames in the Shader.

There is also another option, with rendering only in the block code and post-processing with the Shader, without changing the color of the command directly in the Shader.

изображение

## Suggestions
https://github.com/Anuken/Mindustry-Suggestions/issues/228
https://github.com/Anuken/Mindustry-Suggestions/issues/227
https://github.com/Anuken/Mindustry/issues/805

## Credits
@Summet-V @Remint32

### DIFF
--- a/core/assets/shaders/shield.frag
+++ b/core/assets/shaders/shield.frag
@@ -11,7 +11,6 @@ uniform vec2 u_texsize;
 uniform float u_time;
 uniform float u_dp;
 uniform vec2 u_offset;
-uniform vec4 u_shieldcolor;
 
 varying vec2 v_texCoords;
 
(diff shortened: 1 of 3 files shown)


core/src/mindustry/world/blocks/defense/ForceProjector.java
```diff
@@ -158,7 +158,7 @@ public void draw(){
 
                 Draw.z(Layer.shields);
 
-                Draw.color(Pal.accent);
+                Draw.color(tile.team().color);
```
Quezler, line 161:
```suggestion
                Draw.color(team.color);
```

core/assets/shaders/shield.frag
```diff
@@ -28,8 +27,23 @@ void main(){
 	if(texture2D(u_texture, T).a < 0.9 &&
        		(texture2D(u_texture, T + vec2(0, step) * v).a > 0.0 || texture2D(u_texture, T + vec2(0, -step) * v).a > 0.0 ||
        		texture2D(u_texture, T + vec2(step, 0) * v).a > 0.0 || texture2D(u_texture, T + vec2(-step, 0) * v).a > 0.0)){
-
-		gl_FragColor = mix(u_shieldcolor, vec4(1.0), si);
+                
+		vec4 bordercolor = vec4(0.0, 0.0, 0.0, 0.0);
+
+		if(texture2D(u_texture, T + vec2(0, step) * v).a > 0.9){
```
Anuken, line 33: Why are the statements split up this way?

Anuken:
Implemented in a slightly cleaner way @ `master`.

core/assets/shaders/shield.frag
Remint32: because I focused on your code and code from other shaders and highlighted the variable Declaration for readability.

Remint32: Don't quite understand the claim to margins, what's wrong with them?

PR #2200, opened 2020-06-30
Don't say a person voted to kick if they voted against

Katsute:
### Prerequisites

- [x] I have checked that no other similar pull requests already exists
- [x] Build compiles
- [x] Build runs without exceptions

### Changes Made

- If a player votes to not kick a player then the server will send a message saying that they voted to _not_ kick the player.
  (currently the server says the player voted to kick the player even when they vote no)


Anuken:
> currently the server says the player voted to kick the player even when they vote no

No, it doesn't. It says the player has voted *on kicking* another player, e.g. "voted on the subject of kicking player X".

PR #2212, opened 2020-07-06
Helper turrets

ThePythonGuy3:
I added `helper` boolean to turrets, making the turrets with `helper = true;` follow you mouse on PC, point at enemies on Mobile, and shoot if you are shooting.     

Here you can see "Tracker", the new turret that I added for this PR. (It has no sprite because you can't pack sprites right now, but it is on the PR `assets-raw/sprites/blocks/turrets` folder):

image

The sprite is actually this:

image

To avoid major problems and bugs, the helper turrets work as a normal turret while on multiplayer. (I havent tested this because multiplayer is broken, but it should)

I also added Tracker to the english bundle.


ThePythonGuy3:
Thanks to @Voz-Duh for the resprite of Tracker!

image

Anuken:
I don't see why these should be separate turrets when any turret should (theoretically) be able to be controlled in a formation like a unit.

SnakkiZXZ:
I decided that this sprite is not very suitable and drew my
here turrets itself and texture heat
helper turret
helper turret-heat
`^  she is not visible but she is`

Anuken:
As said previously, I don't think this should be a separate turret. Closing.

PR #2218, opened 2020-07-10
[QOL] SchematicInfoDialog Block Count Implementation

zethnest:
A preview of it:
Screenshot_20200710_102455

Would be cool to see the block count to see ratios are correct or not.

### DIFF
--- a/core/src/mindustry/ui/dialogs/SchematicsDialog.java
+++ b/core/src/mindustry/ui/dialogs/SchematicsDialog.java
@@ -345,9 +345,7 @@ public class SchematicsDialog extends BaseDialog{
                 for(Entry ent : blocks){
                     int v = ent.value;
                     t.image(Vars.content.block(ent.key).icon(Cicon.small)).left();
-                    t.label(() -> {
-                        return "[lightgray]" + v;
-                    }).left().padLeft(4).padRight(6).padTop(4).padBottom(4);
+                    t.add("[lightgray]" + v).left().padLeft(4).padRight(6).padTop(4).padBottom(4);
 
                     if(++i % 6 == 0){
                         t.row();
--- a/core/src/mindustry/ui/dialogs/SchematicsDialog.java
+++ b/core/src/mindustry/ui/dialogs/SchematicsDialog.java
@@ -11,7 +11,10 @@ import arc.scene.ui.ImageButton.*;
 import arc.scene.ui.TextButton.*;
 import arc.scene.ui.layout.*;
 import arc.struct.*;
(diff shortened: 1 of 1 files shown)


core/src/mindustry/ui/dialogs/SchematicsDialog.java
```diff
@@ -333,6 +336,27 @@ public void show(Schematic schem){
...
+            IntIntMap blocks = new IntIntMap();
+            schem.tiles.each(t -> {
+                blocks.increment(t.block.id);
+            });
+            cont.table(t -> {
+                int i = 0;
+                for(Entry ent : blocks){
+                    int v = ent.value;
+                    t.image(Vars.content.block(ent.key).icon(Cicon.small)).left();
+                    t.label(() -> {
```
Anuken, line 348: This should not be a lambda label. Just use `add` with a string.

core/src/mindustry/ui/dialogs/SchematicsDialog.java
```diff
@@ -333,6 +336,27 @@ public void show(Schematic schem){
             cont.add(new SchematicImage(schem)).maxSize(800f);
             cont.row();
 
+            IntIntMap blocks = new IntIntMap();
+            schem.tiles.each(t -> {
```
Anuken, line 340: This should be an expression lambda.

XEN0PHIL:
Energy display is pointless, it will never show an accurate power consumption/generation of a complex schematic

zethnest:
> Energy display is pointless, it will never show an accurate power consumption/generation of a complex schematic

That is from another commit. Implemented by Anuke. Please read PR title and code.

Anuken:
How does this look for massive/diverse schematics? I'm concerned that it will run out of space on mobile landscape orientations.

zethnest:
> How does this look for massive/diverse schematics? I'm concerned that it will run out of space on mobile landscape orientations.

It wraps after 6 blocks. I forgot to show 1 where it has a lot of it. Like this for example:
Screenshot_20200711_231716

Anuken:
Right, but my concern is that all the wrapping will reach the bottom of the screen on landscape. It may need a scrollpane, if it doesn't have one already.

zethnest:
Screenshot_20200712_091344
The picture above is this PR ported to 5.0 to test on mods. Unsure of the behavior of the icons, probably because of the mod. But the schematic display is shrinked. Unsure of master.

zethnest:
I see no significant need for this. Will reopen when there are requests and further ideas for improvements.

PR #2219, opened 2020-07-10
Drills accept Cryofluid as well as water

mnrsU:
Mindustry 7_9_2020 10_09_28 PM (2)

Drills can now accept cryofluid and water as coolant

water cooling is nerfed, bigger drills have better boost effects

probably needs balancing

based on [https://github.com/Anuken/Mindustry-Suggestions/issues/277](https://github.com/Anuken/Mindustry-Suggestions/issues/277)

EDIT: 
Very unsure whether my changes were the best way to implement this (I'm a very new to this stuff)


xXextremeduro69420Xx:
if based on real physics, the drill would freeze if its cooled with cryo, no? my suggestion is to apply this change only for impact drills.

mnrsU:
> if based on real physics, the drill would freeze if its cooled with cryo, no? my suggestion is to apply this change only for impact drills.

True. I'll remove the cryo-cooling for pneumatic and mechanical drills then.

PR #2227, opened 2020-07-14
Item teleport done

SnakkiZXZ:
Why not make a special block that teleports items directly to the core?
I think it should be a very expensive unit and require a lot of energy to work.
 Here is my version of this block: 
**1)** it is expensive 
**2)** in case of incomplete health, it begins to destroy objects with a chance
**3)** if you destroy it, it will blow everything around in a radius of 3 blocks
**4)** requires a lot of energy to work

miniGif_20200714173904
there may be errors in the code, I'm not perfect

`I seem to figure it out`


ValkovY:
Not a bad idea! Briefly and clearly. And most importantly, it is a balanced block

Anuken:
> Why not make a special block that teleports items directly to the core?

Because that makes transportation far too easy. Even if it has many limitations, I don't see the need for such a block. Sorry, I won't merge this.

PR #2232, opened 2020-07-15
Hydro power station

SnakkiZXZ:
There is an idea about a new generator - hydrodynamic! it generates energy if installed on a water tile, if there is no water on the map it can be brought through pipes, but the energy output will be less.
miniGif_20200715204801
it shows how it works.

`A lot has been added in 6.0, but there are no new generators yet.`


DeltaNedas:
very very very low effort, this will only promote stupid checkerboard power cubes of extractors and hydros

ValkovY:
**In principle, this is even a very good idea.** Only depends on how much it will consume and what its craft will be. And so-cool.
And Yes, it would be possible to make the textures move, I know that it is difficult, but it would be beautiful and realistic...

Anuken:
Something like this was on the Trello long ago, but I'm not sure that it is necessary. Will it fill much of a niche that solar panels don't already?

The code posted here isn't valid at all, both in terms of style and implementation - even if it is accepted, I'll be adding it myself.

PR #2239, opened 2020-07-17
Boost pyratite

gen-NA-owl:
Pyratite needs to be buffed. You can find the reasons here: https://github.com/Anuken/Mindustry-Suggestions/issues/80

Its Annoying to Make, and when you make it, there isn't much of it: 
Changed the craft time for mixers to 1 sec. (excluding cryo-mixers)

Energy Sources:
Buffed differentials to make them low maintenance (as no one really uses them). Also increased the energy output for all gens that use pyratite. 

Ammo: 
Made the status effect "burning" much more powerful. Also buffed all "Incendiary"-type bullets. 

These changes probably needs rebalancing, as I probably made pyratite too good. Hopefully in the meantime, people will start to use pyratite as something other than just an ingredient in blast.


Anuken:
The balance of this PR is completely broken; pyratite has already been significantly buffed since 5.0. Taking time into account, large pyratiate bullets deal more damage then thorium, and pierce armor with their burning effect.

Issue #130, opened 2018-06-06
Naming Bug


Gabgab2003:
You can make your Username 127 chars long and still join servers(128 throws an error) by editing C:\Users\\%username%\\.prefs\\io.anuke.moment. Furthermore you can make your name invisible by setting the color value to -0000000000.
![image](https://user-images.githubusercontent.com/33293877/41059731-ad1061ba-69cd-11e8-9a09-4fdeda5b508d.png)

Anuken:
As of the latest current 4.0 commit, the server now strictly validates length of usernames, chat messages and clear tags in names. Closing.

Issue #82, opened 2018-01-22
BlockSync Not Sending


Anuken:
Sometimes, BlockSyncPackets will not be sent to players from the server, resulting in desynced world state. Can only reproduce on non-local servers, so far.

PR #2241, opened 2020-07-18
boost thermal pumps and airblast drills

gen-NA-owl:
If thermal pumps are really "The ultimate pump", as it says in the description, why does it pump less liquid per tile than rotary pumps?

Same goes for airblast drills.

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1013,7 +1013,7 @@ public class Blocks implements ContentList{
 
         thermalPump = new Pump("thermal-pump"){{
             requirements(Category.liquid, with(Items.copper, 80, Items.metaglass, 70, Items.silicon, 30, Items.titanium, 40, Items.thorium, 35));
-            pumpAmount = 1.5f;
+            pumpAmount = 2.0f;
             consumes.power(0.30f);
             liquidCapacity = 40f;
             hasPower = true;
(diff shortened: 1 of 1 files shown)


core/src/mindustry/content/Blocks.java
```diff
@@ -1013,7 +1013,7 @@ public void drawBase(Tile tile){}
 
         thermalPump = new Pump("thermal-pump"){{
             requirements(Category.liquid, with(Items.copper, 80, Items.metaglass, 70, Items.silicon, 30, Items.titanium, 40, Items.thorium, 35));
-            pumpAmount = 1.5f;
+            pumpAmount = 2.0f;
```
gen-NA-owl, line 1016: this makes it about 11% more powerful

Dexapnow:
Because both of them allow player to make use of more tiles.

gen-NA-owl:
> Because both of them allow player to make use of more tiles.

I know, but shouldn't they at least be a bit more powerful?

ILiekMelons:
I agree

itcannotbe:
Why make this a pull? This should be a suggestion not a direct edit to the code.

stuffyAI:
true, for balancing changes your should make suggestions where people vote and link the commit there, or talk in discord as to why and how this is.

gen-NA-owl:
> Why make this a pull? This should be a suggestion not a direct edit to the code.

Very well. I shal close it.

stuffyAI:
there currently are 2 suggestions open regarding this and both are going south, still you can look at them if you want  [#130](https://github.com/Anuken/Mindustry-Suggestions/issues/130), [#82](https://github.com/Anuken/Mindustry-Suggestions/issues/82)

PR #2244, opened 2020-07-19
Forbidden character in autosave filename fix

itcannotbe:
Added a line to replace the character : with - since it is illegal in windows to have a file name with the character :


Anuken:
I fixed this before I saw your PR.

itcannotbe:
It's fine, you implemented it better anyway.

PR #2245, opened 2020-07-19
change Overdrive Projector to Overdriver and Large Overdrive Projector to Overdrive Projector

gen-NA-owl:
A small mend block is a mend(er) and a large one is a mend (projector). I think having the small overdrive device as an overdriv(er) and a large one as a overdrive (projector) would be feasible as "large overdrive projector" needs to be renamed anyway.


Anuken:
Sorry, I'll need to use a different name for these:

- It doesn't make much sense to keep the 2x2 mend-block a "projector" while naming the 2x2 overdrive-block a "mender". If they're the same size, why are their "tier" names different?
- Renaming an existing block or changing its size has serious consequences; older saves will attempt to load older 2x2 "mend projectors" as the new 3x3 projectors, leading to all sorts of problems.

PR #2249, opened 2020-07-21
Change wave targeting

gen-NA-owl:
The main purpose of waves are to apply status effects to all units around them, so shouldn't they prioritize units without the status effect they apply? Also I don't see why waves shouldn't be able to target air, so I changed that too.


gen-NA-owl:
Imma close this and add this to suggestions

PR #2257, opened 2020-07-23
Update Tetrative Reconstructor's Liquid Capacity

Zenith-LY:
I increased the liquid capacity of tetrative reconstructor because when we use boosted overdrive projector or large overdrive projector, tetrative reconstructor stops because i think it requires so much liquid in a sec and the capacity cant handle that (180+/sec for a capacity of 10.) and to make it so it can handle it i increased the capacity. I don't know if this will fix the issue because i couldnt try it.

Default capacity: 10
Suggested capacity: 15

If 15 doesnt work then we may have to increase it even more because im sure the issue is related to liquids and i think its about liquid capacity.


zethnest:
It should be because of throughput and not storage capacity

Anuken:
I already changed its capacity to 60 before you made this PR.

PR #2261, opened 2020-07-24
Add Atanner Gaming Hub to Serverlist

AydenRennaker:
The hub has all of our main servers listed as well as a few that are for verified members. We are working on a few more gamemodes and servers to be added to the hub.


Anuken:
This PR got lost. It's probably best to put this in the v6 list now.

PR #2269, opened 2020-07-27
add 'use strict' to mod scripts

DeltaNedas:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

tl;dr;
`const x = 0; x = 1; print(x);`
will throw an error instead of silently ignoring it and printing 0

### DIFF
--- a/core/src/mindustry/mod/Scripts.java
+++ b/core/src/mindustry/mod/Scripts.java
@@ -93,7 +93,7 @@ public class Scripts implements Disposable{
                 context.evaluateString(scope, "modName = \"" + currentMod.name + "\"\nscriptName = \"" + file + "\"", "initscript.js", 1, null);
             }
             context.evaluateString(scope,
-            wrap ? "(function(){\n" + script + "\n})();" : script,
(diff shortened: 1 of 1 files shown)


zethnest:
Maybe a warning instead of error?

core/src/mindustry/mod/Scripts.java
```diff
@@ -93,7 +93,7 @@ private boolean run(String script, String file, boolean wrap){
                 context.evaluateString(scope, "modName = \"" + currentMod.name + "\"\nscriptName = \"" + file + "\"", "initscript.js", 1, null);
             }
             context.evaluateString(scope,
-            wrap ? "(function(){\n" + script + "\n})();" : script,
+            wrap ? "(function(){'use strict'\n" + script + "\n})();" : script,
```
Shawak, line 96: It really should be `'use strict';\n`

ILiekMelons: you missed a semicolon

Anuken:
image

This is already a thing...?

DeltaNedas:
That specifically is fixed but there are lots of other things specified in the link

PR #2271, opened 2020-07-27
Remove "sortersnek"

Dexapnow:
There is many reasons behind it, but most important (that let to removing other "sneks") is infinite throughput/teleporter like mechanic.
I know that most of players like it and they will never change their mind, so this might be better solution.

### DIFF
--- a/core/src/mindustry/world/blocks/distribution/Sorter.java
+++ b/core/src/mindustry/world/blocks/distribution/Sorter.java
@@ -88,7 +88,7 @@ public class Sorter extends Block{
 
         boolean isSame(Building other){
             //uncomment code below to prevent sorter/gate chaining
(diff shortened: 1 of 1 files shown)


stuffyAI:
Anuken said this won't be changed since he made a vote in discord and it got voted against.
You can look at countless suggestions which all got cancelled because of that, I bet this won't change today either.

DeltaNedas:
item -> snek -> sand
teleporting is too much for an item to take

ILiekMelons:
oh no ^

ILiekMelons:
also if you don't like the snakes just don't use them

Remint32:
Press 🇫

Minecrafter69:
6A640727-489A-4C6F-952B-A7E715026C9A

ILiekMelons:
😭

FatalS117:
HA LOL

PR #2274, opened 2020-07-29
Make snek great again

daniel071:
This commit reverts pull request #2271 

Majority of the people voted **for** keeping the sortersnek in the Mindustry discord server.
(https://discordapp.com/channels/391020510269669376/513178557653188609/695649084895133766)

By removing the sortersnek, a lot of efficient and compact schematics would be broken, so that's why I reverted it.

### DIFF
--- a/core/src/mindustry/world/blocks/distribution/Sorter.java
+++ b/core/src/mindustry/world/blocks/distribution/Sorter.java
@@ -88,7 +88,7 @@ public class Sorter extends Block{
 
         boolean isSame(Building other){
             //uncomment code below to prevent sorter/gate chaining
-            return other != null && (other.block() instanceof Sorter || other.block() instanceof OverflowGate);
+            return other != null && (other.block() instanceof Sorter/* || other.block() instanceof OverflowGate */);
         }
 
         Building getTileTarget(Item item, Building source, boolean flip){


HungryForWater:
Keep the sorter/gate logic the way it is. Although it is a bug. It has been a part of the community for a long time, and has been studied to the peak. It is the complete definition of efficiency. And is a wonderful thing. Yes, It's a bug that has to be removed. But atleast create a proper block that replicates  the sorter chain with the same logic and absolute accuracy. It's like copying and pasting! :D

itcannotbe:
Actually it won't be the same as the underlying issue has been fixed already.

The only thing that this reverts is tha ability to use overflows/underflows next to sorters. The insane bandwith however is not going to return.

Also ask what happened to the teleporters of classic. Then we can argue about "It has been a part of the community for a long time, and has been studied to the peak. It is the complete definition of efficiency. And is a wonderful thing."

itcannotbe:
Actually it won't be the same as the underlying issue has been fixed already.

The only thing that this reverts is the ability to use overflows/underflows next to sorters. The insane bandwith however is not going to return.

Also ask what happened to the teleporters of classic. Then we can argue about "It has been a part of the community for a long time, and has been studied to the peak. It is the complete definition of efficiency. And is a wonderful thing."

Also every mechanic that provided a lot of bandwith has been removed from the game, what makes this so special from the rest of them that can warrant it's return?

DeltaNedas:
Here's what you do
1. abuse snek until 6.0
2. download 6.0 snek mod
3. use it

stuffyAI:
>     2\. download 6.0 snek mod

actually we will get a plastanium conveyor which will give us access to 24 items/sec, but you cannot use it in factories like the snek due to it's behaviour

itcannotbe:
Actually you can overdrive them to get more throughput.

zethnest:
@itcannotbe Teleporters are implemented and works as the way they should. Unlike snek, it is accidental.

itcannotbe:
But. "It has been a part of the community for a long time, and has been studied to the peak. It is the complete definition of efficiency. And is a wonderful thing."

Nobody complain and got the point why they weren't a part of v4 back then.

stuffyAI:
> plastanium conveyor

.

Anuken:
> It is the complete definition of efficiency. And is a wonderful thing.

That's why I'm removing it. Designs should not be utilizing an accidental feature with such ridiculous throughput. You can make the same argument for Classic's teleporters and the old vault snakes - yes, they're very efficient, but they really don't make the game better. 

If I see that the removal of overflow chaining makes too much of an impact on designs in v6, I might consider removing it. Until then, *no*.

stuffyAI:
> 24 items/sec

Well plastanium conveyors just got buffed to 40 items/second, that's 2/3rd of sneks at optimal host performance.
Main difference now is that they are locked behind titanium, however plastanium conveyors are easier to build with.

itcannotbe:
Actually sneks can do much more than 40 i/s

Also before you go crazy with plastanium conveyors.

They are not drop in replacement for other conveyors.

Absolute359:
> yes, they're very efficient, but they really don't make the game better.

I can totally agree with that.
I love using them for compact designs for a harmonious appereance of a factory instead of belts, but less for their high throughput.
However, spectating Multiplayer you can see most of the Players mess around with this snek and others try to fix them all the time instead of improving economy or defense/attack lines. Thats far away from a good gameplay i think.
Talking about defense, that would be probably really game changing for high-end defenses without the snek. But hey, there are good designers out there and they gonna setup good schematics without using the snek.

stuffyAI:
> Actually sneks can do much more than 40 i/s

Yes, sneks usually do 60 items per second, as I said here:

> 40 items/second, that's 2/3rd of sneks at optimal host performance

40x3/2=60, which is usually the host frames unskipped, the speed at which sneks operate.

> They are not drop in replacement for other conveyors.

right now they behave that way, compile current build and test them.
They don't work single very well, but practically they are a straight upgrade since you never really find a single use for them, given that you can just use sorters or gates for fast transport in single blocks.

Absolute359:
> sneks lose speed with tickrate.

well below 60TPS I'm able to squeeze over 7000 items/s through 1 overflow/sorter on singleplayer (PC)
its so many

z0mbiesrock:
> > sneks lose speed with tickrate.
> 
> well below 60TPS I'm able to squeeze over 7000 items/s through 1 overflow/sorter on singleplayer (PC)
> its so many



> > It is the complete definition of efficiency. And is a wonderful thing.
> 
> That's why I'm removing it. Designs should not be utilizing an accidental feature with such ridiculous throughput. You can make the same argument for Classic's teleporters and the old vault snakes - yes, they're very efficient, but they really don't make the game better.
> 
> If I see that the removal of overflow chaining makes too much of an impact on designs in v6, I might consider removing it. Until then, _no_.

> Anuken said this won't be changed since he made a vote in discord and it got voted against.
> You can look at countless suggestions which all got cancelled because of that, I bet this won't change today either.

Democracy schmemocracy, remember:
5c3vz8kmt9j01

stuffyAI:
> well below 60TPS I'm able to squeeze over 7000 items/s through 1 overflow/sorter on singleplayer (PC)

The cap lies in the distributer as input output, quite sure they limit it to 6 full blue belts ( if you try 7 they start to stagger), so if it doesn't go to the core it will be slower

sample-text-here:
I think they should be added back, but limit the throughput and speed to half of a titanium conveyor. So designs will still work, but sneks won't be overpowered.

itcannotbe:
https://github.com/Anuken/Mindustry/pull/2274#issuecomment-665632868

FatalS117:
> I think they should be added back, but limit the throughput and speed to half of a titanium conveyor. So designs will still work, but sneks won't be overpowered.

_N O_

PR #2275, opened 2020-07-29
Shift to prioritize build

blahblahbloopster:
Hold shift to prepend build request to queue instead of appending.  Note: I was unable to make it work for breaking requests.


zethnest:
You mean the boost button? What if the player is boosting?

Anuken:
I'm not a fan of hidden controls like this.

sample-text-here:
Maybe it could be in the tutorial

markozajc:
would be really neat to have

PR #2286, opened 2020-07-30
Merge item void and liquid void into "void".

Catchears:
Simple change to merge Item and Liquid Void. Here's the suggestion: 
https://github.com/Anuken/Mindustry-Suggestions/issues/351

I left it as "ItemVoid" in the code because I can't call it "Void".


Anuken:
I don't see what problem this solves. Having separate voids isn't useful in most cases, but there may be times when you want a block to only accept only items or liquids, e.g. when there's limited space and you want an item void next to a liquid tank.

This would also break saves and schematics with voids in them.

Zenith-LY:
> I don't see what problem this solves. Having separate voids isn't useful in most cases, but there may be times when you want a block to only accept only items or liquids, e.g. when there's limited space and you want an item void next to a liquid tank.
> 
> This would also break saves and schematics with voids in them.

Makes sense.

z0mbiesrock:
Why not a sandbox-only Incinerator that doesn't rewuire power?

Anuken:
> I don't see what problem this solves.

PR #2299, opened 2020-08-02
Add capacity indicator to vaults and containers.

SummetDev:
This is how they will look:
mindustry-vault-capacity
mindustry-container-capacity


itcannotbe:
Have you considered that vaults can actually hold more than 1 item which has it's own limit?

TheSlaylord:
I think indicators will work just like an indicator in the core.

Summet-V:
Ok, now it works the same as with cores.
mindustry-vault-capacity-fixed

itcannotbe:
Wouldn't this cause confusion? Since technically the limit is still 1000 for vaults and 300 for containers.

Doing Capacity * Known Items would confuse more people than help them especially when used not connected to the core.

Summet-V:
Last change: It will show how much vault/ container holds for one resource, and the green bar will show the capacity for all items.

summetdev:
Conflicts resolved.

PR #2304, opened 2020-08-03
update README.md

DeltaNedas:
5.0 throws an error when importing 6.0 maps.
People frequently ask where to download vanilla maps from, so it's mentioned in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,10 @@ _[Javadoc](https://mindustrygame.github.io/docs/)_
 
 See [CONTRIBUTING](CONTRIBUTING.md).
 
+### Maps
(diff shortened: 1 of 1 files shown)


ILiekMelons:
perhaps we could make a thing to convert 5.0 maps to 6.0

Anuken:
> perhaps we could make a thing to convert 5.0 maps to 6.0

Read the text carefully. It's the other way around.

> People frequently ask where to download vanilla maps from, so it's mentioned in the readme.

I don't think the readme should contain anything other than build instructions and links.

ILiekMelons:
> Read the text carefully. It's the other way around.

whoops, sorry

PR #2318, opened 2020-08-07
Update servers_be.json

Quezler:
hi anuke,

i would like to put the 6.0 nydus server on another vps so that it does not interfere with the 5.0 nydus server.

of course i could have done this with a weird firewall bunnyhop, but this solution is a bit more clean. (i hope you don't mind)

and if i ever do move them back to the same device it'll be a simple dns change on my end, since the ports differ.

❤️

### DIFF
--- a/servers_be.json
+++ b/servers_be.json
@@ -3,6 +3,6 @@
     "address": "mindustry.us.to:6568"
   },
   {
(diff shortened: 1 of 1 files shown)


ILiekMelons:
nydustry

PR #2324, opened 2020-08-08
deep water is 2x better to pump

DeltaNedas:
May implement https://github.com/Anuken/Mindustry-Suggestions/issues/383, I can't build it due to annotation processor errors.


annotations/src/main/resources/revisions/Bullet/5.json
```diff
@@ -0,0 +1 @@
+{version:5,fields:[{name:collided,type:arc.struct.IntSeq,size:-1},{name:damage,type:float,size:4},{name:data,type:java.lang.Object,size:-1},{name:lifetime,type:float,size:4},{name:owner,type:mindustry.gen.Entityc,size:-1},{name:team,type:mindustry.game.Team,size:-1},{name:time,type:float,size:4},{name:type,type:mindustry.entities.bullet.BulletType,size:-1},{name:x,type:float,size:4},{name:y,type:float,size:4}]}
```
Anuken, line 1: This shouldn't be in the commit. It might have been added here due to some Java version differences.

PR #2338, opened 2020-08-11
Revert Logic Blocks Back To Orange Color

Zenith-LY:
This PR will probably fail since they were originally orange but then changed to purple.

I made this PR because every block in mindustry is orange themed while logic blocks aren't. Maybe because they are very unique blocks but still they should be orange to match the theme because when they are purple they look like blocks from "purple" team.

I also think purple looks good but for the theme it has to be orange. I know they are very unique blocks but they also have sprites and making it orange would be better for the theme.


Anuken:
> every block in mindustry is orange themed

What are you talking about? Less than half of all blocks are orange themed. There are plenty of red, purple, yellow, brown, blue and black blocks.

summetdev:
I think he meant that the accent color is orange.

Anuken:
> I think he meant that the accent color is orange.

What did you think I was referring to? There is no other color besides the accent that could be orange in this case.

Just look in the drill/crafting/liquids tab - many of the blocks there have no orange in them at all. This doesn't make any sense.

Zenith-LY:
> What are you talking about? Less than half of all blocks are orange themed. There are plenty of red, purple, yellow, brown, blue and black blocks.

Ok but purple felt strange.

Anuken:
I don't agree.

PR #2362, opened 2020-08-17
M A K E S P E C T R E G R E A T E R

laohuaji233:
Specter costs a lot of material, but it’s less useful than ripple, cyclone
Moreover it consumes 120 liquid/second
So players seldom use specter 
I think specter should be greater and more useful

### DIFF
--- a/core/src/mindustry/content/Bullets.java
+++ b/core/src/mindustry/content/Bullets.java
@@ -350,18 +350,21 @@ public class Bullets implements ContentList{
         standardDenseBig = new BasicBulletType(7f, 55, "bullet"){{
             width = 15f;
             height = 21f;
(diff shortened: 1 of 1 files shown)


Dexapnow:
Specter doesn't have knowback, so making bullets price wouldn't be greatest idea.

gen-NA-owl:
how about adding a pierce cap (like I did) to not make it too op?

Dexapnow:
I would say that it is good at it is, there is no need to buff/nerf it, spectre uses some of cheapest ammo types and deals highest damage per that material in game.

Anuken:
Spectre was already buffed. I'll reopen this if it turns out those buffs were not sufficient.

PR #2374, opened 2020-08-19
use 6.0 branch for github import

DeltaNedas:
all 6.0 mods i know of use this for the 6.0 port.
revert when 6.0 is ready to be the default version.

### DIFF
--- a/core/src/mindustry/ui/dialogs/ModsDialog.java
+++ b/core/src/mindustry/ui/dialogs/ModsDialog.java
@@ -106,7 +106,7 @@ public class ModsDialog extends BaseDialog{
                             Core.settings.put("lastmod", text);
 
                             ui.loadfrag.show();
-                            Core.net.httpGet("http://api.github.com/repos/" + text + "/zipball/master", loc -> {
+                            Core.net.httpGet("http://api.github.com/repos/" + text + "/zipball/6.0", loc -> {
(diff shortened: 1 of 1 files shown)


Anuken:
It should only use the 6.0 branch if it's sure that the branch exists.

Remint32:
@Anuken If a mod maker writes a mod for 6.0, then he knows what he is doing and it will not be difficult for him to make a branch 6.0

Anuken:
> @Anuken If a mod maker writes a mod for 6.0, then he knows what he is doing and it will not be difficult for him to make a branch 6.0. At the moment, you need to create two separate repositories if you want to make an existing mod for 6.0.

It's not guaranteed that the person would know what they're doing, especially considering this 6.0-branch switch isn't documented anywhere. If someone happened to make a mod that functions as a file but fails github import with a cryptic 404, it wouldn't be immediately obvious what the problem is.

> We have two versions at the same time. We have a separate list of official servers for versions. Why don't we have such a division for mods, given that developers really need it?

What does this have to do with my message? I didn't say anything about having *no division*. I'm saying that *if* the mod system uses a 6.0 branch by default, it should at least check that the branch exists (e.g. does not 404) to prevent confusion and unnecessary bug reports.

DeltaNedas:
yeah ill add master fallback soon

ILiekMelons:
soon

PR #2382, opened 2020-08-19
Server disable schematic option

J-VdS:
Some servers like to disable schematics ex to prevent (NSFW) pixelart.
And this was also requested by some plugin makers.

It works like the `reactorExplosions` rule. You can disable schematics via:
`rules add schematicAllowed true`

## How does it affect the player?
1. The schematic button is disabled or greyed out.
afbeelding

2. If you use the schematic keybind `T` or copy  keybind `F` an error message gets printed on the screen.
afbeelding


J-VdS:
I could add this to the editor too.

DeltaNedas:
1 line edit for furries with custom clients
alternatively you could have mods to ban anyone placing nsfw schematics
the real reason is that noobs abuse schematics but that would take a community change or good plugin not an easily bypassable boolean

J-VdS:
@DeltaNedas the problem is harder than you think it is. If you really want to try smth. It takes more effort that is all. And you cannot really block schematics via a plugin.

BlueWolf3682:
I support this entirely, indeed I've even mention this idea to Quezler as an idea cause honestly, spamming schematics isn't the way to play mindustry, and I feel like the schematics option should be removed entirely from the game.

ILiekMelons:
I don't think the copy keybind should be on the same rule imo

J-VdS:
> 
> 
> I don't think the copy keybind should be on the same rule imo

Yeah I was thinking about this. But if you plan to go the "block packets"-way a server will not know the difference between copy and schematic.

I will add the change (for now). Should you still be able to save a schematic?

ILiekMelons:
> > 
> > 
> > I don't think the copy keybind should be on the same rule imo
> 
> Yeah I was thinking about this. But if you plan to go the "block packets"-way a server will not know the difference between copy and schematic.
> 
> I will add the change (for now). Should you still be able to save a schematic?

i dont see why not given that you can just download the whole map anyway

markozajc:
this is just one more thing that'll make people use a custom client with this feature disabled since there's no way to make it a server side check, meaning you're just restricting an opensource and easily moddable client

stuffyAI:
> schematics option should be removed entirely from the game

People gotta remember the builds and do them manually you say?
Sure there are some 4heads that spam their massive air to power schematics before you have copper but any efficient factory builder qutilizes schematics due to bridge intersections and block settings, they are also shareable so newer players can utilize more complex ones to understand some of the tricks.



> that'll make people use a custom client with this feature disabled since there's no way to make it a server side check

One first way catch would be denying placing multiple different blocks instantly, this could be done via plug-in even, which would require the custom client to buffer schematics purely client side until the server registers separate placements.

markozajc:
doesn't matter if the idea is good or not, there's absolutely no way to implement it reliably. it's the same as uuid bans, you can just modify the source (or easier, download a custom client) to spoof it.

Shawak:
> air to power schematics before you have copper
I feel like this is more a game balancing problem

J-VdS:
> 
> 
> > schematics option should be removed entirely from the game
> 
> People gotta remember the builds and do them manually you say?
> Sure there are some 4heads that spam their massive air to power schematics before you have copper but any efficient factory builder qutilizes schematics due to bridge intersections and block settings, they are also shareable so newer players can utilize more complex ones to understand some of the tricks.
> 
> > that'll make people use a custom client with this feature disabled since there's no way to make it a server side check
> 
> One first way catch would be denying placing multiple different blocks instantly, this could be done via plug-in even, which would require the custom client to buffer schematics purely client side until the server registers separate placements.

`could be done with a plugin` No it isn't easy to detect it. I tried it. And most of the solutions are not failproof. 
You could check for buildqueue size and detect big grows but a bad internet connection could mess things up pretty quickly.

`new players can learn new things` This true but I see a lot of players just grabbing/copying schematics from discord and some are bad/require a ton of resources.

Quezler:
(responding to the original post, haven't read the rest of it in here yet)

while i do like the idea of the schematic library being unavailable / greyed out, i oppose the idea of not allowing people to `F` (pun intended), since that will only allow you to clone structures already on the map, while this does allow for spamming of stuff initially build by hand i do believe it would be far to annoying to not have access to it at all 🤔

J-VdS:
> 
> 
> (responding to the original post, haven't read the rest of it in here yet)
> 
> while i do like the idea of the schematic library being unavailable / greyed out, i oppose the idea of not allowing people to `F` (pun intended), since that will only allow you to clone structures already on the map, while this does allow for spamming of stuff initially build by hand i do believe it would be far to annoying to not have access to it at all 🤔

I allowed copying in a later commit. But you can't detect the difference between copying/schematics serverside.

Quezler:
well there is no stopping custom clients from using or mimicking schematics anyways, you can only ask the client nicely 🤔

stuffyAI:
> I feel like this is more a game balancing problem.

No, I meant that 4heads try to place their overdriven air to power shematics before you even have the stuff to build it and waste all resources, if there is so much space that air to power is the main power source it's the map design or fault

core/src/mindustry/game/Rules.java
```diff
@@ -38,6 +38,10 @@
     public boolean canGameOver = true;
     /** Whether reactors can explode and damage other blocks. */
     public boolean reactorExplosions = true;
+    /** Whether schematics are allowed */
+    public boolean schematicAllowed = true;
+    /** Whether copying is allowed */
+    public boolean copyAllowed = true;
```
Anuken, line 44: I don't think this should be changeable by servers. Copying is very useful, and doesn't have the same issues that schematics may have.

core/src/mindustry/input/DesktopInput.java
```diff
@@ -402,6 +403,8 @@ void pollInput(){
         if(Core.input.keyTap(Binding.schematic_menu) && !Core.scene.hasKeyboard()){
             if(ui.schematics.isShown()){
                 ui.schematics.hide();
+            }else if(!state.rules.schematicAllowed) {
+                ui.showInfoToast("@schematic.disabled", 3f);
```
Anuken, line 407: People should still be able to save and view schematics, even if the option is disabled. Instead of preventing access to the menu, simply popup a dialog when they click a schematic in the menu instead.

Anuken (changes requested):

core/src/mindustry/game/Rules.java
J-VdS: I will push a new version with some major changes, so I will close this if it is fine.

core/src/mindustry/input/DesktopInput.java
J-VdS: I will push a new version with some major changes, so I will close this if it is fine.

Anuken:
closing, as per comment

J-VdS:
new PR  #2660

PR #2383, opened 2020-08-20
make arc cost silicon

DeltaNedas:
costs 20 here

reasoning: shock mine takes silicon, why shouldn't arc

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1476,7 +1476,7 @@ public class Blocks implements ContentList{
         }};
 
         arc = new PowerTurret("arc"){{
-            requirements(Category.turret, with(Items.copper, 35, Items.lead, 50));
+            requirements(Category.turret, with(Items.copper, 35, Items.lead, 50, Items.silicon, 20));
             shootType = new LightningBulletType(){{
                 damage = 21;
                 lightningLength = 25;


ILiekMelons:
no

ifritdiezel:
I see this as a better way to nerf arc. Increasing power consumption doesn't affect it's overpowered mechanic anyhow and adding silicon cost at least makes it unusable in first waves.

ThatOneBepis:
Congrats
you jxhn'd the concept

ritzip:
hail time

BasedUser:
do keep in mind that arc already costs a buttload of power

Ilya246:
20 silicon is too much - it should be at most 10, considering the recent 3x power consumption increase

ILiekMelons:
it's already going to delete our power and now it's trying to delete our silicon

BasedUser:
> https://github.com/Anuken/Mindustry/pull/2383#issuecomment-677568008
silicon is already used way too much, i'm thinking titanium instead

BlueWolf3682:
the arc is a powerful turret, yes, but if one has to make silicon then might as well just use lancers, as when used right they're much more powerful than arcs. as well as something else: combustion gens, small batteries, power nodes, they only cost copper and lead. the only way that this turret should be further nerfed is simply less damage.

BlueWolf3682:
the arc is a powerful turret, yes, but if one has to make silicon then might as well just use lancers, as when used right they're much more powerful than arcs. as well as something else: combustion gens, small batteries, power nodes, they only cost copper and lead. the only way that this turret should be further nerfed is simply less damage.

BlueWolf3682:
the arc is a powerful turret, yes, but if one has to make silicon then might as well just use lancers, as when used right they're much more powerful than arcs. as well as something else: combustion gens, small batteries, power nodes, they only cost copper and lead. the only way that this turret should be further nerfed is simply less damage.

BlueWolf3682:
the arc is a powerful turret, yes, but if one has to make silicon then might as well just use lancers, as when used right they're much more powerful than arcs. as well as something else: combustion gens, small batteries, power nodes, they only cost copper and lead. the only way that this turret should be further nerfed is simply less damage.

BlueWolf3682:
the arc is a powerful turret, yes, but if one has to make silicon then might as well just use lancers, as when used right they're much more powerful than arcs. as well as something else: combustion gens, small batteries, power nodes, they only cost copper and lead. the only way that this turret should be further nerfed is simply less damage.

ILiekMelons:
blue calm down

BlueWolf3682:
Wasn't supposed to send multiple times, I have encountered this issue before on discord

markozajc:
20 seems a bit much since arc is an early to mid game turret after all, but I'm all in with the idea of making it cost silicon in the first place

ritzip:
the problem is spam and less damage will either make it be spammed more or useless.

ritzip:
this also makes arcs have a long build time(pvp lmao)

gen-NA-owl:
How about make arc cost 5-10 silc and make lancers also take 25 titanium to build(so arcs are unlocked sooner than lancers)

GlennFolker:
> costs much power
battery exists

Anuken:
The power nerf is enough.

PR #2390, opened 2020-08-20
Update UnitTypes.java

Dexapnow:
Nerfed alpha, I'm not sure if that hight nerf in necessary but "dart rushing" is still a thing.


Anuken:
That's a typo, the damage is supposed to be *0.095* (10x less, which should be sufficient).

PR #2412, opened 2020-08-24
make input fields public

DeltaNedas:
a thing was moved from public player to private input handler and broke my mod h


ILiekMelons:
~~doesnt this mean you just need to fix your mod~~

DeltaNedas:
yes which i cant do without this pull
how else can i get player's target?

Anuken:
I'm not a fan of making half the fields public and half of them private with no particular order. It's inconsistent.

DeltaNedas:
ok
also i forgot some cool line drag stuff too so it was a good idea to go over it again

sk7725:
YES THANK YOU

PR #2420, opened 2020-08-25
Indonesian translation

DaGamerFiles:
~~Still hasn't done yet~~
Completed as of September 16th, 2020.


DaGamerFiles:
Translated both bundle_in_ID.properties (some lines from #2513 included) and Indonesian translation for Google Play Store.

PR #2451, opened 2020-08-28
Make unloaders as fast as titanium conveyors

Catchears:
Previously, unloaders were a tiny bit slower than titanium conveyors which annoyed multiple players because it didn't quite match up. So I went ahead and set the speed to match the speed of the titanium conveyors.


Quezler:
_**sad plastanium conveyor noises**_

summetdev:
~~well, add plastanium unloader~~

Quezler:
~~i think one unloader is more than enough, and blue/purple tier speed match is plenty~~

ILiekMelons:
~~inf speed unloader~~

Anuken:
No. I already made them slightly faster in v6.

potatoisagender:
catchears attempts to rectify my note.

PR #2464, opened 2020-08-30
Update servers_be.json

way-zer:
A bleeding-edge server from China, with lots of plugins


RandomTNT:
I think ye

RandomTNT (commented):
I think, if it contains toxic, remove immediately

way-zer:
How do you think that? I don't even want to reply you.

PR #2483, opened 2020-09-04
Update MapResizeDialog.java

BasedUser:
small change for insignificant ram usage

### DIFF
--- a/core/src/mindustry/editor/MapResizeDialog.java
+++ b/core/src/mindustry/editor/MapResizeDialog.java
@@ -8,7 +8,7 @@ import arc.util.*;
 import mindustry.ui.dialogs.*;
 
 public class MapResizeDialog extends BaseDialog{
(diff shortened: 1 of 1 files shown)


imreallybadatnames:
ok

Quezler:
now make minsize a multiple of 2 as well 👀

summetdev:
~~how about remove map size limits~~

Quezler:
_haha ram goes brrr!_

DeltaNedas:
map size limits shouldn't be final so it's easy to change in f8 rather than recompile for a tiny change

BasedUser:
how about I shoot your computer in the CPU

Anuken:
The limit isn't 500 because of RAM limits, it's 500 because that's a nice round number that can easily be split into chunks of 100/50. There's no reason to use 512 here.

PR #2485, opened 2020-09-05
Make payload units able to pick up bigger blocks based on its capacity

sk7725:
Make payload units be able to pick up blocks of sizes that are smaller or equal to its capacity, not a fixed number(2).   
Units will now be able to pick up blocks if `size^2`(area) is less or equal to its current capacity left.   
1. This will impose that the unit and block storage will be shared, which may be logical(if not, just remove `- pay.payloade().size`)   
2. This will have no effect of `mega`'s capability, as its capacity is 4(although if there are units on board it will only be able to pick up 1x1)   
3. This will let modded units be able to carry meltdowns, or give them more freedom

### DIFF
--- a/core/src/mindustry/input/InputHandler.java
+++ b/core/src/mindustry/input/InputHandler.java
@@ -129,7 +129,7 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
         if(tile != null && tile.team == unit.team && pay.payloads().size < unit.type().payloadCapacity
             && unit.within(tile, tilesize * tile.block.size * 1.2f)){
             //pick up block directly
-            if(tile.block().buildVisibility != BuildVisibility.hidden && tile.block().size <= 2 && tile.canPickup()){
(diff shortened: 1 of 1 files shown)


BasedUser:
This is (in my opinion) a good pull, but keep in mind that you need to include checks for core blocks, as carrying those may break the game. I'm not only talking about `Blocks.coreShard` through `Blocks.coreNucleus`, but all `CoreBlock`s.

I'm making (hopefully by the time any other contributor sees this, *have made*) a pull request that appends this condition to your patch, in case Anuke agrees with the matter.

sk7725:
There is `tile.canPickup()` for a reason I guess....

Anuken:
This is a step in the right direction, but I think the mega cargo capacity should be based on square area for *all* types of cargo. 

For example, a capacity of 1 would be 1 block, or 8x8 = 64 units². A hypothetical unit of size 4 would have an area of 16 units², and would be able to fit in the cargo 4 times. 

I'll see if I can implement something like this.

PR #2486, opened 2020-09-05
More wall fields for modding

sk7725:
- flashColor(flashWhite was renamed to flashHit since it is no longer white)
- lightningColor


Gdeft:
Color.pink for everything intensifies

PR #2488, opened 2020-09-05
Fields for Shock Mine and Door

sk7725:
Door: `doorSound` (does exactly what you think it does)
Shock Mine: `lightningColor` (since lots of of colored lightning are used in 6.0, it would be neat)


sk7725:
oh no Arc

PR #2498, opened 2020-09-06
Distribution/PowerNode/TractorBeamTurret fields for modding

sk7725:
- PowerNode/TractorBeamTurret: `laserColor`(since RepairPoint had laserColor, I thought it would work for these laser blocks to have one too(for PowerNode it is split into `laserColor1` and `laserColor2`))
- Plastanium Conveyor: `loadEffect`, `unloadEffect`(effects now customizable)
- MassDriver: `shootSound`(just like the other turrets, Sounds.none by default)

*...This should be my last json-related PR for now, you've done lots well already, it was really hard trying to find PR-worthy stuff!*

### DIFF
--- a/core/src/mindustry/world/blocks/defense/TractorBeamTurret.java
+++ b/core/src/mindustry/world/blocks/defense/TractorBeamTurret.java
@@ -31,6 +31,7 @@ public class TractorBeamTurret extends Block{
     public float scaledForce = 0f;
     public float damage = 0f;
     public boolean targetAir = true, targetGround = false;
+    public Color laserColor = Color.white;
 
     public TractorBeamTurret(String name){
         super(name);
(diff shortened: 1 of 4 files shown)


Quezler:
> nah, its fine, makes sense, and this was made with modding in mind :) ❤️

Quezler (commented):
changing laser colors might be problematic since their sprites were tainted last time i checked 🤔

sk7725:
@Quezler 
Yes, but it is better than nothing, also I wanted to make it use @-laser and @-laser-end instead but I couldn't figure oit how to change the atlas to rename the related sprites...

Quezler:
fair enough 👍

sk7725:
Also they don't look too bad even when tainted, I played with different colored lasers and they seem fine, the taint is ignorable. The only thing that might be a bit off is when different colored nodes connect to each other, but modders will learn to use JS to make colored nodes only connect to themselves or something. This is just so JSON mods will have more asthetics and JS mods less toil.

PR #2499, opened 2020-09-06
Create readme/chinese/CHINESE.md

guozhenduo:


Anuken:
Sorry, but I don't think the README should be translated. It's not something I see being done on other repositories, and the file can quickly become outdated when I change a certain command or requirement (e.g. the minimum Java version). I think the directions should be simple enough, even if you're using a bad auto-translator.

PR #2503, opened 2020-09-06
add SearchBar utility class

DeltaNedas:
unify search bars everywhere :) 
TODO: make everything use it, currently just modsdialog
TODO: fiddle with padding?


Anuken:
Doesn't seem like it will have much use right now, will re-open if opinions change.

PR #2509, opened 2020-09-06
balance sector capture wave

DeltaNedas:
easy = quick capture
ground 0 at 10 -> frozen forest at 40 is way too much imo


stuffyAI:
> easy = quick capture
> ground 0 at 10 -> frozen forest at 40 is way too much imo

Perhaps we could also increase wave ramping, even if someone just started out the waves from 1-15 are just sitting ducks waiting for research to finish so you can spam arcs, then waiting another 21 waves for the special surprise, just to die or know how to deal with it.

Anuken:
Will balance myself when necessary.

PR #2510, opened 2020-09-06
Second pvp server 6.0

Somka000:
Because players like pvp and attack, but i cant host 2 gamemodes in one server.


DeltaNedas:
hub

SummetDev:
~~6.0 hub~~

Somka000:
idk how make it

Anuken:
You don't need a hub when there are so few BE servers.

PR #2511, opened 2020-09-06
ignore indentation in imported assembly

DeltaNedas:
previously it would treat indented lines as invalid instructions


Anuken:
How about `.replace("\t", "").trim()` for each line instead? Tabs in strings don't make sense anyway.

DeltaNedas:
ok

PR #2514, opened 2020-09-07
[QoL] Port selection field.

antsif-a:
It looks like this:
host-valid

When port is invalid:
host-invalid
The valid port is saved in the game settings and uses as default for the next game.

* [x] Help message refactor.

Contributed by @zethnest.


DeltaNedas:
why close

Remint32:
reopen please

Anuken:
It's better to leave this unmerged for now, because local server discovery breaks when the port used is non-default.

summetdev:
What if I just want to host the game on a different port, without even displaying it in the local server list?

Anuken:
Right now, you can't.

summetdev:
Yes I know. This is why I am making this pull.

Anuken:
I'm saying you can't *not* display it on the local server list.

zethnest:
If a host with a different port is found, should it not be displayed or should it be displayed? Reason?

I know currently Mindustry is able to discover hosts with any port. I am trying to mitigate the problem where `local server discovery breaks when the port used is non-default`

Anuken:
> If a host with a different port is found, should it not be displayed or should it be displayed? Reason?

It doesn't matter. Either one works, as long as it isn't broken.

Servers can be discovered on any port, but the connection port is never passed from the discovery system to the client. Clients will see non-default-port servers on their list, but will fail to connect to them.

zethnest:
@summetdev Merge conflicts and make ready for review?

deltanedas:
mindustry

Anuken:
fake

antsif-a:
real

deltanedas:
real

PR #2516, opened 2020-09-07
Less Capacity For Plastanium Conveyor

Zenith-LY:
10 is a lot and it takes long to fill, i think 5 would be better because the max items anything can use in a frame is 4 with Alloy Smelter using Lead, also, lower capacity means better for worse drills so that'd be a buff to plastanium conveyors and makes it usable with other drills. I couldn't test how 5 capacity would do and btw you can change it.


Anuken:
This just halves the throughput.

itcannotbe:
Ah yes. I think you haven't thought of using multiple loading bays or feed a single bay with multiple inputs.

1hu37y:
above big smart

PR #2517, opened 2020-09-08
Update LAccess.java

BasedUser:
This PR adds a constant returning the block's full name, to be able to determine the block type even if a block is prelinked or mixed with alike blocks (i.e. different tier walls, reactors, conveyors).

### DIFF
--- a/core/src/mindustry/entities/comp/BuildingComp.java
+++ b/core/src/mindustry/entities/comp/BuildingComp.java
@@ -1187,6 +1187,7 @@ abstract class BuildingComp implements Posc, Teamc, Healthc, Buildingc, Timerc, 
 
     @Override
     public double sense(LAccess sensor){
+        if(sensor == LAccess.name) return block.name;
         if(sensor == LAccess.x) return x;
         if(sensor == LAccess.y) return y;
         if(sensor == LAccess.team) return team.id;
--- a/core/src/mindustry/entities/comp/UnitComp.java
+++ b/core/src/mindustry/entities/comp/UnitComp.java
@@ -77,6 +77,7 @@ abstract class UnitComp implements Healthc, Physicsc, Hitboxc, Statusc, Teamc, I
         if(sensor == LAccess.health) return health;
         if(sensor == LAccess.x) return x;
         if(sensor == LAccess.y) return y;
+        if(sensor == LAccess.name) return type.name;
         if(sensor == LAccess.team) return team.id;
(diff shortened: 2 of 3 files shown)


sk7725:
block name is different from link name, all walls are wall1 wall2 wall3... for example.

Anuken:
But this doesn't do anything with the access. It's not implemented in `Building`/`Unit`.

BasedUser:
Say no more.

Anuken:
After thinking about it, this PR is problematic.

The sensor itself is fine, but it introduces a non-constant string to the logic system. Why is this a problem? Because strings are heavily discouraged as a data type, and I will not be adding any functions for them. 

People will inevitably ask for things such as character manipulation and capitalization when they see names such as `titanium-conveyor` being returned for names. They may also get confused when they try to compare strings and discover that `conveyor` == `copper-wall` (because operations are only valid for numbers - everything non-null gets coerced to 1).

What is the use case for this PR?

Anuken:
(Also, this doesn't compile because `sense` must return a number.)

BasedUser:
The following reply is based on a citation from direct messages.

> What is the use case for this PR?

Originally, this PR was intended for detecting different types of blocks (i.e. you can't tell a plastanium conveyor from a normal conveyor, or an impact from a nuke) in certain use cases (prelinking, when a processor `wall` could be a real `ripple`, or getlink, the block of which you have literally NO IDEA what type it is), and filtering them out with `jump`s.

The issue is this:

> because operations are only valid for numbers - everything non-null gets coerced to 1

and I'd never realised that since I was comparing strings to certain constants and itself, and not other strings.

> (Also, this doesn't compile because `sense` must return a number.)

At the time this reply was made, I was trying to figure out a workaround for exactly this issue: first by changing the return type (which broke everything and beyond), then by adding a new class (which was such a dumb idea I stopped it in the middle of the making), and last but not least, adding a separate `string senseS(LAccess content)` function, as well as adding a value for every enum to determine their type, individually, in the case that more read-only string constants were added (i.e. producedUnitType, name (players)).

But I'm starting to think this PR's scale is way out of my league. It'll be converted to draft right after this comment, in case the collective contributor base figures out a more efficient way to do it, or if it's cancelled.

Anuken:
> detecting different types of blocks

For that purpose, I'll add a special `type` sensor and implement some sort of object sensing system. `==` and `!=` for objects will probably have to be implemented as well.

PR #2528, opened 2020-09-10
Request, PowerCell element

markdude247:
I'd like to add a powercell object to the game so users can transport/store power other items


Anuken:
You can transport power in v6 by picking up batteries. This isn't necessary.

PR #2538, opened 2020-09-11
Block Fields for Modding

sk7725:
+ `quickRotate`: whether the block can be rotated with R(or Binding.rotateplaced) + Scroll Wheel.
This is because modded blocks are sometimes not meant to be rotated after placed, and a block cannot tell if it has been rotated unless it keeps track of its rotation, so in some cases it is better if the block is just not rotated at all after placed.
+ `rotateArrow`: whether the block placement is guided by the arrow icon.
This is because some modded block are using the `rotation` of the block in an "unintended" way, such as making only whether it was vertically or horizontally placed count(a.k.a only `rotation%2` matters), and in some cases it just makes sense to get rid of the arrow icon.

### DIFF
--- a/core/src/mindustry/input/DesktopInput.java
+++ b/core/src/mindustry/input/DesktopInput.java
@@ -116,7 +116,7 @@ public class DesktopInput extends InputHandler{
         //draw request being moved
         if(sreq != null){
             boolean valid = validPlace(sreq.x, sreq.y, sreq.block, sreq.rotation, sreq);
-            if(sreq.block.rotate){
+            if(sreq.block.rotate && sreq.block.rotateArrow){
                 drawArrow(sreq.block, sreq.x, sreq.y, sreq.rotation, valid);
             }
 
@@ -146,13 +146,13 @@ public class DesktopInput extends InputHandler{
             if(mode == placing && block != null){
                 for(int i = 0; i < lineRequests.size; i++){
                     BuildPlan req = lineRequests.get(i);
-                    if(i == lineRequests.size - 1 && req.block.rotate){
+                    if(i == lineRequests.size - 1 && req.block.rotate && req.block.rotateArrow){
                         drawArrow(block, req.x, req.y, req.rotation);
                     }
                     drawRequest(lineRequests.get(i));
                 }
             }else if(isPlacing()){
-                if(block.rotate){
+                if(block.rotate && block.rotateArrow){
                     drawArrow(block, cursorX, cursorY, rotation);
                 }
                 Draw.color();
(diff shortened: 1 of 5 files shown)


sk7725:
Yes, these are quite specific use cases, but surely some mods will benefit from this, even more that there is no other easy way of achieving this.

Anuken:
Can you provide an example of both types of fields being used in real blocks?

sk7725:
These are fields purely for modding - if you are asking for modding examples, for the first field, an example could be Pistons from https://github.com/sk7725/Commandblocks which uses placeholder blocks as the piston arm once it is extended(to disallow block placement and add collision to the piston arm), which obviously should not be rotated once it is extended. Also, nearly any block that only calculates stuff in placed() and not updateTile() can serve as an example. 
For the second field, these rotatable mirrors could serve as an example:
image

sk7725:
Maybe other modders could come up with other examples...

core/src/mindustry/input/DesktopInput.java
```diff
@@ -306,7 +306,7 @@ public void update(){
                 cursorType = ui.unloadCursor;
             }
 
-            if(cursor.build != null && cursor.interactable(player.team()) && !isPlacing() && Math.abs(Core.input.axisTap(Binding.rotate)) > 0 && Core.input.keyDown(Binding.rotateplaced) && cursor.block().rotate){
+            if(cursor.build != null && cursor.interactable(player.team()) && !isPlacing() && Math.abs(Core.input.axisTap(Binding.rotate)) > 0 && Core.input.keyDown(Binding.rotateplaced) && cursor.block().rotate && cursor.block().quickRotate){
```
Anuken, line 309: This prevents rotation from happening, but it still draws the rotation selection over the block when you hold the rotate button.

Anuken:
> should not be rotated

This is reasonable, but...

> these rotatable mirrors could serve as an example

This is not. Arrows are generally an indication that a block can be rotated, and even if the block is not supposed to be directly pointing at the arrow (e.g. your mirror), I don't see why the arrow should be removed completely.

sk7725:
> This is not. Arrows are generally an indication that a block can be rotated, and even if the block is not supposed to be directly pointing at the arrow (e.g. your mirror), I don't see why the arrow should be removed completely.
It leaves room for mods to give it a new arrow or some icon that fits the block more. But I see your point, should I remove it then?

core/src/mindustry/input/DesktopInput.java
sk7725: Fixed!

Anuken:
> should I remove it then?

Yes.

sk7725:
Done.

PR #2539, opened 2020-09-11
Mud and Quicksand

ThePlayerA:
This pull adds mud and quicksand environmental floors. These floors slow units but don't damage them. These can provide mapmakers with a way to slow player units down without damaging them. They also add more aesthetic and functional variety to the environments of maps. Quicksand and mud provide a boost to water extractors due to their high moisture content. These floors can provide new options for building defenses as players can take advantage of the slowing effects.

demo


DeltaNedas:
good

RogueRedHawk:
Would love to see this be conditional to weather-- if it's raining, random patches of dirt or sand can turn to mud and quicksand (potentially specify "heavy rain" to have this effect"). them if you go back to a sector with no rain, those patches dry up.

genNAowl:
Here is a relevant suggestion? https://github.com/Anuken/Mindustry-Suggestions/issues/487

ThePlayerA:
> Here is a relevant suggestion? [Anuken/Mindustry-Suggestions#487](https://github.com/Anuken/Mindustry-Suggestions/issues/487)

Perhaps water units could travel through these blocks but rather than only travelling a short distance before breaking they can travel through these blocks normally but with a speed decrease? This could provide better versatility for naval units as they could travel on more terrains making them more useful for more maps. Then you could have a mudflats map for example using both naval and ground units for attack and defense. Also is the water extractor buff too much?

Anuken:
> Perhaps water units could travel through these blocks

This doesn't make sense to me. Shallow water, maybe, but mud is *practically solid*. You're not going to be moving a battleship through mud at any speed. Just use shallow water if you want both ground and naval coverage.

> quicksand

I don't think two blocks are necessary. Only having mud is enough for now.

The mud in your screenshot doesn't look like it uses any colors from the dirt palette, and it's practically a direct recolor. It would be faster for me to implement this myself than fix everything in this PR.

PR #2542, opened 2020-09-12
store building class with constructor

DeltaNedas:
Mods can use block.entityClass instead of needing to know the building too.
However, I'm not sure if rhino will be able to extend a Class<T> or if it needs a JavaClass which i have no idea how to get.

Use case:
`routerify(Blocks.router, Router.RouterBuild, "h", {}, {})`
becomes
`routerify(Blocks.router, "h", {}, {})`
and `block.entityClass`

I'll need to read some rhino docs though

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -197,7 +197,7 @@ public class Block extends UnlockableContent{
     public boolean instantTransfer = false;
 
     protected Prov<Building> entityType = null; //initialized later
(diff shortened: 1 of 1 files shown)


SummetDev:
~~what's `routerify`?~~

Anuken:
I'd rather not. If you really need the class, get it from the provided constructor. This adds an implicit requirement for anyone setting the entity to set the class as well.

PR #2547, opened 2020-09-12
Change to draw flushing

LordChrom:
Lets more draw flushes happen at once while also lowering memory usage a bit in most cases.

Before:
Before

After:
After

(Same program being run on processors, but draw flushes didn't all go thru before)


rodrigolitzius:
Leave a comment

imreallybadatnames:
ok

core/src/mindustry/logic/LExecutor.java
```diff
@@ -484,11 +484,12 @@ public void run(LExecutor exec){
...
                 LogicDisplayBuild d = (LogicDisplayBuild)build;
-                if(d.commands.size + exec.graphicsBuffer.size < maxDisplayBuffer){
-                    for(int i = 0; i < exec.graphicsBuffer.size; i++){
-                        d.commands.addLast(exec.graphicsBuffer.items[i]);
-                    }
-                }
+
+                //adds a copy of the graphics buffer, since the it can be changed before it gets drawn
+                LongSeq flush = Pools.obtain(LongSeq.class,LongSeq::new);
+                flush.addAll(exec.graphicsBuffer);
```
Anuken, line 490: You cannot simply add a sequence to the graphics buffer - if no rendering happens, e.g. the display is off-screen, this will lead to infinitely increasing memory usage.

LordChrom: Wait, why does it need to be buffered between the draw flush and the actual drawing anyway? I just tried it where updates everything immediately when it draw flushes and it seems to work fine from what I can tell, even if it like flushes lots of times at once every frame. Is there any particular reason why it needs to be queued?

Anuken: It's queued because there's no way of knowing what the graphics state will be in the *update* loop. Drawing anything (even inside a framebuffer) in the update loop is a bad idea and can lead to bizzare and non-obvious bugs.

LordChrom: I see

imreallybadatnames: if there was a way to check if the display is off-screen then you could just have stopped adding more items in the queue when the display is off-screen

Anuken: That's not viable.

Anuken: Also, I don't see why you're using a *list of lists* here, because it doesn't serve any purpose besides adding extra overhead to the flush system. This is equivalent to removing the display buffer check, just with extra steps. Did you benchmark this and confirm that it uses less memory?

LordChrom: It did show less usage in task manager with lots of displays. It also did the infinite buildup thing offscreen, though that should be hard to fix. The main point though is that it can show all of the flushes, since I was running into issues before. Simply increasing the graphics buffer enough to hold them all increased memory usage dramatically. This way the only memory allocated to the graphics buffer all contains buffered graphics, if done properly, it should only take up more space where previously it wouldn't've worked.

As for how to fix it the way I came up with is to not add anything if the display hasn't been drawn in 10 frames using Core.Graphics.getFrameID(), which prevents things from building up.

LordChrom: Though I should probably just make it limit the queue offscreen, for graphs and stuff.

Anuken: > It did show less usage in task manager with lots of displays

Unless your map was completely full of displays, I don't see how that could be attributed to logic. The max display buffer is (512 * 8) = 4KB per display. You would need at least a hundred displays (assuming extra overhead) for my system to register 1 MB used solely in logic displays - and memory usage supposedly *less* than 1MB would be very difficult to detect reliably even with Java debugging tools. The amount of memory taken up by a Java application and the amount it's actually using are two very different numbers.

imreallybadatnames:
nice

core/src/mindustry/logic/LExecutor.java
LordChrom: Yeah one of the test maps was full of displays so I could see it better, it was a few MB lower on average. the other was the one with lots of flushes. Either way the main point of the changes is that the display wasn't working right before unless I cranked up the max display buffer to 16k (which was very bad for memory usage). This shows the image properly and this definitely better than cranking it up.

imreallybadatnames:
also it could sometimes falsely skip so i think you can just change the if condition from `Core.graphics.getFrameId() - d.lastFrameUpdated < 10 || d.flushes.size < maxOffscreenFlushes` to `!(Core.graphics.getFrameId() - d.lastFrameUpdated > 10 && d.flushes.size < maxOffscreenFlushes)`

core/src/mindustry/logic/LExecutor.java
Anuken: That could just be because of the pre-allocation of the queue. Since the buffer size is low, it resizes it to 512 by default. When you set it to 16k, *every single display* will allocate 16k up front, which it shouldn't.

LordChrom: Yeah. Thats why I wanted to do something other than giving 16k to each display (which still doesnt 100% solve the issue)

LordChrom:
That's the same thing
frame diff | flushes | result
9               |15         | true
10             |15         | true
9               |16         | true
10             |16         | false

core/src/mindustry/logic/LExecutor.java
Anuken: But you can do that by making the starting buffer size 512 (as before, or even less) and setting the *max* buffer to 16k.

RandomTNT (approved):
This goddamm omega mega good

imreallybadatnames:
ok

Anuken:
I'm not going to be increasing the buffer size significantly, and I don't think this PR's implementation makes sense. Allowing some sort of raster texture rendering system that loads bytes and creates a GL texture directly would be a far better solution for the future.

PR #2552, opened 2020-09-14
Mud v2

ThePlayerA:
Revamp of mud pr. Removes naval unit illogicalness. Removes quicksand. New sprites that use colours from dirt palette – sprites in first picture are currently in use, I can change pr to second option if it better suits. This pr cleans up implementation of the new blocks. Hopefully this fixes all the errors in the previous pr!

image
image


RogueRedHawk:
Second option looks more "muddy" in my opinion. More wet, while the first one is drier. Just imo.

Anuken:
Merged.

PR #2559, opened 2020-09-16
Update servers_v6.json

BasedUser:
Move routerchain.ddns.net to v6 server list


Anuken:
1. The name is currently still set to "*BE* logic sandbox".
2. Any particular reason why turrets and units are banned?

PR #2562, opened 2020-09-16
+ commander

Quezler:


imreallybadatnames:
ok

Quezler:
well i was helping someone on [discord](https://discordapp.com/channels/391020510269669376/391020510269669378/755730495530074223) who wondered which mechs could command, and figured they only looked at the comments and not the actual code in the file, so this is for them 🙂

PR #2610, opened 2020-09-19
make techtree funcs public

DeltaNedas:
so mods can add item research, override entire tech tree, etc.
im pretty sure having research in the item json would be better but this will do for now


sk7725:
YES

PR #2630, opened 2020-09-20
Added "char" logic operation

maximmasterr:
Added "char" logic operation, that gets char in string by index, very usefull

Also if i may ask:
what about create class that represents num, if so, it will be easier to manage custom operations for example I also want to create string join op, but it requires one more function smth like: objFunction2Obj (that get two object and return obj) and objFunction2Num(that get two object and return num), it will be much easier to handle only objects and create special class that represents num, so all operations will consist of only two functions function1 (for 1 arg) and function2 (for 2 args)
if you like the idea, ask and i will refactor code

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -411,8 +411,10 @@ public class LExecutor{
             }else{
                 Var va = exec.vars[a];
                 Var vb = exec.vars[b];
-
-                if(op.objFunction2 != null && (va.isobj || vb.isobj)){
+                if(op.objNumFunction2 != null && (va.isobj && !vb.isobj)){
+                    //use objectNum function if provided and first arg is obj and second num 
+                    exec.setobj(dest, op.objNumFunction2.get(exec.obj(a), exec.num(b)));
+                }else if(op.objFunction2 != null && (va.isobj || vb.isobj)){
                     //use object function if provided, and one of the variables is an object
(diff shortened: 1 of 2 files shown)


Anuken:
> very useful

Useful for what purpose? You will not be able to do anything with that string. You shouldn't *need* it, because all strings are constants. You will never be getting a string as a result from an operation. Message blocks are write-only.

>  I also want to create string join op

Absolutely not. Strings **must** remain constants. I will not allow allocations in the logic system, and I have yet to see any good use for strings. If you want to display something in a message block, numbers and constant strings work just fine.

> what about create class that represents num

This is called `java.lang.Double`, and is an unacceptable way of handling numbers, because any boxing from a `double` to a `Double` **results in an allocation.**

maximmasterr:
Ok, I understand about allocations, but the way I want to use it for storong images i multiple strings as hex, so we will be able to draw 80x80 image in single processor instead of 7, and it will allow to do much more with it

Anuken:
> draw 80x80 image in single processor

Don't do this. It's not what logic is designed for.

maximmasterr:
For parsing hex I will just use get char compare it to 0123456789abcdef bit shift and do second time
If you worry about allocations, I can change pr to return double instead of String

Anuken:
Again, you should not do this. Logic is not designed to display raster images or parse strings in any way. If I ever end up supporting it, I'll add a dedicated way to upload textures instead of the hacky system you're proposing.

PR #2643, opened 2020-09-21
Make Pal.java colors changable for texturepack-mods tweaks

Remint32:
> I don't know about other people, but I would like to use these fields for modding.

I noticed this when I was re-texturing logical displays and when they were placed, they displayed the default base color on screen, not the texture color, because it is hardcoded.

I know that it is correct not to change the base classes and options, but to create your own in modifications and use them instead, but this is an exception.


Remint32:
oh yes, thank you

PR #2647, opened 2020-09-21
inconsistencies

BasedUser:
🔥

### DIFF
--- a/core/src/mindustry/content/UnitTypes.java
+++ b/core/src/mindustry/content/UnitTypes.java
@@ -10,6 +10,7 @@ import mindustry.entities.bullet.*;
 import mindustry.gen.*;
 import mindustry.graphics.*;
 import mindustry.type.*;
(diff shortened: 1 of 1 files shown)


core/src/mindustry/content/UnitTypes.java
```diff
@@ -10,6 +10,7 @@
 import mindustry.gen.*;
 import mindustry.graphics.*;
 import mindustry.type.*;
+import mindustry.Vars;
```
summetdev, line 13: why not `import static`?

BasedUser: good point, fixing to match

Anuken:
The reason this was written as `8` and not `tilesize` is because payloads include units as well, and the amount you can carry relates to their hitbox size, not `tilesize` specifically. Not a great reason, admittedly. 

I'll fix this myself.

PR #2648, opened 2020-09-21
Barrel system for turrets so you can set where and the rotation that bullets spawn from.

MEEPofFaith:
[Demonstration](https://youtu.be/bLPek-m7tZY)


MEEPofFaith:
[Now Applies to ChargeTurret](https://youtu.be/MeGuAUGDew0)

MEEPofFaith:
Since the LaserTurret works by updating its bullet constantly while firing, I don't think doing barrel stuff on it is possible, and I won't even try.

GlennFolker:
why not try, for loop exist

BlueWolf3682:
(the video isnt that helpful, us a larger turret like the Wolf's Teeth)

MEEPofFaith:
Note: I have no idea what the player and bullet .json things that changed are, they just happened on their own,

Also, Blue, no.

BlueWolf3682:
then use a larger turret, the barrel code for the swarmer is simply too hard to see cause the turret is small

also, meep, yes

MEEPofFaith:
I'm not gonna record another video, but if I somehow am forced to do it I'll use Spectre.

MEEPofFaith:
If I have a `Time.run` thing inside of a for loop, how can I use the variable in the for loop?
Because what I currently have doesn't work.
```java
for(int i = 0; i < shots; i++){
  Time.run(burstSpacing * i, () ->{
    blah blah, coded uses i
  }
}
```
Has error

MEEPofFaith:
[This is fun to mess with](https://youtu.be/jkaHf1BfqyE)

sk7725:
Increase in calculations is not a good idea at all.

summetdev:
In the future: don't edit the master branch, instead create a new branch and make a pull request.

MEEPofFaith:
> In the future: don't edit the master branch, instead create a new branch and make a pull request.

I accidentally created it on master. I guess when you directly edit a file and let it make the fork for you it automatically puts it on the Master branch instead of creating a master and a separate branch for the suggested edits. Well, at least any branches I make are based on the master branch of `Anuke/Mindustry` instead of `MEEPofFaith/Mindustry`.

MEEPofFaith:
Whoops, thought I was on `Anuken/Mindustry`, but instead I was on `MEEPofFaith/Mindustry`

MEEPofFaith:
Gone stale, closing

PR #2668, opened 2020-09-23
Move ECAN servers to the v6 server list & put all the servers from the BE server list to the v6 one

mse-k:
e


Anuken:
BE servers *are not* compatible with v6 servers.

PR #2675, opened 2020-09-23
Turrets Shadow

Voz-Duh:
In this pull i created shadow for turrets
shadow
shadow-1


Quezler:
potential problem: day-night weather cycle and "position" of the sun during the day, but it looks nice, gj 👍

sk7725:
why not use the unit-weapon-style round shadows?
For consistencies and also it will look less bad.

PR #2684, opened 2020-09-23
save game on android app exit

DeltaNedas:
should implement https://github.com/Anuken/Mindustry-Suggestions/issues/584
no idea if it'll work though


Anuken:
IIRC this runs in the wrong thread, and will probably lead to corruption and crashes.

DeltaNedas:
should it be app.post()ed?
main concern is that it functions in theory, anything else i need to do?

Anuken:
> should it be app.post()ed?

You can't, because the application is dead and probably won't do anything at that point.

Make sure it works, running on a simulator shouldn't be hard.

PR #2702, opened 2020-09-24
Added boss as radar target

LeoDog896:
This allows for the 6.0 radar logic block to target bosses


MEEPofFaith:
if boss in range
have all spectre target boss

successfully erased boss

PR #2704, opened 2020-09-24
remove .bat extension from README.md

summetdev:
all Windows batch files can be executed without extension


Remint32:
oh no i use bat scripts, why

PR #2707, opened 2020-09-24
Make renderer shader final

LeoDog896:
Seriously, it shouldn't be editable.


Anuken:
There are hundreds (if not thousands) of cases like this across the codebase, does it really matter?

LeoDog896:
No, not really, just in the case of purpose. Feel free to close it then.

Anuken:
In this case, it probably wouldn't do any harm, but in general I avoid `final` for many variables as it doesn't affect performance[1] and only leads to problems with modding in the future.


*[1]: yes I am aware that certain `static final` primitive/string fields are inlined, but those cases are quite rare*

sk7725:
*oh no no no no no no*

LeoDog896:
should we reverse the commit?

sk7725:
tbh I wanted to change that to *public* entirely

LeoDog896:
oH yeah feel free to edit it

LeoDog896:
or make a PR with that, I'm sure no one would mind it :D

Anuken:
> oh no no no no no no

Were you really thinking of touching the *IndexedRenderer* shader? What use could that serve?

sk7725:
> Were you really thinking of touching the _IndexedRenderer_ shader? What use could that serve?

Sorry, I guess I misread.

PR #2708, opened 2020-09-24
make logic dialog stuff public

DeltaNedas:
Also made BaseDialog#onResize public.

Most of these fields are table children which are trivially accessed for a specific version, but break easily and are very messy.


Anuken:
why?

DeltaNedas:
because i want to add a quick button to view logic manual in the processor dialog but onResize is protected, along with LCanvas' fields being package private

DeltaNedas:
o fuk

PR #2718, opened 2020-09-24
Make lightning have frag angle properties

MEEPofFaith:
Adds `lightningCone`, which is the full angle of where the lightning direction can be, and `lightningAngle` which is the angle compared to the bullet that the center of the lightning cone is.

### DIFF
--- a/core/src/mindustry/entities/bullet/BulletType.java
+++ b/core/src/mindustry/entities/bullet/BulletType.java
@@ -184,8 +184,7 @@ public abstract class BulletType extends Content{
         }
 
         for(int i = 0; i < lightning; i++){
-            float litA = b.rotation() + Mathf.range(lightningCone/2) + lightningAngle;
-            Lightning.create(b, lightningColor, lightningDamage < 0 ? damage : lightningDamage, b.x, b.y, litA, lightningLength + Mathf.random(lightningLengthRand));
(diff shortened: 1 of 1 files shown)


core/src/mindustry/entities/bullet/BulletType.java
```diff
@@ -181,7 +184,8 @@ public void hit(Bullet b, float x, float y){
         }
 
         for(int i = 0; i < lightning; i++){
-            Lightning.create(b, lightningColor, lightningDamage < 0 ? damage : lightningDamage, b.x, b.y, Mathf.random(360f), lightningLength + Mathf.random(lightningLengthRand));
+            float litA = b.rotation() + Mathf.range(lightningCone/2) + lightningAngle;
```
Anuken, line 187: This should go in the function itself for consistency. No extra variable.

MEEPofFaith: Is there a button for accepting suggested requests, or is it committed already?

Anuken: There is no suggested code change or commit. It's a verbal request.

MEEPofFaith: Ok

PR #2719, opened 2020-09-24
More Shadow

Voz-Duh:
In this pull i created shadow for more blocks
more-shadow


Voz-Duh:
I have not shown bridge conveyors

Anuken:
This makes sense for repair points and mass drivers, but for drills it doesn't look great. They look flatter, ironically.

Voz-Duh:
> This makes sense for repair points and mass drivers, but for drills it doesn't look great. They look flatter, ironically.

Ok 
_I too don't like it_

PR #2723, opened 2020-09-25
New liquid turret - Tsunami

Slava0135:
A lot of new units but only 2 new turrets. Wave is not particularly useful cos of it's range. It's still useful against fire but we need something better in late game.


Slava0135:
oopsie forgot to add it to the research tree

core/src/mindustry/content/Blocks.java
```diff
@@ -1458,6 +1458,26 @@ public void drawBase(Tile tile){}
             shootSound = Sounds.splash;
         }};
 
+        tsunami = new LiquidTurret("tsunami") {{
```
Quezler, line 1461:
```suggestion
        tsunami = new LiquidTurret("tsunami"){{
```

Anuken:
- This turret should be 3x3, not 4x4. 4x4 is reserved only for the highest tier of combat turrets, and the sprite is not detailed enough to fit in with other 4x4 turrets.
- Missing a `-top` sprite for the liquid.
- Should be near the other 3x3 turrets, not next to the wave in building menu.
- Too cheap, should require a higher tier material like thorium at least.

Slava0135:
Ok then. I'll try to fix the issues

Slava0135:
Everything should be fine now. It's fine to me but you always can modify it

Anuken:
Are you going to use EoD's sprite for this?

Slava0135:
Not sure really. The problem with my sprite is that it has more thin lines compared to other turrets.

Slava0135:
But I refused to add pipes to the sides as it looks weird to me.

Anuken:
I prefer EoD's version; I think it's slightly cleaner and more consistent with the Wave sprite.

Slava0135:
I feel the same but those pipes... idk about them

Anuken:
They're fine.

Slava0135:
Everything is ready I guess

Slava0135:
oh no forgot comma in Blocks

SnakkiZXZ:
as I see you cannot solve with the texture, so I specially drew this sprite for this, if there are 2 images.
tsunami
tsunami-liquid
Criticism is welcome.
P.S. you can add ```plastanium``` to craft, it is more expensive than ```titanium```.

SnakkiZXZ:
oh no i am late?

Slava0135:
too late but this looks more like a unit and not a turret

SnakkiZXZ:
oh no

core/src/mindustry/content/Blocks.java
```diff
@@ -1589,6 +1589,26 @@ public void drawBase(Tile tile){}
             reloadTime = 10f;
         }};
 
+        tsunami = new LiquidTurret("tsunami"){{
+            requirements(Category.turret, with(Items.metaglass, 100, Items.lead, 400, Items.titanium, 250, Items.thorium, 100));
+            ammo(
+                    Liquids.water, Bullets.waterShot,
```
Anuken, line 1595: Should use different bullet variants that deal more damage, draw larger and have more knockback.

core/src/mindustry/content/Blocks.java
```diff
@@ -1589,6 +1589,26 @@ public void drawBase(Tile tile){}
...
+                    Liquids.slag, Bullets.slagShot,
+                    Liquids.cryofluid, Bullets.cryoShot,
+                    Liquids.oil, Bullets.oilShot
+            );
+            size = 3;
+            recoilAmount = 0f;
+            reloadTime = 1f;
```
Anuken, line 1602: This won't work. The reload should be at least 2-3; turrets can only shoot once per frame, so any FPS below 60 would make this turret fire slower than it should.

Anuken (changes requested):

Slava0135:
Made the changes

Slava0135:
something went wrong and puddle/bullet size is so small

Slava0135:
nvm everything is fine after repacking. The stats are somewhat reasonable

PR #2727, opened 2020-09-25
Drown event

LeoDog896:
We all need to know when they drown, and it was on TODO. First time making an event, so, I would much prefer any feedback.


core/src/mindustry/entities/comp/FlyingComp.java
```diff
@@ -1,10 +1,12 @@
 package mindustry.entities.comp;
 
+import arc.Events;
```
summetdev, line 3:
```suggestion
import arc.*;
```

PR #2731, opened 2020-09-25
Change Toxopid's Name

Spfwniw:
This is not a necessary change, and I know you don't like necessary changes. I am also fine with the current name, but I think this would be better.

I changed Toxopid's name to Octopid. This would make more sense since "Octo" means 8 (Hence Octa-gon, but "Octo" actually) and Toxopid also has 8 legs. Other units may have 8 legs as well but I don't think there should be more than one unit with "octo" in their name (Btw there will be 3 units with "oct" in their name if this gets merged), also, I only could find a good name with "octo" in it for Toxopid (Random thought btw) and since It's a final tier unit and currently the strongest, Toxopid deserves this name most.

I looked at your commit "risse -> risso" and did similar things to Toxopid, i changed bundles.properties, icons.properties, Blocks.java and UnitTypes.java.

I didn't bother changing Toxopid in other languages because in the future they will be localised anyway, if it would be a problem currently I'd like someone to update them ASAP.

I may could have missed somethings, for example, in your "risse -> risso" commit there isnt a change in weapon name, but in mine there is and thats lucky because I remembered weapons also have name and I also expected one of them to have "toxopid" in their name, and there was, so I changed it. As said, I could've missed that, and I may already have missed somethings.

After all, I hope you appreciate this and I also hope this PR would be successfully merged. As said, not necessary but merging it shouldn't be hard and It'd be a good change.


Anuken:
"Octopid" sounds like a generic name for a spider unit to me. It's just too obvious and boring.

Toxopid was chosen as a shortened form of [Toxopidae](https://en.wikipedia.org/wiki/Toxopidae), and I'd rather it not be changed.

PR #2741, opened 2020-09-26
Mem UI

ThePythonGuy3:
Adds a simple UI to check the memory cell contents. [Image](https://discord.com/channels/391020510269669376/663853133587546138/759503028985790494)

### DIFF
--- a/core/src/mindustry/world/blocks/logic/MemoryBlock.java
+++ b/core/src/mindustry/world/blocks/logic/MemoryBlock.java
@@ -27,7 +27,7 @@ public class MemoryBlock extends Block{
         stats.add(BlockStat.memoryCapacity, memoryCapacity, StatUnit.none);
     }
 
-    public class MemoryBuild extends Building {
+    public class MemoryBuild extends Building{
         public double[] memory = new double[memoryCapacity];
 
         @Override
--- a/core/src/mindustry/world/blocks/logic/MemoryBlock.java
+++ b/core/src/mindustry/world/blocks/logic/MemoryBlock.java
@@ -1,15 +1,12 @@
 package mindustry.world.blocks.logic;
 
-import arc.scene.ui.layout.Table;
-import arc.util.Align;
+import arc.scene.ui.layout.*;
+import arc.util.*;
 import arc.util.io.*;
 import mindustry.gen.*;
-import mindustry.ui.Styles;
-import mindustry.ui.dialogs.BaseDialog;
+import mindustry.ui.*;
 import mindustry.world.*;
 import mindustry.world.meta.*;
-
-import java.awt.*;
 
 public class MemoryBlock extends Block{
     public int memoryCapacity = 32;
@@ -28,7 +25,7 @@ public class MemoryBlock extends Block{
         stats.add(BlockStat.memoryCapacity, memoryCapacity, StatUnit.none);
     }
 
-    public class MemoryBuild extends Building implements mindustry.world.blocks.logic.MemoryBuild {
+    public class MemoryBuild extends Building {
         public double[] memory = new double[memoryCapacity];
 
         @Override
--- a/core/src/mindustry/world/blocks/logic/MemoryBlock.java
+++ b/core/src/mindustry/world/blocks/logic/MemoryBlock.java
@@ -1,9 +1,15 @@
 package mindustry.world.blocks.logic;
 
+import arc.scene.ui.layout.Table;
+import arc.util.Align;
 import arc.util.io.*;
 import mindustry.gen.*;
+import mindustry.ui.Styles;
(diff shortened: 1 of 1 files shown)


ThePythonGuy3:
~~Yes i love messing with ui stuff~~

core/src/mindustry/world/blocks/logic/MemoryBlock.java
```diff
@@ -1,17 +1,24 @@
 package mindustry.world.blocks.logic;
 
+import arc.scene.ui.layout.Table;
+import arc.util.Align;
 import arc.util.io.*;
 import mindustry.gen.*;
+import mindustry.ui.Styles;
+import mindustry.ui.dialogs.BaseDialog;
```
summetdev, line 8: Use `.*` imports.

core/src/mindustry/world/blocks/logic/MemoryBlock.java
```diff
@@ -1,17 +1,24 @@
...
 import mindustry.world.*;
 import mindustry.world.meta.*;
 
+import java.awt.*;
```
summetdev, line 12: Don't use [incompatible Java features](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md#do-not-use-incompatible-java-features-javautilfunction-javaawt).

core/src/mindustry/world/blocks/logic/MemoryBlock.java
```diff
@@ -21,9 +28,21 @@ public void setStats(){
         stats.add(BlockStat.memoryCapacity, memoryCapacity, StatUnit.none);
     }
 
-    public class MemoryBuild extends Building{
+    public class MemoryBuild extends Building implements mindustry.world.blocks.logic.MemoryBuild {
```
summetdev, line 31: Why not import it?

ThePythonGuy3: what i dont get it

core/src/mindustry/world/blocks/logic/MemoryBlock.java
ThePythonGuy3: i absolutely have no idea on why Intellij goes so mad

ThePythonGuy3:
Fixed

core/src/mindustry/world/blocks/logic/MemoryBlock.java
```diff
@@ -21,9 +25,21 @@ public void setStats(){
         stats.add(BlockStat.memoryCapacity, memoryCapacity, StatUnit.none);
     }
 
-    public class MemoryBuild extends Building{
+    public class MemoryBuild extends Building {
```
l-Luna, line 28: why add a space

l-Luna (approved):

core/src/mindustry/world/blocks/logic/MemoryBlock.java
summetdev: ~~Anuke doesn't like spaces.~~

DeltaNedas:
deltanedas/ldb stolen

core/src/mindustry/world/blocks/logic/MemoryBlock.java
ThePythonGuy3: No idea

Anuken:
There are several problems with this:
- It's not dynamic, which means any changes to memory while configuring would not be reflected
  - Making it dynamic would mean spewing garbage everywhere and lagging even more
- It's not very easy to use
- It's going to lag, because 500+ labels in a row is not great for mobile performance, especially considering they're set up every time you click on a memory block
- Is it really necessary? If you really want to debug, why not get and print?

sk7725:
Add a "refresh" button at the top.
This solves:
> It's going to lag, because 500+ labels in a row is not great for mobile performance, especially considering they're set up every time you click on a memory block
And also implies than you have to refresh it yourself in a simple way:
> It's not dynamic, which means any changes to memory while configuring would not be reflected
Also maybe a dedicated FloatingDialog will suit better than just a table.

ThePythonGuy3:
hmmm i will take a look at those points

ThePythonGuy3:
@Anuken now i fixed the things you and others pointed out. [Final result](https://discord.com/channels/391020510269669376/663853133587546138/761301438981996564)

Anuken:
> Add a "refresh" button at the top.
This solves:

Well, no, it doesn't, because you still have to set the labels up and check their culling every frame.

The only valid implementation for this that I would accept is a custom element that:
- doesn't use any elements, only one display
- directly draws using the font
- only displays the numbers in range of the screen by calculating the culling range (nothing outside the range is updated, created or drawn)

PR #2747, opened 2020-09-27
Make oct ACTUALLY have the most durability of any T5 unit

z0mbiesrock:
image
Not true; Reign has the same amount of health. This fixes that


Anuken:
> Not true;
Wrong. Notice that I did not say *health*, I said *durability*. Oct has more armor and a regenerating shield, which the Reign does not.

PR #2749, opened 2020-09-27
Add a keyPress operation to the Jump logic statement

ThePythonGuy3:
Pretty self-explanatory. Jumps if the <value> index key (KeyCode indexes) is pressed.

Test Schematic:
> bXNjaAF4nGPgYOBhZGDJS8xNZWApSS0uYWbgSc4vStXNK03OSS0tZuDPzUwuytctKMpPTi0uzi9i4M3JT89M1k3JLC7ISaxkYGBgZmBgAkIGRiCbmw9IZFfMSU5JSJgxKWhKxLoJcybPiliSkJK1KoFBtXECt8rUacFRcxZOmBQZM2FyROaJqaGzohckLJowN0oyYEpUzJxZixJS1iWkMAqaB3jo6vieCTQIMXhyyOBaQ8NCN49Zje+SUh40nDTe8Pc/M8OUWiNJoMVsDOwMDACu0UTf

The display goes red when you press the 0 key (not the numpad one)

### DIFF
--- a/core/src/mindustry/logic/ConditionOp.java
+++ b/core/src/mindustry/logic/ConditionOp.java
@@ -12,7 +12,6 @@ public enum ConditionOp{
     lessThanEq("<=", (a, b) -> a <= b, false),
     notEqual("not", (a, b) -> Math.abs(a - b) >= 0.000001, (a, b) -> a != b, false);
 
-
     public static final ConditionOp[] all = values();
 
     public final CondObjOpLambda objFunction;
@@ -30,7 +29,6 @@ public enum ConditionOp{
         this.objFunction = objFunction;
         this.alone = alone;
     }
-
 
     @Override
     public String toString(){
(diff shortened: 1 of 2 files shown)


Anuken:
This doesn't make any sense on a headless server or in multiplayer. Use in-game buttons for input.

ThePythonGuy3:
Imo this would be very useful if its limited to singleplayer, and i think a lot of people agree with me.

Legomonster33:
this seems unnecessary and relatively useless, if you need inputs from anywhere you can set up a network of switches (or some other form of input).

Anuken:
Everything in the logic system either uses built-in variables or gets state from the world it's in. Key presses have nothing to do with the world, they're client-side events that the logic system should have no way of knowing. *Logic is not scripting.*

DeltaNedas:
if you only want to use it in singleplayer, make it a dedicated inputs block (or leave as instruction) in its own mod

PR #2764, opened 2020-09-29
control rotation block <0-3>

DeltaNedas:
makes rotation not only sensable but controllable.
also refactored very ugly stuff in sensor code


DeltaNedas:
cat say no

PR #2765, opened 2020-09-29
Save last FileChooser's directory to settings.

summetdev:
I'm not sure about the name of the setting `lastDirectory`, if something is wrong correct me.


DeltaNedas:
oh yes

PR #2770, opened 2020-09-30
Autotiler docs and cleanup

LeoDog896:
Adds lots of docs, and cleans up the messy switch-like if statement code.

### DIFF
--- a/core/src/mindustry/world/blocks/Autotiler.java
+++ b/core/src/mindustry/world/blocks/Autotiler.java
@@ -14,20 +14,40 @@ import java.util.*;
 //TODO documentation
 public interface Autotiler{
 
-    //holds some static temporary variables, required due to some RoboVM bugs
+    /**
+     * Holds some static temporary variables, required due to some RoboVM bugs
+     */
     class AutotilerHolder{
         static final int[] blendresult = new int[5];
         static final BuildPlan[] directionals = new BuildPlan[4];
     }
 
-    /** slices a texture region:
-     * mode == 0 -> no slice
-     * mode == 1 -> bottom
-     * mode == 2 -> top */
-    default TextureRegion sliced(TextureRegion input, int mode){
-        return mode == 0 ? input : mode == 1 ? botHalf(input) : topHalf(input);
+    /**
+     * The mode to slice a texture at.
+     */
+    enum SliceMode{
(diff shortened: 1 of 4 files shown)


core/src/mindustry/world/blocks/Autotiler.java
```diff
@@ -112,24 +144,47 @@ default TextureRegion botHalf(TextureRegion input){
...
-            bits[0] = 2;
-        }else if(num == 3){
-            bits[0] = 2;
-            bits[2] = -1;
-        }else if(num == 4){
-            bits[0] = 1;
-            bits[2] = -1;
-        }else if(num == 5){
-            bits[0] = 1;
+        switch (num){
```
Anuken, line 154: I don't understand why people think `switch` is somehow less messy. This is *more lines of code*, and you need a `break` statement after every case. It's not any better than an `if`-chain.

core/src/mindustry/world/blocks/Autotiler.java
```diff
@@ -14,27 +14,53 @@
...
-     * mode == 0 -> no slice
-     * mode == 1 -> bottom
-     * mode == 2 -> top */
-    default TextureRegion sliced(TextureRegion input, int mode){
-        return mode == 0 ? input : mode == 1 ? botHalf(input) : topHalf(input);
+    /**
+     * The mode to slice a texture at.
+     */
+    enum SliceMode{
+        NO_SLICE,
```
Anuken, line 29: Never use `SCREAMING_SNAKE_CASE`, see [the guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md#follow-the-formatting-guidelines) for why.

Anuken (changes requested):

core/src/mindustry/world/blocks/Autotiler.java
LeoDog896: Really, it just matters as how its implemented throughout the code, and, after checking, this is how its used a lot of the time, so I'll revert it.

core/src/mindustry/world/blocks/Autotiler.java
LeoDog896: Ok, ill fix that now

PR #2771, opened 2020-09-30
Checks for JDK version 14

Slava0135:


summetdev:
Please no

Slava0135:
***why***

J-VdS:
it still works when using jdk 8

Slava0135:
oh no

Anuken:
> it still works when using jdk 8

no, it doesn't; this check was already implemented some time ago

PR #2772, opened 2020-09-30
Remove nearGround in favor of !nearLiquid

LeoDog896:
It was on TODO, so why not get that over with.


Anuken:
I added that TODO because I wasn't sure at the time; looking at it now, I can think of at least one significant situation where a block can be near both ground and liquid.

LeoDog896:
That is very true, I was a little confused but I rolled with it. So, all that's necessary to do is just do the same thing with ground then.

LeoDog896:
Now, this PR literally just removes the TODO since its pretty sure now.

PR #2775, opened 2020-09-30
Keybinds follow CONTRIBUTING

LeoDog896:
Contributing.MD referenced keybinds as one that didn't follow contributing, well now it does.


Anuken:
It's not that simple. Did you test:
- if it breaks translations
- if it breaks the english bundle
- if it breaks keybind saves from older versions

LeoDog896:
No longer breaks the bundles after testing, now to add legacy support

sk7725:
Great, now 5.0 JS mods are now butchered even more, and 6.0 mods are also affected.

Is this *really* that necessary?

Anuken:
Closing due to stale-ness; adding legacy support is a lot of work and will either involve very specific string manipulation upon load, or a large translation table. I don't really think either are worth it.

PR #2778, opened 2020-09-30
Added descriptions for the Tsunami, Parallax and Foreshadow turrets. Also renamed the "Oct" Unit to the octo unit

Colonizor48:
Idk why it says i changed so many things. If someone can tell me why it says that it would be Appreicated. This is one of my first pull requests for an actually big project so please feel free to give me any feedback. Also i don't speak any other languages so if anyone could translate my new lines into other languages it would be good.


Anuken:
Descriptions: I'd rather write these myself when necessary.
Name change: *Why?* There is no need to change the name.

Colonizor48:
I thought the octo being called the oct was a bug because i remember it being called the octo in earlier bleeding edge builds. But it might have not been called that.

Colonizor48:
If you want i can make another pull request with just the descriptions. But if not that is fine.

Anuken:
> because i remember it being called the octo in earlier bleeding edge builds

Sounds like a false memory. Never happened.

> If you want i can make another pull request with just the descriptions

As I said, I'd rather write these myself when necessary.

Colonizor48:
Understandable. Have a nice day

PR #2784, opened 2020-10-01
Tint tsunami pipes

Quezler:
> (partial) implementation of https://github.com/Anuken/Mindustry-Suggestions/issues/655

Screen Shot 2020-10-01 at 11 07 16

`added the top region with a 50% transparency `

(wave left unaltered since imo it looks strange)


Quezler:
so apparently github decided that when i manually had to make a new patch branch to use the master branch of my fork that hasn't been synced for over like 3k commits, hence the merge conflict, not sure how i can cleanly resolve it via the web interface while keeping the branch 👀

Quezler:
looks like it got fixed, but i'm not sure if merging it would not pollute the git history with a humongous merge, i would recommend squashing this pull request or even just adding it manually and closing it 🤔

SnakkiZXZ:
I think it looks strange, if you add it, then you should add it to the `wave`, and if we don't add it to the `wave`, then you shouldn't add it here either

Quezler:
i'll let anuke decide that, if he wants the wave to have it as well, all he needs to do is ask, it just looks better on here imo 🙂

Anuken:
This doesn't look good, in my opinion.

EyeOfDarkness:
looks strange.

Anuken:
...Nah.

PR #1508, opened 2020-02-04
[formatting] add newline to file(s)

Quezler:
> tries to add newlines to please the github diff, not sure if this does not break anything 😗 

Screen Shot 2020-02-04 at 10 45 45
> pull includes the script i used to achieve it, its filter could use improvements 👀

PR #2785, opened 2020-10-01
Run newline script from #1508 again

Quezler:
`sh ./tools/run-newline` (#1508)


Anuken:
This changes a lot of stuff that doesn't really matter, like entity field saves.
It also seems to break schematics, which are mostly binary data.

PR #2792, opened 2020-10-01
3x3 Router: Splitter

z0mbiesrock:
A larger (and faster) router. Useful for late-game distribution.

Knowing my luck with PRs, I'm certain you're going to reject it...


Anuken:
> Knowing my luck with PRs, I'm certain you're going to reject it...

Yep. This is unnecessary. We do not need even more transportation blocks that don't do anything new.

Spfwniw:
> Yep. This is unnecessary. We do not need even more transportation blocks that don't do anything new.

merging a PR shouldn't be hard, therefore, extra things would harm the game.

Anuken:
merging a PR that deletes everything in the repo would not be hard, therefore it would not harm the game

flawless logic

DeltaNedas:
this belongs in a bad version of routorio where there are 3x3 to 16x16 routers

z0mbiesrock:
> this belongs in a bad version of routorio where there are 3x3 to 16x16 routers

no

sk7725:
> > this belongs in a bad version of routorio where there are 3x3 to 16x16 routers
> 
> no

yes, mods exist for a reason

Spfwniw:
> merging a PR that deletes everything in the repo would not be hard, therefore it would not harm the game
> 
> flawless logic

What do you mean? It would harm but this doesn't.

Esnardo:
I don't think we need this. it would be better in most cases to just use several distributors. What i would like to see is some capacity (like 5 items) in distributors

MEEPofFaith:
Distributors already split into 7, we don't need something to split to 11.

PR #927, opened 2019-10-28
🌡 Silicon furnace

Quezler:
Screen Shot 2019-10-28 at 09 40 38

Silicon setups don't scale very well once you need one or more blue belts of titanium 😗 
(and a plethora of other reasons)

This (obviously a draft) pull request adds a `multi-press` equivalent for the silicon smelter 😇 

But instead of requiring water it uses heat from its environment:

- source 1: magma tiles, up to roughly 2/3rd efficiency
- source 2: overheating thorium reactor, up to 3/3 efficiency

(the 2 sources are summed together for the final heat, but the reactor won't heat if its heat is less than or equal to what the magma tiles already provide)

Reasoning behind is that it would be to easy otherwise, these requirements demand some more planing/thought/placement 🔬 

- source 1: magma tiles are often sparse, and a 3x3 square of the _good_ ones tend to be rare, do note that the machine also works on less than perfect tiles, as long as at least something gives heat.
- source 2: having to create a controlled environment that slowly feeds thorium (or limits coolant) to get the reactor to overheat anywhere from 0 to 100% 

So the challenge this block provides is that you'd need to setup an overheating setup, the closer to annihilation the faster the machine becomes, so high risk high reward **or** if you're lucky the map will provide some heat 🏅 

Or of course use the normal ones as a fallback, but lose space in comparison, this would be neat in tight maps 🐁 

Oh and the `i/o` is the same as the smaller one, the batches are just larger, since this is just intended to safe space, but the subject is open for balancing ⚖️ 

Demo of several setups, note that i had disabled thorium reactor explosions locally, 100% efficiency is not exactly safe, but even the lesser speeds (and even more so with phase'd overdrivers) should be sufficient 🤔 
Oct-28-2019 11-39-34
[sprite source](https://github.com/Anuken/Mindustry/files/3778332/gimp-silicon-furnace.xcf.zip) 🖌[gimp](https://www.gimp.org/)

PR #2800, opened 2020-10-02
new factory: plastanium-polypress

SnakkiZXZ:
Now we are starting to use `plastanium` on a massive scale, and a conventional plastanium plant may not be enough for such consumption volumes. So I made an improved plastanium plant, it works like a silicon crucible. But instead of heat, it requires `oil`, and `sand` as an additional material.
Screenshot_20201002_145925

I do factories for the first time, so I do not know all the tricks and there may be mistakes. Criticism is welcome. 

P.s. If not turrets, then factories. Some already wanted to petition those turrets to make XD


unjown:
Noice. We definetly need this

walksanatora:
That feeling when quezler says to checkout aomeones pr

BlueWolf3682:
y e s

core/src/mindustry/content/TechTree.java
```diff
@@ -162,6 +162,7 @@ public void load(){
 
                                             node(Items.plastanium, with(Items.titanium, 10000, Items.silicon, 10000), () -> {
                                                 node(plastaniumCompressor, () -> {
+                                                    node(plastaniumPress, () -> {});
```
Quezler, line 165:
```suggestion
                                                    node(plastaniumPress, () -> {
                                                    
                                                    });
```

Quezler (changes requested):
the emoji for it is missing:
https://github.com/Anuken/Mindustry/blob/master/core/assets/icons/icons.properties
(file auto generates when packing sprites, i would recommend not adding it by hand)

Quezler:
also, the naming, both tiers of graphite factories use the word `press`  so `plastanium press` might sound a bit off key 🤔

Anuken:
- This doesn't compile. **Test your PRs before submitting them!**
- Why does it use oil as an attribute booster? For the crucible, it made *some* sense: more external heat supplied could mean better efficiency. But for this block, it doesn't: the "oil" attribute is meant for oil extractors, which collect oil from below the surface. Making this block be faster on "oily" blocks doesn't make much sense, especially considering it already takes oil as an input.
-  Sand as a input seems a bit generic, and too similar to the oil extractor. Why sand?

l-Luna:
name suggestion?: plastanium polypress (based on plastics being polymers)
could require water or cryofluid as an extra material, mirroring how real-life polymerisation requires extensive cooling.
similarly, could gain from being placed in water.

genNAowl:
How about Plastanium Condensor?

genNAowl:
Interestingly enough, It doesn't seem like you need a new attributeCrafter class, it is possible to use AttributeSmelter and flameColor = Color.clear;

Colonizor48:
If u fix this this would probably make it in

unjown:
@SnakkiZXZ fix it first cuz i see that Anuke is interested. Btw ya we need a nee plast fac

SnakkiZXZ:
I renamed the block, made some edits. 
Why exactly did I use sand and oil? everything is simple, according to my idea, inside this plant there is a small drilling rig that extracts a large amount of oil, and sand is needed just for mining. Now I use water as a cooling water to prevent overheating and oil explosion, now they cannot be used on hot springs to avoid fire. And I use graphite as catalysis

SnakkiZXZ:
I cannot understand the reason for the compilation error, everything works for me

Anuken:
> I cannot understand the reason for the compilation error, everything works for me

That seems very unlikely. You have a class with an incorrect name, check the CI logs.

> plastanium-compositor

This is visually very similar to "compressor" and doesn't make sense as a name.

image

SnakkiZXZ:
hmm

SnakkiZXZ:
everything should work correctly now

SnakkiZXZ:
to be honest, I would make improved types of all standard factories, but some factories run fast enough not to improve them. for example `kiln`

MEEPofFaith:
Why does this thing drill oil itself? _It's already getting oil input._

SnakkiZXZ:
@MEEPofFaith it doesn't matter anymore, I changed the attribute

Quezler:
polymerizer sounds like a mouthful 👀

core/src/mindustry/content/Blocks.java
```diff
@@ -43,9 +43,9 @@
 
     //ores
     oreCopper, oreLead, oreScrap, oreCoal, oreTitanium, oreThorium,
-
+
```
Quezler, line 46:
```suggestion
 
```

core/src/mindustry/content/TechTree.java
```diff
@@ -162,6 +162,9 @@ public void load(){
 
                                             node(Items.plastanium, with(Items.titanium, 10000, Items.silicon, 10000), () -> {
                                                 node(plastaniumCompressor, () -> {
+                           
+                                                node(plastaniumPolymerizer, () -> {
+                                                                                    });
```
Quezler, line 167:
```suggestion
                                                node(plastaniumPolymerizer, () -> {
                                                
                                                });
```

core/src/mindustry/content/Blocks.java
SnakkiZXZ: what?

Quezler: you added a few spaces on a line that was empty, if you accept this comment it reverts it back to an empty line 🙂

core/src/mindustry/content/TechTree.java
Quezler: its not resolved 👀

Quezler (changes requested):

unjown:
Yey, rhanx anuke for planning to add this.

core/src/mindustry/content/Blocks.java
SnakkiZXZ: my mistake, it came out by accident 😅

Quezler:
@SnakkiZXZ instead of pressing `mark as resolved` push the button that actually implements the suggested changes 🙄

SnakkiZXZ:
@Quezler sorry, I still don't know much about github and didn't know it. But thanks for saying, I will definitely remember.

Catchears:
I don't know if it's needed but I made a version with the same sprite, but with different block stats and without the attribute. It's called the Plastanium Polypress as suggested above by @l-Luna. It doesn't need an extra resource, but why should it? Surely larger versions of factories which just increase efficiency are also okay sometimes. If @Anuken thinks that an extra resource is necessary, then I'd be happy to add one. If you think that my PR is useless and that this one should be used instead feel free to react to this with a thumbs down but I thought I'd offer it anyway. (I'm also happy to actually implement good suggested changes)

[Here's the link.](https://github.com/Anuken/Mindustry/pull/2818)

Catchears:
It still fails while packing.

Quezler:
with what error @Catchears ?

Catchears:
Well you see the texture is called "plastanium press" but in blocks.java it is called plastanium-polywhatever

That ain't gonna work

unjown:
noice. :)

SnakkiZXZ:
oh, as if I completely forgot to rename the texture

SnakkiZXZ:
I have some problems now, for some reason I can't add commits. I am already looking for solutions

SnakkiZXZ:
@Quezler due to the fact that I accepted your changes but did not indicate them in my files, I got a `nonfastforvard` error

unjown:
@Catchears hoestly we need this 
im not saying that urs is useless. but all we need is a new plast dource with the increasing demand

core/src/mindustry/content/TechTree.java
```diff
@@ -162,6 +162,10 @@ public void load(){
 
                                             node(Items.plastanium, with(Items.titanium, 10000, Items.silicon, 10000), () -> {
                                                 node(plastaniumCompressor, () -> {
+                           
+                                                node(plastaniumPolymerizer, () -> {
+                                                
+                                                });
```
Quezler, line 168: the indentation was wrong:
```suggestion
                                                    node(plastaniumPolymerizer, () -> {
                                                
                                                    });
```

SnakkiZXZ:
the problem is solved, I can edit the branch again

core/src/mindustry/content/Blocks.java
```diff
@@ -536,6 +536,27 @@ public void drawBase(Tile tile){}
...
+            itemCapacity = 25;
+            boostScale = 0.20f;
+            liquidCapacity = 135f;
+            attribute = Attribute.water;
+
+            consumes.items(new ItemStack(Items.titanium, 4), new ItemStack(Items.graphite, 2));
+            consumes.liquid(Liquids.oil, 0.70f);
+            consumes.power(4.5f);
+            
+        }};
```
Quezler, line 559:
```suggestion
            consumes.power(4.5f);
        }};
```

SnakkiZXZ: please no. after that I get errors. I will do it myself in order to avoid numerous and difficult mistakes.

Quezler: _the art of pulling, then pushing_

MEEPofFaith: ( ͡° ͜ʖ ͡°)

SnakkiZXZ:
Mr. Anuke, I apologize for my impatience, but maybe we should add this to the game before the release of build 108? I just can't figure out what you are waiting for, maybe you only want to add it to build 109?

unjown:
@SnakkiZXZ noice 
but ya Anuke mee too I also want it now :smile:

Anuken:
I'm just not particularly happy with the requirements this block has at the moment. Every other block already uses water as an input (optional or not). I may try and implement this myself using the sprites in the PR.

unjown:
This uses oil :smile:

SnakkiZXZ:
@Anuken And what do you suggest? what is used instead of the water attribute? spore? lights?

unjown:
> @Anuken And what do you suggest? what is used instead of the water attribute? spores? lights?

spore maybe cuz i never saw anything uses spore as a booster

SnakkiZXZ:
@unjown spores using cultivators, didn't you know that?

Colonizor48:
> 
> 
> I'm just not particularly happy with the requirements this block has at the moment. Every other block already uses water as an input (optional or not). I may try and implement this myself using the sprites in the PR.

If every other block already uses water why shouldn't this one?

Anuken:
> If every other block already uses water why shouldn't this one?

because that's boring

unjown:
> @unjown spores using cultivators, didn't you know that?

ik what i mean is optional booster
but spores is good here

unjown:
> > I'm just not particularly happy with the requirements this block has at the moment. Every other block already uses water as an input (optional or not). I may try and implement this myself using the sprites in the PR.
> 
> If every other block already uses water why shouldn't this one?

> > If every other block already uses water why shouldn't this one?
> 
> because that's boring

make sense

SnakkiZXZ:
I can remove the attribute, but it seems to me that although it is boring, it is correct that water is used. After all, it is used not just as a liquid, it is used as an attribute and this is far from all

unjown:
> I can remove the attribute, but it seems to me that although it is boring, it is correct that water is used. After all, it is used not just as a liquid, it is used as an attribute. And the liquid attribute is not used everywhere. It seems that only the pump is using it.

spores pls

SnakkiZXZ:
@unjown why spores? this is not logical because they are not used when creating plastanium.

unjown:
> @unjown why spores? this is not logical because they are not used when creating plastanium.

Why not cuz only spore presses uses spore so why not spore 
btw im reffering for T2 plast

SnakkiZXZ:
then it turns out that there is a mini factory inside the plant

unjown:
> then it turns out that there is a mini factory inside the plant

Kinda

l-Luna:
probably just cryofluid input?

Colonizor48:
Just make it use either water or cryofluid and call it a day please. We need this in the game soon.

Colonizor48:
> 
> 
> > If every other block already uses water why shouldn't this one?
> 
> because that's boring

You could always just have it use water as a placeholder and think of something else(or not) later

unjown:
why not 
titanium,sand,oil,cryo(as coolant)

SnakkiZXZ:
Let's decide something, this request has been here for too long. @Anuken, it's time for you to honor requests, close or add the oldest ones.

Quezler:
don't force anuke to do something

SnakkiZXZ:
well, I mean that something needs to be decided.

Quezler:
@SnakkiZXZ lets not put pressure on @Anuken to force him to do anything, its normal for pull requests to stay open for long:

just look at the plastanium conveyor from pull #1260, it stayed open for four months before being merged into master,
or the silicon furnace from #927, coined even earlier and implemented by anuke even later,
meanwhile this pull request is just slightly older than a week and you expect anuke to add everything into his game or force him to spend time using or closing it asap?

my recommendation is leave this pull request to ferment, anuke has already showed interest in using the sprites for it and indicated he would implement the java for it himself if it were to be added at all, bugging him by constantly asking for it only lowers the chance that it will ever make it into the game, so please just stay patient and play the waiting game.

SnakkiZXZ:
really, okay, I won't make excuses and make a big speech. I'll just say: I'm sorry I hurried.

A5TR0spud:
Please rename to polypress, playing on poly as in polymer and poly as in multiple.

unjown:
Any idea for input?
I already have.

SnakkiZXZ:
the word press is already used in both types of graphite factories, as Quezler has already spoken about

Quezler:
@A5TR0spud @SnakkiZXZ actually i really like the sound of `plastanium polypress` ❤️ 
(the usage of the word `press` here matters less since its part of a bigger word 🤔)

stuffyAI:
In the campaign I can comfortably get along using the normal Plastanium Compressor, as opposed to the silicon crucible which is quite necessary to squeeze the most out of sand. I really don't run out of it and even use it as my preferred ammunition (my only, in fact).

Now given that neither is set in stone I might change my opinion later, but do we really require a higher tier Plastanium factory?
I know that multiplayer can sometimes be tricky with certain players being creative, but even there I can imagine this to be a hazard when (trying to be) used by them, so I don't see the point there either.

SnakkiZXZ:
As you requested, I renamed `Plastanium Polypress`

the-kido (approved):

Dinistor:
I suggest get rid of sand input. it's used in oil extractor for reason, there's no need to use it. Why don't we make it work more efficient, but requre cooling for press parts? Like multipress. My version looks like this (maybe too op): 
Production time: 1.5 seconds
Input: 4 titanium, 16 criofluid, 32oil/sec
Output: 3 plastanium
Power consumption: 300 power/sec
Build cost: 50 plastanium, 40 metaglass, 200 silicon, 250 lead, 180 titanium, 100 graphite.
Description: Tier 2 plastanium compressor. Increased efficiency, but requres cooling.

HaphLife:
I think that this block should use cryofluid as the augment instead of sand because A) It's a tier two material (requires one factory interaction) like pyratite, B) having a block that requires two different fluid inputs would be interesting, and C) it makes more sense, since cryofluid is a titanium-water mixture, and plastanium is made from titanium.

SnakkiZXZ:
many do not like the idea of ​​consuming two liquids, there used to be such a block before, it was not the best. I do not know if this is necessary, because I have already installed cooling from the cold plates, and this is quite logical, because the work capacity of the plant decreases on hot plates. If others also decide that a second fluid is needed, I will make it.

Anuken:
After thinking about it for a while, I've come to the conclusion that this isn't necessary right now. I rarely see significant plastanium deficiencies on servers. Even *silicon*, a resource that is always in demand, isn't produced in 3x3 smelters very often.

PR #2807, opened 2020-10-03
Define payload capacity of payload conveyors on the conveyor itself.

MEEPofFaith:
`payloadLimit` defines the size limit a payload conveyor can carry. 2.5 is the default.


BasedUser:
2x2 blocks have a size of 4.

PR #2810, opened 2020-10-03
make backRegion nullable

DeltaNedas:
null checker broke routorio h


DeltaNedas:
@Anuken

Anuken:
This shouldn't be nullable.

> @Anuken

I get notifications for every PR and every commit anyway.

DeltaNedas:
why did this break turrets and units in every mod then

DeltaNedas:
actually should i just make it default to clear and be normal

PR #2813, opened 2020-10-03
make bullet's backRegion default to clear

DeltaNedas:
fix every mod


Anuken:
`Core.atlas.find` can't return `null`. This won't fix it.

DeltaNedas:
oh right

PR #2818, opened 2020-10-03
Alternative Plastanium Polypress (idea from SnakkiZXZ)

Catchears:
Here is my version of the upgraded plastanium compressor. Idea and textures by [SnakkiZXZ](https://github.com/SnakkiZXZ). As I said in my comment in their PR, this may not be needed / wanted and in that case will be promptly closed.  I also changed the blocks stats and removed the attribute. 

[Link to Snakki's PR](https://github.com/Anuken/Mindustry/pull/2800)


SnakkiZXZ:
I believe that every improved type of plant should consume additional resources, no matter whether it is a liquid or an object. This is more logical and helps balance the game without making it easier to obtain any resource.

Catchears:
In that case I'll add the need for water as suggested by I-Luna.

SnakkiZXZ:
In this case, your version will not be very different from mine, since I added the need for cold tiles, and the need for water itself, in addition to oil, requires a new class of factories operating on 2 liquids.

Catchears:
Turns out that that's true. I'll close this now, seeing as they are now, as you said, similar. However, I will suggest these things to you:
- use the same order for things in blocks.java as every other block 
- consider scaling the liquidcapacity with block area not with the length of one side (135f instead of 90f)

PR #2819, opened 2020-10-03
Simple fix for delayed weapons.

ThePythonGuy3:
Makes the shootSound sound when it *actually* shoots, not when the shoot sequence starts. (PS: sorry for 3 PRs in an hour, i feel somehow inspired today)

### DIFF
--- a/core/src/mindustry/entities/comp/WeaponsComp.java
+++ b/core/src/mindustry/entities/comp/WeaponsComp.java
@@ -167,8 +167,6 @@ abstract class WeaponsComp implements Teamc, Posc, Rotc, Velc, Statusc{
 
         float baseX = this.x, baseY = this.y;
 
-        weapon.shootSound.at(x, y, Mathf.random(0.8f, 1.0f));
-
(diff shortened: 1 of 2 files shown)


Anuken:
For delayed weapons, the idea is that you have a sound effect with a "charge-up" sequence so it matches the charging of the bullet.

ThePythonGuy3:
oh okay. what about 2 sounds?

ThePythonGuy3:
@Anuken maybe you like it now?

Anuken:
This breaks compatibility. Weapons without a delay will use the charge sound.

ThePythonGuy3:
oh true, missed one thing, sorry

ThePythonGuy3:
now its all compatible, and as long as they dont override load completely (and if they do they *should* know how to add chargeSound) just with shootSound is enough.

ThePythonGuy3:
@Anuken if this is way off what you had in mind just let me know

Anuken:
I'll just do this myself.

PR #2828, opened 2020-10-04
tier 2 cryofluid mixer

BasedUser:
To accomodate the new tier 4 and tier 5 unit factories' cryofluid usage, I've made an upgrade to the cryofluid mixer which involves a rotator accelerating titanium processing. This block is intended to use a large amount of power and to have a less efficient; but higher volume output, inheriting a warmup mechanic close to the impact reactor's.

Things that I have no idea how to fix, and so won't include in this PR until further notice:
* sprite loading
* sprite loading


Ninnginni:
I think a cryo fluid mixer more efficient like the t2 graphite press will be better.

Anuken:
"Uses a lot of power" or "takes some time to warm up" is not enough. Every T2 block must have additional requirements (*besides simply using more resources*) to make sure they can't be used as a direct upgrade.

Ninnginni:
The additional requirement could be the cost of manufacturing, instead of the cost of production, or an additional ingredient to make cryoliquid.

Suggestedone9:
maybe make it so it needs to use oil (if it doesn't make sense in v4.0 oil could be used as a coolant and its used irl for heat transfer applications)

Ferociousfeind:
I think using more titanium per unit of cryofluid, along with higher power requirements, answers the question of "but how is it not a direct upgrade?"

However, if not, perhaps consider making it require plastanium. That's a way of requiring oil.

BasedUser:
A good one would be explosive materials, to speed the rotator up.

unjown:
@Ninnginni there is such thing as graph press lol

unjown:
WE NEED THIS T2 CRYO MIXER
my suggestion input
titanium
water
graph??

genNAowl:
Graph doesn't make too much sense, but none of the other ingredients make as much easier. A better suggestion might be silicon?

genNAowl:
@BasedUser Check your PRs

unjown:
@genNAowl nice idea silicon 
but im avoiding fac,etc. to use sili
but ok not bad but i prefer graph
:)

hemisemidemipresent:
maybe let it require a bit of phase?

SnakkiZXZ:
I like the idea of ​​a large cryogenic liquid plant, but I do not like its texture, sorry, but I decided to make my own version.
criofluid-gringer
criofluid-gringer-top
criofluid-gringer-rotator
criofluid-gringer-bottom
criofluid-gringer-liquid

 here are all the textures that are needed for the code.

sk7725:
> "Uses a lot of power" or "takes some time to warm up" is not enough. Every T2 block must have additional requirements (_besides simply using more resources_) to make sure they can't be used as a direct upgrade.

Maybe make it "heat up", needing pauses between intense runs or making its production diminish(so it needs advanced control by processors or manual item distribution(using incinerators&routers or overflow stuff)) or halt ~~or blow up, we need more suicidal blocks in Mindustry~~

stuffyAI:
Throw in Spore pods for t̶h̶e̶ ̶m̶e̶m̶e̶s̶ good measure, lines the progression not only by building recipe due to power costs.  
If you wanna pretend to be accurate you can additionally output coal for biochemical reasons.

Colonizor48:
I have an idea. Why not have it need cryofluid in the first place to make more cryofluid? And you have to setup a positive feedback loop with it.

unjown:
> I have an idea. Why not have it need cryofluid in the first place to make more cryofluid? And you have to setup a positive feedback loop with it.
@Anuken what u think?

stuffyAI:
> I have an idea. Why not have it need cryofluid in the first place to make more cryofluid? And you have to setup a positive feedback loop with it.

This might be cursed, when something has the same input as output it will just output the input.

I actually encountered a lack of Cryofluid during campaign, causing me to wait the full wave timer for it to refill.

To make this block unique we could force People to use at least a bit of the new logic system: 
The T2 cryomixer starts freezing sometimes (random or periodic), requiring you to insert a blast compound. However, should you have more than one blast compound in the mixer at one time, it will slow down or even stop working due to excess heat.
This can be done adding the heat bar from the thorium reactor , but this time you have to try keeping it above 0% and below 100%, with an perfect range of 40-50% for 100% efficiency. The player will not only be forced to finally research logic, but also utilize it to input just enough blast compound for the mixer to function.

unjown:
@Anuken what u think of that new idea?

RogueRedHawk:
@unjown Don't need to ping him, he reads nearly everything.

unjown:
> @unjown Don't need to ping him, he reads nearly everything.

sry

stuffyAI:
> Why not have it need cryofluid in the first place to make more cryofluid?

All this would accomplish is to bother the player to

place the schematic with tier 2 cryo-mixers
replace one of them with a tier 1 one
replace that with a tier 2 one, whilst slowly piping some cryofluid around

after that the cryofluid mixers would sustain themselfes...

unjown:
> > I have an idea. Why not have it need cryofluid in the first place to make more cryofluid? And you have to setup a positive feedback loop with it.
> 
> This might be cursed, when something has the same input as output it will just output the input.
> 
> I actually encountered a lack of Cryofluid during campaign, causing me to wait the full wave timer for it to refill.
> 
> To make this block unique we could force People to use at least a bit of the new logic system:
> The T2 cryomixer starts freezing sometimes (random or periodic), requiring you to insert a blast compound. However, should you have more than one blast compound in the mixer at one time, it will slow down or even stop working due to excess heat.
> This can be done adding the heat bar from the thorium reactor , but this time you have to try keeping it above 0% and below 100%, with an perfect range of 40-50% for 100% efficiency. The player will not only be forced to finally research logic, but also utilize it to input just enough blast compound for the mixer to function.
> 
> Realism excuse: the t2 Cryomixer can work on much lower temperatures for more efficiency, but requires heat conservative machinery, requiring careful temperature maintenance to avoid freezing of parts due to cold or failing of binding cryofluid due to excess heat.

Nice 
Have another Idea 
Make blast storable cuz it will be hard to maintain the flow

stuffyAI:
> Make blast storable cuz it will be hard to maintain the flow

the point is to not store it in the cryomixer, and you already can store it anywhere else.

sample-text-here:
> I have an idea. Why not have it need cryofluid in the first place to make more cryofluid? And you have to setup a positive feedback loop with it.

@Colonizor48 All you need to make this irrelevant is to place two mixers next to each other (or if you only have one, a liquid router)

> (what stuffyai wrote)

Yes, people need better reasons to use logic. Right now most people don’t even bother with it.

sample-text-here:
Here's an idea: `3 titanium + 1 pyrite + 100/sec water = 36/sec cryo + 1 sand`. It's less efficient than a normal cryo mixer (uses more water) and requires an additional ingredient. Also, you need to remove the sand or it will clog. Finally, it will implement @Colonizor48 's feedback loop idea; the sand output can be used to make more pyrite. What do you think?

Xeloboyo:
>  Every T2 block must have additional requirements

Why cant the additional requirement (on top of the ideas suggested above) be a terrain requirement like the oil extractor? e.g Ice/snow tiles. that way it obviously cant be just used in every circumstance and you need to make a decision much like thermal tiles what production blocks would best benefit. in the case of ice/snow I guess it would be water extractors and this new block. Perhaps the new space tile can be utilised this way.

As a contrarian, I disagree with the notion that the logic block should be mandatory in operating this building. Perhaps as a way to improve output, but not whether this block operates at all. Logic is probably one of the _last_ things you actually learn as a person **_without_** a programming background. It already has immense advantages and therefore incentives to using it already as an 'advanced' gameplay mechanic for finer production control/logistics with existing production buildings (e.g. using unit logic to get the output of 2 water-fed airblast drills from a single tile of sand).  Making logic mandatory will not incentivise learning logic, so much as incentivising giving up on the block entirely because you don't understand logic intuitively and you expecting the block to just 'work' given the resources (like every other production block in existence so far) off the bat. There are people I've seen who have a hard time grasping even what the conditional jump does, and in my honest opinion I feel such an approach will be detrimental to the gameplay experience.

stuffyAI:
> Logic is probably one of the _last_ things you actually learn as a person **_without_** a programming background

Now I am not a madman but what I suggested was basically a one-liner to wait x amount of time to achieve full production, if that won't do we can always implement it to work always at 100% but you can reduce the blast consumption by the wait timer or increase production 20 like 200% by using logic, like you described your other examples.

Xeloboyo:
it's deceptively simple for you, but i don't feel its quite enjoyable for someone who doesn't have that background.
And yeh its a two liner optimised, but starting out ur probably gonna end up with some convoluted jump mess.
and yeh like you said, something like a base working speed of 75% and having it in the sweet zone gets it up to 100-125% is a good solution that still encourages logic directly, but doesn't force it on you.

TechFish3000:
I like the idea of using logic to control it, but I believe a nice way to use it would be as it runs it slowly freezes, and slows production to 40-ish percent minimum, but if you disable the mixer, it would heat up again at roughly 1% per second. This makes it so you _can_ use it without logic, but it would only be about 75% the output of a T1 mixer. However, using logic, you can keep it within a range to get maximum efficiency. When being enabled, it could take about a second to get started, so you can't just spam it on and off, and time spent off is time not spent producing.
This doesn't make logic mandatory per se, but it incentivises it to attain higher production.
Otherwise,
Affinities for ice & snow tiles, but instead of production or power, they affect the size of the "buffer" of efficiency.

Size: 3x3
Inputs per operation: 60 Water, 1 Plastanium, 2 Titanium
Outputs: 45 Cryo
Time to produce: maybe 3x T1 at max efficiency, 0.75x at minimum, exponential scale?

the-kido:
> minimum, but if you disable the mixer, it would heat up again at roughly 1% per second. This makes it so you _can_ use it without logic, but it would only be about 75% the output of a T1 mixer. However, using logic, you can keep it within

The use of logic might deter and/or overwhelm players from using it. I would personally find myself just using the t1's even if they aren't as fast as the t2. BUT, I like the idea of it getting cold. So what if it used something like slag or coal and in the texture, there would be a dash of red to signify the factory is warm and not freezing up.

Dinistor:
I think good idea will be make it require additional oil input, "to lubricate moving parts of the mixer", and mechanic, similar to impact reactor - it can't produce crio, when there's no enough power, production will have spin-up time, and it will need part of it's output crio, for cooling itself (so it can't start without crio input). Maybe add small buff on cold surfaces, like ice.
Approximate ratios may look like this (idk what to do with build cost and power consumption):
Production time: 3 seconds
Input: 96 water, 36 oil, 12 criofluid, 2 titanium
Output: 96 criofluid
Description: Experimental tier 2 criofluid mixer. Pulverizes titanium at extremly high rotor speed. Requres oil for lubricating moving rotor parts and a little cooling for them.

Anuken:
I don't think there's a pressing need for this right now; may consider adding it after 6.0 if there is enough demand.

PR #2846, opened 2020-10-06
Add help button to hosting for pause dialog menu

FrankWhoee:
Hi! I'm a beginner to contributing to this repository, so go easy :) 

When I first started playing this game with my friends, I found that hosting was really frustrating because the @host.info button was within the HostDialog, but as soon as you clicked the host button, it would bring you out of the HostDialog back to PausedDialog, and you can't see the @host.info anymore. This made it so everytime I was trying to set hosting up, if I forgot the port number, or had to carefully read the host.info again, I had to stop hosting by quitting the current game, restarting, and then going back to HostDialog. Moving host.info out will help beginners set up hosting easier!

image

### DIFF
--- a/core/src/mindustry/ui/dialogs/HostDialog.java
+++ b/core/src/mindustry/ui/dialogs/HostDialog.java
@@ -50,8 +50,6 @@ public class HostDialog extends BaseDialog{
             runHost();
         }).width(w).height(70f);
 
-        cont.button("?", () -> ui.showInfo("@host.info")).size(65f, 70f).padLeft(6f);
-
         shown(() -> {
             if(!steam){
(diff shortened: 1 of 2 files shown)


rodrigolitzius:
Kool

itcannotbe:
What info would it even include?

FrankWhoee:
> What info would it even include?

It's just the normal info button from the HostDialog panel, moved out to the PausedDialog panel.

Anuken:
In principle, this is a good idea, but the current implementation breaks layout (*by shifting two dialogs to the side*) and doesn't look all that great. I don't think it should be implemented with an extra button in the paused dialog.

PR #2851, opened 2020-10-06
Make Core team color brighten like the Discord logo.

MEEPofFaith:
Let's try this again and hope it works. I don't know why what I did breaks everything, it really looks like it shouldn't, so I'm trying again to see if this works.


MEEPofFaith:
wait why did it keep everything from previously

PR #2854, opened 2020-10-07
Update bundler launcher java

Anubhav-developr:
As there is no need to put // inside the  " " it will work fine with  single / reducing time complexity.

### DIFF
--- a/tools/src/mindustry/tools/BundleLauncher.java
+++ b/tools/src/mindustry/tools/BundleLauncher.java
@@ -44,7 +44,7 @@ public class BundleLauncher{
                 }
             }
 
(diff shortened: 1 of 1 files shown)


Anubhav-developr:
Changed some time consuming elements

SnakkiZXZ:
why is it? how will this affect the gameplay?

SethiPandi:
Might as well just delete the method call at that point? :thinking: :thinking: :thinking:

```java
.replace("\n", "\n")
```

Probably just Hacktoberfest spam.

l-Luna:
dumb spam

PR #2860, opened 2020-10-07
Update servers_be.json

BasedUser:
not a second server, instead, a measure to allow faster switches between BE and alpha (since servers with a connection error are not shown)

the more crash logs the better


Quezler:
iirc it still shows the server, but displays a (red) version & build warning 🤔

BasedUser:
only if that server is online (which it wouldnt be unless i ran a python script to reply with ping data)

PR #2876, opened 2020-10-09
Phase Fabric ammo for Spectre

z0mbiesrock:
The Spectre is a 4x4 all-around turret with a limited selection of projectiles. By adding phase fabric as a form of ammunition, Spectre turrets have more meaning to their name, which results in bullets that do slightly less damage, but pierce through enemy units.

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1747,6 +1747,7 @@ public class Blocks implements ContentList{
             ammo(
             Items.graphite, Bullets.standardDenseBig,
             Items.pyratite, Bullets.standardIncendiaryBig,
+            Items.phasefabric, Bullets.standardPhaseBig,
             Items.thorium, Bullets.standardThoriumBig
             );
             reloadTime = 6f;
--- a/core/src/mindustry/content/Bullets.java
+++ b/core/src/mindustry/content/Bullets.java
@@ -31,7 +31,7 @@ public class Bullets implements ContentList{
 
     //standard
     standardCopper, standardDense, standardThorium, standardHoming, standardIncendiary, standardMechSmall,
-    standardGlaive, standardDenseBig, standardThoriumBig, standardIncendiaryBig,
+    standardGlaive, standardDenseBig, standardThoriumBig, standardPhaseBig, standardIncendiaryBig,
 
     //liquid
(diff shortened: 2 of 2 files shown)


genNAowl:
1. Why?
2. Who would use this anyway?
3. Why should phase fabric be an ammo? (I mean its a fabric. And it has "phase" properties. Using phase fabric as an ammo would be like launching T-shirts at enemies)

genNAowl:
4. This doesn't even compile, you need to replace `reloadMultiplier = 1.1;` with `reloadMultiplier = 1.1f;`

XEN0PHIL:
Agreed with genNAowl, phase should not be used as a projectile ammo.
If you want ammo diversity for Spectre, convince Anuke that titanium is a suitable material for bullets :thinking:

genNAowl:
Or maybe silicon as a possible ammo? (to reflect on the fact that the Spectre is an upgraded duo/salvo)

z0mbiesrock:
> Or maybe silicon as a possible ammo? (to reflect on the fact that the Spectre is an upgraded duo/salvo)

Isn't Phase Fabric made from sand, too?

sk7725:
Oh *yes* phase ammo

genNAowl:
> made from sand

Its a fabric. Not a hard material that you can just shoot. Silicon, on the other hand, is already used as an ammo in the Duo and Salvo, and since the Spectre are a high level version of the two, silicon ammo can be added, if only just for consistency.

stuffyAI:
> hase fabric as a form of ammunition

I beg to differ, using phase to boost your entire factory by 250% is more valuable than wasting it as ammo on a dps turret.

Anuken:
Nah.

z0mbiesrock:
5d61aa2be1369

Colonizor48:
check this out for a turret that uses phase https://github.com/Anuken/Mindustry-Suggestions/issues/721

Colonizor48:
@z0mbiesrock The thing is we already kinda have a 3x3 turret that attacks multible enemies. Its called the meltdown.

genNAowl:
> 3x3 turret that attacks multiple enemies. Its called the meltdown.

Last time I checked it was a 4x4.

Colonizor48:
my mistake

PR #2879, opened 2020-10-09
Console yes command

Quezler:
> implementation of https://github.com/Anuken/Mindustry-Suggestions/issues/718

Screen Shot 2020-10-09 at 09 31 07

> (gets cleared after every valid command)

### DIFF
--- a/server/src/mindustry/server/ServerControl.java
+++ b/server/src/mindustry/server/ServerControl.java
@@ -55,6 +55,8 @@ public class ServerControl implements ApplicationListener{
     private ServerSocket serverSocket;
     private PrintWriter socketOutput;
 
+    private String yes;
+
     public ServerControl(String[] args){
         Core.settings.defaults(
             "bans", "",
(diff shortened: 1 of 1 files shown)


Quezler:
gradle overwrote what i typed, but in order it was:
- `yes`
- `playerz`
- `yes`

server/src/mindustry/server/ServerControl.java
```diff
@@ -937,13 +947,16 @@ private void handleCommandString(String line){
...
+                yes = line.replace(response.runCommand, closest.text);
             }else{
                 err("Invalid command. Type 'help' for help.");
             }
         }else if(response.type == ResponseType.fewArguments){
             err("Too few command arguments. Usage: " + response.command.text + " " + response.command.paramText);
         }else if(response.type == ResponseType.manyArguments){
             err("Too many command arguments. Usage: " + response.command.text + " " + response.command.paramText);
+        }else if(response.type == ResponseType.valid){
+            yes = null;
```
Quezler, line 959: the position of where and how yes gets reset to null is debatable*

PR #2881, opened 2020-10-09
Fixes the * float int error in the space shader (merge to close all 5 issues automatically)

Quezler:
`'*' does not operate on 'float' and 'int'`


Anuken:
fixed before I saw this PR

Quezler:
you are forgiven 😃

PR #2898, opened 2020-10-10
Crawler and Quad

D-player1:
h

### DIFF
--- a/core/src/mindustry/content/UnitTypes.java
+++ b/core/src/mindustry/content/UnitTypes.java
@@ -514,9 +514,9 @@ public class UnitTypes implements ContentList{
         crawler = new UnitType("crawler"){{
             defaultController = SuicideAI::new;
 
-            speed = 0.92f;
+            speed = 1.2f;
             hitSize = 8f;
-            health = 180;
+            health = 220;
             mechSideSway = 0.25f;
             range = 40f;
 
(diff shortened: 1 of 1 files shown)


Anuken:
I don't agree with these values.

D-player1:
Which one

D-player1:
Crawler is a suicide robot, with it's damage, it needs 3 amount of crawlers to kill a single dagger, making them tankier and faster doesn't even change a thing. Their damage should also buffed
And for Quad, it's the weakest T4 unit, even though it has a decent amount of speed, Quad have 6000 health, it's just trading Mega's speed for a bigger HP, Mega have 250% building speed and faster than Quad, which makes Quad pointless, and Quad can't even heal blocks, with it's bomb, it suits more as an attacker, but with it's low HP, Quad just doesnt fit anywhere.

Anuken:
> Crawler is a suicide robot, with it's damage, it needs 3 amount of crawlers to kill a single dagger, making them tankier and faster doesn't even change a thing.

Making it have *even more* health than the dagger when it is smaller and lighter makes no sense.

> And for Quad, it's the weakest T4 unit

It's *supposed* to be a fragile bomber. Its health is very similar to that of Antumbra already.

> Quad can't even heal block

It can.

D-player1:
Antumbra has range, using bomber as an attacker is riskier than sniping enemy defense

PR #2904, opened 2020-10-11
Make deconstructing scrap walls return scrap

Quezler:
found myself deconstructing some scrap walls on ruinous shores just now (ironically to uncover a scrap patch) and it dawned on me that scrap walls return nothing when deconstructing.

but now that the core can contain any item (including scrap) it would make sense that scrap walls return the expected amount.

even though you can't build them (yet) it just seems like a quality of life change.

- to keep the theme of wall costs i have kept the size 1 cost 6 like the other walls, but maybe less would make some more sense since there are chunks missing in the texture, though ironically that would make it even worth more scrap.

- sizes 3 and 4 are also multiplied from size 1 to make the calculation visibly easier to understand.

- technically just a suggestion, feel free to either merge of close this at your discretion: @Anuken ❤️ 

Screen Shot 2020-10-11 at 14 02 54

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -829,27 +829,27 @@ public class Blocks implements ContentList{
         }};
 
         scrapWall = new Wall("scrap-wall"){{
-            requirements(Category.defense, BuildVisibility.sandboxOnly, with());
+            requirements(Category.defense, BuildVisibility.sandboxOnly, with(Items.scrap, 6));
             health = 60 * wallHealthMultiplier;
             variants = 5;
         }};
 
         scrapWallLarge = new Wall("scrap-wall-large"){{
-            requirements(Category.defense, BuildVisibility.sandboxOnly, with());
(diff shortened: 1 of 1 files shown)


Anuken:
I'm not sure what purpose this would serve. It makes sense in terms of item drops, but in most cases the amount of scrap you get is insignificant compared to what you'd get by mining. Players may even be annoyed to discover that their core has a small amount of scrap in it after deconstructing a few blocks.

Quezler:
a few scrap would do no harm, and its more than likely that small amounts of sand/coal/etc will make it in there as well 🤔

Quezler:
i didn't really mean this as a way to farm scrap though, just making it fit the picture of anything being deconstructed giving a form of refund, and scrap walls just felt left out 😅

Arvinderjeet:
what is this

Catchears:
I agree that it would make the most sense if all walls returned some of the material they are made of.

MEEPofFaith:
Makes sense. Not helpful in any way, but it's the little details that matter ;)

PR #2905, opened 2020-10-11
add names to lots of ui elements

DeltaNedas:
Allows for mods to easily modify UI with scene.find rather than obtuse .cells.get(n).get().cells.get(n)...
Possible uses:
- Moving fps around or adding bits to it
- Adding stuff under minimap
- Adding buttons to the main menu
- Planet and tech tree stuff


Anuken:
This doesn't compile.

DeltaNedas:
yes my jdk is borked it couldnt start

sk7725:
you sir are a caring genius

PR #2906, opened 2020-10-11
Visualize maximum amount of power node wires

Quezler:
on the same ruinous shore map as my other pull request i ran into an old friend: 12f1ee95373b0e531c55efcb858e7a5ec92acb88
(that commit ~~lowered~~ nerfed the max amount of power node connections)

people used to the normal wire count might expect to be able to link the water extractor to this node, as i have:

Screen Shot 2020-10-11 at 14 24 34

the only problem that there is no real visual indication of the node being full, clicking does nothing and the only hint as to why is in the core database.

i propose to add a bar to power nodes that list the max and current amount of connections:
Screen Shot 2020-10-11 at 14 57 48

> the naming of its bar & bundle entry doesn't feel quite right, accepting suggestions on how to (re)name those*

### DIFF
--- a/core/src/mindustry/world/blocks/power/PowerNode.java
+++ b/core/src/mindustry/world/blocks/power/PowerNode.java
@@ -112,6 +112,12 @@ public class PowerNode extends PowerBlock{
             (UI.formatAmount((int)entity.power.graph.getLastPowerStored())), UI.formatAmount((int)entity.power.graph.getLastCapacity())),
             () -> Pal.powerBar,
             () -> Mathf.clamp(entity.power.graph.getLastPowerStored() / entity.power.graph.getLastCapacity())));
+
+        bars.add("connections", entity -> new Bar(() ->
+        Core.bundle.format("bar.powerlines", entity.power.links.size, maxNodes),
+            () -> Pal.ammo,
+            () -> (float)entity.power.links.size / (float)maxNodes
+        ));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


core/src/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -112,6 +112,12 @@ public void setBars(){
             (UI.formatAmount((int)entity.power.graph.getLastPowerStored())), UI.formatAmount((int)entity.power.graph.getLastCapacity())),
             () -> Pal.powerBar,
             () -> Mathf.clamp(entity.power.graph.getLastPowerStored() / entity.power.graph.getLastCapacity())));
+
+        bars.add("connections", entity -> new Bar(() ->
+        Core.bundle.format("bar.powerlines", entity.power.links.size, maxNodes),
+            () -> Pal.ammo,
```
Quezler, line 118: @Anuken accept this suggestion to make it use a/the green bar, it might me more fitting given the context:
```suggestion
            () -> Pal.items,
```

Voz-Duh:
Why not `Pal.powerBar`?

Quezler:
@Voz-Duh just didn't seem like a good idea to have 3 bars in a row with the exact same color

Voz-Duh:
Okay

Esnardo:
While you're at it, maybe change power inf to 150? it doesn't seen right that a sandbox item can theoretically run out of connections if you fill it's range with 1x1 blocks like arcs.

PR #2911, opened 2020-10-11
Fix Vars.loops declaration

summetdev:
`mindustry.audio.LoopControl` seems weird, there is `import mindustry.audio.*;` so we don't need to write full package name. Compiles fine.


Anuken:
This was a by-product of IntelliJ refactoring.

PR #2930, opened 2020-10-13
Add New Covered Conveyor

Voz-Duh:
In this pull I create 2 new types - CoveredConveyor and PayloadCoveredConveyor
This conveyors don't move units and use meta glass for make it 
covered-conveyor


DeltaNedas:
stolen from routorio h

Esnardo:
What do they do?

Voz-Duh:
> stolen from routorio h

From Vanilla Upg >:<

Voz-Duh:
> What do they do?

> This conveyors don't move units and use meta glass for make it

Quezler:
this seems bloated

DeltaNedas:
routorio payload conduit

Anuken:
I don't think this is needed. What is the use case?

Voz-Duh:
Screenshot

genNAowl:
Seems to specific. Use bridges or get rid of the copper walls instead.

Voz-Duh:
In some places you won't be able to put the bridge conveyor

Voz-Duh:
> Get rid of the copper walls instead.

*This for example*

core/src/mindustry/content/Blocks.java
```diff
@@ -58,8 +58,8 @@
     scrapWall, scrapWallLarge, scrapWallHuge, scrapWallGigantic, thruster, //ok, these names are getting ridiculous, but at least I don't have humongous walls yet
 
     //transport
-    conveyor, titaniumConveyor, plastaniumConveyor, armoredConveyor, distributor, junction, itemBridge, phaseConveyor, sorter, invertedSorter, router,
-    overflowGate, underflowGate, massDriver, payloadConveyor, payloadRouter,
+    conveyor, titaniumConveyor, plastaniumConveyor, coveredConveryor, armoredConveyor, distributor, junction, itemBridge, phaseConveyor, sorter, invertedSorter, router,
```
Catchears, line 61: you wrote conve**r**yor. Probably just a typo. You did the same in line 954.

SnakkiZXZ:
I agree with Anyuk, I think that there is no point in this, since the units can cope with the usual version of pipelines. Moreover, if the conveyor is closed, then it has a volume, which means that it cannot be crossed,

Voz-Duh:
> Moreover, if the conveyor is closed, then it has a volume, which means that it cannot be crossed,

I wanted to the units to slow down on them but then changed my mind

Voz-Duh:
Ok

Voz-Duh:
I can close it

PR #2943, opened 2020-10-14
Team Indicator For Plastanium Conveyor

Voz-Duh:
In this pull I create team indicator for plastanium conveyor
team-indicator-for-plastanium-conveyor


Quezler:
does it need to look like a tail, or can you make the regions flush with the outer and inner circle?

Voz-Duh:
Hm...
Yes i can

Voz-Duh:
image

Quezler:
oh wow, i'd expected a small dot flush with the end, but i like these wings a lot, ima test them ingame today 👀

Voz-Duh:
/-/

sk7725:
Make the ones on your team not display the wings, so you can enjoy the roomba raw

Quezler:
i've tested it ingame, and they are moving to quick (even when not overdriven at all) to see the detailed wings, i think this pull would only make it in if the default speed of them is significantly slower & their item capacity a lot higher 🤔 (like 1/4th and 25)

Voz-Duh:
> i've tested it ingame, and they are moving to quick (even when not overdriven at all) to see the detailed wings, i think this pull would only make it in if the default speed of them is significantly slower & their item capacity a lot higher 🤔 (like 1/4th and 25)

Press space-bar

Voz-Duh:
> Make the ones on your team not display the wings, so you can enjoy the roomba raw

No

MEEPofFaith:
maybe change the color of the front tip things, or naw.

Quezler:
> maybe change the color of the front tip things, or naw.

those are eyes, and their color kinda needs to stay the same.

MEEPofFaith:
i didn't know those were eyes.
~~Now give them colored laser vision.~~

Voz-Duh:
Magic

PR #2960, opened 2020-10-15
changed private -> public of checkRadius method

sharlotte-mobile:
when i try to change color of forceField by using draw method , the Mindustry is crashed because of that private method.


Quezler (approved):

imreallybadatnames:
ok

PR #2961, opened 2020-10-15
change map to minimap to clarify setting

Catchears:
Simple change as it wasn't completely clear what this setting did before. If Anuke wants me to change all of the other bundles as well, I can, but maybe the tool you use to update the other bundles does that automatically.


Quezler (approved):

Anuken:
I'd rather remove this setting altogether. It doesn't seem very useful.

summetdev:
> It doesn't seem very useful.
Why? Many players uses this feature. For example mobile users when they want to make a screenshot without minimap.

Anuken:
What does this have to do with making screenshots?

PR #2972, opened 2020-10-16
Add colors to StatusEffects

sk7725:
Since there is a color field why not do it justice?
Also where is the StatusEffect `"blasted"` used?


sk7725:
Besides, `muddy` and `wet` for some reason already had colors

ritzip:
Blasted deals 18 armor ignoring damage when the enemy has `freezing`

Anuken:
This doesn't compile. You can't reference other content because of initialization order; use color codes instead.

sk7725:
Fixed,I hope

PR #2974, opened 2020-10-16
New feature request: bullet homing delay

abomb4:
I'm a mod creator, I want a feature that adds a delay for bullet homing, to get some funny effects.
The image below shows the difference between homing delay or not, the left turret has a delay, the right one has no delay.
图片

### DIFF
--- a/core/src/mindustry/entities/bullet/BulletType.java
+++ b/core/src/mindustry/entities/bullet/BulletType.java
@@ -104,6 +104,8 @@ public abstract class BulletType extends Content{
     public float incendChance = 1f;
     public float homingPower = 0f;
     public float homingRange = 50f;
+    /** Use a negative value to disable homing delay. */
+    public float homingDelay = -1f;
 
     public Color lightningColor = Pal.surge;
     public int lightning;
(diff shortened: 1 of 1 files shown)


Quezler:
wouldn't `0` instead of `-1` have the same effect?

sk7725:
not bad tbh

summetdev:
cool

abomb4:
> 
> 
> wouldn't `0` instead of `-1` have the same effect?

Hmm, I saw `splashDamageRadius` default value is `-1`, so I used `-1` as default value.
Now `0` has the same effect as `-1`.

SirMachinegun:
> > wouldn't `0` instead of `-1` have the same effect?
> 
> Hmm, I saw `splashDamageRadius` default value is `-1`, so I used `-1` as default value.

I think the reason 'splashDamageRadius' has '-1' as default value is because '-1' isn't a valid value therefore disabling splash damage. So I think by default, splash damage is disabled on a bullet, but if it is higher than 0 it will enable splash damage and define it's radius.

By setting homing delay to '-1' I think same things will happen.

DeltaNedas:
you had me at
> feature **request**

h

PR #2975, opened 2020-10-16
Update Unit Locate instruction to access located building

z0mbiesrock:
Update the instruction so that when a "building" is specified for locating, it also returns the building itself to allow access to its variables and as a variable. This would allow a controlled unit to directly interact with the block

An example of a logic process that would work from this change is:
```
 ubind @nova
 sensor cnt @unit @totalItems
 sensor icp @unit @itemCapacity
 ucontrol boost 1 0 0 0
 ulocate building core false @copper creX creY found cre
 ucontrol approach creX creY 10 0
 ucontrol itemTake cre @silicon icp 0
```

The current unitLocate command doesn't return the block; currently the nova would just hover endlessly over the core through this process and take no items whatsoever.

image
I'm certain this would probably get rejected, because I am uncertain as to why

### DIFF
--- a/core/src/mindustry/logic/LStatements.java
+++ b/core/src/mindustry/logic/LStatements.java
@@ -820,6 +820,7 @@ public class LStatements{
         public BlockFlag flag = BlockFlag.core;
         public String enemy = "true", ore = "@copper";
         public String outX = "outx", outY = "outy", outFound = "found";
+        public String getOut = "out";
 
         @Override
         public void build(Table table){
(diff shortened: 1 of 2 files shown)


z0mbiesrock:
RIP. A syntax error has spiked the chances of rejection to 100%

core/src/mindustry/logic/LStatements.java
```diff
@@ -905,6 +906,12 @@ void rebuild(Table table){
             table.add(" found ").left();
             fields(table, outFound, str -> outFound = str);
 
+            if(locate == building) {
```
summetdev, line 909:
```suggestion
            if(locate == building){
```

summetdev:
> RIP. A syntax error has spiked the chances of rejection to 100%

Use an IDE like IntelliJ Idea to avoid syntax errors

Anuken:
> because I am uncertain as to why

There are many issues that could occur:
- controlling/reading data from blocks not on your team
- controlling blocks far away from the processor with no visible connection
- saving a reference to a building and controlling it even when the unit near it is gone
- saving and controlling a building which no longer exists/has been destroyed

I will implement this myself when I need to.

PR #2979, opened 2020-10-16
PlanetGrid stuff

DeltaNedas:
This pull does a few things
1. Makes PlanetGrid's constructor protected not pkg private :) 
2. Buildingount -> tileCount
3. Make random calculation helpers public (not that mess at the top though, I can't comprehend it)
4. Make the fields mutable, see 1 for use case.

### DIFF
--- a/core/src/mindustry/graphics/g3d/PlanetGrid.java
+++ b/core/src/mindustry/graphics/g3d/PlanetGrid.java
@@ -221,12 +221,12 @@ public class PlanetGrid{
     }
 
     public static class Ptile{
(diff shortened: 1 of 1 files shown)


Anuken:
Why?

DeltaNedas:
i want to make a planet with 1 sector

DeltaNedas:
orbital ion router

Anuken:
The calculations look very specific and can probably be re-implemented in other code if needed; make the package-private/private again.

DeltaNedas:
okay

PR #2992, opened 2020-10-17
allow modded planets to customize drawing

DeltaNedas:
allows mods to hide planets and to do something other than drawing the mesh (e.g. a giant laser)


core/src/mindustry/type/Planet.java
```diff
@@ -264,4 +264,12 @@ public void displayInfo(Table table){
     public ContentType getContentType(){
         return ContentType.planet;
     }
+
+    public boolean visible(){
```
summetdev, line 268: Why not make it as class property?

DeltaNedas: what if i want it to be visible only when its day irl, or when some block is researched, etc.

DeltaNedas:
@Anuken fixed conflicts

PR #3004, opened 2020-10-18
Add comments

blahblahbloopster:
Did some refactoring and added comments.

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -34,9 +34,13 @@ import java.util.*;
 
 import static mindustry.Vars.*;
 
+/** Defines a type of block, does not store state. */
 public class Block extends UnlockableContent{
+    /** Weather this block can contain items. */
     public boolean hasItems;
+    /** Weather this block can contain liquids. */
     public boolean hasLiquids;
+    /** Weather this block can use or produce power. */
     public boolean hasPower;
 
     public boolean outputsLiquid = false;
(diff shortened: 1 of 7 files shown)


core/src/mindustry/world/Block.java
```diff
@@ -34,9 +34,13 @@
 
 import static mindustry.Vars.*;
 
+/** Defines a type of block, does not store state. */
 public class Block extends UnlockableContent{
+    /** Weather this block can contain items. */
     public boolean hasItems;
+    /** Weather this block can contain liquids. */
```
DeltaNedas, line 41: Pissing rain but it's not too bad

core/src/mindustry/world/Tile.java
```diff
@@ -446,7 +446,7 @@ public Tile getNearby(int rotation){
         return null;
     }
 
-    public Building getNearbyEntity(int rotation){
+    public Building getNearbyBuilding(int rotation){
```
DeltaNedas, line 449: I think just `adjacentBuild` would be better, `getNearbyBuilding` is a mouthful and adjacent describes it better.

DeltaNedas (commented):
j

sk7725:
Whether, not weather

one gives you a choice, and one punishes you for choosing wrong on whether to bring an umbrella with you today.

Anuken:
Some of these have grammatical/spelling issues, and some of them are just wrong. I'd rather do it myself.

PR #3008, opened 2020-10-18
Try fix 'Cannot attach to current VM' problem in IDEA

abomb4:
Seems add a `fork` option to core module solves this problem.

I have tried running `:desktop:run` and `:server:run` in my IDEA or `gradlew` in powershell, no problems were found.

I'm not sure if it will cause other problems, on platforms other than windows。


Anuken:
Seems to fix it, thanks.

PR #3010, opened 2020-10-18
Units payload.

summetdev:
Allows you to drop payloads from all of your controlled units.

preview

Suggested by [Makc#2908](https://github.com/Makc-Carbon)

### DIFF
--- a/core/src/mindustry/input/InputHandler.java
+++ b/core/src/mindustry/input/InputHandler.java
@@ -241,7 +241,11 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
             pay.set(x, y);
             pay.dropLastPayload();
             pay.set(prevx, prevy);
-            pay.controlling().each(u -> Call.payloadDropped(u, u.x, u.y));
+            pay.controlling().each(u -> {
+                if(u instanceof Payloadc){
+                    Call.payloadDropped(u, u.x, u.y);
+                }
+            });
         }
     }
 
--- a/core/src/mindustry/input/InputHandler.java
+++ b/core/src/mindustry/input/InputHandler.java
@@ -241,6 +241,11 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
             pay.set(x, y);
             pay.dropLastPayload();
             pay.set(prevx, prevy);
+            pay.controlling().each(u -> {
+                if (u instanceof Payloadc p){
+                    p.dropLastPayload();
+                }
+            });
         }
     }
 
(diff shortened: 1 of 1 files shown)


YellOw139:
But, uh, what if I want to control where every block goes?

Also, do they all go into a line? Or who is going to guarantee that my megas in a mixed unit formation will drop blocks in specific, predictable spots? I doubt there's anyone who can do that. Hence, your idea is well intentioned but too unpredictable to for me or other players to be able to rely on, which makes it useless IMO.

core/src/mindustry/input/InputHandler.java
```diff
@@ -241,6 +241,11 @@ public static void payloadDropped(Unit unit, float x, float y){
             pay.set(x, y);
             pay.dropLastPayload();
             pay.set(prevx, prevy);
+            pay.controlling().each(u -> {
+                if (u instanceof Payloadc p){
```
Anuken, line 245:
```suggestion
                if(u instanceof Payloadc p){
```

core/src/mindustry/input/InputHandler.java
```diff
@@ -241,6 +241,11 @@ public static void payloadDropped(Unit unit, float x, float y){
...
+                    p.dropLastPayload();
```
Anuken, line 246: This won't sync properly. You must use `Call.payloadDropped(...)` so positions are correct clientside.

Anuken (changes requested):

YellOw139:
Maybe if it was enabled for units only... Dropping units all at once would be a useful thing, I agree. But with in blocks I don't see its use.

I'm so mixed on this lol.

core/src/mindustry/input/InputHandler.java
summetdev: Fixed. https://github.com/Anuken/Mindustry/pull/3010/commits/12e35b643caeb1d19ffd306a884921da93fcc08a

BasedUser:
is it possible to make controlled units take and drop items in sync with your main unit?

summetdev:
> is it possible to make controlled units take and drop items in sync with your main unit?

Yes

summetdev:
>But with blocks I don't see its use.

image

core/src/mindustry/input/InputHandler.java
```diff
@@ -241,11 +241,7 @@ public static void payloadDropped(Unit unit, float x, float y){
             pay.set(x, y);
             pay.dropLastPayload();
             pay.set(prevx, prevy);
-            pay.controlling().each(u -> {
-                if(u instanceof Payloadc p){
-                    p.dropLastPayload();
-                }
-            });
+            pay.controlling().each(u -> Call.payloadDropped(u, u.x, u.y));
```
Anuken, line 244: I know the `instanceof` check technically isn't necessary, but by omitting it you'll be sending extra payload drop packets to clients that won't do anything with it. It's best to keep the check there.

summetdev: Fixed. https://github.com/Anuken/Mindustry/pull/3010/commits/801eca5ca3b2ada72884880636c9344d6ea999ea

PR #3011, opened 2020-10-18
add core unit upgrades to reconstructors, ignore factory

BasedUser:
compatibility with potential server mods that put alphas on conveyors


Anuken:
As the upgrades show up in the reconstructor stats, I'd rather not. People will start asking how they're supposed to put their units on conveyors.

PR #3030, opened 2020-10-19
remove final from item.color (and liquid too)

sk7725:
1. texturepack-style mods can modify color to better suit the textured items
2. setting color via js (as opposed to json only or I just don't know how to extendContent with 2 params)
3. abusing it with color-changing items(will only look cool on the sorters and item sources but whatever)


Quezler (approved):

sk7725:
Liquids have a bonus; their color is used more often so rainbow liquid will be pleasing to look at.

GlennFolker:
> I just don't know how to extendContent with 2 params

Well, `new JavaAdapter(classType, {}, ...params)`, duh.

PR #3041, opened 2020-10-20
Admin chat.

summetdev:
Preview:
preview

I'm not sure about `#FF4000`, so suggest your colors.

P.S: Now it's not \<Admin> but \<A>. (Thanks to @Quezler)

### DIFF
--- a/core/src/mindustry/core/NetServer.java
+++ b/core/src/mindustry/core/NetServer.java
@@ -306,7 +306,7 @@ public class NetServer implements ApplicationListener{
                 return;
             }
 			
-            Groups.player.each(Player::admin, a -> a.sendMessage(args[0], player, "[" + Pal.adminChat + "]<A>" + NetClient.colorizeName(player.id, player.name)));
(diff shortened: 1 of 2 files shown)


Quezler (changes requested):
maybe use a `Pal.<color>` instead of a magic string?

summetdev:
> maybe use a `Pal.<color>` instead of a magic string?

Yes, you're right.

core/src/mindustry/core/NetServer.java
```diff
@@ -299,6 +300,15 @@ private void registerCommands(){
             }
         });
 
+        clientCommands.<Player>register("a", "<message...>", "Send a message only to admins.", (args, player) -> {
+            if(!player.admin){
+                player.sendMessage("[scarlet]You must be admin to use this command.");
+                return;
+            }
+			
+            Groups.player.each(Player::admin, a -> a.sendMessage(args[0], player, "[" + Pal.adminChat + "]<A>" + NetClient.colorizeName(player.id, player.name)));
```
Quezler, line 309: this would be identical to how teamchat does it:
```suggestion
            Groups.player.each(Player::admin, a -> a.sendMessage(args[0], player, "[#" + Pal.adminChat.toString() + "]<A>" + NetClient.colorizeName(player.id, player.name)));
```

PR #3043, opened 2020-10-20
cleanup?

younggam:
how about this


Anuken:
If you're going to change how one block handles stack lists, at least change all the other blocks so it's consistent.

Anuken:
(see previous comment)

PR #3058, opened 2020-10-22
v5 Javelin Movement Lightning Ability

MEEPofFaith:
[The attack of the "Flarelins"](https://youtu.be/UBFy37o0igk)

### DIFF
--- a/core/src/mindustry/entities/abilities/MovementLightningAbility.java
+++ b/core/src/mindustry/entities/abilities/MovementLightningAbility.java
@@ -1,0 +1,32 @@
+package mindustry.entities.abilities;
+
+import arc.util.*;
+import arc.math.*;
+import mindustry.content.*;
+import mindustry.entities.*;
+import mindustry.gen.*;
+
+public class MovementLightningAbility extends Ability{
+    public float damage = 35f, length = 10f, reload = 0.15;
+    public float minSpeed = 0.8f, maxSpeed = 1.2f;
+    public Effect shootEffect = Fx.lightningShoot;
+    
+    MovementLightningAbility(){}
+    
+    public MovementLightningAbility(float damage, float length, float reload, float minSpeed, float maxSpeed){
+        this.damage = damage;
+        this.length = length;
+        this.reload = reload;
+        this.minSpeed = minSpeed;
+        this.maxSpeed = maxSpeed;
+    }
+    
+    @Override
+    public void update(Unit unit){
+        scl = Mathf.clamp((unit.velocity().len() - minSpeed) / (maxSpeed - minSpeed));
+        if(Mathf.chance(Time.delta() * (reload * scl))){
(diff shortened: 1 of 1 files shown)


genNAowl:
You could add a custom color setting to enable things like healing-green colored lightning

MEEPofFaith:
Thing is, they change color based on team,

genNAowl:
I've never seen red lightning before though...

core/src/mindustry/entities/abilities/MovementLightningAbility.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class MovementLightningAbility extends Ability{
+    public float damage = 35f, reload = 0.15f;
+    public int length = 12;
+    public float minSpeed = 0.8f, maxSpeed = 1.2f;
+    public Effect shootEffect = Fx.sparkShoot;
+    public Sound shootSound = Sounds.spark;
+    public Color color = Color.valueOf("a9d8ff");
+    
+    MovementLightningAbility(){}
```
summetdev, line 19: Is this constructor really needed?

MEEPofFaith: Well, The blank constructor is found in the other ability `.java`s.

SnakkiZXZ:
v5?

core/src/mindustry/entities/abilities/MovementLightningAbility.java
```diff
@@ -0,0 +1,47 @@
...
+    //Lightning damage
+    public float damage = 35f;
+    //Ticks between firing
+    public float reload = 15f;
+    //Length of lightning
+    public int length = 12;
+    //Speeds for when to start lightninging and when to stop getting faster
+    public float minSpeed = 0.8f, maxSpeed = 1.2f;
+    //Shoot effect
+    public Effect shootEffect = Fx.sparkShoot;
```
Anuken, line 21: Yes, I can see that the variable is called "shootEffect", you don't need to write "Shoot effect" above it. Either add more descriptive comments (*"effect created at unit position when shooting"*), or don't add anything. Same for the sound.

core/src/mindustry/entities/abilities/MovementLightningAbility.java
```diff
@@ -0,0 +1,46 @@
...
+        this.reload = reload;
+        this.minSpeed = minSpeed;
+        this.maxSpeed = maxSpeed;
+        this.color = color;
+    }
+    
+    @Override
+    public void update(Unit unit){
+        float scl = Mathf.clamp((unit.vel().len() - minSpeed) / (maxSpeed - minSpeed));
+        if(Mathf.chance(Time.delta * (reload/10 * scl))){
```
Anuken, line 40: This isn't correct. "Ticks between firing" is not what reload is. For example, if reload = 15, this will happen *every frame*, because scl = 1 (at max sped) and 15/10 > 1.

PR #3061, opened 2020-10-22
Let logic find the closest block of a certain type (units)

ThePythonGuy3:
This lets you do things like this one: https://discord.com/channels/391020510269669376/742769933926269069/768777006367113236

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -263,11 +263,11 @@ public class LExecutor{
                         }
                         case block -> {
                             if(exec.obj(block) instanceof Block) {
-                                if (exec.bool(enemy)) {
+                                if(exec.bool(enemy)){
                                     Building b = Units.findEnemyTile(unit.team, unit.x, unit.y, Float.MAX_VALUE, e -> !e.dead() && e.block == exec.obj(block));
                                     res = b == null ? null : b.tile;
                                     build = true;
-                                } else {
+                                }else{
                                     Building b = Units.findAllyTile(unit.team, unit.x, unit.y, Float.MAX_VALUE, e -> !e.dead() && e.block == exec.obj(block));
                                     res = b == null ? null : b.tile;
                                     build = true;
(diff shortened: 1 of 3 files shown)


core/src/mindustry/logic/LExecutor.java
```diff
@@ -260,6 +261,19 @@ public void run(LExecutor exec){
                             res = b == null ? null : b.tile;
                             build = true;
                         }
+                        case block -> {
+                            if(exec.obj(block) instanceof Block) {
+                                if (exec.bool(enemy)) {
+                                    Building b = Units.findEnemyTile(unit.team, unit.x, unit.y, Float.MAX_VALUE, e -> !e.dead() && e.block == exec.obj(block));
```
Anuken, line 267: This is very inefficient and not an acceptable solution. You can't scan the whole map every time you want to fetch a block.

core/src/mindustry/logic/LExecutor.java
```diff
@@ -260,6 +261,19 @@ public void run(LExecutor exec){
...
+                                    res = b == null ? null : b.tile;
+                                    build = true;
+                                } else {
```
Anuken, line 270: [Read the style guide!](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md#style-guidelines) I'm getting sick of saying this.

Anuken (changes requested):

core/src/mindustry/logic/LExecutor.java
ThePythonGuy3: im very sorry, fixed

core/src/mindustry/logic/LExecutor.java
ThePythonGuy3: Then what should i do?

Anuken: Not sure what the best solution is. Could try a loop through all building entities, or a cache that tracks buildings by type.

ThePythonGuy3: hmmm ill see ok

Anuken:
Caching blocks by type may be complex, so I'll implement this myself when I have time.

ThePythonGuy3:
okay thank you

PR #3063, opened 2020-10-22
getBlock return if block is constructing

ThePythonGuy3:


Anuken:
This should not be an extra return variable. A sensor would be a better solution.

PR #3093, opened 2020-10-24
add more info to print object

DeltaNedas:
Adds content name to the end of some printed objects.

`print @routorio-solar-router`: [content] -> **routorio-solar-router**
`print distributor1`: [building] -> **distributor**
`ubind @routorio-sexy-router; print @unit` : [unit] -> **routorio-sexy-router**

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -878,6 +878,7 @@ public class LExecutor{
                     v.objval == null ? "null" :
                     v.objval instanceof String ? (String)v.objval :
                     v.objval instanceof MappableContent ? "[content] " + ((MappableContent)v.objval).name :
+                    v.objval instanceof Content ? "[content]" :
                     v.objval instanceof Building ? "[building] " + ((Building)v.objval).block.name :
                     v.objval instanceof Unit ? "[unit] " + ((Unit)v.objval).type.name :
                     "[object]";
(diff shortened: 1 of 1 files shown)


Quezler (changes requested):
no fallback for `Content`? is `MappableContent` the only possible child?

DeltaNedas:
adding a fallback for just Content in case its possible, but I'm pretty sure it's not.

DeltaNedas (commented):
ok

core/src/mindustry/logic/LExecutor.java
```diff
@@ -877,9 +877,10 @@ public void run(LExecutor exec){
                 String strValue =
                     v.objval == null ? "null" :
                     v.objval instanceof String ? (String)v.objval :
+                    v.objval instanceof MappableContent ? "[content] " + ((MappableContent)v.objval).name :
                     v.objval instanceof Content ? "[content]" :
-                    v.objval instanceof Building ? "[building]" :
-                    v.objval instanceof Unit ? "[unit]" :
+                    v.objval instanceof Building ? "[building] " + ((Building)v.objval).block.name :
```
Anuken, line 882: string concatenation -> StringBuffer allocation -> garbage, not acceptable

DeltaNedas:
yes i the silly

core/src/mindustry/logic/LExecutor.java
```diff
@@ -877,9 +877,10 @@ public void run(LExecutor exec){
                 String strValue =
                     v.objval == null ? "null" :
                     v.objval instanceof String ? (String)v.objval :
+                    v.objval instanceof MappableContent ? ((MappableContent)v.objval).name :
```
Anuken, line 880: These three should use pattern variables, not casts.

DeltaNedas:
anuke I don't know how to use pattern variables so if I'm wrong correct it, I'm only familiar with the normal if version.

PR #3094, opened 2020-10-24
New bundle key: Block can't build.

summetdev:
When translating into russian, we got a problem that the bundles key is the same for blocks and for units, so I decided to add a new one due to the ability to control the router.

image


Anuken:
I don't think this is necessary; the message could just be changed to "can't build" by default.

summetdev:
> I don't think this is necessary; the message could just be changed to "can't build" by default.

It looks better and clearer for the players.

BasedUser:
@summetdev you never know when anuke will add a block that *can* build

summetdev:
> @summetdev you never know when anuke will add a block that *can* build

There is "Unit can't build" string, but some units can build.

BasedUser:
yes

l-Luna:
also for turret control

Anuken:
I still don't think this is necessary.

PR #3096, opened 2020-10-24
make ending direction follow the path

azureblue:
Current logic is annoying (make final tile rotation as relation between start and end) and IMHO rarely useful.
Examples (before and after fix):
example1
example2

### DIFF
--- a/core/src/mindustry/input/InputHandler.java
+++ b/core/src/mindustry/input/InputHandler.java
@@ -1213,7 +1213,14 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
             line.x = point.x;
             line.y = point.y;
             if(!overrideLineRotation || diagonal){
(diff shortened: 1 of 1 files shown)


core/src/mindustry/input/InputHandler.java
```diff
@@ -1213,7 +1213,14 @@ void iterateLine(int startX, int startY, int endX, int endY, Cons<PlaceLine> con
             line.x = point.x;
             line.y = point.y;
             if(!overrideLineRotation || diagonal){
-                line.rotation = next != null ? Tile.relativeTo(point.x, point.y, next.x, next.y) : baseRotation;
+                if (next != null){
```
Anuken, line 1216: incorrect formatting here and below, see the guidelines

azureblue: fixed

DeltaNedas:
i think using rotate with pathfinding conveyors should change the end direction.
changing all of them is only useful with straight lines, but this can satisfy all use cases and is intuitive to use.

PR #3103, opened 2020-10-24
[Modding] Custom Base Support

MEEPofFaith:
If you have a `sprite-base` file, it'll use that as the base, otherwise, it'll default to the regular bases.


DeltaNedas:
`import mindustry.gen.*;` be like

MEEPofFaith:
No one told me where the load textures function comes from, so I just made it similar to how it is in `Conduit.java`.

MEEPofFaith:
hm ok

PR #3104, opened 2020-10-25
[Aesthetic] Turret idle turning

MEEPofFaith:
I'm just making this so that I can get an error message from Travis. The error I have on my end seems to not do with my code.

### DIFF
--- a/core/src/mindustry/world/blocks/defense/turrets/LaserTurret.java
+++ b/core/src/mindustry/world/blocks/defense/turrets/LaserTurret.java
@@ -96,7 +96,7 @@ public class LaserTurret extends PowerTurret{
 
         @Override
         protected void turnToTarget(float targetRot){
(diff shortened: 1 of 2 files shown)


MEEPofFaith:
The mass of reverts is from me checking previous code since it did work previously. I still got the same error, so whatever it is it doesn't have to do with my code.

Anuken:
This introduces a lot of configuration and visual noise, so no thanks.

MEEPofFaith:
oh

MEEPofFaith:
A closed request won't stop me from at figuring out why it doesn't work when it used to.

MEEPofFaith:
It seems to affect all my branches since I can't compile a previously working thing anymore.

MEEPofFaith:
and it's not a duplicate Arc this time

MEEPofFaith:
> This introduces a lot of configuration and visual noise

what does this mean?

PR #3115, opened 2020-10-25
Add Vietnamese characters to font.woff

YellOw139:
Added Vietnamese characters. They properly show up in-game (see photo) and match nicely with the others.

image


braainsiogame:
xin nhập nó vào

PR #3118, opened 2020-10-25
Make run_server.sh executable.

summetdev:
Just a good practice.


Quezler (approved):

Shawak (changes requested):
#!/usr/bin/env bash

Arkanic:
even so people will still need to chmod the file

PR #3125, opened 2020-10-26
Make bloom be accessible

sk7725:
Alternative way: `public Bloom bloom;`  

Make the bloom that is being used in the renderer be public or accessible.
+ The bloom's bloomIntensity, originalIntensity, threshold can be set.  
+ Certain mods can modify it, such as a "high-res" mod or realistic mod(like minecraft's).  

(Also, after tinkering with the bloom I would PR something that implements https://github.com/Anuken/Mindustry-Suggestions/issues/918 later... even if you reject that I can at least make it so in my mods)

### DIFF
--- a/core/src/mindustry/core/Renderer.java
+++ b/core/src/mindustry/core/Renderer.java
@@ -100,6 +100,10 @@ public class Renderer implements ApplicationListener{
     public float weatherAlpha(){
         return weatherAlpha;
     }
+    
+    public Bloom renderedBloom(){
+        return bloom;
+    }
 
     public float landScale(){
         return landTime > 0 ? landscale : 1f;


sk7725:
also, the function name is renderedBloom() and not bloom() so rhino js does not go brrrrr

Quezler:
can't you just read it directly from the config? or do mods not have access to that?

sk7725:
> can't you just read it directly from the config? or do mods not have access to that?

config? I do not think so, since unless something is explicitly public, js mods do not have access to it(exceptions are the weird comps that lurk in gen).

sk7725:
Correct me if I'm wrong, but do JAVA mods have access to private variables in classes?

Also please explain what is "config" that you are talking about? I have yet to see a mod that uses something remotely similar.
Also utilizing your own Bloom class is different from utilizing **the** bloom instance that the Renderer uses.

Anuken:
Bloom should just be public, I frown upon getters.

PR #3127, opened 2020-10-26
allowing customized info for mods

younggam:
since displayInfo for mods disappeared,
there is no way to customize info table.
So this change gives access for info Table.


sk7725:
"mod" -> "modded content", or "mods"

PR #3129, opened 2020-10-26
minor serverside optimisation in roomba

DeltaNedas:
ignore blendprox as it's only used in draw, reduce blends() calls


Anuken:
I doubt this affects server performance much. What lead you to make this PR? Was `blend` causing performance issues?

DeltaNedas:
rtfsc when making upcoming router, just noticed it.
in phase and fusion router i check for client before running blend stuff

PR #3130, opened 2020-10-26
LegsUnit can play step sound if given one

GlennFolker:
It plays `stepSound` if not null, at corresponding foot with given pitch and volume


GlennFolker:
Are my prs actually annoying ~~because my anxiety tells so~~

GlennFolker:
nop, not a good idea to do this

PR #3138, opened 2020-10-26
[political] remove router control size check

DeltaNedas:
i hate to get political but uh router


Quezler:
this probably will output on wrong sides

Anuken:
no

Quezler:
also, what would the intended behaviour be?
a) output on all sides of the edge
b) output only in the exact row you're looking

DeltaNedas:
the latter, it just changes the output direction not tile

PR #3139, opened 2020-10-26
click player icon to respawn

DeltaNedas:
Equivalent of V key on mobile, for when you dont want to swipe over to core and double tap.


DeltaNedas:
@Anuken another commit called h for master please

Quezler:
and in case of an accidental swipe/press?

DeltaNedas:
then you have to go back to your unit.
i would make it respawn on double click but i don't know how to, cleanly

Anuken:
Will implement myself.

PR #3141, opened 2020-10-26
Allow control of bigger routers

sk7725:
Please tell me this compiles & is valid

The size should be under 5(2/(1.41-1) = 4.87) so it doesn't send items to non-adjacent blocks.


sk7725:
(Also delta what were you thinking #3138)

DeltaNedas:
j

Anuken:
Again, I see no reason for this. The math looks like it would break in certain cases, and I'm not going to bother testing a feature that was only supposed to work for one specific block anyway.

PR #3146, opened 2020-10-27
Adds Stat.none

younggam:
Stat.none is for customized info table.
input:    <--this part would be confusing
bla-table-bla-table
bla-table-bla-table
bla-table-bla-table


Quezler:
as per your example would it not still show the `:` part?

younggam:
ohno you are genius. which method can be alternative?
`info:`
makes sense?

core/assets/bundles/bundle.properties
```diff
@@ -649,6 +649,7 @@ stat.minetier = Mine Tier
 stat.payloadcapacity = Payload Capacity
 stat.commandlimit = Command Limit
 stat.abilities = Abilities
+stat.none =
```
summetdev, line 652: Put an empty key to each bundle is pretty strange.

younggam: it's valid for .propeties

summetdev: Why do you even need an *empty* bundle key, bundles are for translations.

younggam: becuz Stat.bla matches with bundle
if there is no bundle for it, it will show ???stat.none???

genNAowl: or you can hardcode it into the display to make it just display nothing

younggam: hmm.. ok

younggam:
currently, if Stat.none, nothing will be shown except StatValue

Quezler:
👀

PR #3151, opened 2020-10-27
even more mod support

GlennFolker:
dunno why, but when i tried using `Draw.vert`, it says that rhino js array can't be converted into `float[]`, so i made this.


DeltaNedas:
pretty sure is only needed because rhino thinks that dropping the 0.0000000002 is consider a lossy conversion and is prohibited

GlennFolker:
Yeah I prefer `.jar` mods, but there _is_ a `.js` mod I need to get over with

Anuken:
I'll just add newFloats. Everything else seems unnecessary.

PR #3163, opened 2020-10-28
"Controlled" property for logic

summetdev:
Closes Anuken/Mindustry-Suggestions#945.
Suggested by @Slava0135.


core/src/mindustry/logic/LAccess.java
```diff
@@ -30,6 +30,7 @@
     team,
     type,
     flag,
+    bounded,
```
summetdev, line 33: Possible alternatives are: `bound`, `isBound`, `isBounded`.

Quezler:
it doesn't seem very descriptive, what about `controlled` or something a little more self explanatory ?

Slava0135:
> it doesn't seem very descriptive, what about `controlled` or something a little more self explanatory?
> (personally it reminds me a lot of boundaries like (map) borders or the max radius of the processor)

agreed

summetdev:
@Quezler changed, thanks.

Anuken:
This should be used for buildings as well. Turrets (and routers?) should return 1 if they are controlled.

summetdev:
@Anuken now it works with blocks too.

Issue #937, opened 2019-10-29
Linux Bumblebee and Mindustry fail


Gorock:
Using Linux Bumblebee optirun and primusrun 

fails starting Mindustry i get following crash report
[crash-report-10_29_2019_14_48_37.txt](https://github.com/Anuken/Mindustry/files/3783837/crash-report-10_29_2019_14_48_37.txt)

i suspect an error by parsing the GLVersion
when i run glinfo primusrun glinfo

i get following Output

GL_VERSION: 4.6.0 NVIDIA 430.40
GL_RENDERER: GeForce GTX 1050/PCIe/SSE2
GL_VENDOR: NVIDIA Corporation
.... many lines probably not nessesarry

LionelBergen:
If you want to debug this issue yourself, you can build the related projects yourself: https://github.com/Anuken/Arc/blob/master/arc-core/src/io/anuke/arc/graphics/glutils/GLVersion.java

This is returning null:
    gl20.glGetString(GL20.GL_VERSION);

I tested using the version string "4.6.0 NVIDIA 430.40", no error is thrown. Based on your stacktrace, the 'versionString' for GL was null.

If you force the versionString to have that value (E.g hard coding it), does everything else run fine?

Anuken:
Some searching gives me two possible reasons for why it's null:
1) I'm not initializing the openGL context before calling it. This is wildly unlikely, as my initialization+version code works fine on every system to date and this is the only crash I've seen of this sort across thousands of reports.
2) Your system's OpenGL system is messed up in some way. This seems far more likely, and as it has never happened for me or any other person that I know of, you'll have to look into the issue yourself.

PR #3165, opened 2020-10-28
Add transportation group to payload conveyors.

summetdev:
So they can replace each other.

preview

*Partially* implements https://github.com/Anuken/Mindustry-Suggestions/issues/937.


scarf-generic:
> _Partially_ implements [Anuken/Mindustry-Suggestions#937](https://github.com/Anuken/Mindustry-Suggestions/issues/937).

Well, not really. When placing over the other conveyor, you still need to build the new conveyor. If this happened with reconstructors, mobile users would still be at a disadvantage, pc users can use r + scrollwheel to rotate already existing buildings. While this is a great addition, sadly it does not implement #937 at all.

PR #3186, opened 2020-10-30
keep ground() in Effect so older code is still valid

sk7725:
#3184

### DIFF
--- a/core/src/mindustry/entities/Effect.java
+++ b/core/src/mindustry/entities/Effect.java
@@ -52,6 +52,17 @@ public class Effect{
         this.layerDuration = duration;
         return this;
     }
+    
+    public Effect ground(){
+        layer = Layer.debris;
+        return this;
+    }
+    
+    public Effect ground(float duration){
+        layer = Layer.debris;
+        this.layerDuration = duration;
+        return this;
+    }
 
     public void at(Position pos){
         create(this, pos.getX(), pos.getY(), 0, Color.white, null);


core/src/mindustry/entities/Effect.java
```diff
@@ -52,6 +52,17 @@ public Effect layer(float l, float duration){
...
     }
+    
+    public Effect ground(){
+        layer = Layer.debris;
+        return this;
+    }
+    
+    public Effect ground(float duration){
+        layer = Layer.debris;
+        this.layerDuration = duration;
```
summetdev, line 63:
```suggestion
        layerDuration = duration;
```

sk7725: I don't know, the original code before *and* after #3184 has the "this.".

summetdev: There is to difference between them, just without `this` is shorter.

MEEPofFaith: `this` is used where the input value and the variable are called the same thing.
If it were `ground(float layerDuration)`, then you would have to do `this.layerDuration = layerDuration` for it to work properly.
I took a 2 week basic Java camp during the summer.

L0615T1C5-216AC-9437: since input `duration` is not named `layerDuration` the `this.` is redundant

sk7725: then what about the functions right above it? the layer() of line 45.

MEEPofFaith:
You do realize that in `js` mod scripting you don't even use `ground()`...
`effect.ground()` turns up invalid and instead you just did `effect.ground = true`

Instead of doing `effect.layer(float)` you just do `effect.layer = float`...

sk7725:
h

PR #3201, opened 2020-10-31
bugFixing

younggam:
multipling side makes ejectEffect weird


Anuken:
Removing the side breaks ejecting even more.

PR #3207, opened 2020-10-31
fix modded bullets

DeltaNedas:
lightningType is checked for being null by content parser when it's set in init anyways


braainsiogame:
tinanium

PR #3209, opened 2020-10-31
Show Core Database Button in Custom Game.

summetdev:
As some people said, core database is very needed in custom or multiplayer game, so I decided to add it.

In custom game:
custom

In campaign:
campaign

Actually suggested to me by @Makc-Carbon.


Makc-Carbon:
Юху! Наконец датабаза данных на пк можно будет открыть в обычной игре

PR #3211, opened 2020-10-31
[maps] Author: Anuke.

summetdev:
Added "Author: Anuke" to each non-authored map. It is cleaner now than "Unknown".

*Yes you have to compile this pull and check it in the game to see difference.*


summetdev:
*Travis why do you need to check I didn't changed source code.*

Anuken:
This is actually a bug. It's supposed to display "Anuke" as the author of any built-in map if it has no author specified.

PR #3215, opened 2020-10-31
[QoL] Add Unit Factories to the Common Block Group.

summetdev:
Closes Anuken/Mindustry-Suggestions#1017. (see more info in issue)

Yet another QoL feature.

Preview:
preview


summetdev:
Closes Anuken/Mindustry-Suggestions#1016 too, it's the same.

Anuken:
That group should include reconstructors.

summetdev:
> That group should include reconstructors.

Would it be better to add group to `UnitBlock`?

Anuken:
Probably.

PR #3233, opened 2020-11-01
New 6.0 server

PAP6:
New server. Nothing special.


Pietro303HD:
Stealing information from: 194.156.120.78
Operation 50% || Fetching the ip address

UltraBlackLinux:
Do not add this. This IP doesn't even have a ddns or any other static IP adress. The server will be lost before even merged.

PR #3239, opened 2020-11-01
properly log logic build errors

DeltaNedas:
instead of just printing the stack trace to stderr, print to log too


Anuken:
if you're going to log it, don't print the stack trace too, that's redundant.

DeltaNedas:
oh i thought that Log.err only showed the head of it not the whole trace

DeltaNedas:
@Anuken

PR #3265, opened 2020-11-04
[editor] Searchable Blocks Implementation

summetdev:
It would make it easier to find the block you want. There are lots of blocks and it takes time to look for the one you want, especially because they're all placed in one large uncategorized all-inclusive selection with no indication if they're power blocks, terrain, drills, etc. (@RedstonerKBC)

Little demonstration:

preview

Closes Anuken/Mindustry-Suggestions#1012.

### DIFF
--- a/core/src/mindustry/editor/MapEditorDialog.java
+++ b/core/src/mindustry/editor/MapEditorDialog.java
@@ -666,7 +666,7 @@ public class MapEditorDialog extends Dialog implements Disposable{
         ui.showConfirm("@confirm", "@editor.unsaved", this::hide);
     }
 
-    private void addBlockSelection(Table table){
+    private void addBlockSelection(Table cont){
         blockSelection = new Table();
         pane = new ScrollPane(blockSelection);
         pane.setFadeScrollBars(false);
@@ -677,16 +677,15 @@ public class MapEditorDialog extends Dialog implements Disposable{
             }
         });
 
-        Table searchBar = new Table();
-        searchBar.image(Icon.zoom);
-        searchBar.field("", this::rebuildBlockSelection)
-        .name("editor/search").maxTextLength(maxNameLength).get().setMessageText("@players.search");
-
-        table.add(searchBar).pad(10);
-        table.row();
-        table.table(Tex.underline, extra -> extra.labelWrap(() -> editor.drawBlock.localizedName).width(200f).center()).growX();
-        table.row();
-        table.add(pane).expandY().top().left();
+        cont.table(search -> {
+            search.image(Icon.zoom).padRight(8);
+            search.field("", this::rebuildBlockSelection)
+            .name("editor/search").maxTextLength(maxNameLength).get().setMessageText("@players.search");
+        }).pad(5);
+        cont.row();
+        cont.table(Tex.underline, extra -> extra.labelWrap(() -> editor.drawBlock.localizedName).width(200f).center()).growX();
+        cont.row();
+        cont.add(pane).expandY().top().left();
 
         rebuildBlockSelection("");
     }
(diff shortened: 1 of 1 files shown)


skykatik (approved):

summetdev:
Update: If there are no results found, the "none" label appears.
none

DeltaNedas:
good

Voz-Duh (approved):
This function will be very useful for both map creators and ordinary people.

core/src/mindustry/editor/MapEditorDialog.java
```diff
@@ -666,18 +667,34 @@ private void tryExit(){
...
         });
-        ButtonGroup<ImageButton> group = new ButtonGroup<>();
 
-        int i = 0;
+        Table searchBar = new Table();
+        searchBar.image(Icon.zoom);
+        searchBar.field("", this::rebuildBlockSelection)
+        .name("editor/search").maxTextLength(maxNameLength).get().setMessageText("@players.search");
+
+        table.add(searchBar).pad(10);
```
Anuken, line 685: Should be defined with `.table(t -> {/** add field, image, etc*/})`

Also, 10 is too much padding. Try something like 4-5.

Anuken (changes requested):

summetdev:
Upd: Padding.

image

core/src/mindustry/editor/MapEditorDialog.java
summetdev: https://github.com/Anuken/Mindustry/pull/3265/commits/275395b5536586a90ce0b79fe20ce90f198963b9

core/src/mindustry/editor/MapEditorDialog.java
```diff
@@ -691,28 +707,32 @@ private void addBlockSelection(Table table){
...
 
+        int i = 0;
+
         for(Block block : blocksOut){
             TextureRegion region = block.icon(Cicon.medium);
 
-            if(!Core.atlas.isFound(region) || !block.inEditor || (block.buildVisibility == BuildVisibility.debugOnly)) continue;
+            if(!Core.atlas.isFound(region) || !block.inEditor
+                    || block.buildVisibility == BuildVisibility.debugOnly
+                    || (!block.localizedName.toLowerCase().contains(searchText.toLowerCase()) && !searchText.isEmpty())
```
Anuken, line 717: `!searchText.isEmpty()` should be put first so it short-circuits and prevents unnecessary `contains()` checks.

core/src/mindustry/editor/MapEditorDialog.java
```diff
@@ -665,19 +666,34 @@ private void tryExit(){
...
+        cont.row();
+        cont.add(pane).expandY().top().left();
+
+        rebuildBlockSelection("");
+    }
+
+    private void rebuildBlockSelection(String searchText){
+        blockSelection.clear();
+
+        Seq<Block> filteredBlocks = new Seq<>();
```
Anuken, line 696: Instead of a list, how about this:

- instead of `filteredBlocks.isEmpty()`, check `i == 0`
- instead of `filteredBlocks.first();`, just set `drawBlock` to the current when `i == 0` and you're iterating

core/src/mindustry/editor/MapEditorDialog.java
summetdev: Done.

core/src/mindustry/editor/MapEditorDialog.java
summetdev: Good, resolved in https://github.com/Anuken/Mindustry/pull/3265/commits/255d1c511ef07c6e7cbb9cf3353f3ff616026e62.

PR #3269, opened 2020-11-05
[Bundle] Power void is sandbox only

YellOw139:
I assumed this must have been a mistake.

Either that or power voids coming to campaign confirmed.


QmelZ:
it doesn't have to mention that it's only for sandbox, but i think this should be merged

QmelZ (approved):

summetdev (approved):

PR #3281, opened 2020-11-07
Building self-heal ability

MEEPofFaith:
Allows buildings to heal themselves.


Anuken:
This should not be in Building.

MEEPofFaith:
Then where should it be?

Anuken:
Nowhere, honestly. I don't think self-healing buildings should exist outside of specializations like mend projectors.

MEEPofFaith:
_shrugs_

PR #3298, opened 2020-11-08
Flying AI in the case that a unit has MoveLightningAbility but no weapons

MEEPofFaith:


Anuken:
This doesn't compile, and I don't think it should be a special case.

PR #3301, opened 2020-11-09
Add new logic categories

sk7725:
As far as I know, js has no way of tempering with `enum`s.
This adds 2 new `LCategory`s that mods can use.


Anuken:
Why? They're just colors anyway. Most modded instructions should use the vanilla categories based on what they do.

sk7725:
There are some instructions that do not fi, because you mostly concentrate on Buildings and Units.
In fact, I will use these for *Sounds*, *Effects* and *Bullets*(Accent) and *Js-related*(Pink).

Anuken:
How about removing LCategory altogether and just using colors?

sk7725:
> How about removing LCategory altogether and just using colors?

@Anuken That would be awesome too, in fact it is actually better than my proposal!

Anuken:
Removed.

PR #3316, opened 2020-11-09
[settings] Discord RPC Setting

summetdev:
Allow players to disable/enable Discord RPC. ~~aww big preview~~

preview-rpc-setting

### DIFF
--- a/desktop/src/mindustry/desktop/DesktopLauncher.java
+++ b/desktop/src/mindustry/desktop/DesktopLauncher.java
@@ -266,6 +266,11 @@ public class DesktopLauncher extends ClientLauncher{
         }
 
         if(useDiscord){
+            if(!Core.settings.getBool("discordrpc")){
+                DiscordRPC.INSTANCE.Discord_ClearPresence();
+                return;
+            }
+
             DiscordRichPresence presence = new DiscordRichPresence();
 
             if(inGame){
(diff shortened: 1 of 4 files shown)


Anuken:
Why? Can't you configure that in Discord?

desktop/src/mindustry/desktop/DesktopLauncher.java
```diff
@@ -266,6 +266,11 @@ public void updateRPC(){
         }
 
         if(useDiscord){
+            if(!Core.settings.getBool("discordrpc")){
+                DiscordRPC.INSTANCE.Discord_ClearPresence();
+                return;
```
Quezler, line 271: there should **not** be a return here since there is another `if` below it 🤔

Quezler (changes requested):

summetdev:
> Why? Can't you configure that in Discord?

You can configure it for **all** games, but what if the player only wants Mindustry? I think the setting would be neat.

DeltaNedas:
what if you want to disable discord rpc stuff from sending anything on start

summetdev:
> what if you want to disable discord rpc stuff from sending anything on start

Just disable it in settings, it saves.

DeltaNedas:
does this still work

summetdev:
@DeltaNedas, yes.

AqilCont:
When will it be merged??? It's been so long and Discord holds a pretty big Mindustry/Gaming centered community that can be better supported in the few minutes it would take to merge.

PR #3321, opened 2020-11-10
Offset lightning spawn location in ability

MEEPofFaith:


MEEPofFaith:
I'ma make this better

PR #3326, opened 2020-11-10
Update bundle_tr.properties

Yakuzet:
I fixed the misstakes


QmelZ (commented):
not really useful, everything is already descriptive enough and you aren't fixing any issues

PR #3333, opened 2020-11-10
Fix bleeding edge updater

buthed010203:
The updater previously didn't update the original jar that was launched due to the use of the wrong function lol


summetdev:
BE updater doesn't supposed to update the original .jar, it downloads a new one into `builds` directory.

buthed010203:
What is the purpose of the entire becopy system property then? It is only located in that one spot in the code and is meant to overwrite the old jar when an update is installed.

PR #3360, opened 2020-11-12
Any unit builder

MEEPofFaith:
Similarly to what was done to mining, now if `buildSpeed > 0`, your unit becomes able to mine.


Anuken:
This isn't implemented properly; it breaks older saves.

PR #3367, opened 2020-11-12
add some boolean logic ops, tweak others

DeltaNedas:
- rename != from not to =/=, a lot of players mix it up with bitwise not
- add section comments
- add ||, && and ! equivalents under the same names
- move pow from bitwise stuff to arithmetic

### DIFF
--- a/core/src/mindustry/logic/LogicOp.java
+++ b/core/src/mindustry/logic/LogicOp.java
@@ -20,7 +20,7 @@ public enum LogicOp{
     greaterThan(">", (a, b) -> a > b ? 1 : 0),
     greaterThanEq(">=", (a, b) -> a >= b ? 1 : 0),
     lnot("!", (a, b) -> a < 0.000001 ? 1 : 0, (a, b) -> a == null ? 1 : 0),
-    lor("||", (a, b) -> a < 0.000001 || b < 0.000001 ? 1 : 0, (a, b) -> a != null || b != null ? 1 : 0),
+    //logical or is essentially the same as bitwise or
     land("&&", (a, b) -> a < 0.000001 && b < 0.000001 ? 1 : 0, (a, b) -> a != null && b != null ? 1 : 0),
     //lxor is just notEqual
     //section bitwise ops
--- a/core/src/mindustry/logic/LogicOp.java
+++ b/core/src/mindustry/logic/LogicOp.java
@@ -14,7 +14,7 @@ public enum LogicOp{
     pow("^", Math::pow),
     //section boolean logic
     equal("==", (a, b) -> Math.abs(a - b) < 0.000001 ? 1 : 0, (a, b) -> Structs.eq(a, b) ? 1 : 0),
-    notEqual("=/=", (a, b) -> Math.abs(a - b) < 0.000001 ? 0 : 1, (a, b) -> !Structs.eq(a, b) ? 1 : 0),
+    notEqual("!=", (a, b) -> Math.abs(a - b) < 0.000001 ? 0 : 1, (a, b) -> !Structs.eq(a, b) ? 1 : 0),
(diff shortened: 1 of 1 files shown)


core/src/mindustry/logic/LogicOp.java
```diff
@@ -4,24 +4,32 @@
 import arc.util.*;
 
 public enum LogicOp{
+    //section arithmetic
```
summetdev, line 7: how about this?
```
//region arithmetic
...
//endregion
```

core/src/mindustry/logic/LogicOp.java
```diff
@@ -4,24 +4,32 @@
...
     sub("-", (a, b) -> a - b),
     mul("*", (a, b) -> a * b),
     div("/", (a, b) -> a / b),
     idiv("//", (a, b) -> Math.floor(a / b)),
     mod("%", (a, b) -> a % b),
+    pow("^", Math::pow),
+    //section boolean logic
     equal("==", (a, b) -> Math.abs(a - b) < 0.000001 ? 1 : 0, (a, b) -> Structs.eq(a, b) ? 1 : 0),
-    notEqual("not", (a, b) -> Math.abs(a - b) < 0.000001 ? 0 : 1, (a, b) -> !Structs.eq(a, b) ? 1 : 0),
+    notEqual("=/=", (a, b) -> Math.abs(a - b) < 0.000001 ? 0 : 1, (a, b) -> !Structs.eq(a, b) ? 1 : 0),
```
summetdev, line 17: maybe `!=`? since equal is `==` so it makes sense.

core/src/mindustry/logic/LogicOp.java
```diff
@@ -4,24 +4,32 @@
...
     lessThan("<", (a, b) -> a < b ? 1 : 0),
     lessThanEq("<=", (a, b) -> a <= b ? 1 : 0),
     greaterThan(">", (a, b) -> a > b ? 1 : 0),
     greaterThanEq(">=", (a, b) -> a >= b ? 1 : 0),
-    pow("^", Math::pow),
+    lnot("!", (a, b) -> a < 0.000001 ? 1 : 0, (a, b) -> a == null ? 1 : 0),
+    lor("||", (a, b) -> a < 0.000001 || b < 0.000001 ? 1 : 0, (a, b) -> a != null || b != null),
```
Anuken: Why is this needed?

core/src/mindustry/logic/LogicOp.java
Anuken, line 17: This is not clear at all.

DeltaNedas: the mathematical symbol is = with a slash through it

core/src/mindustry/logic/LogicOp.java
DeltaNedas: if (!cond || !ptr)
if (block || enemy)

DeltaNedas:
changed =/= to != as requested

core/src/mindustry/logic/LogicOp.java
Anuken: That's exactly what `or` already does.

DeltaNedas: technically correct, but can be neither 0 nor 1, e.g. 0 | 2 = 2

Anuken: Why is that a problem? `0 | 2 = 2 != 0` -> will be evaluated as `true` in any condition

core/src/mindustry/logic/LogicOp.java
DeltaNedas: ok

Anuken:
Decided to use `b-and` for *binary and* instead, as that's more clear than just using "and".

PR #3372, opened 2020-11-12
Add Chaotic Neutral's name and new attack server

L0615T1C5-216AC-9437:


L0615T1C5-216AC-9437:
also anuke is it possible to list server groups by ping? it would be slightly annoying having servers with high ping first

PR #3374, opened 2020-11-13
Formation of the CAMS

BasedUser:
After talking a bit with pandorum's head admin, Skin, we decided to unite the servers under one category.


summetdev:
P.S: CAMS - Coalition of Advanced Mindustry Servers.

PR #3383, opened 2020-11-13
Lets duo use lead. Also lets salvos use lead and titanium

Bohaska:
Tell me if there is anything wrong with this


Anuken:
There's no reason to do this.

PR #3384, opened 2020-11-13
Lead and titanium bullets for duos and salvos

Bohaska:
duos do not have titanium bullets


Anuken:
Again, I don't need this. There is enough ammo variety as is.

Bohaska:
Ok

PR #3386, opened 2020-11-13
Added affinity for solar panels

Bohaska:
I'm not sure if this works though... I haven't tested it yet. Can anyone give me feedback about this?

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -270,12 +270,14 @@ public class Blocks implements ContentList{
             itemDrop = Items.sand;
             playerUnmineable = true;
             attributes.set(Attribute.oil, 0.7f);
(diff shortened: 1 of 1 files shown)


Anuken:
This doesn't make any sense, and doesn't compile. Why would *solar* panels generate power more on sand?

MEEPofFaith:
If this were October, this would have the Hacktoberfest of Shame slapped onto it.

Legomonster33:
> This doesn't make any sense, and doesn't compile. Why would _solar_ panels generate power more on sand?

thermal generators working on sand might make sense (at like 25% speed) but solar panels makes 0 sense.

Anuken:
> at like 25% speed

Even that doesn't make sense. Sand isn't hot by itself, and it's definitely not going to be once you put a large generator on top that blocks out the sun.

PR #3388, opened 2020-11-14
protected->public for mod supports

younggam:
validating `new Conveyor, foo.requirements(), bar.ammo()`
and
how about using below way instead of double braces initialization?
```
public <T extends Block> Block newBlock(Prov<T> constructor, Cons<T> setter){
    Block target = constructor.get();
    setter.get((T) target);
    return target;
}

foo=newBlock(()->new Conveyor("foo"), (Conveyor t)->{
    t.requirements(...);
    t.health=100;
    .
    ..
});
```

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -556,17 +556,17 @@ public class Block extends UnlockableContent{
         return cacheLayer == CacheLayer.walls;
     }
 
-    protected void requirements(Category cat, ItemStack[] stacks, boolean unlocked){
+    public void requirements(Category cat, ItemStack[] stacks, boolean unlocked){
         requirements(cat, BuildVisibility.shown, stacks);
         this.alwaysUnlocked = unlocked;
     }
 
-    protected void requirements(Category cat, ItemStack[] stacks){
+    public void requirements(Category cat, ItemStack[] stacks){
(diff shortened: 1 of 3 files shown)


DeltaNedas:
why not
````
foo = new Conveyor("h") {{
    requirements();
};
```
its much shorter

younggam:
it is but it creates an anonymous class

GlennFolker:
Who cares? Anuke uses it all the time

younggam:
It is well known that anonymous classes cause performance decrease.
But yes, nobody seems care about it.
and main content of this pr is just changing access level.

Anuken:
> It is well known that anonymous classes cause performance decrease.

What is your source for that?

younggam:
https://stackoverflow.com/questions/924285/efficiency-of-java-double-brace-initialization
I heard it affects startup and garbage collections from here. eh, i dunno how much it affects in mindustry.

Anuken:
In my case, none of those objects should be garbage collected anyway, and while there *is* a startup cost to creating so many anonymous classes, it should not affect runtime performance at all.

younggam:
Oh, I understand.
then will you reject thus pr?

Anuken:
No, it seems simple enough.

PR #3397, opened 2020-11-14
Enum for Editor Bindings

summetdev:
Same as Bindings, for more convenience.


Anuken:
This complicates maintenance, so I'd rather not.

PR #3406, opened 2020-11-15
sprite load mod support

younggam:
just overriding sprite is dangerous if multiple mods are overriding same sprite.
So I pr this as solution of the problem.
this allows `mod-name-block-4` instead `block-4` as fallback


DeltaNedas:
turret-name-base.png exists

younggam:
in turret and maybe payloadAccepter case, there is much chance to reuse same sprites. Imagine about adding 139848592 size 5 turrets

Anuken:
This isn't necessary. Just use a custom named base for each turret.

> Imagine about adding 139848592 size-5-turrets

Don't?

younggam:
ohno

Issue #3404, opened 2020-11-15
@shooting always returns 0 for uncontrolled units


Slava0135:
**Platform**: *Any*

**Build**: *113.2*

**Issue**: *You can't check if bound unit is shooting at the moment (while not being controlled)*

**Steps to reproduce**: *Try to read @shooting field when bound unit is actually shooting*

---

*Place an X (no spaces) between the brackets to confirm that you have read the line below.*  
- [X] **I have searched the closed and open issues to make sure that this problem has not already been reported.**

DeltaNedas:
oh so you tested on every supported platform cool
nice code you posted

Slava0135:
nvm before you can check vars from bound unit you need to give him some command...

Slava0135:
nvm it's not right

```ubind @dagger
set fraction 0.8
sensor health @unit @health
sensor max @unit @maxHealth
op mul limit fraction max
jump 12 lessThan health limit
sensor ammo @unit @ammo
jump 12 lessThanEq ammo 1
sensor shooting @unit @shooting
jump 11 notEqual shooting true
ucontrol stop 0 0 0 0 0
end
sensor x @this @x
sensor y @this @y
ucontrol approach x y 4 0 0
```

Slava0135:
basically a retreat system, if unit is low on either ammo or health it retreats. ***But I can't check if it is shooting, it's always 0...***

DeltaNedas:
sense shooting before jump, i guarantee it will work

Slava0135:
![Uploading Peek 2020-11-15 15-24.gif…]()

DeltaNedas:
ok yes i replicate too
**Build**: 11289
**Platform**: Android 9
**Issue**:
Display does not turn green when mega shoots enemies
**Steps to reproduce**:
1. load schem
2. spawn mega
3. spawn enemy sexy routers or make a cage that enemies cant harm the processor in
4. display is still red
5. manually shooting makes it green

[...]

PR #3407, opened 2020-11-15
potential bugfix

DeltaNedas:
for https://github.com/Anuken/Mindustry/issues/3404
makes default ai controller set isShooting to true if any weapon is shooting
no idea if its the right place but oh well


Anuken:
This would always set it to true.

PR #3412, opened 2020-11-15
say to press [c] when first toggling ui

DeltaNedas:
implements https://github.com/Anuken/Mindustry-Suggestions/issues/1205
uses UI#announce


Anuken:
image

core/assets/bundles/bundle.properties
```diff
@@ -290,6 +290,7 @@ cancelbuilding = [accent][[{0}][] to clear plan
 selectschematic = [accent][[{0}][] to select+copy
 pausebuilding = [accent][[{0}][] to pause building
 resumebuilding = [scarlet][[{0}][] to resume building
+showui = [accent][[{0}][] to show UI again
```
summetdev, line 293: How about `Press [accent][[{0}][] to show UI again`?

Anuken: should also mention "UI hidden." as the first line

PR #3456, opened 2020-11-18
add @mapx/y to logic processor

DeltaNedas:
implements https://github.com/Anuken/Mindustry-Suggestions/issues/1251
use to e.g. iterate 0 <= x < @mapx to fill map with routers, for unit patrolling, etc,


Anuken:
x/y are coordinates, not sizes - use w/h

Legomonster33:
priorities
>  fill map with routers

DeltaNedas:
ok

PR #3463, opened 2020-11-19
Finishing pt br translation

JeanJPNM:
I finished translating everything i could and i also polished a few sentences


Pietro303HD:
prepare to die

Pietro303HD:
everything is wrong

JeanJPNM:
> everything is wrong

oh shi-

JeanJPNM:
guess i will die

JeanJPNM:
I would like to say that the translation back to english could not have quite the same meaing because i had to use the names in portuguese defined before in the file rather than just pasting the text into an translator

PR #3470, opened 2020-11-19
Allow for plugins to trigger the same alerts as sectors and guardians

Quezler:
> as discussed with @Anuken on [discord](https://discord.com/channels/391020510269669376/653294015969624064/778606437857099816)

not sure if `Fonts.def` is the correct font to use, but my guess is that it controls the size at which it shows so i kept it 🤔 

(if i am not mistaken this allows the use of both the `Icon` and `Iconc` classes, but not sure if the above plays a role ^)

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -21,6 +21,7 @@ import mindustry.net.Administration.*;
 import mindustry.net.Net.*;
 import mindustry.net.*;
 import mindustry.net.Packets.*;
+import mindustry.ui.*;
 import mindustry.world.*;
 import mindustry.world.modules.*;
 
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -339,6 +339,13 @@ public class NetClient implements ApplicationListener{
 
         ui.showInfoToast(message, duration);
     }
+    
+    @Remote(variants = Variant.both)
+    public static void warningToast(int unicode, String text){
+        if(text == null || mindustry.ui.Fonts.def.getData().getGlyph((char)unicode) == null) return;
+
+        ui.hudfrag.showToast(mindustry.ui.Fonts.getGlyph(mindustry.ui.Fonts.def, (char)unicode), text);
+    }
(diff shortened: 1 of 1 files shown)


summetdev:
> as discussed with Anuken on discord

:eyes:

image

DeltaNedas:
roomba linked the good development channel

core/src/mindustry/core/NetClient.java
```diff
@@ -339,6 +339,13 @@ public static void infoToast(String message, float duration){
 
         ui.showInfoToast(message, duration);
     }
+    
+    @Remote(variants = Variant.both)
+    public static void warningToast(int unicode, String text){
+        if(text == null || mindustry.ui.Fonts.def.getData().getGlyph((char)unicode) == null) return;
```
Anuken, line 345: Why use the whole package name?

Quezler: oh i copy pasted it from `Icon` 👍 

`fileTextFillSmall = mindustry.ui.Fonts.getGlyph(mindustry.ui.Fonts.def, (char)61788);`

(but `.def` is fine? shall i import that one, or the `Fonts` class?)

Anuken: Import `Fonts`. `Icon` uses the whole package name because it's generated code.

Quezler: done 👍 

(`Fonts` is already available there without an extra import it seems)

core/src/mindustry/core/NetClient.java
```diff
@@ -339,6 +340,13 @@ public static void infoToast(String message, float duration){
 
         ui.showInfoToast(message, duration);
     }
+    
+    @Remote(variants = Variant.both)
+    public static void warningToast(int unicode, String text){
+        if(text == null || Fonts.def.getData().getGlyph((char)unicode) == null) return;
```
Anuken, line 346: Wait, you should use `Font.icon` for this, as that's where the large icons are.

Quezler: like this?

PR #3471, opened 2020-11-19
hitColor colored beam effects.

MEEPofFaith:


MEEPofFaith:
aaaand something with the second commit broke everything

PR #3473, opened 2020-11-19
show public ip when hosting in-game

DeltaNedas:
uses icanhazip.com, nice plaintext ip thing

### DIFF
--- a/core/src/mindustry/ui/dialogs/HostDialog.java
+++ b/core/src/mindustry/ui/dialogs/HostDialog.java
@@ -2,6 +2,7 @@ package mindustry.ui.dialogs;
 
 import arc.*;
 import arc.scene.ui.*;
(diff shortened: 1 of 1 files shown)


summetdev:
So technically if you change your IP, it doesn't update in game.

DeltaNedas:
yes

Anuken:
This breaks UI layout.

*now that it comes with artifacts, you have no excuse not to...*

image

DeltaNedas:
yes but i dont know why
i added row and all

Anuken:
1. This is still broken, you haven't used colspan() or any nested tables.
2. Seems like a bad idea. What happens if you're on a cellular network, or it's IPv6?
3. Not a big fan of making HTTP requests to another site, even if it fails gracefully.

PR #3475, opened 2020-11-20
Disable certain condition-based turning.

MEEPofFaith:


Anuken:
I don't see the purpose of this. This may not work in all cases, and requires custom rendering to look decent anyway.

PR #3485, opened 2020-11-20
Remove two stray spaces

Quezler:


Anuken:
Trailing spaces in Markdown have special needed; it may not be needed here but unless it breaks something I'd rather not remove it.

PR #3503, opened 2020-11-21
Only command if there are units to command

Quezler:
Currently this if toggles even when the command list is empty:
https://github.com/Anuken/Mindustry/blob/master/core/src/mindustry/input/InputHandler.java#L373

not sure if this is the best way to fix this, nor if the command list resets when all the commanded units die.

maybe `if(commander.isCommanding()){` should (also) check the count of how many are alive, who knows.

leaving this as a draft for now until @Anuken decides where he would like what. 🤔 

code used to test: (triggers only once every two command keybind presses)
Screen Shot 2020-11-21 at 10 00 01


Anuken:
Implemented in next commit.

PR #3504, opened 2020-11-21
Prevent RhinoJS from having a stroke when you try to use `range()` on bullets

MEEPofFaith:


GlennFolker:
java superior

Anuken:
It's better to change the property name instead.

MEEPofFaith:
?

Issue #3488, opened 2020-11-20
tile.setFloorNet(<liquid>) not updating visually


Quezler:
**Platform**: `mac`

**Build**: `11414`

**Issue**: `the tar floor doesn't show until the user uses /sync:`

(doesn't seem to work when replacing both ground & other liquids)

**Steps to reproduce**: `tried to make server sided oil spills in the puddle class:`

```
                p.lastRipple(Time.time());

//                if(tile.floor().liquidDrop == Liquids.water){
                    tile.setFloorNet(Blocks.tar);
//                }
```

(minimap does in fact show the update, so is at least received by the clients, its just not re-rendering the floor)

---

*Place an X (no spaces) between the brackets to confirm that you have read the line below.*  
- [x] **I have searched the closed and open issues to make sure that this pr oil blem has not already been refined.**

Anuken:
Automatic updating of floors is disabled indefinitely. It cause massive memory usage problems on low-end devices even when updates didn't occur.

Quezler:
ah, unfortunate to hear, gotta scrap some ideas then 😓 

should/shall i make a pull request to remove all references to `setFloorNet` since you've mentioned `indefinitely` ?

PR #3507, opened 2020-11-21
Add remote call for world.endMapLoad

Quezler:
so as discussed in #3488 floor (& wall/darkness) setnet's can't visually update it due to performance reasons.

however, `world.endMapLoad()` gives the desired effect as well.

currently (& in version `5`) the only way around this restriction is to send **the entire map** over again.

this call would simply allow for the (bulk) of changes already sent via setnets to visually update at once.

---

on the version 5 nydus server it got "solved" by tagging the player object with an array of tiles and forcing a sync if the player got close enough to one, aka: "being in an area that significantly changed", so the sync was entirely optional if you worked on another spot of the map, due to fully syncing taking a good solid second most of the time.

with this (on my desktop) it visually updates in the blink of an eye, and i doubt mobile users will have to much problems if this is used few and far in between (maybe several calls per minute, not per second to update small details)

---

i am not sure whether or not this has some side effects, but it just seems to do what i need: reloading world visuals :) 🌍

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -370,6 +370,11 @@ public class NetClient implements ApplicationListener{
             netClient.disconnectQuietly();
         });
     }
+    
+    @Remote(variants = Variant.both)
+    public static void endMapLoad(){
+        world.endMapLoad();
+    }
 
     @Remote(variants = Variant.one)
     public static void setPosition(float x, float y){


Quezler:
unintended side effect 1: the building menu resets back to the first category (though normal with a sync)

Anuken:
I don't think this is an acceptable solution.
- Main issue: It'll cause multi-second lag with no obvious cause. With a /sync, at least there's a loading screen so you know what's happening.
- It updates a *lot* of unnecessary stuff, where only re-caching and recalculating darkness would suffice.
- It might have (even more) unforeseen side effects. At least we know that /sync *works*, with this, I'm not so sure...

PR #3508, opened 2020-11-21
unified extend and extendContent to work for any class

DeltaNedas:
changed extend to be extend(class, constructorargs..., def) which works with old extend(class, def) and deprecated extendContent(class, name, def) which is an alias.
pros:
- backwards compatible
- you now never need to use JavaAdapter unless working with interfaces
- supports any number of constructor args (think planets, bullets)
- also merges non-functions removing need for `a.b = c` over and over
cons:
- none

example:
```js
// Backwards compat alias
extendContent(Block, "name", {test(){print("i am " + this.name)}}).test();
// New variadic args stuff
extend(Block, "egg", {test(){print("i am " + this.name)}}).test();
extend(BasicBulletType, 10, 0.5, {test(){print("my dmg " + this.damage)}}).test();
// Setting variables for you - old extendContent didn't do this.
print("Speed: " + extend(UnitType, "fast", {speed: 1000}).speed);
```

### DIFF
--- a/core/assets/scripts/base.js
+++ b/core/assets/scripts/base.js
@@ -1,6 +1,6 @@
 "use strict";
 
-const log = function(context, obj){
+function log(context, obj){
     Vars.mods.scripts.log(context, String(obj))
 }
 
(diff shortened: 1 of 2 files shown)


summetdev:
don't use `arguments` in js, it's very bad practice. In ES6 you can use [Rest Parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Rest_parameters) whose much more better. (i hope they're supported by Rhino)

DeltaNedas:
yes
yes

DeltaNedas:
@summetdev rhino only has some es6 features, function a(...args){} is a syntax error

Anuken:
Have you tested this with existing mods?

DeltaNedas:
35

DeltaNedas:
it would only break them if they added a predefined field to extendContent then did something assumed it wasnt set and somehow crashed
a simple
```js
const block = extendContent(Block, "h", {
    coolNewField: 0
});
block.coolnewfield = 4;
```
wouldn't change.
if a mod explicitly checks that it doesnt get added then chooses to crash then perhaps it should *not*

PR #3539, opened 2020-11-23
Increase instructions limit

owler0954:
1k instructions not enought


DeltaNedas:
this is done because java short pimits
use more processors for maximum anukage

Anuken:
No.

PR #3561, opened 2020-11-24
Update contributors

Volas171:


summetdev:
Why put a new name at the top? It's better to put it at the end.

Angel-24:
> Why put a new name at the top? It's better to put it at the end.

Why? Does it has an order?

summetdev:
> > Why put a new name at the top? It's better to put it at the end.
> 
> Why? Does it has an order?

Yes.

Angel-24:
> > > Why put a new name at the top? It's better to put it at the end.
> > 
> > 
> > Why? Does it has an order?
> 
> Yes.

ok

Volas171:
ok

PR #3577, opened 2020-11-25
unit movement properties - add velX, velY, hovering, flying, dead variables

BasedUser:
based on https://github.com/Anuken/Mindustry-Suggestions/issues/637


Anuken:
- Velocity: Won't work properly on servers and is completely unreliable. Don't use it.
- Dead state: Why? A unit can be invalid even if it is not dead. This is misleading and not particularly useful.
- Flying/hovering: Why not just return the unit elevation, 0-1?

PR #3583, opened 2020-11-25
automatic reconnect

Quezler:
> this pull request will allow servers to request clients to automatically reconnect when kicked with the right reason.

primary use case: not losing players when updating server sided java (plugins / mods / minigames / etc)

- the last ip & port are these just for ease of use, since its buried [way deep within arc](https://github.com/Anuken/Arc/blob/master/extensions/arcnet/src/arc/net/Client.java#L121). ⛏️ 
- the ping task has no `repeatCount` since i did not see a clean solution to close the loadfrag afterwards. (e.g. `i++`)
- haven't tested what happens if you get kicked on connect with this message.
- bundle entry `server.kicked.serverRestarting` doesn't show anymore.
- not very practical for `BeControl` since the version changes.
- example usage for in minigames or custom commands: 

Screen Shot 2020-11-25 at 10 49 28


core/src/mindustry/ui/dialogs/JoinDialog.java
```diff
@@ -445,13 +450,29 @@ public void connect(String ip, int port){
...
                 hide();
                 add.hide();
             });
         });
     }
 
+    public void reconnect(){
+        ui.loadfrag.show("@reconnecting");
+
+        this.ping = Timer.schedule(() -> {
```
summetdev, line 463: maybe use `ping = ....` here? As `this` is unnecessary

Quezler: no, i tried that, but then java throws several huge tantrums about it not being initialized / not final / several other weird errors / etc, making it a class property was the cleanest working solution

core/src/mindustry/ui/dialogs/JoinDialog.java
```diff
@@ -445,13 +450,34 @@ public void connect(String ip, int port){
...
-            net.connect(ip, port, () -> {
+            net.connect(lastIp = ip, lastPort = port, () -> {
                 hide();
                 add.hide();
             });
         });
     }
 
+    public void reconnect(){
+        if(lastIp.isEmpty()) return;
```
Anuken, line 461:
```suggestion
        if(lastIp == null || lastIp.isEmpty()) return;
```

PR #3596, opened 2020-11-25
Fix ai schematics with power weapons and steam gens

Catchears:
This partially implements [this](https://github.com/Anuken/Mindustry-Suggestions/issues/1312). It adds batteries to all schems which use power weapons and don't generate more power then they could use, and fixes the water extractor ratio issue. Also some of the schems had the steam gens missing, I went ahead and added them as well.


genNAowl:
yes gud

PR #3608, opened 2020-11-26
EntityMapping.map(this.name)

younggam:
kinda mod separation?


summetdev (commented):
`this.name` is equal to `name`, so actually this PR change nothing

younggam:
no, name is local variable and this.name is field.
this.name gets mod-name prefix contrast to name

Anuken:
This doesn't actually accomplish anything. Constructors for modded units should be assigned manually.

younggam:
it is but there can be possibility that confilicts with different mods' units.
Being assigned maunally is not point in this pr

PR #3653, opened 2020-11-29
use JumpI for end

DeltaNedas:
since its equivalent to `jump 0 always`, may as well make it that
removes EndI as it's redundant.

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -782,14 +782,6 @@ public class LExecutor{
         }
     }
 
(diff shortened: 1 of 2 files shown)


Legomonster33:
> since its equivalent to `jump 0 always`, may as well make it that
> removes EndI as it's redundant.

Iirc end clears all variables, jump doesnt

DeltaNedas:
1. no it doesnt, don't believe reddit
2. check what i deleted

Anuken:
Not really necessary and has the possibility of breaking something.

PR #3662, opened 2020-11-30
Fixes #3661

### LINKED ISSUE #3661: Blocks with quickRotate = false still quick-rotates when a new block with a different rotation is placed on top of it.
**Platform**: *Windows*

**Build**: *20141*

**Issue**: *Blocks with `quickRotate = false` quick-rotates when the same block with a different rotation is placed on top. This is the updated behavior, but it shouldn't do so for blocks with quickRotate disabled, since the point of that field is to make the building's rotation persist and be constant until it is broken(Reasons were given on my PR at that time). A suggested solution is to revert the behavior to what it was like before for blocks that has !`quickRotate`.*

**Steps to reproduce**: *Make a block with `quickRotate = false`. Easiest way is to just open console and type `Blocks.conveyor.quickRotate = false;`.*

**Link(s) to mod(s) used**: *null*

**Save file**: *codewise issue*

If you remove the line above without reading it properly and understanding what it means, I will reap your soul. Even if you're playing on someone's server, you can still save the game to a slot.   
*I doubt I have a soul inside me. Besides, what happens at (supposedly) InputHandler.java surely doesn't require a save file(or I hope so).*

**Crash report**: *null*

---

*Place an X (no spaces) between the brackets to confirm that you have read the line below.*  
- [X] **I have updated to the latest release (https://github.com/Anuken/Mindustry/releases) to make sure my issue has not been fixed.**
- [X] **I have searched the closed and open issues to make sure that this problem has not already been reported.**

Quezler:
#3661

### DIFF
--- a/core/src/mindustry/world/Build.java
+++ b/core/src/mindustry/world/Build.java
@@ -58,7 +58,7 @@ public class Build{
         if(tile == null) return;
 
         //auto-rotate the block to the correct orientation and bail out
-        if(tile.team() == team && tile.block == result && tile.build != null){
+        if(tile.team() == team && tile.block == result && tile.build != null && tile.block.quickRotate){
             if(unit != null && unit.isPlayer()) tile.build.lastAccessed = unit.getPlayer().name;
             tile.build.rotation = Mathf.mod(rotation, 4);
             tile.build.updateProximity();


sk7725:
Wait its *that* simple?

sk7725 (approved):

Quezler:
> Wait its _that_ simple?

yes

PR #3664, opened 2020-11-30
shock mine improvements

Slava0135:
Now shock mine won't trigger if it's on low hp (less than 1 charge). They will blink if so.
Peek 2020-11-30 18-18


Slava0135:
*not my fault

core/src/mindustry/world/blocks/defense/ShockMine.java
```diff
@@ -37,20 +40,30 @@ public void drawTeam(){
...
+            if(isActive() || progress < 20f){
+                Draw.color(team.color);
+                Draw.alpha(0.5f);
+                Fill.rect(x, y, 2f, 2f);
+                Draw.color();
+            }
+            if(!isActive()){
+                progress += delta();
+                progress %= blinkDelay;
+            }
```
summetdev, line 52: it's better to use `else` here.
```suggestion
            }else{
                progress += delta();
                progress %= blinkDelay;
            }
```

core/src/mindustry/world/blocks/defense/ShockMine.java
```diff
@@ -29,6 +29,9 @@ public ShockMine(String name){
 
     public class ShockMineBuild extends Building{
```
summetdev, line 31: empty line in top of class.
```suggestion
```

core/src/mindustry/world/blocks/defense/ShockMine.java
```diff
@@ -37,20 +40,30 @@ public void drawTeam(){
...
+            if(!isActive()){
+                progress += delta();
+                progress %= blinkDelay;
+            }
         }
 
         @Override
         public void unitOn(Unit unit){
-            if(enabled && unit.team != team && timer(timerDamage, cooldown)){
+            if(enabled && unit.team != team && timer(timerDamage, cooldown) && isActive()){
```
summetdev, line 57: maybe put `isActive()` after `enabled`? so it doesn't need to execute more height statements.

summetdev (changes requested):

core/src/mindustry/world/blocks/defense/ShockMine.java
Slava0135: this will break stuff

DeltaNedas:
good

joshuaptfan:
The square light is ugly, how about a smaller circle light, same size as the 4 holes? And it should be red, no need for team color as it will already have the team marker when it belongs to another team.

sample-text-here:
[I don't think shock mines are supposed to show their team color](https://github.com/Anuken/Mindustry/blob/d64f603e08ea27804bdc25026392ad30b1470563/core/src/mindustry/world/blocks/defense/ShockMine.java#L33-L35)

Bohaska:
@Slava0135 update this pr

sk7725 (changes requested):
1. Why not only render the blink by `if Time.time % certainamount > certainamount / 2`?
2. Shock mines are supposed to be hard to spot. Make the blink only draw for mines on your team, and keep the led "off" by default, so it is hard to tell your shock mines from the enemy's(which is why team indicators are nonexistant on this block)
3. Why not use a textureregion instead of a solid square, for mods and aesthetics?

Slava0135:
> 1. Why not only render the blink by `if Time.time % certainamount > certainamount / 2`?
> 2. Shock mines are supposed to be hard to spot. Make the blink only draw for mines on your team, and keep the led "off" by default, so it is hard to tell your shock mines from the enemy's(which is why team indicators are nonexistant on this block)
> 3. Why not use a textureregion instead of a solid square, for mods and aesthetics?

2 is Pretty reasonable. I will move the blink effect to drawTeam. Idk about 3. People could just override draw methods if they want. And 1 - there is 2 divisions, while

Slava0135:
Oh I forgot what drawTeam does. It is called on block if you are on enemy team not ally.

Slava0135:
image
idk if it's bright enough to be noticable at close look but not too obvious

Slava0135:
Peek 2020-12-09 20-16

Is there a way to remove "breaking" layer?

Slava0135:
Peek 2020-12-16 14-18
no cracks for now

Slava0135:
*perhaps I should just copy unit blinking health bar thing

PR #3667, opened 2020-11-30
coloured item bubble

Slava0135:
Peek 2020-11-30 20-24

### DIFF
--- a/core/src/mindustry/type/UnitType.java
+++ b/core/src/mindustry/type/UnitType.java
@@ -512,7 +512,7 @@ public class UnitType extends UnlockableContent{
 
             Draw.mixcol();
 
-            Lines.stroke(1f, Pal.accent);
+            Lines.stroke(1f, unit.team.color);
             Lines.circle(
             unit.x + Angles.trnsx(unit.rotation + 180f, itemOffsetY),
             unit.y + Angles.trnsy(unit.rotation + 180f, itemOffsetY),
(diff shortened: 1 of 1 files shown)


Legomonster33:
yis uwu

summetdev:
actually it's a part of UI, that shouldn't be colored

Legomonster33:
> it's a part of UI, that shouldn't be colored

its part of the ui yes, but it isnt like the rest of the ui and its more of an actual gameplay thing rather than a  proper ui element

Slava0135:
> > it's a part of UI, that shouldn't be colored
> 
> its part of the ui yes, but it isnt like the rest of the ui and its more of an actual gameplay thing rather than a proper ui element

facts

Slava0135:
The reason I want it to be changed is because there is colored engine below the bubble. Other UI elements don't have this.

Remint32:
> its part of the ui yes, but it isnt like the rest of the ui and its more of an actual gameplay thing rather than a proper ui element

I agree with this, but then this element itself should be visually redesigned if it is going to be colored in the team color. Because the colors of the teams are not always bright and the bubble merges with the engine fire and even the ship's hull.

Anuken:
I'd rather not.

Legomonster33:
> I'd rather not.

Sad

Voz-Duh:
@Slava0135 don't copy paste my code, stop it

PR #3675, opened 2020-11-30
typo

Remint32:


Anuken:
Aside from the missing brackets, these aren't typos - I intentionally placed the icons next to the text, as they looked better that way.

Remint32:
> Aside from the missing brackets, these aren't typos - I intentionally placed the icons next to the text, as they looked better that way.

Я тоже так подумал, но всё же исправил. Сейчас верну.

PR #3676, opened 2020-11-30
Fix BlockForge progress bar & recenter payload

Quezler:
> experimenting with the block forge for custom plugins, so kinda fixing the stuff i'm running into, this has 2 fixes:

- when it has constructed a block it moves from the center to the edge, but it never resets for future output.

- progress bar was ~~wrong~~ inaccurate, now it stays at 100% when it has stuff in it, and shows actual progress.

### DIFF
--- a/core/src/mindustry/world/blocks/experimental/BlockForge.java
+++ b/core/src/mindustry/world/blocks/experimental/BlockForge.java
@@ -41,7 +41,7 @@ public class BlockForge extends PayloadAcceptor{
     public void setBars(){
         super.setBars();
 
-        bars.add("progress", entity -> new Bar("bar.progress", Pal.ammo, () -> ((BlockForgeBuild)entity).progress));
+        bars.add("progress", (BlockForgeBuild entity) -> new Bar("bar.progress", Pal.ammo, () -> entity.payload != null ? 1f : entity.recipe == null ? 0f : (float)(entity.progress / entity.recipe.buildCost)));
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/blocks/experimental/BlockForge.java
```diff
@@ -41,7 +41,7 @@ public BlockForge(String name){
     public void setBars(){
         super.setBars();
 
-        bars.add("progress", entity -> new Bar("bar.progress", Pal.ammo, () -> ((BlockForgeBuild)entity).progress));
+        bars.add("progress", (BlockForgeBuild entity) -> new Bar("bar.progress", Pal.ammo, () -> entity.payload != null ? 1f : entity.recipe == null ? 0f : (float)(entity.progress / entity.recipe.buildCost)));
```
Quezler, line 44: accept this suggestion in case you want the progress bar to reset to 0 before the block finishes moving out:

```suggestion
        bars.add("progress", (BlockForgeBuild entity) -> new Bar("bar.progress", Pal.ammo, () -> entity.recipe == null ? 0f : (float)(entity.progress / entity.recipe.buildCost)));
```

sk7725 (commented):
Also, the do not seem to sync in servers(aka configuration is not set to the host) last time i tried to experiment with them. Can you be kind enough to fix that too? (Or was that already fixed when I wasn't looking?)

Quezler:
@sk7725 yea that got fixed by a pull earlier on the same day as this one ^

Quezler:
putting this on hold for now since payload vector doesn't sync, which desyncs serverside dump logic fixing nothing 🤔

Quezler:
nvm i'm blind, the vector gets reset when the recipe finishes, not when its dumped, so it would work fine 😅

PR #3707, opened 2020-12-01
Changes to water extractor requirements

Bohaska:
It is a common complaint that the water extractor shouldn't require you to beat Salt Flats, so I changed the requirements for water extractor to completing Ruinous Shores. I think this pull request should also be accompanied by a slightly harder Salt flats (steam gens, some salvos near the base etc.)

### DIFF
--- a/core/src/mindustry/content/TechTree.java
+++ b/core/src/mindustry/content/TechTree.java
@@ -121,7 +121,7 @@ public class TechTree implements ContentList{
 
                             });
 
-                            node(waterExtractor, Seq.with(new SectorComplete(saltFlats)), () -> {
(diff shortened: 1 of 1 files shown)


GreyMario:
Smells like "balance changes".

I'd reject this, personally, due to basic common sense. Salt Flats isn't even that bad. People are just scared.

Bohaska:
> Smells like "balance changes".

And what's wrong with that?

> I'd reject this, personally, due to basic common sense. Salt Flats isn't even that bad. People are just scared.

Salt flats isn't easy without water extractors. I said I want this suggestion to be accompanied by harder Salt Flats to make up for adding water extractors.

Anuken:
I think it's fine.

PR #3715, opened 2020-12-02
Keep items inside reconstructor when construction booped

Quezler:
its a pretty common occurrence that the reconstructor is deconstructed partially for a split second to clear out the unit currently stuck in there, the only problem is that it also gets rid of any items still remaining in there.

this pull does two things:

- returns you all the items if you complete the partially deconstructed reconstructor again (must be the same block)
- tags all deconstructing blocks with what they were so they can be restored if they have custom overwrote logic

during testing i found that you can place a larger reconstructor over a smaller one (and that does not transfer any compatible items), but i am leaving that outside the scope of this pull request (+ who even does that?)

Dec-02-2020 12-18-19

### DIFF
--- a/core/src/mindustry/world/blocks/units/Reconstructor.java
+++ b/core/src/mindustry/world/blocks/units/Reconstructor.java
@@ -115,6 +115,13 @@ public class Reconstructor extends UnitBlock{
         }
 
         @Override
+        public void overwrote(Seq<Building> builds){
+            if(builds.first().block == block){
+                items.add(builds.first().items);
+            }
+        }
+
+        @Override
         public void draw(){
             Draw.rect(region, x, y);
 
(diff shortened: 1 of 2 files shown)


core/src/mindustry/world/blocks/units/Reconstructor.java
```diff
@@ -114,6 +114,13 @@ public int getMaximumAccepted(Item item){
             return capacities[item.id];
         }
 
+        @Override
+        public void overwrote(Seq<Building> builds){
+            if(builds.first().block == block){
```
Anuken, line 119: Can be empty.

```suggestion
            if(builds.any() && builds.first().block == block){
```

Quezler: u sure? https://github.com/Anuken/Mindustry/blob/b63501b1c7cea96146b6664ba66b3f748bb71ddd/core/src/mindustry/world/blocks/ConstructBlock.java#L78

Anuken: Hmm, didn't see that.

PR #3722, opened 2020-12-02
Fast forward boulder deconstruction

Quezler:
(all types of) boulder(s) still need to deconstruct ~35% more when visibly it has already reached the stage of emptyness.

this pull fast forwards the deconstruction once it has reached that point.

notable:
- static wall extends the boulder class, but it cannot be constructed. (`previous.getClass() == Boulder.class` failed)
- 0.35 is just my `Log.info(progress)` guess where i compared it visually, the sweet spot may lie slightly higher or lower.

old:

Dec-02-2020 16-26-58

new:

Dec-02-2020 16-24-18

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -126,7 +126,7 @@ public class Block extends UnlockableContent{
     public BlockGroup group = BlockGroup.none;
     /** List of block flags. Used for AI indexing. */
     public EnumSet<BlockFlag> flags = EnumSet.of();
-    /** Targeting priority of this block, as seen by enemies.*/
+    /** Targeting priority of this block, as seen by enemies .*/
     public TargetPriority priority = TargetPriority.base;
     /** How much this block affects the unit cap by.
      * The block flags must contain unitModifier in order for this to work. */
@@ -139,9 +139,9 @@ public class Block extends UnlockableContent{
     public boolean consumesTap;
     /** Whether to draw the glow of the liquid for this block, if it has one. */
     public boolean drawLiquidLight = true;
-    /** Whether to periodically sync this block across the network.*/
+    /** Whether to periodically sync this block across the network. */
     public boolean sync;
-    /** Whether this block uses conveyor-type placement mode.*/
+    /** Whether this block uses conveyor-type placement mode. */
     public boolean conveyorPlacement;
     /**
      * The color of this block when displayed on the minimap or map preview.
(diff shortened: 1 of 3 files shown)


Quezler:
test map with the different boulder types:
[boulders.msav.zip](https://github.com/Anuken/Mindustry/files/5630435/boulders.msav.zip) (borders are dark tho)

core/src/mindustry/world/blocks/ConstructBlock.java
```diff
@@ -288,6 +289,9 @@ public void deconstruct(Unit builder, @Nullable Building core, float amount){
             }
 
             progress = Mathf.clamp(progress - amount);
+            
+            // fast forward boulder deconstruction when visually already gone
+            if(progress <= 0.35f && previous instanceof Boulder) progress = 0;
```
Anuken, line 294: Instead of hardcoding it for boulders, the number should be a special property that is 0 for normal blocks.

Quezler: ok, I'll change it within the hour 👍

Quezler: any name suggestions?

- deconstructionFinishesAt
- deconstructTrigger
- demolishThreshold

they just don't sound quite right 🤔

Anuken: `deconstructThreshold` seems appropriate

Quezler:
also took the liberty to add some comment changes in here while i'm in there, since the bulk has a space in there ^ 😊

core/src/mindustry/world/blocks/ConstructBlock.java
```diff
@@ -289,7 +289,7 @@ public void deconstruct(Unit builder, @Nullable Building core, float amount){
 
             progress = Mathf.clamp(progress - amount);
 
-            if(progress <= 0 || state.rules.infiniteResources){
+            if(progress <= previous.deconstructThreshold || state.rules.infiniteResources){
```
Anuken, line 292: I believe it's possible for `previous` to be null, so check that just in case.

core/src/mindustry/world/Block.java
```diff
@@ -126,7 +126,7 @@
     public BlockGroup group = BlockGroup.none;
     /** List of block flags. Used for AI indexing. */
     public EnumSet<BlockFlag> flags = EnumSet.of();
-    /** Targeting priority of this block, as seen by enemies.*/
+    /** Targeting priority of this block, as seen by enemies .*/
```
genNAowl, line 129: Whitespace in the wrong place?

Quezler: whoops

PR #3759, opened 2020-12-03
Nerf water extractor (Not my idea)

Bohaska:
Note: This isn't my idea. It is BlueWolf3682 and  itcannotbe's idea and I personally disagree, but I am not sure about Anuke's opinion on this idea, which I need. If all of you hate this idea, then I will close this pr immediately.

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1312,13 +1312,13 @@ public class Blocks implements ContentList{
         waterExtractor = new SolidPump("water-extractor"){{
             requirements(Category.production, with(Items.copper, 25, Items.graphite, 25, Items.lead, 20));
             result = Liquids.water;
-            pumpAmount = 0.11f;
+            pumpAmount = 0.07f;
             size = 2;
             liquidCapacity = 30f;
             rotateSpeed = 1.4f;
             attribute = Attribute.water;
 
-            consumes.power(1.5f);
+            consumes.power(3f);
         }};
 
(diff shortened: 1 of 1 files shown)


BlueWolf3682 (changes requested):
it should be `2f`, which is 120pu/second, not `3f`, which is 180.

D-player1:
Bruh, if you guys want a nerf so badly, why don't you make it output water lower than 6/s, there, we'll straight up killed it

GreyMario:
See also: the entire discussion at https://github.com/Anuken/Mindustry-Suggestions/issues/1402, which has summarized quite well--and repeatedly--why the Water Extractor is in a good place and does not need any changes. If you as a mapmaker want to discourage the Water Extractor, either ban the block, provide better water sources near higher-density material processing areas, or make the floor out of Salt. Don't completely destroy the block just because you can't handle it being used in PVP for fire-and-forget schematics.

Bohaska:
Everyone doesn't like the ner, so...

the-kido:
when the nerf happened i went from "i use this all the time" to "i only use if it i _have to_". I think the nerf was effective but if you guys insist on a nerf how about we lower both the energy usage, the water output, and increase the cost of graphite (which can be a pain to get early-game) to make it an overall worse unit but less energy demanding than it is right now. IMO I think that's the most effective way to "nerf" them.

Bohaska:
>  and increase the cost of graphite (which can be a pain to get early-game) 

I can't understand what you mean. Are you referring to PVP, which I don't play?

PR #3772, opened 2020-12-04
bundle_th.properties

CxZxRainzzz:
Language for thai


Anuken:
This is just the existing thai bundle with a misspelled name...?

PR #3773, opened 2020-12-04
Complete bundle_it

Diegator04:
I compared the **@GioIacca9's bundle** with mine so I was able to notice **mistakes** and **missing translations** which I corrected. We worked together and we discussed about a suitable words and phrases in italian. Furthermore I added the **first version** of _unit's names_, we'll improve them in the future.


Diegator04:
@Anuken, I think it's time to merge this

PR #3776, opened 2020-12-04
support <name>-pod (fallback to launch-pod)

DeltaNedas:
for lunch pad

-----
Place an X in the boxes below to confirm you have read the following:
* [X] I have **not** tested this is any way shape or form


Anuken:
This doesn't actually change what the launched pod looks like, making it inconsistent.

> I have not tested this is any way shape or form

Why make a PR if you don't want to test it? I'm not going to do that for you; it would be easier for me to implement it myself if I have to test it all anyway.

DeltaNedas:
not tested *yet*
apologies, i thought it would be a simple change

PR #3781, opened 2020-12-04
potentially catch strange server list errors

DeltaNedas:
Sometimes the server list won't load, the log says it is fetching it and nothing else.
I believe that the exception ignorer (t -> {}) is catching unexpected errors.

This also gives the stack trace if an error is caught.


DeltaNedas:
anuke

PR #3789, opened 2020-12-05
Custom emoji for Mod(javascript & java) support

AmateurPotion:


core/src/mindustry/ui/Fonts.java
```diff
@@ -202,6 +202,12 @@ public Font loadSync(AssetManager manager, String fileName, Fi file, FreeTypeFon
             ((Font)f).getData().down *= 1.5f;
         };
     }
+    
+    /** Mod's custom emoji support **/
```
summetdev, line 206:
```suggestion
    /** Mod's custom emoji support */
```

PR #982, opened 2019-11-03
Implement number and directional category/block selection hotkeys

joshuaptfan:
This PR implements [this hotkeys suggestion](https://www.reddit.com/r/Mindustry/comments/dlmklh/hotkeys_suggestions_for_building/).

When I had the idea for this suggestion, I didn't know that Mindustry already had hotkeys at the time (that were [removed soon after](https://github.com/Anuken/Mindustry/commit/2e0e3512283afa923a2b5fb4a09a3b4977f5f5f0)). The success or failure of this new hotkey system depends on making it discoverable to players via hints/tutorial.

## Added
* 16 new key bindings (10 numbers, 2 category prev/next, 4 directions)
* Dynamically generated hotkey combo hint (e.g. "Conveyor Key: [3,1]") in each recipe's hover info
* Hotkey tutorial in tutorial mission
* 2 new desktop-only settings (block selection timeout, show block select keys)
* PlacementFragment.java method:
  * `Block getSelectedBlock(Category)`
* Category.java public methods:
  * `Category prev()`
  * `Category next()`

## Modified
* Blocks.java: Moved sandbox blocks to end of each category so hotkey muscle memory will carry over to sandbox
* PlacementFragment.java method:
  * `Array<Block> getByCategory(Category)`

## Cleaned Up
* Moved "Dash" binding to under "Move X/Y" in settings
* Capitalized "X"/"Y" in "Move X/Y" for consistency with "Flip Schematic X/Y"
* Made "# seconds" lowercase for "save interval"
* Added 10 second step to "save interval" slider

PR #3791, opened 2020-12-05
Enable double-click to play sector

joshuaptfan:
I have frequently double-clicked on a sector expecting to go to it immediately, only to realize I have to then click the "Go" button on the bottom of the screen.

This PR makes clicking on a sector that is already selected invoke the same action as the "Go/Resume/Launch" button. It is a QoL change that makes it easier to rapidly switch between sectors, without having to first click on a sector then move the mouse to the "Go" button. It changes very little code; the bulk of the changes was just moving the contents of a lambda function outside into its own function.

I originally intended to make it work only via double-click, but 2 separate clicks also work.


core/src/mindustry/ui/dialogs/PlanetDialog.java
```diff
@@ -412,6 +412,10 @@ void setup(){
                     public void tap(InputEvent event, float x, float y, int count, KeyCode button){
                         if(showing()) return;
 
+                        if(selected == hovered){
```
Anuken, line 415: This should check for a double-tap. If I click a sector once, then click it once again later, it should not send me there.

joshuaptfan:
I've also added myself to contributors at the place in the list where I would have been at the time that my [hotkeys PR](https://github.com/Anuken/Mindustry/pull/982) was merged.

core/src/mindustry/ui/dialogs/PlanetDialog.java
```diff
@@ -405,19 +405,22 @@ void setup(){
...
+
             {
                 //add listener to the background rect, so it doesn't get unnecessary touch input
                 addListener(new ElementGestureListener(){
                     @Override
                     public void tap(InputEvent event, float x, float y, int count, KeyCode button){
                         if(showing()) return;
 
-                        if(selected == hovered){
+                        if(selected == hovered && Time.timeSinceMillis(tapMillis) < 500){
```
Anuken, line 418: `int count` already stored the amount of taps done, you shouldn't need to do your own calculations.

PR #3828, opened 2020-12-06
Add korea server again

Kieaer:
Testing plugin and consulting with the community
Server maintence every Saturday and Sunday.

Addition of server list has been planned for a long time, but is pending due to fast build updates.


ritzip:
r-e.kr when

PR #3830, opened 2020-12-07
Yea remember what happened to `range`?

MEEPofFaith:
Conflict between `collides` and `collides()` gives Rhino a stroke.


Anuken:
Change the *method*, not the property.

MEEPofFaith:
It's better to change the property name instead.

_Originally posted by @Anuken in https://github.com/Anuken/Mindustry/issues/3504#issuecomment-732188981_

But you said the property name when I tried to do the same thing to `range`...

PR #3846, opened 2020-12-07
display online admins draft

Quezler:
> this is a draft, just the concept + some barely working code, **merging discouraged**, discussion welcome.

suggested by `N4B3R1US` on the nydus discord server, displaying the amount of online admins.

pros:
- players will know which servers are currently "safe" to play on due to prescence of admins.

cons:
- griefers will know which servers are currently susceptible to griefers due to lack of staff.

Screen Shot 2020-12-07 at 16 20 48

(maybe a darker color or something, its visually a bit bulky atm)

(and maybe without the parenthesis it should just subtract from the first count?)


braainsiogame:
also make it affect the server's list

Quezler:
> also make it affect the server's list

currently it is sorted by ping, i don't think the admin count should play a role (maybe inside the section, but not outside)

sk7725:
make the color as dark as the ping, and the color of the number should also be darker(Pal.adminChat or sth?)

Legomonster33:
> > also make it affect the server's list
> 
> currently it is sorted by ping, i don't think the admin count should play a role (maybe inside the section, but not outside)

server list sorting options when

braainsiogame:
> > > also make it affect the server's list
> > 
> > 
> > currently it is sorted by ping, i don't think the admin count should play a role (maybe inside the section, but not outside)
> 
> server list sorting options when

yes

Anuken:
I believe I made a comment on this PR weeks ago, but my internet ended up cutting out. The gist of it: I think the server info box is already cluttered as is, and this information doesn't seem important enough to list there. Since admins are displayed on the top of the player list, you can already check for admin presence easily enough after connecting.

PR #3855, opened 2020-12-08
[QOL] Ability to Disable Weather Effects 

AFFalcons:
Small Quality of Life change to allow users to disable weather effects in the settings menu. 

Still, a draft sill going to disable weather-related code in other areas of the game to stop "unused functions" from running while the user has weather disabled. I am currently only disabling the rendering of weather, the weather still exists just not visible to the user. 

As this is my first time contributing looking for suggestions or comments.

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -832,6 +832,7 @@ setting.chatopacity.name = Chat Opacity
 setting.lasersopacity.name = Power Laser Opacity
 setting.bridgeopacity.name = Bridge Opacity
 setting.playerchat.name = Display Player Bubble Chat
+setting.enableweather.name = Enable Weather
 public.confirm = Do you want to make your game public?\n[accent]Anyone will be able to join your games.\n[lightgray]This can be changed later in Settings->Game->Public Game Visibility.
 public.beta = Note that beta versions of the game cannot make public lobbies.
 uiscale.reset = UI scale has been changed.\nPress "OK" to confirm this scale.\n[scarlet]Reverting and exiting in[accent] {0}[] seconds...
--- a/core/src/mindustry/ui/dialogs/SettingsMenuDialog.java
+++ b/core/src/mindustry/ui/dialogs/SettingsMenuDialog.java
@@ -404,6 +404,7 @@ public class SettingsMenuDialog extends SettingsDialog{
         graphics.checkPref("fps", false);
         graphics.checkPref("playerindicators", true);
         graphics.checkPref("indicators", true);
+        graphics.checkPref("enableweather", true);
         graphics.checkPref("animatedwater", true);
         if(Shaders.shield != null){
             graphics.checkPref("animatedshields", !mobile);
--- a/core/src/mindustry/type/Weather.java
+++ b/core/src/mindustry/type/Weather.java
@@ -314,7 +314,7 @@ public class Weather extends UnlockableContent{
 
         @Override
         public void draw(){
-            if(renderer.weatherAlpha() > 0.0001f){
+            if(renderer.weatherAlpha() > 0.0001f && Core.settings.getBool("enableweather")){
                 Draw.draw(Layer.weather, () -> {
                     weather.rand.setSeed(0);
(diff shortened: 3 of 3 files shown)


sk7725:
Maybe when a weather render is disabled, we can have a "rain" or "snow" icon pop up below the minimap to indicate that the specific weather is ongoing instead(like those icons on weather reports).

sk7725:
the sandstorm really lags, the others are not that bad(although the spore and fog still shave a full 10fps off)

AFFalcons:
@sk7725 One idea I had to make it more efficient was to just not add the weather to the `rules.weather` if it is toggled off. In this case, it wouldn't even try to simulate weather. It would only check the setting once when the sector is loaded. 

I would have to figure out of to add the possible weather elements back to `rules.weather` if the player toggled it while inside of a sector.

AFFalcons:
This would be what the setting would stop from happening.
`
        //TODO bad code
        boolean hasSnow = floors[0].name.contains("ice") || floors[0].name.contains("snow");
        boolean hasRain = !hasSnow && content.contains(Liquids.water) && !floors[0].name.contains("sand");
        boolean hasDesert = !hasSnow && !hasRain && floors[0] == Blocks.sand;
        boolean hasSpores = floors[0].name.contains("spore") || floors[0].name.contains("moss") || floors[0].name.contains("tainted");


        if(hasSnow){
            state.rules.weather.add(new WeatherEntry(Weathers.snow));
        }


        if(hasRain){
            state.rules.weather.add(new WeatherEntry(Weathers.rain));
            state.rules.weather.add(new WeatherEntry(Weathers.fog));
        }


        if(hasDesert){
            state.rules.weather.add(new WeatherEntry(Weathers.sandstorm));
        }


        if(hasSpores){
            state.rules.weather.add(new WeatherEntry(Weathers.sporestorm));
        }
`
This would make it impossible to show what weather condition should be rendered with an icon though.

EyeOfDarkness:
What if i told you, that some weathers are meant to obstruct your vision?

AFFalcons:
> What if I told you, that some weathers are meant to obstruct your vision?

I know that that is the primary purpose of the weather effects, but while watching a youtube video about the game the Youtuber was trying to find a way to turn off the weather for video reasons.

sk7725:
> @sk7725 One idea I had to make it more efficient was to just not add the weather to the `rules.weather` if it is toggled off. In this case, it wouldn't even try to simulate weather. It would only check the setting once when the sector is loaded.
> 
> I would have to figure out how to add the possible weather elements back to `rules.weather` if the player toggled it while inside of a sector.

Weather is there to make the game harder, as it effects certain blocks. The visuals can be turned off, but the effects should remain. Also, why not make weather a UnlockableContent, so they have icons?

core/assets/bundles/bundle.properties
```diff
@@ -832,6 +832,7 @@ setting.chatopacity.name = Chat Opacity
 setting.lasersopacity.name = Power Laser Opacity
 setting.bridgeopacity.name = Bridge Opacity
 setting.playerchat.name = Display Player Bubble Chat
+setting.enableweather.name = Enable Weather
```
YellOw139, line 835:
```suggestion
setting.enableweather.name = Show Weather Graphics
```
You aren't actually toggling the weather system itself, just the graphics. This should make it clear.

core/src/mindustry/ui/dialogs/SettingsMenuDialog.java
```diff
@@ -404,6 +404,7 @@ void addSettings(){
         graphics.checkPref("fps", false);
         graphics.checkPref("playerindicators", true);
         graphics.checkPref("indicators", true);
+        graphics.checkPref("enableweather", true);
```
Voz-Duh, line 407: Rename it to "showweather"

QmelZ:
I tested all weathers, and these are the results:

No weather: 60 fps

Snow: 55-60 fps

Rain: 50-55 fps

Sandstorm: 10 fps

Sporestorm: 60 fps

Fog: 50 fps

QmelZ:
thank you anuke

PR #3857, opened 2020-12-08
"[E] to enable building" hint while building is paused

joshuaptfan:
Currently, when "auto-pause building" is off, no UI hint is shown when players press E to toggle pause/resume building, until they try to build something. Thus, they are left to guess what the current state of building is if they have pressed E multiple times. If they guess wrong, this can cause them to accidentally start building something when they only wanted to create a plan.

This PR makes it so that an "[E] to enable building" hint is a persistent indicator of the state of building (always displayed when building is paused) when auto-pause is off, regardless if building plans exist. It does not change how the hint is displayed when auto-pause is on.

### DIFF
--- a/core/src/mindustry/input/DesktopInput.java
+++ b/core/src/mindustry/input/DesktopInput.java
@@ -61,18 +61,33 @@ public class DesktopInput extends InputHandler{
         group.fill(t -> {
             t.bottom();
             t.visible(() -> {
-                t.color.a = Mathf.lerpDelta(t.color.a, !isBuilding && !buildWasAutoPaused || player.unit().isBuilding() ? 1f : 0f, 0.15f);
+                t.color.a = Mathf.lerpDelta(t.color.a, player.unit().isBuilding() ? 1f : 0f, 0.15f);
 
                 return ui.hudfrag.shown && Core.settings.getBool("hints") && selectRequests.isEmpty() && t.color.a > 0.01f;
             });
             t.touchable(() -> t.color.a < 0.1f ? Touchable.disabled : Touchable.childrenOnly);
             t.table(Styles.black6, b -> {
                 b.defaults().left();
-                b.label(() -> Core.bundle.format(!isBuilding ?  "resumebuilding" : "pausebuilding", Core.keybinds.get(Binding.pause_building).key.toString())).style(Styles.outlineLabel);
+                b.label(() -> Core.bundle.format(isBuilding ? "pausebuilding" : "resumebuilding", Core.keybinds.get(Binding.pause_building).key.toString())).style(Styles.outlineLabel);
                 b.row();
                 b.label(() -> Core.bundle.format("cancelbuilding", Core.keybinds.get(Binding.clear_building).key.toString())).style(Styles.outlineLabel);
                 b.row();
                 b.label(() -> Core.bundle.format("selectschematic", Core.keybinds.get(Binding.schematic_select).key.toString())).style(Styles.outlineLabel);
+            }).margin(10f);
+        });
+
+        //build enable hint
+        group.fill(t -> {
+            t.bottom();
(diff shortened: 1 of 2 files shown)


joshuaptfan:
Ideally, "[Q] to clear plan, [F] to select+copy" would only display when building plans exist, making the building pause indicator less obtrusive. I'm currently having trouble implementing this.

joshuaptfan:
New smaller hint
enable building

core/src/mindustry/input/DesktopInput.java
```diff
@@ -68,14 +68,28 @@ public void buildUI(Group group){
...
                 b.row();
                 b.label(() -> Core.bundle.format("cancelbuilding", Core.keybinds.get(Binding.clear_building).key.toString())).style(Styles.outlineLabel);
                 b.row();
                 b.label(() -> Core.bundle.format("selectschematic", Core.keybinds.get(Binding.schematic_select).key.toString())).style(Styles.outlineLabel);
             }).margin(10f);
         });
 
+        //build enable hint
+        group.fill(t -> {
+            t.bottom();
```
Anuken, line 81: You should probably initialize this table with `t.color.a = 0` so it starts out hidden.

joshuaptfan: Then shouldn't that also be applied to the general building hint table above?

Anuken: The difference is that this one starts out visible, while the other hints dialog does not.

joshuaptfan: Testing without `t.color.a = 0`, that line doesn't seem to be necessary. Tested these scenarios:

* Starting fresh custom game with auto-pause off: build enable hint not visible until pressing E.
* Resuming saved game with building disabled: build enable hint visible as expected, disappears when pressing E.
* Resuming saved game with building enabled: build enable hint not visible as expected, appears when pressing E.
* Changing auto-pause setting during game: some anomalous behavior, like having to press E twice to get hint to appear.

[`isBuilding` starts out `true`](https://github.com/Anuken/Mindustry/blob/0cc9b0b0a3b7a30ae1dab30d3f194ac04f7152fc/core/src/mindustry/input/InputHandler.java#L60), which means the build enable hint is invisible at the start of the game.

joshuaptfan: Fixed inconsistencies by replacing `!buildWasAutoPaused` with `!Core.settings.getBool("buildautopause")`. Now the build enable hint always reflects that state of `isBuilding` accurately, when auto-pause is disabled.

joshuaptfan:
All edge cases have been fixed as far as I can tell, with minimal increase in code. The PR currently passes all these tests:
* Press E to toggle building multiple times
  * "[E] to enable building" should fade out
* Build something
  * "[E] to pause building" should fade out
* Pause game, create build plan, toggle building multiple times, unpause, build
  * "[E] to pause building" should fade out
* Enable auto-pause and build
  * "[E] to pause/resume building" should appear as expected
  * "[E] to enable building" should never appear
* Disable auto-pause
  * "[E] to enable building" should appear immediately, and disappear when E is pressed

This partial state machine may help with understanding some of the logic. It shows the necessity of adding the `wasBuilding` variable to track the previous state of `isBuilding`.

> build disabled
> player not building
> "[E] to enable building"
> 
> build disabled -> build enabled
> player not building
> "[E] to enable building" fade out
> 
> build enabled
> player building
> "[E] to pause/resume building"
> 
> build enabled
> player building -> player not building
> "[E] to pause/resume building" fade out
> 
> build enabled
> player not building
> (hidden)

Anuken:
As reported in #4642, this draws over the respawn text. I don't see an easy fix for this, so I'm reverting this change until further notice.

PR #3858, opened 2020-12-08
Add block region for the lighting rule

Quezler:
different build visibility than the other sandbox blocks (amongst other things), would make some sense to give it its own 🤔 

(debating on whether or not to give the resupply point an ammo region or not, but keeping it in units makes the most sense)


Anuken:
But there's only one light-related block, and there are unlikely to be any more.

Quezler:
alternatively they could be grouped together, but that'll probably mess up the content id orders? (feel free to close)

Quezler:
the diff already looks horrible, i'll save you the trouble 👍

PR #3859, opened 2020-12-08
Update bundle_es.properties

Angel-24:
Those new hints


Angel-24:
So, when will this be merged?

PR #3860, opened 2020-12-08
Add ALEX v6 servers to servers_v6.json

alexpvpmindustry:
brand new PVP/survival server for bleeding edge and custom builds :D 🦾 🦾 🦾


kiVts (commented):
Is this the only file you need to modify? are you sure?
also I`m  not good in network and thing s

alexpvpmindustry:
yes. im only adding the server ip address. im not sure why the "java14 build" failed.

should i repull the request?

Anuken:
> bleeding edge and custom builds

This is the wrong list for that. Modify `severs_be` instead.

PR #3870, opened 2020-12-08
add @commanded for units

DeltaNedas:
allows processors to detect units commanded by a player (`controller instanceof FormationAI`)
implements https://github.com/Anuken/Mindustry-Suggestions/issues/1490, can be used alongside @controlled to detect players or their squads


Quezler (approved):

Erisfiregamer1:
Can we just merge this, so I can make an enemy controlled unit killer?

joshuaptfan:
This still does not allow logic to distinguish player-controlled units separately from logic-controlled units, so it does not implement that suggestion.

PR #3885, opened 2020-12-09
Update servers_v6.json

TNT331:
SubZero

minty-server.ddns.net

testing server


Volas171 (changes requested):
add the recently added servers so it doesn’t conflict

TNT331:
> add the recently added servers so it doesn’t conflict

huh

TNT331:
> add the recently added servers so it doesn’t conflict

also HI VOLAS what i do

PR #3886, opened 2020-12-09
Shadow For Item

Voz-Duh:
image
image
It is not very visible but slightly improves the game


Voz-Duh:
Summet! Don't say me what this don't compile

DeltaNedas:
> items levitate above conveyors

Voz-Duh:
> > items levitate above conveyors

No, but they still don't let light through

PR #3900, opened 2020-12-10
Add moveTarget for Mobile

sk7725:
Add moveTarget, for modded blocks that have mech(unit) pad-like properties, or require a certain closeness between the player and the block, which may be inconvenient to achieve on mobile.


sk7725:
I will wait for the travis to go brrrrrr

sk7725:
Sorry, but I forgot how to convert to draft.

Anuken:
I'd rather not have mod-specific input features. I won't maintain or use this in any way, and when it inevitably breaks, I'm not going to fix it because I won't even know what the use case is supposed to be.

sk7725:
Can you at least make it so the input's targetPos can be manipulated? Right now it just sets to the camera, peroid.

PR #3908, opened 2020-12-10
Delete old bleeding edge builds on update

buthed010203:
Lots of people have bleeding edge folders that look something like this lol

### DIFF
--- a/core/src/mindustry/net/BeControl.java
+++ b/core/src/mindustry/net/BeControl.java
@@ -50,7 +50,7 @@ public class BeControl{
                 Fi dest = Fi.get(System.getProperty("becopy"));
                 Fi self = Fi.get(BeControl.class.getProtectionDomain().getCodeSource().getLocation().toURI().getPath());
                 
-                for(Fi file : self.parent().findAll(f -> f != self)) file.delete();
+                self.parent().findAll(f -> f != self).forEach(Fi::delete);
 
                 self.copyTo(dest);
             }catch(Throwable e){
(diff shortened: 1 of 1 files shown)


core/src/mindustry/net/BeControl.java
```diff
@@ -49,6 +49,8 @@ public BeControl(){
             try{
                 Fi dest = Fi.get(System.getProperty("becopy"));
                 Fi self = Fi.get(BeControl.class.getProtectionDomain().getCodeSource().getLocation().toURI().getPath());
+                
+                self.parent().findAll(f -> f != self).forEach(Fi::delete);
```
Anuken, line 53: 1. **Do not** use forEach, that is a Java 8 feature.
2. Have you tested this?
3. Do not use reference comparison for files.

buthed010203: I have tested this and it works fine, how would you suggest that I avoid deleting the file without comparing references? If it's built with java 14 then whats the issue with using java 8 features?

Anuken: `findAll` creates new file objects, which means reference comparison is meaningless. You're just deleting the entire contents of the folder. [Use equals.](https://stackoverflow.com/questions/7520432/what-is-the-difference-between-and-equals-in-java)

> If it's built with java 14 then whats the issue with using java 8 features?

[Read the contributing guidelines. ](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md#do-not-use-incompatible-java-features-javautilfunction-javaawt) The version built with is irrelevant.

Anuken:
I'll assume you've tested this and it doesn't explode; I'm not maintaining the BE build update system currently, so if it breaks things for other platforms, I won't be fixing it.

PR #3910, opened 2020-12-10
Respawning player won't respawn in a formation

buthed010203:
There is nothing more infuriating than someone being an idiot and dying on loop and stealing all of your formation units, the purpose of this pr is to reduce the likelihood of that happening. Players wont automatically respawn as a unit that is part of someone's formation anymore.


Quezler:
and what if the unit is part of the formation the user is currently controlling?

buthed010203:
> and what if the unit is part of the formation the user is currently controlling?
> (depends on the death race condition*)

I uh forgot to test on servers and ur right, it doesnt switch to one of your own commanded units

PR #3912, opened 2020-12-10
Conveyor upgrade system

Slava0135:
Peek 2020-12-10 22-04

No additional keybinds. Just use Diagonal placing.

### DIFF
--- a/core/src/mindustry/input/InputHandler.java
+++ b/core/src/mindustry/input/InputHandler.java
@@ -32,6 +32,7 @@ import mindustry.ui.fragments.*;
 import mindustry.world.*;
 import mindustry.world.blocks.*;
 import mindustry.world.blocks.ConstructBlock.*;
+import mindustry.world.blocks.distribution.ChainedBuilding;
 import mindustry.world.blocks.payloads.*;
 import mindustry.world.blocks.power.*;
 import mindustry.world.blocks.storage.CoreBlock.*;
@@ -1161,7 +1162,15 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
         }
 
         if(diagonal){
-            points = Placement.pathfindLine(block != null && block.conveyorPlacement, startX, startY, endX, endY);
+            Tile start = world.tile(startX, startY);
+            Tile end = world.tile(endX, endY);
+            if(block != null && block instanceof Autotiler
+                    && start.build instanceof ChainedBuilding && end.build instanceof ChainedBuilding
+                    && block.canReplace(end.build.block) && block.canReplace(start.build.block)){
+                points = Placement.upgradeLine(startX, startY, endX, endY);
+            }else{
+                points = Placement.pathfindLine(block != null && block.conveyorPlacement, startX, startY, endX, endY);
+            }
         }else{
             points = Placement.normalizeLine(startX, startY, endX, endY);
         }
--- a/core/src/mindustry/input/InputHandler.java
+++ b/core/src/mindustry/input/InputHandler.java
@@ -32,7 +32,7 @@ import mindustry.ui.fragments.*;
 import mindustry.world.*;
 import mindustry.world.blocks.*;
 import mindustry.world.blocks.ConstructBlock.*;
-import mindustry.world.blocks.distribution.ChainedBuilding;
+import mindustry.world.blocks.distribution.*;
 import mindustry.world.blocks.payloads.*;
 import mindustry.world.blocks.power.*;
 import mindustry.world.blocks.storage.CoreBlock.*;
--- a/core/src/mindustry/input/Placement.java
+++ b/core/src/mindustry/input/Placement.java
@@ -5,7 +5,7 @@ import arc.math.*;
 import arc.math.geom.*;
 import arc.struct.*;
 import arc.util.pooling.*;
-import mindustry.gen.Building;
+import mindustry.gen.*;
 import mindustry.world.*;
 import mindustry.world.blocks.distribution.*;
 
@@ -60,7 +60,7 @@ public class Placement{
         points.clear();
         Building building = world.tile(startX, startY).build;
         points.add(Pools.obtain(Point2.class, Point2::new).set(startX, startY));
-        while((building.tile.x != endX || building.tile.y != endY)){
+        while(building.tile.x != endX || building.tile.y != endY){
             ChainedBuilding chained = (ChainedBuilding)building;
             if(chained.next() == null) return pathfindLine(true, startX, startY, endX, endY);
             building = chained.next();
--- a/core/src/mindustry/input/Placement.java
+++ b/core/src/mindustry/input/Placement.java
@@ -5,7 +5,9 @@ import arc.math.*;
 import arc.math.geom.*;
 import arc.struct.*;
 import arc.util.pooling.*;
+import mindustry.gen.Building;
 import mindustry.world.*;
+import mindustry.world.blocks.distribution.*;
 
 import static mindustry.Vars.*;
 
@@ -23,7 +25,6 @@ public class Placement{
     /** Normalize a diagonal line into points. */
     public static Seq<Point2> pathfindLine(boolean conveyors, int startX, int startY, int endX, int endY){
         Pools.freeAll(points);
-
         points.clear();
         if(conveyors && Core.settings.getBool("conveyorpathfinding")){
             if(astar(startX, startY, endX, endY)){
@@ -50,6 +51,20 @@ public class Placement{
             for(int i = 0; i <= Math.abs(startY - endY); i++){
                 points.add(Pools.obtain(Point2.class, Point2::new).set(startX, startY + i * Mathf.sign(endY - startY)));
             }
+        }
+        return points;
+    }
+
+    public static Seq<Point2> upgradeLine(int startX, int startY, int endX, int endY){
+        Pools.freeAll(points);
+        points.clear();
+        Building building = world.tile(startX, startY).build;
+        points.add(Pools.obtain(Point2.class, Point2::new).set(startX, startY));
+        while((building.tile.x != endX || building.tile.y != endY)){
+            ChainedBuilding chained = (ChainedBuilding)building;
+            if(chained.next() == null) return pathfindLine(true, startX, startY, endX, endY);
+            building = chained.next();
+            points.add(Pools.obtain(Point2.class, Point2::new).set(building.tile.x, building.tile.y));
         }
         return points;
     }
--- a/core/src/mindustry/world/blocks/distribution/ChainedBuilding.java
+++ b/core/src/mindustry/world/blocks/distribution/ChainedBuilding.java
@@ -1,0 +1,7 @@
+package mindustry.world.blocks.distribution;
+
+import mindustry.gen.Building;
+
+public interface ChainedBuilding{
+    Building next();
+}
--- a/core/src/mindustry/world/blocks/distribution/ChainedBuilding.java
+++ b/core/src/mindustry/world/blocks/distribution/ChainedBuilding.java
@@ -1,6 +1,6 @@
 package mindustry.world.blocks.distribution;
 
-import mindustry.gen.Building;
+import mindustry.gen.*;
 
 public interface ChainedBuilding{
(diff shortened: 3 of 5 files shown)


DeltaNedas:
good

Slava0135:
since I am not nerd there might be a mess ~~and unintentional Java.util imports~~

core/src/mindustry/input/InputHandler.java
```diff
@@ -1161,7 +1162,17 @@ void iterateLine(int startX, int startY, int endX, int endY, Cons<PlaceLine> con
...
         if(diagonal){
-            points = Placement.pathfindLine(block != null && block.conveyorPlacement, startX, startY, endX, endY);
+            if(block != null && block instanceof Conveyor){
+                Tile start = world.tile(startX, startY);
+                Tile end = world.tile(endX, endY);
+                if(start.block() instanceof Conveyor && end.block() instanceof Conveyor){
+                    points = Placement.upgradeConveyorLine(startX, startY, endX, endY);
+                }else{
+                    points = Placement.pathfindLine(true, startX, startY, endX, endY);
+                }
```
Quezler, line 1172: these don't seem to be used anywhere other than the if right below it`:

```suggestion
                if(world.tile(startX, startY).block() instanceof Conveyor && world.tile(endX, endY).block() instanceof Conveyor){
                    points = Placement.upgradeConveyorLine(startX, startY, endX, endY);
                }else{
                    points = Placement.pathfindLine(true, startX, startY, endX, endY);
                }
```

Slava0135:
Peek 2020-12-10 22-22

Slava0135:
Peek 2020-12-10 22-27
everything works fine with other blocks

Slava0135:
by "fine" I mean it doesn't mess stuff up.  Placement of armored versions could be improved.

Slava0135:
so now it checks if block you are holding is different from the block you point to

Anuken:
> everything works fine with other blocks

That doesn't look fine to me. What if I want to drag a line across and replace all the blocks in way? The line of upgrades should not be followed unless it contains both endpoints.

Slava0135:
> > everything works fine with other blocks
> 
> That doesn't look fine to me. What if I want to drag a line across and replace all the blocks in way? The line of upgrades should not be followed unless it contains both endpoints.

So you want to build a path check if it you can get to the other end and only then return it?

Slava0135:
But the upgrading works only if you hold not the same block as one at the starting position. So I don't see big problems with that. Also you can use straight placement mode for this. Anyway we can try this version first and then get some feedback.

Slava0135:
@Anuken so now it works as you said
Peek 2020-12-11 10-08

core/src/mindustry/input/Placement.java
```diff
@@ -54,6 +55,32 @@
         return points;
     }
 
+    public static Seq<Point2> upgradeConveyorLine(int startX, int startY, int endX, int endY){
```
Anuken, line 58: These two functions are nearly the same, and use conduit/conveyor-specific implementation details. This isn't acceptable.

core/src/mindustry/input/InputHandler.java
```diff
@@ -1161,7 +1163,15 @@ void iterateLine(int startX, int startY, int endX, int endY, Cons<PlaceLine> con
         }
 
         if(diagonal){
-            points = Placement.pathfindLine(block != null && block.conveyorPlacement, startX, startY, endX, endY);
+            Tile start = world.tile(startX, startY);
+            Tile end = world.tile(endX, endY);
+            if(block != null && block instanceof Conveyor && start.block() instanceof Conveyor && end.block() instanceof Conveyor && start.block() != block){
```
Anuken, line 1168: Conduits/conveyors should not be hardcoded into the system with special functions to handle them. The logic for each type of block is practically identical.

core/src/mindustry/input/Placement.java
Slava0135: Yeah but Conduit and Conveyor is not the same thing. And also there won't be any new such blocks.

core/src/mindustry/input/InputHandler.java
Slava0135: I really don't know how to put them under one roof. Perhaps do something with AutoTiler?

DeltaNedas: add it to an interface

core/src/mindustry/world/blocks/liquid/Conduit.java
```diff
@@ -119,6 +119,16 @@ public void onProximityUpdate(){
...
         }
 
+        @Nullable
+        public ConduitBuild nextConduit(){
+            Tile next = tile.nearby(rotation);
+            if(next != null && next.build instanceof ConduitBuild){
+                return (ConduitBuild) next.build;
+            }else{
+                return null;
+            }
```
joshuaptfan, line 129: All this could be shortened by using ternary:

`return next != null && next.build instanceof ConduitBuild ? (ConduitBuild) next.build : null;`

Slava0135:
@Anuken and it's done, I guess?

Slava0135:
and I forgot about style, yes

Slava0135:
Peek 2020-12-12 22-23

Slava0135:
Peek 2020-12-12 22-31
*there is one little problem*

core/src/mindustry/input/InputHandler.java
```diff
@@ -1165,10 +1164,13 @@ void iterateLine(int startX, int startY, int endX, int endY, Cons<PlaceLine> con
         if(diagonal){
             Tile start = world.tile(startX, startY);
             Tile end = world.tile(endX, endY);
-            if(block != null && block instanceof Conveyor && start.block() instanceof Conveyor && end.block() instanceof Conveyor && start.block() != block){
-                points = Placement.upgradeConveyorLine(startX, startY, endX, endY);
-            }else if(block != null && block instanceof Conduit && start.block() instanceof Conduit && end.block() instanceof Conduit && start.block() != block){
-                points = Placement.upgradeConduitLine(startX, startY, endX, endY);
+            if(block != null && block instanceof Autotiler
+                    && start.block() != block
```
Slava0135, line 1168: Actually this line is kinda redundant but idk if's good for performance or not.

Slava0135:
squashed all commits

core/src/mindustry/world/blocks/distribution/AutotilerBuilding.java
```diff
@@ -0,0 +1,9 @@
+package mindustry.world.blocks.distribution;
+
+import mindustry.gen.Building;
+import mindustry.world.Block;
+
+public interface AutotilerBuilding{
```
Anuken, line 6: Why is this called `AutotilerBuilding` when it has nothing to do with tiling itself...?

core/src/mindustry/world/blocks/distribution/AutotilerBuilding.java
```diff
@@ -0,0 +1,9 @@
...
+    Building next();
+    Block getBlockBasedOn();
```
Anuken, line 8: Why not just use `build.block`?

core/src/mindustry/input/Placement.java
```diff
@@ -54,6 +55,20 @@
         return points;
     }
 
+    public static Seq<Point2> upgradeLine(int startX, int startY, int endX, int endY){
+        Pools.freeAll(points);
+        points.clear();
+        Building building = world.tile(startX, startY).build;
+        points.add(Pools.obtain(Point2.class, Point2::new).set(startX, startY));
+        while((building.tile.x != endX || building.tile.y != endY)){
+            AutotilerBuilding autotiler = (AutotilerBuilding) building;
```
Anuken, line 64: unnecessary space before cast

core/src/mindustry/input/Placement.java
```diff
@@ -5,7 +5,9 @@
 import arc.math.geom.*;
 import arc.struct.*;
 import arc.util.pooling.*;
+import mindustry.gen.Building;
 import mindustry.world.*;
+import mindustry.world.blocks.distribution.AutotilerBuilding;
```
Anuken, line 10: imports should be *

core/src/mindustry/input/InputHandler.java
```diff
@@ -1161,7 +1162,18 @@ void iterateLine(int startX, int startY, int endX, int endY, Cons<PlaceLine> con
...
 
         if(diagonal){
-            points = Placement.pathfindLine(block != null && block.conveyorPlacement, startX, startY, endX, endY);
+            Tile start = world.tile(startX, startY);
+            Tile end = world.tile(endX, endY);
+            if(block != null && block instanceof Autotiler
+                    && start.block() != block
+                    && start.build instanceof AutotilerBuilding
+                    && end.build instanceof AutotilerBuilding
+                    && block.canReplace(((AutotilerBuilding) end.build).getBlockBasedOn())
```
Anuken, line 1171: unnecessary spaces before cast, line below too

Anuken (changes requested):

core/src/mindustry/world/blocks/distribution/AutotilerBuilding.java
Slava0135: Because I check if block is Autotiler in InputHandler. Both conduits and conveyors are inherited from it.

Slava0135:
e V e R y T h I n G i S f I n E
Except maybe some naming issues. Conveyors/Conduits are inherited from Autotiler and I decided to rename my interface to ChainedBuilding as it make sense.

core/src/mindustry/input/Placement.java
```diff
@@ -54,6 +55,20 @@
         return points;
     }
 
+    public static Seq<Point2> upgradeLine(int startX, int startY, int endX, int endY){
+        Pools.freeAll(points);
+        points.clear();
+        Building building = world.tile(startX, startY).build;
+        points.add(Pools.obtain(Point2.class, Point2::new).set(startX, startY));
+        while((building.tile.x != endX || building.tile.y != endY)){
```
summetdev, line 63: why place double brackets here?

```suggestion
        while(building.tile.x != endX || building.tile.y != endY){
```

core/src/mindustry/input/Placement.java
```diff
@@ -5,7 +5,9 @@
 import arc.math.geom.*;
 import arc.struct.*;
 import arc.util.pooling.*;
+import mindustry.gen.Building;
```
summetdev, line 8:
```suggestion
import mindustry.gen.*;
```

core/src/mindustry/input/InputHandler.java
```diff
@@ -32,6 +32,7 @@
 import mindustry.world.*;
 import mindustry.world.blocks.*;
 import mindustry.world.blocks.ConstructBlock.*;
+import mindustry.world.blocks.distribution.ChainedBuilding;
```
summetdev, line 35: imports should be *

```suggestion
import mindustry.world.blocks.distribution.*;
```

core/src/mindustry/world/blocks/distribution/ChainedBuilding.java
```diff
@@ -0,0 +1,7 @@
+package mindustry.world.blocks.distribution;
+
+import mindustry.gen.Building;
```
summetdev, line 3:
```suggestion
import mindustry.gen.*;
```

core/src/mindustry/world/blocks/liquid/Conduit.java
```diff
@@ -16,6 +16,7 @@
 import mindustry.type.*;
 import mindustry.world.*;
 import mindustry.world.blocks.*;
+import mindustry.world.blocks.distribution.ChainedBuilding;
```
summetdev, line 19: and here

```suggestion
import mindustry.world.blocks.distribution.*;
```

summetdev (changes requested):

Slava0135:
and now it is ready. I guess

pixaxeofpixie:
wow, i wait this feature from 5.0,its cool.

summetdev (approved):

PR #3920, opened 2020-12-11
Tanks are able to replace conduits

Slava0135:
Peek 2020-12-11 10-24

I it is always irritating me that you have to remove conduits before placing a tank - the liquid is going to leak. Also when the said liquid is slag it might catch on fire the conduit.

### DIFF
--- a/core/src/mindustry/world/blocks/liquid/LiquidRouter.java
+++ b/core/src/mindustry/world/blocks/liquid/LiquidRouter.java
@@ -2,6 +2,8 @@ package mindustry.world.blocks.liquid;
 
 import mindustry.gen.*;
 import mindustry.type.*;
+import mindustry.world.Block;
+import mindustry.world.meta.BlockGroup;
 
 public class LiquidRouter extends LiquidBlock{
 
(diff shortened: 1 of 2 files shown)


core/src/mindustry/world/blocks/liquid/LiquidRouter.java
```diff
@@ -2,6 +2,8 @@
 
 import mindustry.gen.*;
 import mindustry.type.*;
+import mindustry.world.Block;
+import mindustry.world.meta.BlockGroup;
```
summetdev, line 6: use wildcard imports:

```suggestion
import mindustry.world.*;
import mindustry.world.meta.*;
```

summetdev (changes requested):

core/src/mindustry/world/blocks/liquid/LiquidRouter.java
```diff
@@ -24,4 +26,10 @@ public boolean acceptLiquid(Building source, Liquid liquid){
             return (liquids.current() == liquid || liquids.currentAmount() < 0.2f);
         }
     }
+
+    @Override
+    public boolean canReplace(Block other){
+        if(other.alwaysReplace) return true;
+        return this.group != BlockGroup.none && other.group == this.group && other instanceof Conduit;
```
Anuken, line 33: This prevents tanks from replacing liquid routers.

Slava0135:
so now it works as it should

Slava0135:
Peek 2020-12-11 19-17

Anuken:
Using `instanceof` checks still seems messy, I'll implement this with special group flags.

Slava0135:
ok then

PR #3929, opened 2020-12-11
copy-paste tool for map editor

jakubDoka:
In this pr i will (maybe) implement copy-paste for map editor. I attempted this before, but i chosen a wrong time and i also wasn't very experienced (absolute fool) with java and github. 

To describe this shortly, pr will add copy-paste tool, that will draw rectangle with adjustable size (size may be limited but that depends on your decision). You would then move with selection by clicking to selected area and dragging it. 

Now there is a problem with pasting and changing a selection. My idea is to bind paste to a key and deselect by simply clicking of the rectangle. It's not really consistent with other tools, though, I do not see any other option, other than making select and paste mode, which can slow down editing process for a user.

### DIFF
--- a/core/src/mindustry/editor/CopyData.java
+++ b/core/src/mindustry/editor/CopyData.java
@@ -1,0 +1,6 @@
+package mindustry.editor;
+
+// Class that holds data about selected tiles.
+// It handles rotation of selection.
+public class CopyData {
+}


summetdev:
maybe mark this PR as draft? :eyes:

DeltaNedas:
you know you can just not make a pr before writing a single line of code

jakubDoka:
why would i write anything before even knowing if its not pointless. I once wasted lot of time by writing code before masking pr

l-Luna:
you're supposed to write the code behind a PR first, then open the PR. ask on discord if you're not sure if its useful.

if you're not done with a PR (don't want it merged yet), mark it as a draft.

PR #3939, opened 2020-12-12
bring back block select timeout

DeltaNedas:
not sure why it was removed...

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -809,6 +809,7 @@ setting.conveyorpathfinding.name = Conveyor Placement Pathfinding
 setting.sensitivity.name = Controller Sensitivity
 setting.saveinterval.name = Save Interval
 setting.seconds = {0} seconds
+setting.blockselecttimeout.name = Block Select Timeout
 setting.milliseconds = {0} milliseconds
 setting.fullscreen.name = Fullscreen
 setting.borderlesswindow.name = Borderless Window[lightgray] (restart may be required)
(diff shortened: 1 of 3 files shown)


Anuken:
How about setting it at a default sane value instead?

DeltaNedas:
the default is the same as it was before, no?

Anuken:
Yes, but it's not good if people need to change it.

AbbyNode:
Personally, I prefer to keep it around 300-400ms but I'm not sure if that's the same with everyone. A hard-coded value that feels comfortable to one person may seem too fast or too slow to someone else. I think having the option is nice, at least for accessibility if nothing else.

DeltaNedas:
ok how about 500ms? nice and round

AbbyNode:
That would be better than 700ms, but I would really like to have the option to fine-tune it on the client. :)

Kyllingene:
>Yes, but it's not good if people need to change it.

@Anuken, why is that? Was there some performance issue? This purely affects client-side, so no multiplayer problems are apparent. Why was it removed?

Anuken:
> why is that?

Because defaults should be reasonable values. Does anybody use it at 700ms?

> Why was it removed?

I never used it, I wasn't the one who added it, and it wasn't clear if anyone needed it.

Kyllingene:
Well, I guess unused features are simply RAM and storage space waste.

Anuken:
It was just a big obvious slider at the top of my list of options with no clear purpose.

Kyllingene:
That's fair.

Anuken:
I'll keep the default at 400ms and see if someone else has issues with it.

PR #3947, opened 2020-12-12
Typo

rodrigolitzius:
H


Bohaska:
Why do you add typos...

firstcreator48:
how

rodrigolitzius:
> Why do you add typos...

some problem? I only fixed some Portuguese errors. nothing wrong.

rodrigolitzius:
> how

How what?

Anuken:
Google Translate seems to think that this is correct.

PR #3956, opened 2020-12-13
update KO bundles

sharlotte-mobile:
*update steam description

*update steam achievements

*update android description

*update android recent change-log

*update android summary

*update bundles KO


sk7725:
강강술래 뭔데

PR #3967, opened 2020-12-13
disallow descriptions after mod version

DeltaNedas:
This mod.json
```json
{
	"name": "aaa",
	"version": "1.0\nyes i am version",
	"minGameVersion": "105"
}
```

has a "description" `yes i am version` below the version.
the issue is that this has no bounds checking and can break the dialog, along with being a hack.
having `shortDescription` would be fine.

this pr addresses this by removing any extra lines in a mod's version field.

### DIFF
--- a/core/src/mindustry/mod/Mods.java
+++ b/core/src/mindustry/mod/Mods.java
@@ -647,6 +647,14 @@ public class Mods implements Loadable{
                 meta.hidden = true;
             }
 
+            //disallow putting a description after the version
+            if(meta.version != null){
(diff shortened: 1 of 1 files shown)


DeltaNedas:
before:
Screenshot_19:26:30
after:
Screenshot_19:45:07

DeltaNedas:
@Anuken do you think shortDescription should be a thing?

Anuken:
> do you think shortDescription should be a thing?

No.

PR #3974, opened 2020-12-14
Ignore Control Weapons

MEEPofFaith:
Allows weapons to ignore mass control, such as `singleTarget = true`, commander control, or player control.
[Video](https://youtu.be/HKTzDHnVBgQ)


Violet-Scarelli:
Time to add point defense to the Oct

MEEPofFaith:
> Time to add point defense to the Oct

ohyes

Anuken:
Interesting idea, but has the same problem as #4281: the code isn't great; weapon overrides would be preferable.

MEEPofFaith:
Yea it's a bit too much duplicated code.

PR #3977, opened 2020-12-14
retire CN from v5 Mindustry

L0615T1C5-216AC-9437:


L0615T1C5-216AC-9437:
reason: Planning to reuse ports and don't want v5 players to have a v6 server in community server lsit

PR #3991, opened 2020-12-15
allow controlling @this

DeltaNedas:
- `control enabled @this 0` yes yes
- not really any other uses


Anuken:
`control enabled @this 0`

why? that's an instant deadlock.

GreyMario:
It enables a processor to run once and done without forcing itself into an infinite loop of noop. I had to stop and think about it.

Anuken:
Not exactly. Once it disables itself, the action will time out eventually, since it isn't applied continuously. This leads to inconsistently periodic behavior.

GreyMario:
Oh. Right. That's a thing.

🤔

..The hell is the use case, then?

PR #3992, opened 2020-12-15
new statements are put in center of screen

DeltaNedas:
implements https://github.com/Anuken/Mindustry-Suggestions/issues/1555

TODO
- [X] initiate drag n drop
- [ ] actually put it in the center of the screen


DeltaNedas:
hold up

DeltaNedas:
ok so currently the issue is that i have no idea how to turn center of screen into a position for the new statement element

Overblob:
Hi, I really don't want to bother you, but have you considered my proposal (written in your suggestion) ? I think it would be easier to implement and more user-friendly when it comes to add new statements in the middle of the script .

DeltaNedas:
would require having a selection system which i dont even want to try

Anuken:
Stale, re-open later if implemented.

PR #4002, opened 2020-12-16
Update bundle_es.properties

Angel-24:
I made some clean up in my fork of Mindustry, so the changes to Spanish translations in that fork will be compared from the master branch, instead of any patch. Also, Spanish translations were already finished, for now. Thanks


Angel-24:
The older "Update bundle_es.properties" PR was closed.

PR #4006, opened 2020-12-16
allow mods to set zoom limits

DeltaNedas:
added Vars.renderer.{minZoom,maxZoom}.
should be set to old limits by default, need to check.

**when testing, minZoom is the limit for zooming out, maxZoom for zooming in**

### DIFF
--- a/core/src/mindustry/core/Renderer.java
+++ b/core/src/mindustry/core/Renderer.java
@@ -312,11 +312,11 @@ public class Renderer implements ApplicationListener{
     }
 
     public float minScale(){
-        return Scl.scl(minZoom);
+        return Math.round(Scl.scl(minZoom));
     }
 
     public float maxScale(){
-        return Scl.scl(maxZoom);
+        return Math.round(Scl.scl(maxZoom));
     }
 
     public float getScale(){
--- a/core/src/mindustry/core/Renderer.java
+++ b/core/src/mindustry/core/Renderer.java
@@ -31,7 +31,7 @@ public class Renderer implements ApplicationListener{
     public FrameBuffer effectBuffer = new FrameBuffer();
     public float laserOpacity = 1f;
     /** minZoom = zooming out, maxZoom = zooming in */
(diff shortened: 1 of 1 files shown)


core/src/mindustry/core/Renderer.java
```diff
@@ -302,16 +304,19 @@ public void scaleCamera(float amount){
...
         return camerascale;
     }
 
     public float minScale(){
-        return Scl.scl(1.5f);
+        return Scl.scl(minZoom);
+    }
+
+    public float maxScale(){
+        return Scl.scl(maxZoom);
```
Anuken, line 319: Where did the `Math.round` go?

DeltaNedas: \x7F

core/src/mindustry/core/Renderer.java
```diff
@@ -302,16 +304,19 @@ public void scaleCamera(float amount){
...
+        targetscale = Mathf.clamp(targetscale, minScale(), maxScale());
     }
 
     public float getDisplayScale(){
         return camerascale;
     }
 
     public float minScale(){
-        return Scl.scl(1.5f);
+        return Math.round(Scl.scl(minZoom));
```
Anuken, line 315: minScale wasn't rounded before.

DeltaNedas: aaaa

core/src/mindustry/core/Renderer.java
```diff
@@ -30,6 +30,8 @@
     public @Nullable Bloom bloom;
     public FrameBuffer effectBuffer = new FrameBuffer();
     public float laserOpacity = 1f;
+    /** minZoom = zooming out, maxZoom = zooming in */
+    public float minZoom = 1.5f, maxZoom = 1.5f * 6;
```
Anuken, line 34: hold up, why is this 1.5*6 instead of 6

DeltaNedas: oops i thought it was Scl.scl(1.5f) * 6f

PR #4007, opened 2020-12-16
Update speed of junctions

dhoell:
With the reduction of junction capacity their max transfer speed was also reduced, which breaks may factory designs. Made speed faster to compensate. Due to 121.3 patch notes I assume that the transfer speed change was unintentional? Please ignore if not.


Anuken:
> Due to 121.3 patch notes I assume that the transfer speed change was unintentional?

No, it was intentionally nerfed until it had about the same throughput as a titanium conveyor. ~12/s.

Issue #3966, opened 2020-12-13
Bridge conveyors


rodrigolitzius:
**Platform**: *Android and other platforms too*

**Build**: *20303.*

**Issue**: *[this.](https://1drv.ms/v/s!AnYf7oxSZjN8gUEU7O5ZlArVI7Zx)*

**Steps to reproduce**: *h.*

**Link(s) to mod(s) used**: *h.*

**Save file**: *i don't think that's required, because you can reproduce on any map easily.*

**(Crash) logs**: *h.*

---

*Place an X (no spaces) between the brackets to confirm that you have read the line below.*  
- [X] **I have updated to the latest release (https://github.com/Anuken/Mindustry/releases) to make sure my issue has not been fixed.**
- [X] **I have searched the closed and open issues to make sure that this problem has not already been reported.**

Quezler:
"bridge doesn't link in the way it gets shown during planning if it gets build in reverse due to distance or other causes"

rodrigolitzius:
That means what?

PR #4012, opened 2020-12-17
Bridge priority fix

Slava0135:
***BEFORE:***
before
***AFTER:***
after

### DIFF
--- a/core/src/mindustry/entities/comp/BuilderComp.java
+++ b/core/src/mindustry/entities/comp/BuilderComp.java
@@ -18,7 +18,7 @@ import mindustry.type.*;
 import mindustry.world.*;
 import mindustry.world.blocks.*;
 import mindustry.world.blocks.ConstructBlock.*;
-import mindustry.world.blocks.distribution.ItemBridge;
+import mindustry.world.blocks.distribution.*;
 
 import java.util.*;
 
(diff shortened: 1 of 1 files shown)


Slava0135:
however this prevents you from building other stuff if the bridge out of range.

itcannotbe:
How is that a non-issue in mobile?

Slava0135:
> How is that a non-issue in mobile?
> 
> The behavior is like after. Android 121.4

wdym

itcannotbe:
The behavior on android is not like your before gif. But the behavior is your after gif on mobile.

Slava0135:
> The behavior on android is not like your before gif. But the behavior is your after gif on mobile.

Behaviour on mobile is the same

rodrigolitzius:
I made a issue saying the same thing, but it was closed... #3966

itcannotbe:
> > The behavior on android is not like your before gif. But the behavior is your after gif on mobile.
> 
> Behaviour on mobile is the same

Quite literally can't reproduce the before gif on android though.

The behavior is the after gif.

Slava0135:
> > > The behavior on android is not like your before gif. But the behavior is your after gif on mobile.
> > 
> > 
> > Behaviour on mobile is the same
> 
> Quite literally can't reproduce the before gif on android though.
> 
> The behavior is the after gif.

You have to be outside of the range or have not enough resources.

Anuken:
I don't like this implementation - it's hacky and prevents requests from being prioritized properly. Bridges should somehow store their configs upon placement.

PR #4015, opened 2020-12-17
added conduit color indications

LottieVixen:
Noticed all the conduit 4 way junctions had no indication of type so added some color notes to it.

Original Left, this PR, right
image


Quezler:
my opinion is that the darker 1 pixel line needs to go, very visible on the armored conduit but not so much on pulse: 

Screen Shot 2020-12-17 at 14 56 19

not sure if the darker part of the palette should make an appearance elsewhere on the edge, but just not as 1 line 🤔

LottieVixen:
That is an artifact I believe because I haven't alpha blended them (as the packer does that I believe) 
image

Anuken:
How did you test this?

LottieVixen:
Loaded into a mod for testing, I realised now via pulls that bleeding was needed and has been done, will commit to the branch

Anuken:
> Loaded into a mod for testing

The mod texture packer doesn't bleed the same way the vanilla texture packer does, so it's unlikely that the dark artifacts would occur in the vanilla game. I don't bleed any of my sprites manually.

LottieVixen:
> 
> 
> > Loaded into a mod for testing
> 
> The mod texture packer doesn't bleed the same way the vanilla texture packer does, so it's unlikely that the dark artifacts would occur in the vanilla game. I don't bleed any of my sprites manually.

ah okay, well in that case it should be fine, will compile to check and then verify it is fine in a comment following this one

LottieVixen:
compiled with new sprite in this commit and PR, seems fine (sorry for not testing properly sooner)
image

I am confident this is ready for merge or closure.

PR #4028, opened 2020-12-18
Logic processor hollographic label

Quezler:
> not fit to be merged without further modification: imports / casting / comments / etc (mostly just a proof of concept)

**problem**
currently there is no way of knowing what processors do without opening them one by one.

**exhibit 1**
computer with a keyboard, tons of buttons that need their own message block:
(labeling anything other than processors is currently outside the scope)
https://cdn.discordapp.com/attachments/663063622791987203/789197420864602133/Mindustry_cYK2bm36BQ.png

**exhibit 2**
a thing that retreats units when their health is low, but no real sense as to which logic proccessor controls which type:
Screen Shot 2020-12-18 at 13 58 57

**exhibit 3**
a t5 feeder with 1 or 2 logic processors per resource, again no idea which one does what aside from bruteforce opening:

**concept**
being able to label processors (and possibly some other logic blocks?) with a sort of hologram of what they can do:
(ignore the display screens, was just debugging the drawing sizes with them)
Screen Shot 2020-12-18 at 13 58 46

Screen Shot 2020-12-18 at 13 59 11

Screen Shot 2020-12-18 at 14 08 47

**backstory**
took place on the nydus discord, shows some initial thoughts going on, somewhat different to what this resulted into:

Screen Shot 2020-12-18 at 14 09 22

Screen Shot 2020-12-18 at 14 09 50

Screen Shot 2020-12-18 at 14 10 01

**technical**
- first time i tried adding logic, so i probably did it wrong
- intelij updated and broke a lot of stuff (red), so had to code blind, imports & castings might be sub-optimal
- enabled the commented out comment block to make the dialog look nice (and gave them both a new color)

**future**
no idea actually, its more of a concept at this point and it might be neat if the icons could be put on e.g. buttons to, or even persist in schematic previews or maybe even cycle through several, who knows, for now ima leave this as a discussion 🍿

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -16,6 +16,7 @@ import mindustry.gen.*;
 import mindustry.type.*;
 import mindustry.world.*;
 import mindustry.world.blocks.logic.*;
+import mindustry.world.blocks.logic.LogicBlock.*;
 import mindustry.world.blocks.logic.LogicDisplay.*;
 import mindustry.world.blocks.logic.MemoryBlock.*;
 import mindustry.world.blocks.logic.MessageBlock.*;
@@ -176,6 +177,23 @@ public class LExecutor{
 
     public interface LInstruction{
         void run(LExecutor exec);
+    }
+
+    public static class LabelI implements LInstruction{
+        public int p1;
+
+        public LabelI(int p1){
+            this.p1 = p1;
+        }
+
+        @Override
+        public void run(LExecutor exec){
+            int icon = exec.obj(p1) instanceof UnlockableContent u ? u.iconId : 0;
+
+            if(exec.building(varThis) instanceof LogicBuild){
+                ((LogicBuild) exec.building(varThis)).label = icon;
+            }
+        }
     }
 
     /** Binds the processor to a unit based on some filters. */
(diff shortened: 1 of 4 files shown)


DeltaNedas:
ohyes

joshuaptfan:
Perhaps make the `print flush` command work for processors and switches as well, giving them message block functionality. Or heck, make it work for every logic block... or every block... the ability to label anything would be generally useful.

But leave that functionality like it currently is, message shows only on mouseover (desktop) or tap (mobile) to prevent abuse.

core/src/mindustry/logic/LExecutor.java
```diff
@@ -178,6 +179,23 @@ public Var(String name){
...
+            this.p1 = p1;
+        }
+
+        @Override
+        public void run(LExecutor exec){
+            int icon = exec.obj(p1) instanceof UnlockableContent u ? u.iconId : 0;
+
+            if(exec.building(varThis) instanceof LogicBuild){
+                ((LogicBuild) exec.building(varThis)).label = icon;
+            }
```
summetdev, line 195:
```suggestion
            if(exec.building(varThis) instanceof LogicBuild build){
                build.label = icon;
            }
```

Quezler: ah yes, instanceof magic, hard to get used to 👍

Anuken:
- On one hand, this is a neat way to show the status/purpose of a processor without utilizing additional blocks or resources.
- On the other hand, this will lead to a potentially-significant increase in visual noise, as processors everywhere will be flashing different icons as their status changes. (*No, a setting to hide the labels doesn't really solve the problem.*)
- 'holographic labels' on processors encroach upon the territory of displays. I'm not sure if one block should be able to do two things at once.
- I'm not sure how I feel about the label itself. It doesn't look all that great.

Anuken:
Closing for now.

PR #3889, opened 2020-12-09
draw mined item for laser and blast drills

Slava0135:
I know that Anuke some day removed this indicator from the said drills. But imo it's quite handy and players will be happy about having it.

If you don't like it for some reason I can do some changes to this.

Screenshot from 2020-12-09 20-23-52

PR #4030, opened 2020-12-18
Fix the two largest drills previewing ores in build requests

Quezler:
mildly related to #3889 , this won't be needed anymore if that passes, but this would still be good for other modded drills:

Screen Shot 2020-12-18 at 16 21 20

> this pic shows how it currently looks before fixing*


BlueWolf3682:
laser and airblasts drills, quez

PR #4042, opened 2020-12-19
Ask double pop-ups to create a public server

Kieaer:
# Mindustry steam server is friend only?
Not all players will play normally. griefers, kick command abuse...
This affects all users on that server.

But what if the user who first played the game gets into the steam server?
Most of them will be kicked out or ban as soon as they connect.
Even though players not doing anything.

So most users play the server on community servers..

I hope this resolves the server problem, but unfortunately, most community servers require authentication.
Because it's easy to cheating and too many griefers.

# so why need this?
This's a problem that has been around since the launch of Steam.
image

It seemed to have been resolved by floating a public host check popup, but I understand that almost all players can't do it with friends unless they press **YES**.
image

I've seen a lot of servers that abuse kick and bans today, and to solve this, I fundamentally changed the sequence of buttons and created a double popup to solve the yes man.


core/src/mindustry/ui/dialogs/HostDialog.java
```diff
@@ -69,8 +69,13 @@ public void runHost(){
...
                         ui.showCustomConfirm("@setting.publichost.name", "@public.confirm", "@yes", "@no", () -> {
-                            Core.settings.put("publichost", true);
-                            platform.updateLobby();
+                            ui.showCustomConfirm("@setting.publichost.name", "@public.confirm.really", "@no", "@yes", () -> {
+                                Core.settings.put("publichost", true);
+                                platform.updateLobby();
+                            }, () -> {
+                                Core.settings.put("publichost", false);
+                                platform.updateLobby();
+                            }
```
Quezler, line 78: this might fix the ci:
```suggestion
                            });
```

PR #4055, opened 2020-12-20
Draw item in the drill center when in place mode

Quezler:
drills show the preview item when built, or while in the build queue, but not yet while you hover with a fresh one:

before:
Screen Shot 2020-12-20 at 18 30 39

during: (during development, this was when using the drawxy on the tile instance*)
Screen Shot 2020-12-20 at 18 32 29

after:
Screen Shot 2020-12-20 at 18 41 30

### DIFF
--- a/core/src/mindustry/world/blocks/production/Drill.java
+++ b/core/src/mindustry/world/blocks/production/Drill.java
@@ -124,6 +124,12 @@ public class Drill extends Block{
             Draw.rect(returnItem.icon(Cicon.small), dx, dy - 1);
             Draw.reset();
             Draw.rect(returnItem.icon(Cicon.small), dx, dy);
+
+            if(drawMineItem){
+                Draw.color(returnItem.color);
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/blocks/production/Drill.java
```diff
@@ -124,6 +124,12 @@ public void drawPlace(int x, int y, int rotation, boolean valid){
             Draw.rect(returnItem.icon(Cicon.small), dx, dy - 1);
             Draw.reset();
             Draw.rect(returnItem.icon(Cicon.small), dx, dy);
+
+            if(drawMineItem){
+                Draw.color(returnItem.color);
+                Draw.rect("drill-top", tile.worldx() + offset, tile.worldy() + offset);
```
Quezler, line 130: patch for in case #4056 gets merged (separate pull request since technically its 2 different things entirely)

```suggestion
                Draw.rect(itemRegion, tile.worldx() + offset, tile.worldy() + offset);
```

PR #4023, opened 2020-12-18
Update bundle_zh_TW.properties

RedRadiation:
New translations.
Modified machine translates.

PR #4072, opened 2020-12-22
[Translate] zh-TW

simba-fs:


simba-fs:
It's too similar to  #4023, we will merge our branch together and then PR to Anuke's repo

PR #4074, opened 2020-12-22
Scrap Walls Textures Cleanup

Voz-Duh:
image


Quezler:
can we have both a before and after pic?

Voz-Duh:
@Quezler 
Before:
image
After:
image

Quezler:
its an improvement 👍

PR #4088, opened 2020-12-23
Rebrand power infinite as power source

Quezler:
just some things that have been bothering me a bit:

**naming**
all other sandbox blocks are named either source or void, infinite is the odd one out

**power**
600k is just a weird number, this changes it to 10k per node, since its normal to need several item/liquid sources somewhere.
(calculation happens once in blocks.java since it results in `166.666666667` which would look ugly hardcoded in)

**modding**
any modded power sources (though i doubt (m)any of them exist) will default to what it was

**verdict**
i think this be a nice chance, the new power number might still be "high" compared to what e.g. the item source can output, but i don't wanna break existing saves that rely on a ton of power all to much, they probably didn't use all power anyways 🤔


QmelZ:
> **modding**
> any modded power sources (though i doubt (m)any of them exist) will default to what it was

yes time to fix hackustry v2

PR #4098, opened 2020-12-24
Prevents plastanium conveyor unload exploit

Quezler:
a strange instachain design popped up on the discord overnight:
(unloading straight from the loading dock itself / very fast / bad)

this pull adds an additional low performance (boolean instead of a function) check on the building entity itself.

currently that value is always true since the block itself needs to be unloadable anyways, and i wasn't sure about whether or not to add it to the create, init, or both (of buildingcomp), but that would be a mirrored redundant check, dunno about it 🤔

### DIFF
--- a/core/src/mindustry/entities/comp/BuildingComp.java
+++ b/core/src/mindustry/entities/comp/BuildingComp.java
@@ -63,7 +63,6 @@ abstract class BuildingComp implements Posc, Teamc, Healthc, Buildingc, Timerc, 
     transient boolean enabled = true;
     transient float enabledControlTime;
     transient String lastAccessed;
-    transient boolean unloadable = true;
 
     PowerModule power;
     ItemModule items;
@@ -384,6 +383,10 @@ abstract class BuildingComp implements Posc, Teamc, Healthc, Buildingc, Timerc, 
 
     //endregion
     //region handler methods
+
+    public boolean unloadable(){
+        return block.unloadable;
+    }
 
     /** Called when an unloader takes an item. */
     public void itemTaken(Item item){
--- a/core/src/mindustry/entities/comp/BuildingComp.java
+++ b/core/src/mindustry/entities/comp/BuildingComp.java
@@ -384,7 +384,7 @@ abstract class BuildingComp implements Posc, Teamc, Healthc, Buildingc, Timerc, 
     //endregion
     //region handler methods
 
-    public boolean unloadable(){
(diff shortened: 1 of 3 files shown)


QmelZ:
why and how is this an "exploit"

Quezler:
> why and how is this an "exploit"

unintended feature then, this shouldn't happen

QmelZ:
it actually should happen but is annoying.

BasedUser:
i don't see how this is useful in a real game, looks limited to one unloader

Quezler:
> it actually should happen but is annoying.

unloading from a loading dock makes total sense 🙃

core/src/mindustry/entities/comp/BuildingComp.java
```diff
@@ -63,6 +63,7 @@
     transient boolean enabled = true;
     transient float enabledControlTime;
     transient String lastAccessed;
+    transient boolean unloadable = true;
```
Anuken, line 66: This should be an overridable `boolean` method instead of an instance variable, to be in line with how everything else is implemented.

Quezler: ok, though this was better for performance given how often its called and only for this single purpose atm, I'll get it changed 👍

core/src/mindustry/entities/comp/BuildingComp.java
```diff
@@ -384,6 +384,10 @@ public byte version(){
     //endregion
     //region handler methods
 
+    public boolean unloadable(){
+        return block.unloadable;
```
Quezler, line 388: redundant check inside `Unloader.java` though, should this just be hardcoded to `true` or simply remove the block level check in unloader and route it all through this function?

core/src/mindustry/world/blocks/storage/Unloader.java
```diff
@@ -61,7 +61,7 @@ public void updateTile(){
                     int pos = (offset + i) % proximity.size;
                     var other = proximity.get(pos);
 
-                    if(other.interactable(team) && other.block.unloadable && other.block.hasItems
+                    if(other.interactable(team) && other.block.unloadable && other.unloadable() && other.block.hasItems
```
Quezler, line 64: press to implement the above comment, not sure if this will break on blocks with no building entity or something?

```suggestion
                    if(other.interactable(team) && other.unloadable() && other.block.hasItems
```

PR #4099, opened 2020-12-24
Update UnitTypes.java

Kycb42148:
legSpeed = 0.1 or 0.2? (Arkyid)

### DIFF
--- a/core/src/mindustry/content/UnitTypes.java
+++ b/core/src/mindustry/content/UnitTypes.java
@@ -701,7 +701,6 @@ public class UnitTypes implements ContentList{
             legSpeed = 0.1f;
             legLengthScl = 0.96f;
             rippleScale = 2f;
-            legSpeed = 0.2f;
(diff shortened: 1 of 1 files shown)


core/src/mindustry/content/UnitTypes.java
```diff
@@ -701,7 +701,6 @@ public void load(){
             legSpeed = 0.1f;
             legLengthScl = 0.96f;
             rippleScale = 2f;
-            legSpeed = 0.2f;
```
QmelZ, line 704: 0.2 was the value used before, i suggest you remove the 0.1 one

QmelZ (changes requested):

core/src/mindustry/content/UnitTypes.java
Ilya246: yes, remove 0.1, not 0.2

PR #4100, opened 2020-12-24
mod version check fix

younggam:


zethnest:
What significance does this PR pose? When will 122 be lower than 121.4?

younggam:
currently, mindy refuse to apply mods that minGameVersion is vXXX.Y even if actullay it is appliable. this pr prevents it

zethnest:
Okay, I see what this PR is solving. However, it is unclear. What you need to do is fix the title of this PR. At least `Version check fix`. And then you need to tell why it needs to be fixed and what was broken.

PR #4105, opened 2020-12-24
unit leave formation if it is too far away / acceleration fix

Slava0135:
Peek 2020-12-24 16-06
Peek 2020-12-24 16-00

### DIFF
--- a/core/src/mindustry/ai/types/FormationAI.java
+++ b/core/src/mindustry/ai/types/FormationAI.java
@@ -27,7 +27,7 @@ public class FormationAI extends AIController implements FormationMember{
     @Override
     public void updateUnit(){
 
-        if(leader == null || leader.dead){
+        if(leader == null || leader.dead || !unit.within(leader.x, leader.y, 10*formation.pattern.spacing)){
             unit.resetController();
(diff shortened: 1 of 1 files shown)


Slava0135:
I guess the accel = 1.0 was made on purpose. But it felt weird. Also this way you could accelerate yourself before.

core/src/mindustry/ai/types/FormationAI.java
```diff
@@ -27,7 +27,7 @@ public void init(){
     @Override
     public void updateUnit(){
 
-        if(leader == null || leader.dead){
+        if(leader == null || leader.dead || !unit.within(leader.x, leader.y, 10*formation.pattern.spacing)){
```
Quezler, line 30:
```suggestion
        if(leader == null || leader.dead || !unit.within(leader.x, leader.y, 10 * formation.pattern.spacing)){
```

Anuken:
> But it felt weird. Also this way you could accelerate yourself before.

Why would you want your commanded units to have movement delay? This only breaks things.

Slava0135:
ok

PR #4106, opened 2020-12-24
Reload Button

QmelZ:
adds a reload button at the bottom of mod menu that reloads the game.
image

(i made this with very little java knowledge so its probably bad, but i tested it and it works)

### DIFF
--- a/core/src/mindustry/ui/dialogs/ModsDialog.java
+++ b/core/src/mindustry/ui/dialogs/ModsDialog.java
@@ -23,6 +23,7 @@ import mindustry.mod.Mods.*;
 import mindustry.ui.*;
 
 import java.io.*;
+import java.net.URISyntaxException;
 
 import static mindustry.Vars.*;
 
@@ -35,6 +36,25 @@ public class ModsDialog extends BaseDialog{
         addCloseButton();
 
         buttons.button("@mods.guide", Icon.link, () -> Core.app.openURI(modGuideURL)).size(210, 64f);
+        
+        buttons.button("@mods.reload", Icon.refresh, () -> {
+            
+            try {
+            
+            String jarPath = getClass().getProtectionDomain().getCodeSource().getLocation().toURI().getPath().replaceFirst("/", "").replaceAll("/", "\\" + File.separator).toString();
+            
+            Runtime rt = Runtime.getRuntime();
+            String[] commandAndArguments = {"java","-jar", jarPath};
+            Process p = rt.exec(commandAndArguments);
+            
+            } catch (IOException i) {
+                i.printStackTrace();
+            } catch (URISyntaxException u) {
+                u.printStackTrace();
+            };
+            
+            Core.app.exit();
+        }).size(210, 64f);
 
         shown(this::setup);
 
--- a/core/src/mindustry/ui/dialogs/ModsDialog.java
+++ b/core/src/mindustry/ui/dialogs/ModsDialog.java
@@ -37,24 +37,26 @@ public class ModsDialog extends BaseDialog{
 
         buttons.button("@mods.guide", Icon.link, () -> Core.app.openURI(modGuideURL)).size(210, 64f);
         
-        buttons.button("@mods.reload", Icon.refresh, () -> {
+        if(Vars.mobile = false) {
+            buttons.button("@mods.reload", Icon.refresh, () -> {
             
-            try {
+                try {
             
-            String jarPath = getClass().getProtectionDomain().getCodeSource().getLocation().toURI().getPath().replaceFirst("/", "").replaceAll("/", "\\" + File.separator).toString();
+                String jarPath = getClass().getProtectionDomain().getCodeSource().getLocation().toURI().getPath().replaceFirst("/", "").replaceAll("/", "\\" + File.separator).toString();
             
-            Runtime rt = Runtime.getRuntime();
-            String[] commandAndArguments = {"java","-jar", jarPath};
-            Process p = rt.exec(commandAndArguments);
+                Runtime rt = Runtime.getRuntime();
+                String[] commandAndArguments = {"java","-jar", jarPath};
+                Process p = rt.exec(commandAndArguments);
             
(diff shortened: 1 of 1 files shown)


core/src/mindustry/ui/dialogs/ModsDialog.java
```diff
@@ -23,6 +23,7 @@
 import mindustry.ui.*;
 
 import java.io.*;
+import java.net.URISyntaxException;
```
LottieVixen, line 26: why is this added?

Quezler:
did you copy the code from the BeControl.java?

if yes, then i didn't recognize it, but not having duplicate code might be cleaner

if no, take a peek there at how its done

Quezler (changes requested):
probably won't work on mac as it is currently

core/src/mindustry/ui/dialogs/ModsDialog.java
QmelZ: because of the catching URISyntaxException (i dont really know how it works)

QmelZ:
> did you copy the code from the BeControl.java?

no, i did not actually. if i remembered it existed i would have copied from there

QmelZ:
> probably won't work on mac as it is currently

i dont really have an idea how it works except "it gets the file path and runs `java -jar filepath`"

core/src/mindustry/ui/dialogs/ModsDialog.java
```diff
@@ -35,6 +36,27 @@ public ModsDialog(){
         addCloseButton();
 
         buttons.button("@mods.guide", Icon.link, () -> Core.app.openURI(modGuideURL)).size(210, 64f);
+        
+        if(Vars.mobile = false) {
```
summetdev, line 40: maybe you mean `==` here?

core/src/mindustry/ui/dialogs/ModsDialog.java
```diff
@@ -35,6 +36,27 @@ public ModsDialog(){
...
+            buttons.button("@mods.reload", Icon.refresh, () -> {
+            
+                try {
+            
+                String jarPath = getClass().getProtectionDomain().getCodeSource().getLocation().toURI().getPath().replaceFirst("/", "").replaceAll("/", "\\" + File.separator).toString();
+            
+                Runtime rt = Runtime.getRuntime();
```
summetdev, line 47: unnecessary variable, use `Runtime.getRuntime().exec(...)`

core/src/mindustry/ui/dialogs/ModsDialog.java
```diff
@@ -35,6 +36,27 @@ public ModsDialog(){
...
+                String[] commandAndArguments = {"java","-jar", jarPath};
+                Process p = rt.exec(commandAndArguments);
+            
+                } catch (IOException i) {
+                    i.printStackTrace();
+                } catch (URISyntaxException u) {
+                    u.printStackTrace();
+                };
```
summetdev, line 55: handle any error (i.e `Throwable`), then you don't need to import `URISyntaxException`.

core/src/mindustry/ui/dialogs/ModsDialog.java
```diff
@@ -35,6 +36,27 @@ public ModsDialog(){
         addCloseButton();
 
         buttons.button("@mods.guide", Icon.link, () -> Core.app.openURI(modGuideURL)).size(210, 64f);
+        
+        if(Vars.mobile = false) {
+            buttons.button("@mods.reload", Icon.refresh, () -> {
+            
+                try {
+            
+                String jarPath = getClass().getProtectionDomain().getCodeSource().getLocation().toURI().getPath().replaceFirst("/", "").replaceAll("/", "\\" + File.separator).toString();
```
summetdev, line 45: indentation. it should be:
```
try {
    // code
}
```

Quezler:
mac needs extra flags (see becontrol)

i would probably suggest to rewrite it to make restart a public void function on becontrol, and just call that via the static vars for your reload button

core/src/mindustry/ui/dialogs/ModsDialog.java
```diff
@@ -35,6 +36,27 @@ public ModsDialog(){
...
+            
+                Runtime rt = Runtime.getRuntime();
+                String[] commandAndArguments = {"java","-jar", jarPath};
+                Process p = rt.exec(commandAndArguments);
```
summetdev, line 49: unused `p` variable.

Quezler:
though that code is intended for desktop standalone, no idea how it'll behave for steam / ios / android*

core/src/mindustry/ui/dialogs/ModsDialog.java
```diff
@@ -35,6 +36,27 @@ public ModsDialog(){
...
+        
+        if(Vars.mobile = false) {
+            buttons.button("@mods.reload", Icon.refresh, () -> {
+            
+                try {
+            
+                String jarPath = getClass().getProtectionDomain().getCodeSource().getLocation().toURI().getPath().replaceFirst("/", "").replaceAll("/", "\\" + File.separator).toString();
+            
+                Runtime rt = Runtime.getRuntime();
+                String[] commandAndArguments = {"java","-jar", jarPath};
```
summetdev, line 48: why create new variable instead of passing directly to `exec()`?

summetdev (changes requested):

QmelZ:
at least someone told me what's wrong with my bad code, will fix asap

also quezler i will check the thing about macs too

Anuken:
This implementation:
- Doesn't preserve VM arguments, e.g. `-XstartOnFirstThread`
- Doesn't preserve game command-line arguments
- Doesn't work on Android/iOS
- Doesn't preserve the Java version that launched it
- Doesn't work on systems where Java is not installed (which is quite a few machines)
- Doesn't follow the formatting guidelines
- Is unlikely to work properly, as most of these problems are nearly impossible to fix

QmelZ:
fair enough i guess

PR #4112, opened 2020-12-25
Block Forge improvements

sk7725:
image
image
image

- When hovered over, displays the current block, drill-style.
- `minBlocksize` and `maxBlockSize` for mods.
- Removed fixed size(already set in Blocks.java).

### DIFF
--- a/core/src/mindustry/world/blocks/experimental/BlockForge.java
+++ b/core/src/mindustry/world/blocks/experimental/BlockForge.java
@@ -23,8 +23,7 @@ import static mindustry.Vars.*;
 
 public class BlockForge extends PayloadAcceptor{
     public float buildSpeed = 0.4f;
-    public int minBlockSize = 1;
-    public int maxBlockSize = 2;
+    public int minBlockSize = 1, maxBlockSize = 2;
 
     public BlockForge(String name){
         super(name);
--- a/core/src/mindustry/world/blocks/experimental/BlockForge.java
+++ b/core/src/mindustry/world/blocks/experimental/BlockForge.java
@@ -1,5 +1,6 @@
 package mindustry.world.blocks.experimental;
 
+import arc.graphics.*;
 import arc.graphics.g2d.*;
 import arc.math.*;
 import arc.scene.ui.layout.*;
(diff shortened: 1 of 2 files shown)


core/src/mindustry/world/blocks/experimental/BlockForge.java
```diff
@@ -18,13 +19,16 @@
...
+
 public class BlockForge extends PayloadAcceptor{
     public float buildSpeed = 0.4f;
+    public int minBlockSize = 1;
+    public int maxBlockSize = 2;
 
     public BlockForge(String name){
         super(name);
 
-        size = 3;
```
Anuken, line 27: I don't see a good reason to remove this; block forges should still be 3x3 by default, as anything smaller wouldn't be particularly useful - I'm not sure if it would even work with the payload system. Likewise with block loaders.

core/src/mindustry/world/blocks/experimental/BlockForge.java
```diff
@@ -18,13 +19,16 @@
 import mindustry.world.blocks.production.*;
 import mindustry.world.consumers.*;
 
+import static mindustry.Vars.*;
+
 public class BlockForge extends PayloadAcceptor{
     public float buildSpeed = 0.4f;
+    public int minBlockSize = 1;
```
Anuken, line 26: also these two vars should be on the same line

core/src/mindustry/world/blocks/experimental/BlockForge.java
sk7725: It allows you to make *bigger* block forges(T2 or higher) that will craft larger blocks.

> I'm not sure if it would even work with the payload system.

Well it is experimental, and besides, in vanilla the only versions will be the 3x3 ones. higher ones are for mods!

core/src/mindustry/world/blocks/experimental/BlockForge.java
sk7725: Done.

core/src/mindustry/world/blocks/experimental/BlockForge.java
sk7725: I'l commit follow-up PRs if the higher ones misbehave. Also by the same line, you also allow 2x2 or higher conveyors, despite them never working.

Anuken: > It allows you to make bigger block forges(T2 or higher) that will craft larger blocks.

It changes nothing. Super-constructors run before any user code can interact with the object.

sk7725: > It changes nothing. Super-constructors run before any user code can interact with the object.
I literally do not know what that means but ok

core/src/mindustry/world/blocks/experimental/BlockLoader.java
```diff
@@ -21,7 +21,7 @@
 
     public BlockLoader(String name){
         super(name);
-
+
```
Quezler, line 24: 👀

sk7725: 👀

PR #4127, opened 2020-12-26
Block (un)loader major improvements

sk7725:
image
*(help save me I'm a vault and I'm stuck)*
+ Adds max block size because ^ this is just downright silly
+ Made it do pattern matching instead of casting over and over(my IDE did not error this out, so hopefully it works)
+ Make it take consValid() into account(it does require power but completely ignores it)

*(Side note: I wanted to make max block size automatic and also at the PayloadAcceptor class, but it might jam the unit recon lines so I decided not to touch that)*

### DIFF
--- a/core/src/mindustry/world/blocks/experimental/BlockLoader.java
+++ b/core/src/mindustry/world/blocks/experimental/BlockLoader.java
@@ -100,7 +100,7 @@ public class BlockLoader extends PayloadAcceptor{
 
                 //load up items
                 if(payload.block().hasItems && items.any()){
-                    if(consValid() && timer(timerLoad, loadTime / efficiency())){
+                    if(consValid() && efficiency() > 0.01f && timer(timerLoad, loadTime / efficiency())){
                         //load up items a set amount of times
                         for(int j = 0; j < itemsLoaded && items.any(); j++){
 
--- a/core/src/mindustry/world/blocks/experimental/BlockLoader.java
+++ b/core/src/mindustry/world/blocks/experimental/BlockLoader.java
@@ -99,7 +99,7 @@ public class BlockLoader extends PayloadAcceptor{
 
                 //load up items
                 if(payload.block().hasItems && items.any()){
-                    if(timer(timerLoad, loadTime)){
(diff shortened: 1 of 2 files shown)


BasedUser:
there might be riots everywhere when they realize the tetrative reconstructor they spawned into a conveyor is unable to take input from a (probably also spawned in) block loader

sk7725:
> there might be riots everywhere when they realize the tetrative reconstructor they spawned into a conveyor is unable to take input from a (probably also spawned in) block loader

conveyors have a strict payload size limit, im just enforcing it to other blocks minus the reconstructors(so your T2 can be as big as you want).

core/src/mindustry/world/blocks/experimental/BlockLoader.java
```diff
@@ -99,7 +100,7 @@ public void updateTile(){
 
                 //load up items
                 if(payload.block().hasItems && items.any()){
-                    if(timer(timerLoad, loadTime)){
+                    if(consValid() && timer(timerLoad, loadTime)){
```
Anuken, line 103: Instead of checking for consValid(), you should take efficiency into account. 50% power should mean 50% unload speed.

sk7725: did a hacky math, at least it should work.

core/src/mindustry/world/blocks/experimental/BlockLoader.java
```diff
@@ -99,7 +100,7 @@ public void updateTile(){
 
                 //load up items
                 if(payload.block().hasItems && items.any()){
-                    if(timer(timerLoad, loadTime)){
+                    if(consValid() && efficiency() > 0.01f && timer(timerLoad, loadTime / efficiency())){
```
Anuken, line 103: consValid() check seems redundant when you're already checking for efficiency.

sk7725: enabled check, or is the enavled also already counted in the efficiency()?

sk7725: it works

PR #4131, opened 2020-12-26
Disable publichost on every startup

Kieaer:
Updated the public server description, but to no avail if players don't see the message.
This resets the settings from the previous build and lets you see the popup again.


Anuken:
I already changed the key that it checks for, so it should reset.

Kieaer:
It won't be enough to just send a message once.
I was kicked immediately after entering 3 minutes ago.

PR #4132, opened 2020-12-27
Update servers_v6.json

jixishi:
增加两个中国地区服务器，主要为挑战性地图。
Add two China region servers, mainly for challenging maps.


WilloIzCitron (approved):
JSON its valid and server its online

PR #4134, opened 2020-12-27
Return large launchpad item costs

Quezler:
would be neat to be able to build 4x4 launchpads through #4017

this pull adds the original material costs back in:
https://github.com/Anuken/Mindustry/commit/bd349803d34740cdc5258e624ac3621ee1fa7241#diff-5fcb58432c6c68d32771609241d663343e35cbd86baa58c85923fdb9890748beL1329

though it still keeps it hidden since its tagged for removal, but it would be nice if it had a build cost until that happens 🤔


Quezler:
might make further improvements to the sprite / graphics of the large launch pad if this pull ends up making it in ^

Anuken:
What would the use case for this block be?

Quezler:
> What would the use case for this block be?

as described in #4017 it would be nice to have it available to/through custom plugins, it won't change anything vanilla 🤔

Quezler:
(e.g. on nydus bleeding edge the 3x3 launchpad periodically sends stuff to the core as a sort of long range mass driver, having the 4x4 one available as well might make some designs look prettier and/or more aligned, or perhaps for other functions)

Anuken:
I mean, what would you use it for *in a custom plugin*? Unlike the forge/unloader/etc, it doesn't offer any new functionality.

Quezler:
for example stuff like this, synced server sided magic:

Screen Shot 2020-12-27 at 20 41 18

Quezler:
that is basically where the revealed blocks rule addition was for, being able to construct normally hidden blocks that still have construction item requirements in order for the server to do custom stuff with it, and the large launchpad has no item requirements ^

Quezler:
the full picture:

nydus has kinda always been about creating a modded experience without requiring players to install any mods,
this is accomplished via anything the vanilla game allows (revealed blocks / call packets / bullet particles / etc)

previously any "special" block hidden in the code of the game had to be either spawned in via a sort of multiblock, or overriding the construction of any other block with the same size, which is a disaster for usability since its vague.

thanks to the revealedblocks rule launchpads & block forges can already be constructed via the normal build menu,
but the large launchpad has no item requirements at all, it would be nice if this pull could reset them to what they were.

i'm not trying to turn vanilla into my own mod or anything, but adding the item cost back to this block doesn't seem to cause any harm, since this was what it was before, even though its currently hidden away from users.
(funny enough the scrap wall cost was way before the revealedblocks idea was even conceived)

so in short, this pull is just to make constructing the large launchpad via revealedblocks more logical, since currently its a free instaplace while the normal launchpad is not, having this merged would open up some room for new serversided code.

if you'd rather close this since the block is marked for deletion that's fine by me, it would just be neat if it could have item requirements until that eventually happens, and when it does eventually get removed if this got merged that's also fine, don't see this pull as a reason to keep the large launchpad around for longer than you had originally planned.

**tldr** by giving the large launchpad a build cost i can put it to some good use via server sided plugins/modifications ❤️

Anuken:
I'm just hesitant to merge this, as the large launch pad will almost certainly be removed in 7.0 (*to be replaced with other structures*), although I have no plans to change it in 6.0. If you're fine with your custom mechanics potentially getting broken next major release, I can go ahead with merging.

Quezler:
> don't see this pull as a reason to keep the large launchpad around for longer than you had originally planned.

yea its fine, when you eventually change it for 7.0 I'll just find another place to hause it, but it would be nice for it to have a build cost for now 👍

PR #4139, opened 2020-12-27
Remove 4 square blocks from the accelerator

Quezler:
They conflict annoyingly with the 4 blinking arrows on that level:

Screen Shot 2020-12-27 at 12 11 25
Screen Shot 2020-12-27 at 12 23 57


Quezler:
additional change for this thing i just spotted:

Screen Shot 2020-12-27 at 12 26 32

PR #4146, opened 2020-12-28
Launchpad Launch Sound

MEEPofFaith:


core/src/mindustry/world/blocks/campaign/LaunchPad.java
```diff
@@ -119,6 +121,7 @@ public void updateTile(){
 
             //launch when full and base conditions are met
             if(items.total() >= itemCapacity && efficiency() >= 1f && timer(timerLaunch, launchTime / timeScale)){
+                launchSound.at(x, y, Mathf.random(0.9f, 1.1f));
```
Anuken, line 124: Pitch shouldn't be randomized (it's not a gun!), and the default should be `none`, as artillery is not a good fit.

catatagun:
Epac

PR #4160, opened 2020-12-29
outputSize stat for Block Forge

sk7725:
image
*( This but .replace("Output", "Output Size"); )*
+ Adds stat.outputSize and used it for Block Forge.
+ Name is not *too* specific (outputSize and not outputBlockSize) so it can be utilized by mods, or later with a different purpose, such as payload size or unit size.
+ Comes with ~~brazil~~ korean.

### DIFF
--- a/core/src/mindustry/world/blocks/experimental/BlockForge.java
+++ b/core/src/mindustry/world/blocks/experimental/BlockForge.java
@@ -18,6 +18,7 @@ import mindustry.world.blocks.*;
 import mindustry.world.blocks.payloads.*;
 import mindustry.world.blocks.production.*;
 import mindustry.world.consumers.*;
+import mindustry.world.meta.*
 
 import static mindustry.Vars.*;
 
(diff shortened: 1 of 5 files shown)


Quezler (changes requested):
`import mindustry.world.meta.*`

sk7725:
> `import mindustry.world.meta.*`

OH GOD

core/src/mindustry/world/blocks/experimental/BlockForge.java
```diff
@@ -18,6 +18,7 @@
 import mindustry.world.blocks.payloads.*;
 import mindustry.world.blocks.production.*;
 import mindustry.world.consumers.*;
+import mindustry.world.meta.*
```
Quezler, line 21:
```suggestion
import mindustry.world.meta.*;
```

sk7725: why why why
I swear, the semicolon WAS THERE

Quezler (approved):

Erisfiregamer1 (approved):
yes very good but you are losing your mind

Anuken:
Why not use "Output" instead? I'd rather avoid unnecessary bundle keys, and it seems clear enough.

Anuken:
*implemented, see comment above*

PR #4163, opened 2020-12-29
Display unit cap modifier for core blocks

Quezler:
> core shard set to negative just for the screenshot*

- not sure if a new category like "special" is needed, housing it under general for now
- there is probably a shortcut for `< 0 ? '-' : '+'` somewhere that i haven't found yet
- maybe a slight bundle change is needed, not sure if the current value reflects it

Screen Shot 2020-12-29 at 10 22 56
Screen Shot 2020-12-29 at 10 23 02
Screen Shot 2020-12-29 at 10 23 08

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -376,7 +376,7 @@ public class Block extends UnlockableContent{
             bars.add("items", entity -> new Bar(() -> Core.bundle.format("bar.items", entity.items.total()), () -> Pal.items, () -> (float)entity.items.total() / itemCapacity));
         }
         
-        if(flags.contains(BlockFlag.unitModifier)) stats.add(Stat.maxUnits, Strings.format("@ @", unitCapModifier < 0 ? '-' : '+', Math.abs(unitCapModifier)));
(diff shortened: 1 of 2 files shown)


core/src/mindustry/world/Block.java
```diff
@@ -375,6 +375,8 @@ public void setBars(){
         if(hasItems && configurable){
             bars.add("items", entity -> new Bar(() -> Core.bundle.format("bar.items", entity.items.total()), () -> Pal.items, () -> (float)entity.items.total() / itemCapacity));
         }
+        
+        if(flags.contains(BlockFlag.unitModifier)) stats.add(Stat.maxUnits, Strings.format("@ @", unitCapModifier < 0 ? '-' : '+', Math.abs(unitCapModifier)));
```
Anuken, line 379: This is a very simple case, you shouldn't need to use Strings.format here. Just concatenate.
I don't think the space between `+ num` is necessary.

Quezler: oh it kinda went wrong the first time and the numbers were in the 40-50 range, probably due to the char?

Anuken: Use double-quotes for string concatenation instead.

PR #4187, opened 2020-12-30
Fix block status of repair point and battery

Quezler:
previously the repair point was always orange, this makes it red while it has no power:

Screen Shot 2020-12-30 at 18 24 33
Screen Shot 2020-12-30 at 18 24 36

previously batteries always showed a green status, now its green when full, red when empty & orange anywhere in between:
Screen Shot 2020-12-30 at 18 31 58

### DIFF
--- a/core/src/mindustry/world/blocks/power/Battery.java
+++ b/core/src/mindustry/world/blocks/power/Battery.java
@@ -45,8 +45,8 @@ public class Battery extends PowerDistributor{
 
         @Override
         public BlockStatus status(){
-            if(power.status == 0f) return BlockStatus.noInput;
-            if(power.status == 1f) return BlockStatus.active;
+            if(Mathf.equal(power.status, 0f, 0.01f)) return BlockStatus.noInput;
+            if(Mathf.equal(power.status, 1f, 0.01f)) return BlockStatus.active;
             return BlockStatus.noOutput;
         }
     }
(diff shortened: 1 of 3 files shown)


core/src/mindustry/world/blocks/power/Battery.java
```diff
@@ -42,5 +42,12 @@ public void overwrote(Seq<Building> previous){
                 }
             }
         }
+
+        @Override
+        public BlockStatus status(){
+            if(power.status == 0f) return BlockStatus.noInput;
```
Anuken, line 48: Use Mathf.equals() with a range; comparing floats directly is a bad idea.

BlueWolf3682:
wait, should they have status indicators at all? they have a visual charge indicator sprite from v5, and the v6 status diamond obscures the smaller ones charge indicator

Quezler:
the status indicator is easier to spot than the dots

BlueWolf3682:
the status diamond doesnt show fullness by its shade, if its not full but has charge its just orange no matter how full it is

Quezler:
so? if you want an exact measurement you can look at the dots or hover your cursor over it, the 3color status indicator is just for first glance

BlueWolf3682:
you cant see the small dot though, its kinda obscured

Quezler:
same with any 1x1 block that has a status light

BlueWolf3682:
fair enough
though my final opinion is that either the status should not be there due to an existing status indicator, or the status diamond should at least have a gradient due to it obscuring the smaller battery

joshuaptfan:
Block status can be toggled. Toggle it off if you want to see 1x1 blocks' own indicators. Block status should remain simple (no gradients) for easiest reading at a glance.

PR #4203, opened 2020-12-31
Custom Launch Pad Pod Textures

Voz-Duh:
For example:image
_Yes, Anuken sayed what large launch pad does not needed, and I make it /-/_
_Sorry from it_

### DIFF
--- a/core/src/mindustry/world/blocks/campaign/LaunchPad.java
+++ b/core/src/mindustry/world/blocks/campaign/LaunchPad.java
@@ -32,7 +32,7 @@ public class LaunchPad extends Block{
     public Sound launchSound = Sounds.none;
 
     public @Load("@-light") TextureRegion lightRegion;
-    public @Load("launchpod") TextureRegion podRegion;
+    public @Load(value = "@-pod", fallback = "launchpod") TextureRegion podRegion;
     public Color lightColor = Color.valueOf("eab678");
 
     public LaunchPad(String name){
@@ -125,6 +125,7 @@ public class LaunchPad extends Block{
                 LaunchPayload entity = LaunchPayload.create();
                 items.each((item, amount) -> entity.stacks.add(new ItemStack(item, amount)));
                 entity.set(this);
+                entity.setRegion(podRegion);
                 entity.lifetime(120f);
                 entity.team(team);
                 entity.add();
@@ -175,6 +176,7 @@ public class LaunchPad extends Block{
 
         Seq<ItemStack> stacks = new Seq<>();
         transient Interval in = new Interval();
+        private TextureRegion region;
 
         @Override
         public void draw(){
@@ -221,6 +223,10 @@ public class LaunchPad extends Block{
 
         float cy(){
             return y + fin(Interp.pow5In) * (100f + Mathf.randomSeedRange(id() + 2, 30f));
+        }
+
+        public void setRegion(TextureRegion region){
+            this.region = region;
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/blocks/campaign/LaunchPad.java
```diff
@@ -175,6 +176,7 @@ public void buildConfiguration(Table table){
 
         Seq<ItemStack> stacks = new Seq<>();
         transient Interval in = new Interval();
+        private TextureRegion region;
```
Anuken, line 179: Regions cannot be serialized. Save a reference to the `Block` instead, cast it to a `LaunchPad` and use its region. Don't make this private, there's no reason to use getters here.

Anuken (changes requested):

core/src/mindustry/world/blocks/campaign/LaunchPad.java
Voz-Duh: You suggested a more complex and costly variant, because while it searches for tiles, checks LaunchPad it or no and looks for podRegion in it, it will take a lot of unnecessary MB

Voz-Duh: Maybe I misunderstood you?

Anuken: No. `Block block;` and `((LaunchPad)block).region` is not any more memory intensive.

You also need to check for block/region being null, because older versions won't have it serialized.

Voz-Duh: This work but need fix 
image

Voz-Duh:
@Anuken, I make it

Voz-Duh:
Okay, it does compile

z0mbiesrock:
No it doesn't.
PR **rejected**

XEN0PHIL:
This sprite is ugly and overdetailed

valeciawolf34:
https://raw.githubusercontent.com/nvm-sh/nvm/v0.37.2/install.sh

valeciawolf34:
183add4bcb00fce798244aee9215ba1d79576f91https://github.com/Anuken/Mindustry/commit/183add4bcb00fce798244aee9215ba1d79576f91

SnakkiZXZ:
I can redraw this texture, but only if Anuken approves of it

Voz-Duh:
> I can redraw this texture, but only if Anuken approves of it

This textures maked only for tests

core/src/mindustry/world/blocks/campaign/LaunchPad.java
```diff
@@ -200,7 +200,7 @@ public void draw(){
 
             Draw.z(Layer.weather - 1);
 
-            TextureRegion region = Core.atlas.find("launchpod");
+            TextureRegion region = ((mindustry.world.blocks.campaign.LaunchPad)blockOn()).podRegion;
```
Anuken, line 203:
```suggestion
            TextureRegion region = blockOn() instanceof LaunchPad p ? p.podRegion : Core.atlas.find("launchpod");
```

Voz-Duh: You don't see it?
`public @Load(value = "@-pod", fallback = "launchpod") TextureRegion podRegion;`

Voz-Duh: This good idea but - no

Anuken: What if the launch pod flies up and the pad gets deleted? Your code will crash.

Voz-Duh: Hm... Okay

PR #4242, opened 2021-01-03
Allow for core incineration via rules

Quezler:
i would like to be able to play with core incineration on the nydus bleeding edge server:

- once the core is full of something that production chain turns off, possibly causing power issues when some turn back on.
- makes mixed conveyors (e.g. lead with glass) less of a problem when the core eventually fills up with something.
- adds a nice challenge of having to balance production: not to little but also not too much
- possibly brings some more logic into play, like disabling a conveyor once the core is full
- no more need for impractical overflow + incinerator setups near the core
- makes it more in like with what is expected through playing the campaign
- maybe a few unforseen meta changes due to sideeffects, will be interesting
- currently a base level rule instead of team rule, maybe some map makers would like it overridable per team though?
- generally just interesting to experiment with, would be nice to have this for the time being on bleeding edge
- yes, i could have just written a plugin that constantly sets it back to like 100 below the cap and spawn particles, but spamming all those packets is just not very efficient while a relatively harmful rule addition is objectively better

(false by default of course, its an opt-in)

Screen Shot 2021-01-03 at 11 44 36

> naming might not be the best, could imply on whether or not the core is susceptible or not to catching fire 🔥

### DIFF
--- a/core/src/mindustry/game/Rules.java
+++ b/core/src/mindustry/game/Rules.java
@@ -97,6 +97,8 @@ public class Rules{
     public Team waveTeam = Team.crux;
     /** name of the custom mode that this ruleset describes, or null. */
     public @Nullable String modeName;
+    /** Whether cores incinerate items when full, just like in the campaign. */
+    public boolean coreIncinerates = false;
     /** special tags for additional info. */
     public StringMap tags = new StringMap();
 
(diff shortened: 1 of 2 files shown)


XEN0PHIL:
> once the core is full of something that production chain turns off, possibly causing power issues when some turn back on.
> adds a nice challenge of having to balance production: not to little but also not too much

I'd like the core incineration because of those.

hribik100:
Very good idea
I think when IT is Global rule IT is better

joshuaptfan:
This is great as just a harmless convenience option, and to align with the expectations established in the campaign for new players.

> currently a base level rule instead of team rule, maybe some map makers would like it overridable per team though?

Nah, that's overthinking it.

PR #4245, opened 2021-01-03
Amendments prototype 

Quezler:
currently when you `/sync` the `lastAccessed` property gets wiped since its not part of the block io stream.

this pull attempts to introduce a counterpart to rules, that allows for syncing a bunch of data that isn't meant to be saved.

currently the code looks hideous (looking at you `Iterator<Entry<String>>`), and i have yet to figure out the proper places to read/write it, since currently it just feels a tad wrong.

possibly it will be housed in `Administration.java` with some saving of some kind to persist between restarts? not sure how that'll work yet, and i might have to figure out what side effects it has on the clients.

obviously a lot needs to be done to make this work better (team checks / building still exists checks / etc), but in its current state the last accessed stays intact while syncing.

Screen Shot 2021-01-03 at 16 22 03

goals of this pull:
- get last accessed to persist when syncing, getting a filled in copy when joining & persist during server restarts
- find more stuff to pipe through amendments, for example stuff like what was suggested in this pull: #4028


XEN0PHIL:
I approve, like any server moderator will

DeltaNedas:
good roomba

Quezler:
Soooo, i kinda reworked a bit how last accessed got stored, and i've accidentally opened up the way to see who/what got removed from a tile, so i've implemented a thing with middle click:
(mouse is over the center conveyor)

Screen Shot 2021-01-03 at 20 33 18
Screen Shot 2021-01-03 at 20 33 23

its not exactly user friendly for mobile + that you need to move the mouse first to force a re-render of the tile preview, but it would be neat to have a vanilla way to see deleted tiles, what they were & who deleted em 🙂

Quezler:
@Anuken i believe its ready for a slight review, it works but some points might need adressing:
- the `display` method in tile is a bit messy since it'll probably change how deleted is previewd
- i'm not sure where to best clear it, i recall block configurations limits being stored somewhere, but i've searched and that code has possibly already been deleted, now it listens to the world load event but i couldn't find a definitive answer if that also triggers for hosting from ingame / host in the console / gameover in the console / load in the console
- middle mouse button as a picker is referenced 3 times, because i had no other idea on how to trigger it

Quezler:
note to self: possibly include brokenblocks?

core/src/mindustry/io/TypeIO.java
```diff
@@ -365,6 +365,19 @@ public static Rules readRules(Reads read){
         return JsonIO.read(Rules.class, string);
     }
 
+    public static void writeAmendments(Writes write, Amendments amendments){
+        String string = JsonIO.write(amendments);
```
Anuken, line 369: Writing large amounts of data with JSON is risky; with the 32/64k character limit, you could easily run into IO issues.

Quezler: so i would have to send it in parts similar to block snapshots or something? (assuming the limit is packet size and not your jsonio thing)

royalkhai:
dont send me message

Vào Th 6, 8 thg 1, 2021 vào lúc 01:34 Quezler <notifications@github.com>
đã viết:

PR #4246, opened 2021-01-03
Add defaultName for FileChooser when exporting game data

lorenuars19:
from Anuken/Mindustry-Suggestions#1715.

Modified FileChooser for accepting defaultName parameter
Set defaultName using DateTimeFormatter
Modified ShowFileChooser to accept defautlName parameter

### DIFF
--- a/core/src/mindustry/ui/dialogs/SettingsMenuDialog.java
+++ b/core/src/mindustry/ui/dialogs/SettingsMenuDialog.java
@@ -26,6 +26,7 @@ import mindustry.ui.*;
 
 import java.io.*;
 import java.time.Instant;
+import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.util.Locale;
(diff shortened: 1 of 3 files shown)


core/src/mindustry/ui/dialogs/SettingsMenuDialog.java
```diff
@@ -25,6 +25,10 @@
 import mindustry.ui.*;
 
 import java.io.*;
+import java.time.Instant;
```
Anuken, line 28: `java.time` is a Java 8 API. Do not use it.

Quezler: linking a potential alternative @lorenuars19 can copy from that is already in use elsewhere in the mindustry codebase:

https://github.com/Anuken/Mindustry/blob/master/server/src/mindustry/server/ServerControl.java#L207

lorenuars19:
As anuken said JAVA 8 bad, so I removed it.
Using LocalDataTime <15> instead.

core/src/mindustry/ui/dialogs/SettingsMenuDialog.java
Anuken: You're still using `java.time` here, which is a Java 8 API. The server uses it, yes, but that's because the headless server code **never** runs on Android and hence always has access to this API.

lorenuars19: could I use [SimpleDateFormat](https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/text/SimpleDateFormat.html)

which is JAVA 15 ?

lorenuars19:
this time it should be using only JAVA 15 like you asked

Anuken:
Android and other platforms now use different file chooser implementations, so this is no longer relevant.

PR #4247, opened 2021-01-03
Make rotation via construction emit effect & sound

Quezler:
currently when you build over something with a different rotation it just shows the beams and it happening, but no visual ques.

this pull makes it so that when its being rotated via construction that it shows the construction effect & sound.
(rotation via the mouse wheel is exempt)

this was the cleanest place to add the sound hooks in (due to shouldplay and pitch being private), instead of moving the entire function over here, and on first glance it might look as though scroll wheel rotate triggers this too, which it does not.

in short, this will make it feel better when you rotate lots of stuff like a bunch of conveyors in bulk.

### DIFF
--- a/core/src/mindustry/world/Build.java
+++ b/core/src/mindustry/world/Build.java
@@ -69,6 +69,7 @@ public class Build{
         if(tile.team() == team && tile.block == result && tile.build != null && tile.block.quickRotate){
             if(unit != null && unit.getControllerName() != null) tile.build.lastAccessed = unit.getControllerName();
             tile.build.rotation = Mathf.mod(rotation, 4);
+            ConstructBlock.rotateFinish(tile);
             tile.build.updateProximity();
             tile.build.noSleep();
             return;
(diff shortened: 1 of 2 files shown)


Anuken:
Rotating with E+Scrollwheel doesn't shown any visual effects either. I don't think it should play a sound or make an effect here.

Quezler:
it just feels "wrong" to build a conveyor partially over another conveyor and not hear or see anything just because it only rotates

XEN0PHIL:
If something is easy to do and it will make players happy without affecting anything in a bad way then why not?
Building something should feel like it's being built.

Anuken:
In this case, it's *not* being built. The block is not being re-created. It should *not* look like a standard building being placed.

XEN0PHIL:
I mean it should make you feel like you changed something. I'd even add an effect of rotating something for the same feeling. And as an easier detection of stealthy block rotation griefing

XEN0PHIL:
Perhaps the block placement sound but without the building process humming?

grannybullet:
Yeah this would be a nice change, it does feel a little dull just flipping the orreintation of a block with no feedback.

Anuken:
If an effect is added, it should happen with both types of rotation, and not look identical to the block place animation.

Quezler:
Made obsolete by 69eb8faf67e49d83f4b3c18278937f373c08b012

PR #4248, opened 2021-01-03
Fix mirror being offset

TranquillyUnpleasant:
Because apparently it isn't a valid map even though its been a known issue for months.
P.S. Notice how in the original code `in.width/2` is being added twice... and so is `in.height/2` what is the point of that?


TranquillyUnpleasant:
Before and after for readers convenience:
image 
image

PR #4255, opened 2021-01-04
Add boost bars to overdrive projectors

Quezler:
would be nice to see the current boost percentage of the overdrive projectors, since cross referencing the core database doesn't give you a clear enough picture:

Screen Shot 2021-01-04 at 11 29 25
Screen Shot 2021-01-04 at 11 29 33
Screen Shot 2021-01-04 at 11 29 53
Screen Shot 2021-01-04 at 11 30 03
Screen Shot 2021-01-04 at 11 30 52
Screen Shot 2021-01-04 at 11 36 24

> consume moved down due to a consvalid race condition with the overdrive dome*

Screen Shot 2021-01-04 at 11 32 26

### DIFF
--- a/core/src/mindustry/world/blocks/defense/OverdriveProjector.java
+++ b/core/src/mindustry/world/blocks/defense/OverdriveProjector.java
@@ -1,14 +1,15 @@
 package mindustry.world.blocks.defense;
 
+import arc.*;
 import arc.graphics.*;
 import arc.graphics.g2d.*;
 import arc.math.*;
 import arc.util.*;
 import arc.util.io.*;
 import mindustry.annotations.Annotations.*;
-import mindustry.gen.*;
 import mindustry.graphics.*;
 import mindustry.logic.*;
+import mindustry.ui.*;
 import mindustry.world.*;
 import mindustry.world.meta.*;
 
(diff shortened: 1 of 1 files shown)


grannybullet:
Yass pls

XEN0PHIL:
Nice. The block that boosts should show its effectiveness, why it's still not merged? :cat:

Voz-Duh (approved):

core/src/mindustry/world/blocks/defense/OverdriveProjector.java
```diff
@@ -88,17 +96,20 @@ public void updateTile(){
...
+                indexer.eachBlock(this, realRange, other -> true, other -> other.applyBoost(realBoost(), reload + 1f));
             }
+
+            if(timer(timerUse, useTime) && efficiency() > 0 && consValid()){
+                consume();
+            }
+        }
+
+        public float realBoost(){
+            return consValid() ? (speedBoost + phaseHeat * speedBoostPhase) * efficiency() : 0f;
```
Anuken, line 112: Any particular reason why you added `consValid()` here?

Quezler: yes, without that the overdrive dome will show a full boost bar if it has power without the required items present

Quezler: (which is also why i moved the consume down, to prevent this function from returning 0 when its using its last set of items)

PR #4256, opened 2021-01-04
Highlight all linked inventories on core selection

Quezler:
maps with several cores (placed any distance apart) mess up core highlighting, this pull tries to remedy that:
(in both pics the center core block was hovered over)

> before

Screen Shot 2021-01-04 at 11 55 45

> after

Screen Shot 2021-01-04 at 12 18 34

to make the code look clean i've added the storage flag to the core blocks as well, not sure if this has side effects 🤔

### DIFF
--- a/core/src/mindustry/world/blocks/storage/CoreBlock.java
+++ b/core/src/mindustry/world/blocks/storage/CoreBlock.java
@@ -48,7 +48,7 @@ public class CoreBlock extends StorageBlock{
         update = true;
         hasItems = true;
         priority = TargetPriority.core;
-        flags = EnumSet.of(BlockFlag.core, BlockFlag.unitModifier);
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/blocks/storage/CoreBlock.java
```diff
@@ -48,7 +48,7 @@ public CoreBlock(String name){
         update = true;
         hasItems = true;
         priority = TargetPriority.core;
-        flags = EnumSet.of(BlockFlag.core, BlockFlag.unitModifier);
+        flags = EnumSet.of(BlockFlag.core, BlockFlag.unitModifier, BlockFlag.storage);
```
Anuken, line 51: This may mess up logic. Consider looping through every core and connected container instead.

Quezler: will do tomorrow 👍

PR #4263, opened 2021-01-05
Update servers_v6.json

PAP6:
Reason: This server is currently running version 104.10, so it shouldn't be here.


PAP6:
Thank you

PR #4264, opened 2021-01-05
Add core incineration to custom rules dialog

Quezler:
https://github.com/Anuken/Mindustry-Suggestions/issues/1727#issuecomment-754167207
Screen Shot 2021-01-05 at 11 20 19
Screen Shot 2021-01-05 at 11 23 48

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -935,7 +935,7 @@ mode.custom = Custom Rules
 
 rules.infiniteresources = Infinite Resources
 rules.reactorexplosions = Reactor Explosions
-rules.coreincinerates = Core incinerates overflow
+rules.coreincinerates = Core Incinerates Overflow
 rules.schematic = Schematics Allowed
 rules.wavetimer = Wave Timer
 rules.waves = Waves
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -935,6 +935,7 @@ mode.custom = Custom Rules
 
 rules.infiniteresources = Infinite Resources
 rules.reactorexplosions = Reactor Explosions
+rules.coreincinerates = Core incinerates overflow
 rules.schematic = Schematics Allowed
 rules.wavetimer = Wave Timer
 rules.waves = Waves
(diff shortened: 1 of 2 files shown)


core/assets/bundles/bundle.properties
```diff
@@ -935,6 +935,7 @@ mode.custom = Custom Rules
 
 rules.infiniteresources = Infinite Resources
 rules.reactorexplosions = Reactor Explosions
+rules.coreincinerates = Core incinerates overflow
```
YellOw139, line 938:
```suggestion
rules.coreincinerates = Core Incinerates Overflow
```

XEN0PHIL:
Weird english rules require each word capitalized :eyes:

YellOw139 (commented):
For consistency.

YellOw139:
Lol, so I'm not the only one who noticed.

Quezler:
@XEN0PHIL @YellOw139 thx for pointing it out

SneakIrwin:
Sorry to revive this after it was implemented, but can we add these options to the campaign? I'd really like to be able to turn off core incineration. I miss launchpad/incinerator overflow networks.

YellOw139:
See your own suggestion for the reason it cannot be implemented and go away.

SneakIrwin:
The only reason I was ever given was that it makes it simple to calculate passive income but that is literally so easy to find any other way.

YellOw139:
Oh yeah? Then I'm eagerly waiting for your simple implementation of passing income calculation in a PR.

SneakIrwin:
Why are you so rude and salty? No, I don't know Java and I'm not trying to imply the code behind it isn't intricate, but my request's halfway implementation seemed to be quite speedy in the latest multiplayer update. I'm not convinced just asking for the *option* to make resources incinerate at the core vs. somewhere else is a game-breaking mountain of a request.

Anuken:
> can we add these options to the campaign

The issue tracker is not a place for you to inject suggestions. This PR had nothing to do with the campaign, and the option you request won't be added.

> that is literally so easy to find any other way.

I went and looked up the thread, and your "literally so easy solution" seems to be...

> utilizing export buildings

This is an *awful* idea:
- Export buildings? Export *where*, exactly? There is no global storage, everything is stored  in your core.
- This doesn't remove incineration at all, it just makes another block do it. I fail to see how that addresses your problem.
- This is a significant change of core campaign mechanics. It requires adding **special new blocks**, and is not remotely trivial to implement.
- Adding an 'option' to the campaign just means I have to maintain two different versions of the core mechanics at the same time. No.

> I'm not trying to imply the code behind it isn't intricate

Except, you did. I'm not sure how else "literally so easy" is supposed to be interpreted. 

> I'm not convinced just asking for the option to make resources incinerate somewhere else besides the core is a game-breaking mountain of a request.

It is. End of story.

PR #4271, opened 2021-01-05
Make engine trails a variable.

MEEPofFaith:
https://user-images.githubusercontent.com/54301439/103712432-abaa6580-4f6e-11eb-8614-d31017f20b7d.mp4

Simple true or false. False by default.
Uses the same `trailLength` as water units, though maybe that should be separated in case your water unit can boost. Wait can water units even boost if `canBoost = true`?

### DIFF
--- a/core/src/mindustry/entities/comp/TrailComp.java
+++ b/core/src/mindustry/entities/comp/TrailComp.java
@@ -7,20 +7,22 @@ import mindustry.graphics.*;
 import mindustry.type.*;
 
 @Component
-abstract class TrailComp implements Unitc{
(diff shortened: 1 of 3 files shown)


Anuken:
Trail rendering is broken. Not adding it unless it's fixed.

MEEPofFaith:
ohno

BlueWolf3682:
this should be true by default if fixed, it would look pretty cool

MEEPofFaith:
Well, you can't really see it one slow flyers like boosting novas

MEEPofFaith:
Q: What's wrong with trails?
A:

PR #4273, opened 2021-01-06
allow player to transfer items with no items if in formation

Slava0135:
https://user-images.githubusercontent.com/40753025/103786687-c30c5000-504d-11eb-8aff-9c6e1660c506.mp4

Also fixes some issues while trying to transfer items with units that have no items or their item doesn't fit container requirements

### DIFF
--- a/core/src/mindustry/graphics/OverlayRenderer.java
+++ b/core/src/mindustry/graphics/OverlayRenderer.java
@@ -171,7 +171,7 @@ public class OverlayRenderer{
             Draw.reset();
 
             Building tile = world.buildWorld(v.x, v.y);
-            if(input.canDropItem() && tile != null && tile.interactable(player.team()) && tile.acceptStack(player.unit().item(), player.unit().stack.amount, player.unit()) > 0 && player.within(tile, itemTransferRange)){
+            if(input.canDropItem() && tile != null && tile.interactable(player.team()) && (tile.acceptStack(player.unit().item(), player.unit().stack.amount, player.unit()) > 0 || player.unit().formation != null) && player.within(tile, itemTransferRange)){
                 Lines.stroke(3f, Pal.gray);
                 Lines.square(tile.x, tile.y, tile.block.size * tilesize / 2f + 3 + Mathf.absin(Time.time, 5f, 1f));
                 Lines.stroke(1f, Pal.place);
--- a/core/src/mindustry/graphics/OverlayRenderer.java
+++ b/core/src/mindustry/graphics/OverlayRenderer.java
@@ -165,13 +165,15 @@ public class OverlayRenderer{
         if(input.isDroppingItem()){
             Vec2 v = Core.input.mouseWorld(input.getMouseX(), input.getMouseY());
             float size = 8;
-            Draw.rect(player.unit().item().icon(Cicon.medium), v.x, v.y, size, size);
+            if(player.unit().formation == null){
+                Draw.rect(player.unit().item().icon(Cicon.medium), v.x, v.y, size, size);
+            }
             Draw.color(Pal.accent);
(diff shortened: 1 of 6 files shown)


Slava0135:
Uploading Peek 2021-01-06 18-46.mp4…

core/src/mindustry/graphics/OverlayRenderer.java
```diff
@@ -171,7 +171,7 @@ public void drawTop(){
             Draw.reset();
 
             Building tile = world.buildWorld(v.x, v.y);
-            if(input.canDropItem() && tile != null && tile.interactable(player.team()) && tile.acceptStack(player.unit().item(), player.unit().stack.amount, player.unit()) > 0 && player.within(tile, itemTransferRange)){
+            if(input.canDropItem() && tile != null && tile.interactable(player.team()) && (tile.acceptStack(player.unit().item(), player.unit().stack.amount, player.unit()) > 0 || player.unit().formation != null) && player.within(tile, itemTransferRange)){
```
Anuken, line 174: This is buggy. I can be in a formation with no items, and yet the unload cursor will still display and I will be able to drag a non-existent item to a container.

If the player has no items and tries to drag a formation member's item into a container, the item displayed will be the last item the player used, not the actual item that will be deposited.

core/src/mindustry/input/InputHandler.java
```diff
@@ -109,9 +109,9 @@ public static void clearItems(Building build){
 
     @Remote(called = Loc.server, unreliable = true)
     public static void transferItemTo(@Nullable Unit unit, Item item, int amount, float x, float y, Building build){
-        if(build == null || build.items == null) return;
+        if(build == null || build.items == null || unit == null || unit.stack.amount <= 0) return;
```
Anuken, line 112: Buildings should still handle items even if the unit is null. Not doing this will cause opportunities for desync.

core/src/mindustry/input/InputHandler.java
```diff
@@ -930,7 +934,7 @@ boolean tryTapPlayer(float x, float y){
     }
 
     boolean canTapPlayer(float x, float y){
-        return player.within(x, y, playerSelectRange) && player.unit().stack.amount > 0;
+        return player.within(x, y, playerSelectRange) && (player.unit().stack.amount > 0 || player.unit().formation != null);
```
Anuken, line 937: The player should not be tappable when there are no items in the formation.

core/src/mindustry/ui/fragments/BlockInventoryFragment.java
```diff
@@ -142,7 +142,7 @@ private void rebuild(boolean actions){
 
                 container.add(i);
 
-                Boolp canPick = () -> player.unit().acceptsItem(item) && !state.isPaused() && player.within(tile, itemTransferRange);
```
Anuken, line 145: This means that the cursor will still display even when the player doesn't have room for the item. Not desirable.

Anuken (changes requested):

Slava0135:
Made that now item is not being drown while trying to transfer items in formation. I am not really sure what is the best alternative.
 
https://user-images.githubusercontent.com/40753025/104124406-c0696d80-5361-11eb-8bd3-2827e3ede9f4.mp4

Transigence:
>  Peek.2021-01-06.18-29.mp4 
> Also fixes some issues while trying to transfer items with units that have no items or their item doesn't fit container requirements
I would like to suggest a more elegant way to solve this problem. Not elegant for the person implementing it, but more elegant for the player.

Instead of allowing the player to drop from wingmen, modify the behavior of putting/dropping such that an amount, up to the space available in the container, is taken evenly from all of the members of the wing, with any remainders taken finally from the player, so that the player is always the last one left with any items.

Anuken:
(formations gone)

PR #4284, opened 2021-01-08
Add block info keybind

joshuaptfan:
This PR adds a keybind (default: backtick) that invokes the same action as clicking the "?" button while a block is selected or hovered in the placement menu. It is only usable in the same situations as when the "?" button is visible.

It saves players the effort of moving the cursor to that corner of the screen when picking blocks with middle-click or block select keys, and allows the functionality to be mapped to controller buttons if needed. It also allows the block info dialog to be opened without actually selecting a block (by just hovering and pressing backtick), thereby saving players 2 clicks.

It does add _yet another_ keybind, and provides quite a small QoL improvement, so I understand if you think this is superfluous and should not be added.

## Added
* 1 new keybind


zethnest:
What about mobile users?

joshuaptfan:
Mobile users are able to use keybinds if they connect a keyboard to their device. This PR doesn't affect or concern players on touchscreens.

zethnest:
I'll assume you have never tried playing Mindustry on mobile with keyboard
and mice peripherals. They're practically useles.

Anuken:
Mindustry's mobile mouse/keyboard support is basically dead. I don't play Mindustry on mobile, and I *definitely* don't test it with keyboards or mice. It's only functional because libGDX had some code to support it.

joshuaptfan:
In response to criticisms on the Discord, I've changed the default keybind to backtick/tilde. That key is next to the number row, which is used for block select, thus grouping these related keys together and making the keybind easier to reach.

Anuken:
F1 might actually be a better choice... I personally haven't merged this because I use the tilde/backtick key for a debug menu. Other may be using it as well.

PR #4287, opened 2021-01-08
Update servers_v6.json for alex servers

alexpvpmindustry:
A brand new game mode to share with the public! Turbo PVP


unjown:
Add pls i tried it it isvery fun

VERY

PR #4292, opened 2021-01-09
Override ModMeta#toString

summetdev:
Just a useful thing for console output (plugin in preview):

image

I'm not sure which fields should be returned and which shouldn't. If I am wrong, please correct me.


Quezler (changes requested):
use "'" instead of '\''?

summetdev:
> use "'" instead of '''?

well, I don't know. Anuke uses ''':
https://github.com/Anuken/Mindustry/blob/master/core/src/mindustry/mod/ModListing.java#L10

PR #4299, opened 2021-01-10
1mil source output

BlueWolf3682:
Just gave the power source an extra zero cause right now i can generate more power than it outputs super easily in campaign. And honestly the last thing the source needed was a nerf.


joshuaptfan:
Good. Gameplay needs should take higher priority than pretty numbers.

BlueWolf3682:
(your link doesnt work)

Quezler:
> Good. Gameplay needs should take higher priority than [pretty numbers](https://github.com/Anuken/Mindustry/pull/4088).

jokes on you, 1 million is still a pretty number 👀

PR #4302, opened 2021-01-10
add resolveDependencies gradle task

petabyteboy:
This task will download all dependecies, including transitive dependencies. This is very useful when packaging Mindustry for a distribution, as we can first download the dependencies in a seperate step, archive them, and then disable network access to have a more reproducible build.
I added this in https://github.com/NixOS/nixpkgs/pull/108274, and someone suggested it could be useful for others.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -410,3 +410,13 @@ task deployAll{
     dependsOn "server:deploy"
     dependsOn "android:deploy"
 }
+
+task resolveDependencies{
+    doLast{
+        rootProject.allprojects{project ->
+            Set<Configuration> configurations = project.buildscript.configurations + project.configurations
+            configurations.findAll{c -> c.canBeResolved}
+                .forEach{c -> c.resolve()}
+        }
+    }
+}


Quezler:
why exactly? you have no description as to why we need it or why its beneficial?

petabyteboy:
I clicked a bit too fast and am currently writing the description :smile:

build.gradle
```diff
@@ -410,3 +410,13 @@ task deployAll{
     dependsOn "server:deploy"
     dependsOn "android:deploy"
 }
+
+task resolveDependencies {
```
Anuken, line 414: Incorrect style - space before brace, see the [guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md#style-guidelines).

Anuken (changes requested):

build.gradle
petabyteboy: Of course, sorry for not noticing first. Is it better now?

PR #4308, opened 2021-01-11
makes parallax be able to attack enemies with boosters while they are flying

Slava0135:
https://user-images.githubusercontent.com/40753025/104159151-dcbbe780-53ff-11eb-8e24-7f365cdf2754.mp4

### DIFF
--- a/core/src/mindustry/world/blocks/defense/turrets/TractorBeamTurret.java
+++ b/core/src/mindustry/world/blocks/defense/turrets/TractorBeamTurret.java
@@ -73,7 +73,7 @@ public class TractorBeamTurret extends BaseTurret{
 
             //retarget
             if(timer(timerTarget, retargetTime)){
-                target = Units.closestEnemy(team, x, y, range, u -> u.checkTarget(targetAir, targetGround));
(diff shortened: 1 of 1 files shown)


Quezler:
why not use one of these functions?
https://github.com/Anuken/Mindustry/blob/7ea12e97ecc1216fcdbdf7812dfdb25c7d56670e/core/src/mindustry/entities/comp/FlyingComp.java#L35

core/src/mindustry/world/blocks/defense/turrets/TractorBeamTurret.java
```diff
@@ -73,7 +73,7 @@ public void updateTile(){
 
             //retarget
             if(timer(timerTarget, retargetTime)){
-                target = Units.closestEnemy(team, x, y, range, u -> u.checkTarget(targetAir, targetGround));
+                target = Units.closestEnemy(team, x, y, range, Flyingc::isFlying);
```
Quezler, line 76: do ground units implement that too? if not it might crash 🤔

Anuken:
The check for flying units should not be hard-coded.

PR #4315, opened 2021-01-12
Updated editor.errorlegacy text

XEN0PHIL:
The button has changed to **Import Image File** long ago


Anuken:
That text needs to be changed altogether. 3.5 maps can't be imported at all.

PR #4317, opened 2021-01-13
Scale impact reactor warmup time with boost

Quezler:
Screen Shot 2021-01-13 at 08 17 53

- uses `timeScale()` instead of `timeScale` to make it consistent with the rest of the file
- spin down time left unaffected, it would probably just not look good at all

### DIFF
--- a/core/src/mindustry/world/blocks/power/ImpactReactor.java
+++ b/core/src/mindustry/world/blocks/power/ImpactReactor.java
@@ -72,7 +72,7 @@ public class ImpactReactor extends PowerGenerator{
             if(consValid() && power.status >= 0.99f){
                 boolean prevOut = getPowerProduction() <= consumes.getPower().requestedPower(this);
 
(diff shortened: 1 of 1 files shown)


Anuken:
Not sure if this will give the desired effect. Using `lerp` with the same progress-value each time doesn't give you a linear curve; multiplying the progress speed by, say, 2, will likely speed up warmup by more than 2x.

Quezler:
so what should i do?

Anuken:
Can you compare warmup times with a normal reactor and a dome to see if the difference is significant?

Quezler:
dome is in range of the top one:

https://user-images.githubusercontent.com/3179271/104470952-8e067d00-55ba-11eb-8c1b-68d2f302824f.mp4

Quezler:
just ran an experiment with this, and its basically the same increase as inlining the timescale, no noticable difference:
`warmup += (Mathf.lerpDelta(warmup, 1f, warmupSpeed) - warmup) * timeScale();`

PR #4324, opened 2021-01-14
makes unit pathfind in formation if they are too far away from the leader

Slava0135:
https://user-images.githubusercontent.com/40753025/104589827-8064fc00-567b-11eb-851a-ab2925ba323f.mp4

### DIFF
--- a/core/src/mindustry/ai/types/FormationAI.java
+++ b/core/src/mindustry/ai/types/FormationAI.java
@@ -50,11 +50,7 @@ public class FormationAI extends AIController implements FormationMember{
 
         Vec2 next = vec.set(target).add(leader.vel);
         if(unit.isGrounded() && !unit.within(leader, 4 * formation.pattern.spacing)){
-            if(unit instanceof WaterMovec){
-                next = vec.set(Astar.pathfind(unit.tileOn(), leader.tileOn(), FormationAI::navalPathfindCost, t -> !t.solid() && t.floor().isLiquid).first());
-            }else{
-                next = vec.set(Astar.pathfind(unit.tileOn(), leader.tileOn(), FormationAI::pathfindCost, t -> !(t.solid())).first());
-            }
+            next = vec.set(Astar.pathfind(unit.tileOn(), leader.tileOn(), FormationAI::pathfindCost, t -> unit.canPass(t.x, t.y)).first());
         }
 
         float speed = unit.realSpeed() * unit.floorSpeedMultiplier() * Time.delta;
@@ -109,9 +105,5 @@ public class FormationAI extends AIController implements FormationMember{
 
     private static float pathfindCost(Tile tile){
         return 1f + (tile.floor().isLiquid ? 10f : 0f);
-    }
-
-    private static float navalPathfindCost(Tile tile){
-        return 1f;
     }
 }
--- a/core/src/mindustry/ai/types/FormationAI.java
+++ b/core/src/mindustry/ai/types/FormationAI.java
@@ -3,9 +3,11 @@ package mindustry.ai.types;
 import arc.math.*;
 import arc.math.geom.*;
 import arc.util.*;
+import mindustry.ai.*;
 import mindustry.ai.formations.*;
 import mindustry.entities.units.*;
 import mindustry.gen.*;
+import mindustry.world.*;
 import mindustry.world.blocks.storage.CoreBlock.*;
 
 public class FormationAI extends AIController implements FormationMember{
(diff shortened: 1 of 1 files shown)


Slava0135:
*not working with naval units atm

Slava0135:
https://user-images.githubusercontent.com/40753025/104593669-0f284780-5681-11eb-9f85-bc808f532a10.mp4

Slava0135:
*I was experimenting with unit acceleration. I don't like that it's 1 ALL the time. Should it be scalable? The closer the unit to the leader - the bigger the acceleration or whatever.

DeltaNedas:
very good

Slava0135:
I am not really sure about implementation details. I could put the pathfinder thing into abstract AI class to use it in Logic.

core/src/mindustry/ai/types/FormationAI.java
```diff
@@ -97,4 +106,13 @@ public boolean isBeingControlled(Unit player){
     public Vec3 formationPos(){
         return target;
     }
+
+    private static float pathfindCost(Tile tile){
+        return 1f +
+                (tile.floor().isLiquid ? 10f : 0f);
```
Quezler, line 112:
```suggestion
        return 1f + (tile.floor().isLiquid ? 10f : 0f);
```

core/src/mindustry/ai/types/FormationAI.java
```diff
@@ -46,10 +48,17 @@ public void updateUnit(){
             unit.lookAt(unit.vel.angle());
         }
 
-        Vec2 realtarget = vec.set(target).add(leader.vel);
+        Vec2 next = vec.set(target).add(leader.vel);
+        if(unit.isGrounded() && !unit.within(leader, 4 * formation.pattern.spacing)) {
+            if(unit instanceof WaterMovec) {
```
Quezler, line 53:
```suggestion
            if(unit instanceof WaterMovec){
```

core/src/mindustry/ai/types/FormationAI.java
```diff
@@ -46,10 +48,17 @@ public void updateUnit(){
             unit.lookAt(unit.vel.angle());
         }
 
-        Vec2 realtarget = vec.set(target).add(leader.vel);
+        Vec2 next = vec.set(target).add(leader.vel);
+        if(unit.isGrounded() && !unit.within(leader, 4 * formation.pattern.spacing)) {
```
Quezler, line 52:
```suggestion
        if(unit.isGrounded() && !unit.within(leader, 4 * formation.pattern.spacing)){
```

Quezler (changes requested):

summetdev:
make units fly if they can?

Slava0135:
> make units fly if they can?

Idk really - sometimes you don't want it, if there are AA turrets.

core/src/mindustry/ai/types/FormationAI.java
```diff
@@ -97,4 +106,12 @@ public boolean isBeingControlled(Unit player){
     public Vec3 formationPos(){
         return target;
     }
+
+    private static float pathfindCost(Tile tile){
+        return 1f + (tile.floor().isLiquid ? 10f : 0f);
+    }
+
+    private static float navalPathfindCost(Tile tile){
```
summetdev, line 114: what's the point of making a separate private method that just returns one? as I see there is only one usage in code.

Slava0135: Yes but I am not sure about euristics. It can be modified later. But atm yes - it doesn't do anything.

core/src/mindustry/ai/types/FormationAI.java
```diff
@@ -46,10 +48,13 @@ public void updateUnit(){
             unit.lookAt(unit.vel.angle());
         }
 
-        Vec2 realtarget = vec.set(target).add(leader.vel);
+        Vec2 next = vec.set(target).add(leader.vel);
+        if(unit.isGrounded() && !unit.within(leader, 4 * formation.pattern.spacing)){
+            next = vec.set(Astar.pathfind(unit.tileOn(), leader.tileOn(), FormationAI::pathfindCost, t -> unit.canPass(t.x, t.y)).first());
```
Anuken, line 53: You cannot simply use A* to pathfind here. This is **horribly** inefficient and spews garbage every frame.

Slava0135: but nothing else would work unless... tell the unit to move next to walls until it's free...

Slava0135: won't work really. But can we at least calculate the path once and follow it unless it ends or we are close enough

Anuken: > calculate the path once

- What if the player unit is inaccessible?
- What if it's on the other side of the map? 
- What if it moves behind an obstacle?
- What if the formation unit gets shoved or otherwise displaced somewhere?
- What if this is running on any number of devices that can't pathfind from point A to point B in <16ms?
- What if there are many units that need to calculate the path?

Anuken:
(Closing, as these issues are unlikely to be addressed)

PR #4327, opened 2021-01-14
plastanium - now colorblind friendly (was too similar to surge)

Slava0135:
*waffle*

item-plastanium


ThatOneBepis:
_waffle_

BlueWolf3682:
im trying and failing to find something other to say than "h" but silence isnt an option

Anuken:
I don't like the look of this version at all. It's better to discuss/decide in #pulls rather than here.

PR #4338, opened 2021-01-15
Readme: Formatting

summetdev:
**Before:**
... [here.](https://youtube.com/watch?v=dQw4w9WgXcQ)

**After:**
... [here](https://youtube.com/watch?v=dQw4w9WgXcQ).

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Server builds are bundled with each released build (in Releases). If you'd rathe
 
 ### Android
 
-1. Install the Android SDK [here.](https://developer.android.com/studio#downloads) Make sure you're downloading the "Command line tools only", as Android Studio is not required.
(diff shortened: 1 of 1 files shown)


Anuken:
Why?

summetdev:
In my opinion, this is more correct, because the link refers to the word (in this case, "here") and the period does not related to the link or word, it is just the end of the sentence.

Quezler (approved):

BlueWolf3682:
thanks for the double rickroll

Anuken:
> In my opinion, this is more correct

That's just nitpicking. Removing the period may seem 'more correct', but it also makes the markdown slightly less readable and the link a little harder to click. In general, I'm fine with formatting changes, but making an entire PR just to move a period is "*hacktoberfest hall of shame*" material.

PR #4341, opened 2021-01-16
Update getUsid: prefer use hostname

way-zer:
As cookie is saved based on host. Usid should based on hostname rather than ip address.


way-zer:
should ```Core.settings.put("usid-" + ip, usidByName);``` before return?

PR #4353, opened 2021-01-17
Team colored trails

MEEPofFaith:


Anuken:
This isn't something that the base game will use, so it doesn't belong here.

PR #4354, opened 2021-01-17
conductsTo() returning whether to conduct itself to adjacent builds

sk7725:
+ Allows certain blocks to only conduct on 1 or 2 sides, serving as an insulated wire
+ Allows modded blocks to use adjacent blocks as both inputs and outputs(ex. a logic gate, comparator etc.)
+ Insulated blocks will no longer conduct even if it hasPower; you need to manually connect it via nodes too!

image
+ Practical, real modded use example: make piston not take(accept) power from its "arm" side

### DIFF
--- a/core/src/mindustry/entities/comp/BuildingComp.java
+++ b/core/src/mindustry/entities/comp/BuildingComp.java
@@ -789,8 +789,8 @@ abstract class BuildingComp implements Posc, Teamc, Healthc, Buildingc, Timerc, 
 
         for(Building other : proximity){
             if(other != null && other.power != null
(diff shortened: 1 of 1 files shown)


sk7725:
Some additional notes:
Q. You can just add the whole insulated check in the getPowerConnections() and not use a new function conductsTo()
A. Yes, but modded blocks are still stuck with problems mentioned above. Adding conductsTo() will grant even further freedom to mods, which is always a good thing. Currently, while a modded block can *refuse* to add a Solar Panel to its grid, the Solar Panel will just not give a thought and shove the modded block in its grid no matter what.

PR #4356, opened 2021-01-17
update plastanium conveyor description to include unloading.

Catchears:
Seeing as there's nothing that tells the player that they can use unloaders (idea from [here](https://github.com/Anuken/Mindustry-Suggestions/issues/1809#issuecomment-761621137)), I thought I'd add that (and the german translation). If you don't like this idea or how I've written it, feel free to suggest changes / close this / implement it yourself.


Quezler (approved):

Anuken:
This should not be in the "purpose" section. It is not related to the core function of the block.

Catchears:
I'm not sure how to put it in any other block description areas or if you even want this, but if you do, I'll let you or someone else implement it.

PR #4365, opened 2021-01-18
mass driver player transport

Quezler:
> prototype

getting around on large maps is a chore, it would be nice if some sort of highways can be established for players.

this draft uses mass drivers for it, which will just send you to the next driver in the chain instead of allowing you to vent to any in range.

mass drivers just for player transport may be a bit overkill, so other methods like traversing power nodes somehow could/should also be considered. (to avoid unused mass drivers everywhere, and power poles are already abundant)

aside from figuring that out, this thing would probably also have to rotate or something to make it feel a bit more smooth, currently it just teleports you to the next whenever the previous soul particle has expired.

https://user-images.githubusercontent.com/3179271/104901041-2450ef00-597d-11eb-8a61-baaff1adadb3.mp4

### DIFF
--- a/core/src/mindustry/input/InputHandler.java
+++ b/core/src/mindustry/input/InputHandler.java
@@ -876,7 +876,7 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
         boolean consumed = false, showedInventory = false;
 
         //check if tapped block is configurable
-        if(build.block.configurable && build.interactable(player.team())){
+        if(build.block.configurable && build.interactable(player.team()) && !Core.input.keyDown(Binding.control)){
             consumed = true;
             if(((!frag.config.isShown() && build.shouldShowConfigure(player)) //if the config fragment is hidden, show
             //alternatively, the current selected block can 'agree' to switch config tiles
--- a/core/src/mindustry/world/blocks/distribution/MassDriver.java
+++ b/core/src/mindustry/world/blocks/distribution/MassDriver.java
@@ -16,6 +16,7 @@ import mindustry.gen.*;
 import mindustry.graphics.*;
 import mindustry.type.*;
 import mindustry.world.*;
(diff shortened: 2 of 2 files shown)


summetdev:
maybe make mass drivers consume energy for this? :eyes:

BlueWolf3682:
nice concept but i dont get the point of it, and it doesnt make game sense at all and its just adding bloat to the game

and the current units are fast enough, map transit isnt a problem for me and i dont know why it would be one for anyone

additionally, you cannot do anything once you reach the destination driver unless theres a handy unit around for you to control and use effectively

you could also use routers to achieve this affect by having them positioned close enough to reach but far enough to warrant quick moving

this is my blunt opinion, unfiltered and with sleep deprivation but ill stand by it even once ive gotten sleep, its cool but not the most sensical thing

Violet-Scarelli:
I actually tried to make something like this with logic, making a unit available at either end of a router line. Personally, using mass drivers to do this seems like a step in the right direction, but it needs a little more. How I would execute it is as follows: A secondary power-style network that requires thorium to make, but a player can possess a node and zip through it with the arrow keys, getting ejected back into a default mech when moving out of the network.

Anuken:
This idea seems absolutely bizzare to me. Mass drivers and power networks are only for distribution of items/power, they shouldn't have extra "player respawning" features tacked on to them.

Perhaps something like a ship-building "checkpoint" could be considered in the future, but honestly I don't think this is something the (vanilla) game needs. If you really need to get somewhere fast, why not build support units at the location and pan over with the camera? It's much faster than standard movement, and you get the benefit of automatic re-posession of similar units upon death.

DeltaNedas:
use router possession network instead

Quezler:
fair enough anuke, and it was just a prototype anyways 👍

PR #4378, opened 2021-01-19
Add kr server

Kieaer:
There're many steam servers that ban and kick users who still enter even though warnings popup have been added. https://github.com/Anuken/Mindustry/pull/4042

I looked for other measures, change warning texts and it was https://github.com/Anuken/Mindustry/pull/4131.
but didn't work!

I recently found out that most of them are Koreans.
kr server is still under development, but it may be able to solve this problem.

The server check/maintence every Saturday/Sunday.
At the time of writing, the server was not opened because the CPU was being upgraded.

### DIFF
--- a/servers_v6.json
+++ b/servers_v6.json
@@ -58,5 +58,9 @@
   {
     "name": "Minigames",
     "address": ["shizashizashiza.ml"]
+  },
+  {
+    "name": "Korea"
+    "address": ["mindustry.kr"]
   }
 ]
--- a/servers_v6.json
+++ b/servers_v6.json
@@ -60,7 +60,7 @@
     "address": ["shizashizashiza.ml"]
   },
   {
-    "name": "Korea"
+    "name": "Korea",
     "address": ["mindustry.kr"]
   }
 ]


servers_v6.json
```diff
@@ -58,5 +58,9 @@
   {
     "name": "Minigames",
     "address": ["shizashizashiza.ml"]
+  },
+  {
+    "name": "Korea"
```
Blockzilla101, line 63: missing comma
```suggestion
    "name": "Korea",
```

Blockzilla101 (changes requested):

servers_v6.json
DeltaNedas: bruh

Kieaer: 🤦

Kieaer:
The server is open and playable, but server information can't be checked from outside.
Maybe it's presumed to be a carrier bug.

Anuken:
> The server is open and playable, but server information can't be checked from outside.
Maybe it's presumed to be a carrier bug.

That means it won't show up in the server list.

Kieaer:
@Anuken Fixed.

PR #4379, opened 2021-01-19
Make tests fail if any server json is invalid

Quezler:
its happening a little more often than never that a server pull request contains invalid json, but the tests still pass.

this pull validates the 3 server json files, and fails the test if there's an issue:

Screen Shot 2021-01-19 at 16 30 52

Screen Shot 2021-01-19 at 16 31 03

keeping this in draft mode because:
- the way i accessed the files is probably a weird workaround (landing in the core assets and .. ing up)
- using the rhino json parser because i could not find another (built in) thing that could throw a json exception
- probably not the right way to catch an exception within a test, but this is unfamiliar terrain for me
- doesn't use a commandline call or something to validate json externally since that relies on the os its built on

### DIFF
--- a/tests/src/test/java/ApplicationTests.java
+++ b/tests/src/test/java/ApplicationTests.java
@@ -19,6 +19,11 @@ import mindustry.net.Net;
 import mindustry.type.*;
 import mindustry.world.*;
 import org.junit.jupiter.api.*;
+import rhino.*;
+import rhino.json.*;
+import rhino.json.JsonParser.*;
+
+import java.io.*;
 
 import static mindustry.Vars.*;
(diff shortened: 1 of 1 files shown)


Anuken:
Good idea, but using the rhino json parser seems.. strange.

See commit.

Quezler:
> Good idea, but using the rhino json parser seems.. strange.

well i couldn't really get the others to throw an exception when reading

Anuken:
True, they don't throw exceptions with "technically invalid" trailing commas. However, these are the same parsers I use *in the client*, so if they don't throw errors, it shouldn't matter.

PR #4383, opened 2021-01-20
fix "exit confirmation when saved"

liuxilu:
If it's intended, should we change the text to `please comfirm` but not `not saved`?

### DIFF
--- a/core/src/mindustry/editor/MapEditorDialog.java
+++ b/core/src/mindustry/editor/MapEditorDialog.java
@@ -665,7 +665,7 @@ public class MapEditorDialog extends Dialog implements Disposable{
 
     private void tryExit(){
         if(saved){
-            this::hide.run();
+            hide.run();
         }else{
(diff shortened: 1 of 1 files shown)


core/src/mindustry/editor/MapEditorDialog.java
```diff
@@ -664,7 +664,11 @@ private void doInput(){
     }
 
     private void tryExit(){
-        ui.showConfirm("@confirm", "@editor.unsaved", this::hide);
+        if(saved){
+            this::hide.run();
```
Quezler, line 668:
```suggestion
            hide.run();
```

Anuken:
This is intended. The english text is "*Are you sure you want to exit?\n[scarlet]Any unsaved changes will be lost.*". It doesn't say that it's unsaved.

There is no "saved" check, because many actions like changing rules or waves don't activate it.

liuxilu:
Oh, time to update the translation!

PR #4384, opened 2021-01-20
New team indication

Slava0135:
image
image

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -841,7 +841,6 @@ setting.playerlimit.name = Player Limit
 setting.chatopacity.name = Chat Opacity
 setting.lasersopacity.name = Power Laser Opacity
 setting.bridgeopacity.name = Bridge Opacity
-setting.teamoutlineopacity.name = Team Outline Opacity
 setting.playerchat.name = Display Player Bubble Chat
 setting.showweather.name = Show Weather Graphics
 public.confirm = Do you want to make your game public?\n[accent]Anyone will be able to join your games.\n[lightgray]This can be changed later in Settings->Game->Public Game Visibility.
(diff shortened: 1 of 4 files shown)


Slava0135:
*not sure about turret bases

BlueWolf3682:
green has no lower area and i see z fighting

joshuaptfan:
The concept has problems on various terrain types of varying color, such as spore moss, grass, etc.

Slava0135:
image

unjown:
this is actually a good idea but i think its too bright?? maybe make it more transparent

Slava0135:
ok, I am having problems with parallax base coloring. Probably should revert the effect,,,

Slava0135:
image

Slava0135:
image
*alpha 75%

Slava0135:
> green has no lower area and i see z fighting

It's because I am green lol

Slava0135:
> The concept has fundamental problems on various terrain types of varying color, such as sand, spore moss, grass, etc.

But it's better than nothing. Without it it's even harder to see...

Slava0135:
If you don't like it - tell me why. Everything can be changed.

Slava0135:
ok, I removed the turret base coloring. Also I added a setting to configure the outline opacity.

Slava0135:
https://user-images.githubusercontent.com/40753025/105174902-45a80b80-5b34-11eb-9621-b15654a86dcd.mp4

Slava0135:
image

Slava0135:
image

Slava0135:
*removed opacity option until intersections will be solved

Slava0135:
image

Anuken:
Tested this out a bit on some servers, and while it doesn't look *awful* for most blocks, it's not really an improvement either.

PR #4386, opened 2021-01-20
Better shock mine

Slava0135:
Now it blinks and turns off on low health

https://user-images.githubusercontent.com/40753025/105178286-ded92100-5b38-11eb-8056-00cceb84f95b.mp4


sk7725 (changes requested):
Draw only for your team, aka
if(team != player.team) return;

Slava0135:
https://user-images.githubusercontent.com/40753025/105180821-18f7f200-5b3c-11eb-9774-8111a335b205.mp4

joshuaptfan:
Still think the simple square is ugly and doesn't fit in with the sprite. It would look good if it was the 4 semicircles on the edge that had team color and blinked.

Anuken:
- Changed the team sprite myself.
- I don't think shock mines should stay around, they should just be rebuildable if necessary.

PR #4388, opened 2021-01-20
fix votekick button becoming unfocused 3x a second

buthed010203:
The votekick button will unclick itself every time the player list rebuilds, this is a janky way to stop it from doing so.

### DIFF
--- a/core/src/mindustry/ui/fragments/PlayerListFragment.java
+++ b/core/src/mindustry/ui/fragments/PlayerListFragment.java
@@ -2,7 +2,7 @@ package mindustry.ui.fragments;
 
 import arc.*;
 import arc.graphics.g2d.*;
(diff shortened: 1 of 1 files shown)


core/src/mindustry/ui/fragments/PlayerListFragment.java
```diff
@@ -2,6 +2,7 @@
 
 import arc.*;
 import arc.graphics.g2d.*;
+import arc.input.KeyCode;
```
summetdev, line 5:
```suggestion
import arc.input.*;
```

Anuken:
This is arguably more janky than letting the button get deselected. I've never focused the button long enough to even notice this issue - I don't think it's a real problem.

PR #4392, opened 2021-01-21
Charge stat for Item

sk7725:
+ Makes electric discharge, similar to batteries, when destroyed.

- [x] Add it to surge alloy
- [x] Tweak onDestroyed()
- [x] Add Stat.discharge
- [x] Bundle mundle
- [x] Rename to "charge"
- [x] Surge unit explosions


core/src/mindustry/type/Item.java
```diff
@@ -17,6 +17,8 @@
     public float flammability = 0f;
     /** how radioactive this item is. 0=none, 1=chernobyl ground zero */
     public float radioactivity;
+    /** how electrically potent this item is.*/
```
summetdev, line 20:
```suggestion
    /** how electrically potent this item is. */
```

Erisfiregamer1:
haha yes surge crate go zap zap

Anuken:
This should be called "charge", not "discharge". "discharge" is more of a verb. 

Charge should also be part of unit explosion calculations.

sk7725:
> Charge should also be part of unit explosion calculations.

Good idea. Surge suicide bombing would be a tactic but a fairly expensive one at least.

sk7725:
I think I'm done.

PR #4400, opened 2021-01-21
Replace logic to clipboard icon with export

Quezler:
haven't played around with logic all to much, but initially i couldn't really find the export button on first glance, both buttons only interact with the clipboard yet for one the import icon is used, and for the other the copy/clipboard one, would be nice if both icons matched (since usually the export button is the one you press before you select any of a few methods)

"Copy to clipboard" may have to change to "Export to clipboard" to be more in line with the message above it 🤔 

schematic export:
Screen Shot 2021-01-21 at 15 44 45

logic export:
Screen Shot 2021-01-21 at 15 44 57

schematic import:
Screen Shot 2021-01-21 at 15 47 26


Anuken:
I don't think that's a good reason to change it. The schematic export dialog already uses the copy icon for the clipboard, which is consistent with the logic dialog.

PR #4412, opened 2021-01-22
Add controlling count to logic processors

Quezler:
after asking around in the #logic channel on discord it seems to be quite an overhead/chore to check for when an unit dies / gets commanded / playered / etc, personally it seems a bit messy to me and it would just be easier to have the logic block itself gain access to the count of units under its control. (basically what you get when hovering over it with your cursor)

maybe i'm overreacting and this isn't needed and there's a really simple solution i'm not seeing, but being able to jump over an unit binder block if the controlled count is above a certain amount would seemingly make it way cleaner to scale stuff, as well as easily stock the controlled units if an unit gets ripped from its grasp in any way shape or form.

i'd like to invite "pro logic users" to give their reaction to this, whether it would be useful or not / etc, imo it is. 🤔 

Screen Shot 2021-01-22 at 16 36 04
Screen Shot 2021-01-22 at 16 36 08
Screen Shot 2021-01-22 at 16 36 13
Screen Shot 2021-01-22 at 16 36 23

### DIFF
--- a/core/src/mindustry/world/blocks/logic/LogicBlock.java
+++ b/core/src/mindustry/world/blocks/logic/LogicBlock.java
@@ -486,6 +486,12 @@ public class LogicBlock extends Block{
         }
 
         @Override
+        public double sense(LAccess sensor){
+            if(sensor == LAccess.controlling) return Groups.unit.count(u -> u.controller() instanceof LogicAI ai && ai.controller == this);
+            return super.sense(sensor);
+        }
+
+        @Override
         public byte version(){
             return 1;
         }
--- a/core/src/mindustry/logic/LAccess.java
+++ b/core/src/mindustry/logic/LAccess.java
@@ -33,6 +33,7 @@ public enum LAccess{
     team,
     type,
     flag,
+    controlling,
     controlled,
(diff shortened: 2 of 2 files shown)


core/src/mindustry/world/blocks/logic/LogicBlock.java
```diff
@@ -485,6 +485,12 @@ public boolean onConfigureTileTapped(Building other){
             return super.onConfigureTileTapped(other);
         }
 
+        @Override
+        public double sense(LAccess sensor){
+            if(sensor == LAccess.controlling) return Groups.unit.count(u -> u.controller() instanceof LogicAI ai && ai.controller == this);
```
Anuken, line 490: Two issues with this:
- O(n) search through all units is very slow and not an acceptable solution
- This should be a global constant like `@unit`, sensing controlled count of self doesn't fit how other variables are fetched

Anuken:
Implemented with a different approach.

Anuken:
*EDIT: Turned out to be too flaky and unreliable even when it worked properly, I'll likely try another solution.*

PR #4414, opened 2021-01-23
Make schematics build insulated blocks first, power nodes last

joshuaptfan:
This PR sorts the build order of schematics when placed, so that small plastanium walls will be pushed to the beginning of the build queue, and all power nodes to the end.

This change fixes the following issues:
* Even though schematics in 6.0 remember power connections, that didn't matter at all as power nodes built halfway through the build order would often max themselves out with automatic connections before they could connect to their saved connections.
* For schematics surrounded by plastanium walls, power nodes outside the schematic could still make automatic connections to blocks inside the schematic during construction because some blocks would be built before the insulator walls were.
* Likewise, power nodes inside could make automatic connections out.
* Blocks outside a schematic would often steal power nodes' automatic connections so that by the time a schematic finished building, no connections would be left and some blocks in the schematic would be left unpowered.
* Schematics that used batteries to reduce the number of power connections needed would find some blocks left unpowered due to power nodes being constructed before some batteries, causing multiple automatic connections to be wasted on blocks that could have shared 1 connection via batteries.

Thus, schematics' power connections will look much neater after being built, and players will no longer have to waste time manually fixing bad automatic power connections.

As with anything involving build order, this PR does not address build order not being respected due to the player unit being out of build range. It also doesn't alter the build order of manually built designs.

Forgive these unrealistic, contrived examples.

Schematic Build Order 1

Schematic Build Order 2

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -788,6 +788,7 @@ public class Blocks implements ContentList{
             health = 130 * wallHealthMultiplier;
             insulated = true;
             absorbLasers = true;
+            schematicPriority = 10;
         }};
 
         plastaniumWallLarge = new Wall("plastanium-wall-large"){{
--- a/core/src/mindustry/game/Schematics.java
+++ b/core/src/mindustry/game/Schematics.java
@@ -285,7 +285,8 @@ public class Schematics implements Loadable{
     /** Creates an array of build requests from a schematic's data, centered on the provided x+y coordinates. */
     public Seq<BuildPlan> toRequests(Schematic schem, int x, int y){
         return schem.tiles.map(t -> new BuildPlan(t.x + x - schem.width/2, t.y + y - schem.height/2, t.rotation, t.block, t.config).original(t.x, t.y, schem.width, schem.height))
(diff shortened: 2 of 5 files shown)


Quezler:
and what if it runs out of plastanium / metaglass during construction, will it wait for that or build the rest first?

Quezler:
also, in the spirit of not hardcoding blocks, what about adding `int schematicPriority` to blocks, have it 0 for default, 1 for insulated blocks and -1 for power nodes? (or larger spacing like the layers to allow for future stuff to slide in between, but i would avoid floats since prescision could cause blocks of seemingly the same priority to shuffle amongst themselves)

joshuaptfan:
> and what if it runs out of plastanium / metaglass during construction, will it wait for that or build the rest first?

Current behavior is not changed. Blocks that can't be immediately built are skipped over.

> also, in the spirit of not hardcoding blocks, what about adding int schematicPriority to blocks, have it 0 for default, 1 for insulated blocks and -1 for power nodes?

Great idea, and that simplifies the sort too.

Quezler:
don't forget the other insulated block: the diode

maybe put that at 10 or 9?

Quezler (approved):

PR #4417, opened 2021-01-23
Colored construct

MEEPofFaith:
Adds `Drawf.construct` but with custom colors. Useful for script modders who want to construct something in a specific color without having to do the whole shader thing themself.


Quezler:
a description would be nice, preferably with before/after screenshots or a gif

MEEPofFaith:
Good point.

PR #4419, opened 2021-01-23
Mention Foreshadow's special targeting in its description.

MEEPofFaith:
Since there's no way to tell without watching it attack at the moment.


core/assets/bundles/bundle.properties
```diff
@@ -1446,7 +1446,7 @@ block.ripple.description = Shoots clusters of shells at ground enemies over long
 block.cyclone.description = Fires explosive clumps of flak at nearby enemies.
 block.spectre.description = Fires large armor-piercing bullets at air and ground targets.
 block.meltdown.description = Charges and fires a persistent laser beam at nearby enemies. Requires coolant to operate.
-block.foreshadow.description = Fires a large single-target bolt over long distances.
+block.foreshadow.description = Fires a large single-target bolt over long distances. Prioritieses units with higher max health.
```
Quezler, line 1449: should be either prioritises or prioritizes:

```suggestion
block.foreshadow.description = Fires a large single-target bolt over long distances. Prioritizes units with higher max health.
```

joshuaptfan: Perhaps should be "enemies", not "units". That's the same wording [the original dev preview](https://discord.com/channels/391020510269669376/513178557653188609/760712452777574410) used.

catatagun:
Hmmm yes it does not say foreshadow target highest hp unit

PR #4433, opened 2021-01-24
Reliable disabling of accepting payloads from payload routers.

FrederickDoering:
Without this, payload conveyors would accept 1 payload 1/8th of the time depending on the time at which they were disabled. Reconstructors would always  accept a payload regardless of if they were disabled.
This does not effect the behavior of dropping units on disabled conveyors.


FrederickDoering:
The use case/ why this matters is for precision control of payload routers. If you want a payload to go north, so you disable the conveyors in the other direction(s), there's a chance* the payload will get stuck trying to enter a disabled conveyor anyway. Similarly, for reconstructors, except they currently always accept payloads.

* if the conveyor was disabled during the first 5 ticks of it's 40 tick cycle it would still accept a payload from a router.

PR #4438, opened 2021-01-26
targetable() for Units

sk7725:
+ Main targetable() is placed in UnitType so it is not necessary to create a new constructor
+ targetable() is also in Unit so other future features can be implemented; such as an "Untargetable core units" rule, currently it just calls the targetable() of the type
+ nearby() and the similar functions in Units.java are intentionally unchanged, meaning that it does not consider targetable(), since not being targetable does not always imply invincibility

- [x] UnitType targetable(Unit unit, Team team)
- [x] Unit targetable(Team team)
- [x] Units.java

### DIFF
--- a/core/src/mindustry/entities/Units.java
+++ b/core/src/mindustry/entities/Units.java
@@ -93,7 +93,7 @@ public class Units{
      * @return whether the target is invalid
      */
     public static boolean invalidateTarget(Posc target, Team team, float x, float y, float range){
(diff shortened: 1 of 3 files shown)


sk7725:
Should I add a statusEffect that cloakes a given unit?

sk7725:
update: might as well add intangible(), or not, depends on whether isValid() manipulation will be enough

Anuken:
implement, see discord convo

PR #4439, opened 2021-01-26
Set locale when loading external bundle

liuxilu:
Trying to fix https://github.com/DeltaNedas/rtfm/issues/14


Anuken:
Shouldn't this be handled by the mod itself..? I don't understand how this is related to the main game.

PR #4447, opened 2021-01-27
Secondary bloom layers.

MEEPofFaith:
Some layers that bloom above flying units, allowing for bullets and effects that bloom above them, as well as making launchpad engine draw over flying units while still blooming.

https://user-images.githubusercontent.com/54301439/105931969-06c8f880-6001-11eb-80cd-32095cd5f033.mp4

(Note: I have no idea what performance issues, if any, this'll cause.)

### DIFF
--- a/core/src/mindustry/content/Fx.java
+++ b/core/src/mindustry/content/Fx.java
@@ -752,7 +752,7 @@ public class Fx{
         randLenVectors(e.id, 3, 2f + e.fin() * 7f, (x, y) -> {
             Fill.circle(e.x + x, e.y + y, 0.1f + e.fout() * 1.4f);
         });
-    }),
+    }).layer(Layer.highEffect),
 
     fire = new Effect(50f, e -> {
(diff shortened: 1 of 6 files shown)


sk7725 (changes requested):
About half of current effects need to be moved above the units (ex. all StatusEffect's effects). 
Some bullets need to be moved above too (don't forget to move its shootEffect and trailEffect up the layer too!).
And maybe rename air and sky to something that implies that this layer contains bloom.

MEEPofFaith:
I had no idea what to name them so I just gave them "high altitude" sounding names.

MEEPofFaith:
java_tIS5tpetht

https://user-images.githubusercontent.com/54301439/105947471-02f79f00-601e-11eb-9916-ee460fd1d63f.mp4

sk7725:
what why?

MEEPofFaith:
wjoops

Anuken:
Very stale PR; I think the potential performance impact and complexity isn't worth the minor improvements.

PR #4451, opened 2021-01-27
Make revealedBlocks also affect breakability

Ilya246:
Yes, i have actually compiled this small change and made sure it works properly, which it does

### DIFF
--- a/core/src/mindustry/world/Tile.java
+++ b/core/src/mindustry/world/Tile.java
@@ -375,7 +375,7 @@ public class Tile implements Position, QuadTreeObject, Displayable{
     }
 
     public boolean breakable(){
-        return block.destructible || block.breakable || block.update;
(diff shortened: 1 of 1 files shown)


Quezler:
ehmm, for which blocks is this needed?

Ilya246:
if you want to make players able to build pines, you might want to make them breakable

Quezler:
seems like a valid thing to want, but hijacking the `revealedBlocks` rule to make blocks breakable is a bad idea 🤔

Ilya246:
I mean, you cooould add another rule that, instead, prevents blocks from being deconstructed

Quezler (changes requested):
well in its current state i don't think this will/should be merged

Anuken:
Blocks such as pines **should not** be broken. It is unsupported.

PR #4455, opened 2021-01-27
Make the line consistent :^)

buthed010203:
I fixed the line in CustomGameDialog since you won't.

### DIFF
--- a/core/src/mindustry/ui/dialogs/CustomGameDialog.java
+++ b/core/src/mindustry/ui/dialogs/CustomGameDialog.java
@@ -24,8 +24,6 @@ public class CustomGameDialog extends BaseDialog{
     }
 
     void setup(){
-        clearChildren();
-        add(titleTable);
         row();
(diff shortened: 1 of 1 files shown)


core/src/mindustry/ui/dialogs/CustomGameDialog.java
```diff
@@ -24,8 +24,6 @@ public CustomGameDialog(){
     }
 
     void setup(){
-        clearChildren();
-        add(titleTable);
```
Quezler, line 28: doesn't sound perfectly safe, then again i don't know the gui code that well 🤔

Quezler (approved):

core/src/mindustry/ui/dialogs/CustomGameDialog.java
Anuken: Indeed. It completely breaks layout.

Issue #4454, opened 2021-01-27
Inconsistent lines


buthed010203:
**Platform**: Windows

**Build**: 123.1

**Issue**: The custom game dialog has a line that doesn't fill the screen horizontally. This fills the screen in every other dialog and it makes the game absolutely 100% unplayable when it doesn't. Please fix this game breaking bug ASAP.
![](https://yagpdb.is-ne.at/DLjDlQ.png) ![](https://aethex.is-a.fail/DhKgwd.png)

**Steps to reproduce**: Open custom game dialog, notice that the separator doesn't fill the screen.

---

- [X] **I have updated to the latest release (https://github.com/Anuken/Mindustry/releases) to make sure my issue has not been fixed.**
- [X] **I have searched the closed and open issues to make sure that this problem has not already been reported.**

Anuken:
No.

bvcbvc:
lol agreed

PR #4457, opened 2021-01-28
Custom Game Underline 2: Electric Boogaloo

Quezler:
> #4454 -> #4455 -> #4457

Spent some time digging in several dialog classes, compared differences and stumbled upon the default adder:
https://github.com/Anuken/Arc/blob/master/arc-core/src/arc/scene/ui/Dialog.java#L70

Screen Shot 2021-01-28 at 09 13 06

### DIFF
--- a/core/src/mindustry/ui/dialogs/CustomGameDialog.java
+++ b/core/src/mindustry/ui/dialogs/CustomGameDialog.java
@@ -26,7 +26,6 @@ public class CustomGameDialog extends BaseDialog{
     void setup(){
         clearChildren();
         add(titleTable).growX().row();
-        row();
         stack(cont, buttons).grow();
         buttons.bottom();
         cont.clear();
--- a/core/src/mindustry/ui/dialogs/CustomGameDialog.java
+++ b/core/src/mindustry/ui/dialogs/CustomGameDialog.java
@@ -25,7 +25,7 @@ public class CustomGameDialog extends BaseDialog{
 
     void setup(){
         clearChildren();
-        add(titleTable);
+        add(titleTable).growX().row();
         row();
         stack(cont, buttons).grow();
         buttons.bottom();


sk7725:
Can I have a before and after screenshot?

I am also surprised how you had a hard time doing this, it should've been easy.

Quezler:
> Can I have a before and after screenshot?

before in the first linked issue, after is everywhere.

> I am also surprised how you had a hard time doing this, it should've been easy.

keep in mind that those first 2 aren't from me, i just did this one but gui stuff is unfamiliar territory.

core/src/mindustry/ui/dialogs/CustomGameDialog.java
```diff
@@ -25,7 +25,7 @@ public CustomGameDialog(){
 
     void setup(){
         clearChildren();
-        add(titleTable);
+        add(titleTable).growX().row();
```
joshuaptfan, line 28: You add a `.row()`, which is like a newline in tablespeak, yet directly under this is another `row()`. Methinks this `.row()` is not necessary, try without it and see.
```suggestion
        add(titleTable).growX();
```

core/src/mindustry/ui/dialogs/CustomGameDialog.java
```diff
@@ -25,7 +25,7 @@ public CustomGameDialog(){
...
         row();
```
joshuaptfan, line 29: You add a `.row()`, which is like a newline in tablespeak, yet directly under this is another `row()`. Methinks one of these `.row()`s is unnecessary, try without it and see.

`.growX()` is the important part, it makes the element expand to fit the width of its container.
```suggestion
```

Quezler: removing the `.row()` below it for consistency to keep it inline with the class found in arc.

PR #355, opened 2018-11-25
Update bundle_fr.properties [Incomplete]

JustYanns:
Some improvements on the text (capital letters, plurals, ...)
More later !

PR #4469, opened 2021-01-30
Double-tap to mine, tap anywhere to cancel

joshuaptfan:
This PR makes (mobile only) the first tap anywhere while mining stop mining, and adds an option to make mining require double-tap/click. This first tap does not interact with blocks, so that you won't accidentally invoke anything else.

You can still move while mining. Mobile double-tap prioritizes units, so it is easy to control units on top of ore.

Improvements to single-tap mining:

* Mobile: Prevents mining by accident when you attempt to place a block on terrain with an obstruction.
* Mobile: Prevents mining by accident when you tap on ore while breaking blocks.
* Mobile: Tap anywhere to cancel mining. Mining was previously difficult to cancel, as you had to tap precisely on the tile you're mining, or move out of mining range.

Double-tap mining resolves the following issues:

* Prevents mining by accident when you intend to tap/click on something else.
* Prevents mining by accident when you intend to target/shoot an enemy. (Resolves [suggestion #355](https://github.com/Anuken/Mindustry-Suggestions/issues/355).)

Desktop right-click to cancel mining is unchanged.

[Discord discussion](https://discord.com/channels/391020510269669376/663853133587546138/803312974596800574)
@BlueWolf3682 @GreyMario

Additionally, on mobile, the second tap of a double-tap will not configure blocks if a unit was tapped, and double-tap detects units on the first tap. This resolves the following issues:

* Mobile: If a unit is on top of a bridge or power node, double-tapping the unit will not disconnect the bridge/node.
* Mobile: Makes it easier to take control of moving units.

[Discord discussion](https://discord.com/channels/391020510269669376/663853133587546138/806696277256437770)

## Implementation notes
The desktop implementation perhaps could be cleaner and replace `Time.millis()` with `count`, but may require adding method(s) to Arc equivalent to `tap()` used for mobile, with support for keybinds. Introducing easy-to-use double-click methods to desktop may be useful for other things in the future.

## Added
* "Game → Double-Tap to Mine" setting (default unchecked)

## Needs localization
* "Double-Tap to Mine" setting name


sk7725 (changes requested):
>  left-click does the same thing,

no, left click should not cancel the mining, you don't want *cancelling* the mining to be too easy, either.

joshuaptfan:
I could revert the desktop left-click mining cancel, but before I do that, answer me this:

Are you concerned with canceling mining by accident? If so, how often do you left-click by accident?

Or, do you want to do things like configure blocks while mining? You can still build while mining (left-click does not cancel mining if you have a block selected).

IMO, you should test out the PR yourself and see if the issues you're concerned about are really so severe.

sk7725:
> Are you concerned with canceling mining by accident? 

yes

> If so, how often do you left-click by accident?

as often as you start mining by accident

> Or, do you want to configure blocks while mining? You can still build while mining (left-click does not cancel mining if you have a block selected).

of course, you need to do at least *something* while mining for it to be less tedious

joshuaptfan:
Reverted desktop left-click mining cancel. Right-click to cancel mining, or move out of mining range.

unjown:
use foos

joshuaptfan:
[Foo's client](https://github.com/blahblahbloopster/mindustry-client) is for 5.0. Its auto-mine feature is very different from this PR, it seems a bit cheaty with the unlimited deposit range.

And to suggest that we all just stop trying to improve the vanilla game, and instead switch to an alternate client, is not helpful, especially seeing as that (now stale) alternate client was built on top of all the work put in to improve 5.0 by many of the same people still working on improving 6.0.

buthed010203:
If you are adding double click to mine, why not also make sand mineable?

joshuaptfan:
That's not the purview of this PR. Sand's manual mineability is simply a property of sand tiles, and I'm not going to make that change in this PR because it may jeopardize this PR's likelihood of being merged.

buthed010203:
> That's not the purview of this PR. Sand's manual mineability is simply a property of sand tiles, and I'm not going to make that change in this PR because it may jeopardize this PR's likelihood of being merged.

Fair enough, I was just suggesting it as a minor QoL thing as based on the comments, the only reason sand cant be mined is because it is annoying when trying to click off of things and you start mining.

joshuaptfan:
I've made double-tap mining optional, with single-tap mining remaining the default.

I've also made the following improvements to single-tap mining:
* Prevents mining by accident when (mobile only) you attempt to place a block on terrain with an obstruction.
* Prevents mining by accident when (mobile only) you tap on ore while breaking blocks.
* Tap anywhere to cancel mining. Mining was previously difficult to cancel, as you had to tap precisely on the tile you're mining, or move out of mining range.
* Mobile: if a unit is on top of a bridge or power node, double-tapping the unit will not disconnect the bridge/node.
* Mobile: taking control of moving units is easier.

PR #4472, opened 2021-01-30
Custom shadow offset

MEEPofFaith:
Allows for customization of the turret shadow offset.

### DIFF
--- a/core/src/mindustry/world/blocks/defense/turrets/Turret.java
+++ b/core/src/mindustry/world/blocks/defense/turrets/Turret.java
@@ -78,6 +78,7 @@ public class Turret extends ReloadTurret{
 
     public @Load(value = "@-base", fallback = "block-@size") TextureRegion baseRegion;
     public @Load("@-heat") TextureRegion heatRegion;
+    public float elevation = -1f;
 
     public Cons<TurretBuild> drawer = tile -> Draw.rect(region, tile.x + tr2.x, tile.y + tr2.y, tile.rotation - 90);
     public Cons<TurretBuild> heatDrawer = tile -> {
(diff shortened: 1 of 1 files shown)


joshuaptfan:
Wouldn't this lead to graphical inconsistencies if mods start putting weird values in here? With too high of an elevation, shadows become impossible to associate with their sources.

MEEPofFaith:
Well then that's the mod creator being an idiot.

joshuaptfan:
There are certain things that should not be moddable to maintain Mindustry's graphical style.

MEEPofFaith:
But    **c u s t o m i s a t i o n**

sk7725:
> There are certain things that should not be moddable to maintain Mindustry's graphical style.

Mindustry's graphical style was ***thrown out of the window*** the moment we could draw our own sprites for our mods.

MEEPofFaith:
> > There are certain things that should not be moddable to maintain Mindustry's graphical style.
> 
> Mindustry's graphical style was _**thrown out of the window**_ the moment we could draw our own sprites for our mods.

*cough* heresy sprites *cough*

Anuken:
Where would this actually be used?

joshuaptfan:
I think OP intends for this to be used to make blocks look like they are levitating.

MEEPofFaith:
> Where would this actually be used?

Just one less thing that needs scripting to do. It may or may not look bad if a modder uses this, but that's their issue.

Anuken:
> Just one less thing that needs scripting to do. It may or may not look bad if a modder uses this, but that's their fault for putting it up too high.

That doesn't really answer my question.

MEEPofFaith:
If a modder wants to make their turret appear to sit slightly higher they can increase the shadow offset.

Voz-Duh:
I think this not needed
You can make custom shadow in JS script

MEEPofFaith:
What if whoever is doing it doesn't know how to script.

PR #4478, opened 2021-01-31
Stop buttons from wrapping when no players found

buthed010203:
This has bothered me for ages and my solution to the problem is scuffed but at least it doesn't look terrible now I guess?
Before:

After:


Anuken:
This is better fixed by enforcing a minimum width.

PR #4485, opened 2021-01-31
Make drag calculation no longer clamped

MEEPofFaith:
Reason: Some built-in bullets have negative `drag`, but due to the clamp they don't actually accelerate.
Also allows `drag` to be above 1 if you want your object to move backward for some weird reason. Who knows what modders will do with this.


Anuken:
It's clamped for a good reason. Scenario:
1. Drag is set to something high, like 0.5
2. User FPS drops to below 30
3. `1 - drag * delta` becomes *negative*
4. All affected objects start moving backwards

PR #4487, opened 2021-02-01
Allow negative drag to accelerate objects.

MEEPofFaith:
Built-in bullets have negative drag, but that does nothing. This allows drag calculation to go from -1 to 1 instead of 0 to 1.


Anuken:
This doesn't fix anything.

`vel.scl(Mathf.clamp(1f - drag * Time.delta, -1f, 1f));`

What if drag is 0.5 * 10 = 5? *The object is still moving backwards*.

PR #4501, opened 2021-02-02
add rotation to ParticleEffect

QmelZ:
``ParticleEffect``s didn't use ``e.rotation`` when drawing the texture region, this pr makes it use rotation.


Anuken:
If you want to use rotation in your particle effect sprites, you should also add an offset angle - rotation without offset is useless for many sprites.

QmelZ:
making this a draft until i add and do the tests for that

BlueWolf3682:
please merge this anuke, this would fix an issue with json effects that can be described in one image:
Screenshot_20210202-083621

PR #4502, opened 2021-02-02
Show schematic description in schematic info

joshuaptfan:
Schematic descriptions typically contain usage instructions, so it makes sense to display it in the schematic's info dialog. Better than having to open the name edit panel to view the description.

Has the same 500px wrap width as [content descriptions](https://github.com/Anuken/Mindustry/blob/master/core/src/mindustry/ui/dialogs/ContentInfoDialog.java#L46).

How it looks on desktop:
Schematic description desktop

Mobile:
Schematic description mobile

### DIFF
--- a/core/src/mindustry/ui/dialogs/SchematicsDialog.java
+++ b/core/src/mindustry/ui/dialogs/SchematicsDialog.java
@@ -405,6 +405,10 @@ public class SchematicsDialog extends BaseDialog{
                     }
                 });
             }
+            if(!schem.description().isEmpty()){
+                cont.row();
+                cont.add("[lightgray]" + schem.description()).wrap().width(500f).padTop(20);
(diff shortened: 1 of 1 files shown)


Quezler (approved):

Anuken:
How do long descriptions look in landscape on mobile? (first schematic)

joshuaptfan:
The image gets squished to ensure that everything fits on screen. Should the dialog be made to scroll?
Schematic description mobile landscape

Anuken:
Yeah, that's exactly what I was worried about. I'm not sure how you would make it scroll while enforcing a reasonable size for the schematic image.

joshuaptfan:
Perhaps the description should have a capped height, and scroll if it goes over that height.

PR #4525, opened 2021-02-03
Hi

mindustrykilla:


mindustrykilla:
mindustry is my favourite game because I just love to create fun things, its super inspiring for me and I love the idea of making something out of nothing.

PR #4532, opened 2021-02-03
Hello

mindustrykilla:


mindustrykilla:
Do you remember when you were in fourth or fifth grade and every week you went to "Manila Pizza" with your family to watch American movies?

DeltaNedas:
yes now you have convinced anuke to add .ru to the server list

PR #4586, opened 2021-02-04
Saves rotation of repair points

Quezler:
> resolves https://github.com/Anuken/Mindustry-Suggestions/issues/1920

storing the rotation of repair points seems to be welcomed, this pr hopefully does that correctly.

in terms of backwards compatibility it should probably be merged before 124 since it would break 123.2 (right?)
(but then again, the block sync thing has already broken compatibility, so it doesn't really make much difference)

Screen Shot 2021-02-04 at 16 06 18

> originally made this a suggestion instead of a pull since i had some vague memories about the version/revision stuff being changed/removed like long ago or something, wasn't sure if my own attempt would/will break it or not*

### DIFF
--- a/core/src/mindustry/world/blocks/units/RepairPoint.java
+++ b/core/src/mindustry/world/blocks/units/RepairPoint.java
@@ -6,6 +6,7 @@ import arc.math.*;
 import arc.math.geom.*;
 import arc.struct.*;
 import arc.util.*;
+import arc.util.io.*;
 import mindustry.annotations.Annotations.*;
 import mindustry.entities.*;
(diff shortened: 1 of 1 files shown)


joshuaptfan:
You can word a mention of a suggestion so that [merging the PR will auto-close the suggestion](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

Quezler:
hmm, resolves sounds less snappy as implements, but very well 🍿

PR #4588, opened 2021-02-04
Draws team regions during schematic placement

Quezler:
found it a bit strange to see sharded previews while playing as another team, this pr tries to remedy that:
(looks like no team != null check is needed since the player's team defaults to sharded)

Screen Shot 2021-02-04 at 21 17 40
Screen Shot 2021-02-04 at 21 17 30

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -428,7 +428,7 @@ public class Block extends UnlockableContent{
         TextureRegion reg = getRequestRegion(req, list);
         Draw.rect(reg, req.drawx(), req.drawy(), !rotate ? 0 : req.rotation * 90);
 
-        if(teamRegion.found()){
+        if(req.worldContext && teamRegion.found()){
             Team team = player != null ? player.team() : Team.sharded;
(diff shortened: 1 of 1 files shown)


Quezler:
sidenote: the schematic dialog images get rendered once you first open them (and thus the previews in there lock to that team), might require some thought on whether to refresh them if your team changes or just lock them to sharded 24/7 🤔

Anuken:
This should check for `player != null` for headless, and also make sure `req.worldContext` so schematics don't have team regions.

Quezler:
> This should check for player `!= null` for headless

since when does headless draw schematic previews?

Anuken:
> since when does headless draw schematic previews?

CoreBot draws them, specifically.

Quezler:
allright, i'll make em fall back to sharded :)

but i don't understand what you mean with this exactly:
> and also make sure req.worldContext so schematics don't have team regions.

Anuken:
Schematics shouldn't display team regions, because there's no dynamic way to change them based on player team.

If `req.worldContext` is false, that means it's in a schematic, and the region should not be drawn.

PR #4607, opened 2021-02-06
Removed same if statement.

Xasmedy:
Removed an if statement that was the same of the one up.


Kieaer:
It's great to find something like this!

Xasmedy:
> It's great to find something like this!

Honesty I'm not the one who found it, I just had the idea to make the pull request.

PR #4612, opened 2021-02-07
Add sources jar

Kieaer:
Allow source download from Intellij IDEA
It will be of great help to plugin/mod developers.

before:
image

after:
image


Anuken:
Shouldn't this be in the core project?

Kieaer:
@Anuken

PR #4613, opened 2021-02-07
plastanium-wall-large schematicPriority should also be 10

LeoSko:
plastanium-wall has it, makes sense to put for large one as well

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -797,6 +797,7 @@ public class Blocks implements ContentList{
             size = 2;
             insulated = true;
             absorbLasers = true;
+            schematicPriority = 10;
         }};
 
         thoriumWall = new Wall("thorium-wall"){{


Quezler (approved):

joshuaptfan:
Why?

Large plastanium walls are mostly used for defense against enemies with lasers/electric arcs. Small plastanium walls are mostly used for blocking automatic power connections. There is some overlap between their roles, but each is more ideal for their own role.

Leaving `schematicPriority` off large plastanium walls was not an oversight, it was deliberate. Large walls take significantly longer to build than small walls, so forcing them to have priority can be a significant time waster for players if that is not what players wanted. Large walls used for defense should not have priority the same way no other defensive walls have priority. When attack is imminent, you may build turrets before walls or walls before turrets, depending on your own priorities, the game should not override that for you.

LeoSko:
@joshuaptfan, that simply doesn't make much sense since the function itself - blocking power connections - is still present into large walls. Those **also can be used to block certain power connections**.

When attack is imminent, I doubt you will be building ***any*** large walls. If you care about "turrets before walls" - then you simply place turrets before walls. If you place schematic, game should be consitent on desicions via features of items, but definitely not on their size.

PR #4622, opened 2021-02-08
Use Jsdelivr for serverJson

shugen002:
raw.githubusercontent.com sometimes does not work in China in some area.
use jsDelivr instead and it should be work well in global include China.


shugen002:
Wait for another commit

PR #4624, opened 2021-02-08
Check if `teamRegion` is null

MEEPofFaith:
Check if `teamRegion` actually exists before attempting to `teamRegion.found()`.
Mods with scripting may override load and never load `teamRegion`, making it null and therefore giving a NullPointerException error when attempting to perform `teamRegion.found()`.


Quezler:
while its nice to prevent a crash, mods should just learn to call the parents load function instead, it exists for a reason 🤔

Anuken:
I agree with Quezler - mods should always call `super.load()` - , but if this is a *new* crash, it should be avoided if possible.

PR #4626, opened 2021-02-08
Update servers_v6.json

CxZxRainzz:


CxZxRainzz:
sos i done it in the middle it was to hard to to it at the end

Anuken:
image

1. Please use "our", not "are"
2. Don't use bright primary colors for the title/description. Even no colors at all would be preferable.

CxZxRainzz:
Is that all

PR #4629, opened 2021-02-08
Fix null == 0

BasedUser:
`//use object function if provided, and`*`one of the variables is an object`*

Surely not both?


Anuken:
Null **should be** == 0. This is a direct revert of my previous changes.

PR #4634, opened 2021-02-08
60 * 60 > Time.toMinutes

buthed010203:
It already exists in other places in the code, doesn't make sense to not use it.


Quezler:
this can't possibly be the only place where 60 gets multiplied with something, can it?

buthed010203:
> this can't possibly be the only place where 60 gets multiplied with something, can it?
@Quezler I'm sure you are correct but I am too lazy to find every occurrence of this, I just happened to see this and it bothered me so I fixed it. On another note, why does Time.toSeconds even exist if its never even used? There are tons of spots where it is just using 60 rather than Time.toSeconds

PR #4048, opened 2020-12-19
fix pt br translation

DeltaNedas:
pls give me translation role on discord thanks

PR #4636, opened 2021-02-08
Delete bundle_pt_BR.properties

Lord2treva:
- There are no actual translators, some of them get the role just to flex
- Bad translation, full of google translate and typos 
- Everyone can change it, that allows 8y old kids to screw it up again
- There are no motivation (imo)
- Some "translators'" are just 13 years old mentally broken 
- Too messy
- Brazil


DeltaNedas:
point 3 is terrible:
1. that goes for every bundle silly
2. wikis exist and thrive

GreyMario:
Duplicate of #4048

YellOw139:
Ok. It was fun while it lasted, but now all I see is picking on people.

How the BR translators manage their translation is their problem IMO. If that's the BR playerbase, let it be. If you are speaking the language and it bothers you, go fix it instead of complaining. That way you fix both the translation and how the game is perceived by Brazilian audiences.
If you are not even speaking the language, just stop.

BTW if you care so much about proper localization, go check other bundles. There's half English ones out there. If deleting bad translations is what you do, then at least finish your job.

PR #4637, opened 2021-02-08
60 * 60 > Time.toMinutes 2: Electric Boogaloo

buthed010203:
They are bothering me ok


genNAowl:
Well... I will say that the length of `Time.toSeconds` is substantially longer than `60f`

Anuken:
Implemented a few of these, but some are better left without changes.

PR #4647, opened 2021-02-10
friendly attacker range

liuxilu:


Anuken:
?

liuxilu:
Unit won't shot on current max range by logic, this returns a shootable range.

PR #4648, opened 2021-02-10
add heat&broken sensor for force projector

liuxilu:
Because of 
image

Can't copy the link for unknown reasons.


Anuken:
`broken` seems redundant - you could check for heat=1 instead.

Anuken:
*implemented myself*

liuxilu:
I was sleeping💤. But `buildup=1` doesn't mean broken or not? Maybe we could add another sensor for current shield health.

PR #4649, opened 2021-02-10
Add missing properties

Kieaer:
image
oh no


build.gradle
```diff
@@ -322,6 +322,10 @@ project(":core"){
         classifier = 'sources'
         from sourceSets.main.allSource
     }
+    
+    artifacts {
```
Anuken, line 326: Incorrect formatting, should not have a space before {.

PR #4661, opened 2021-02-12
Update bundle_vi.properties

caiban123bo:
Miscellaneous translation updates ,changing some vietnamese words so it meaning can be more clearly


Anuken:
Outdated PR.

PR #4663, opened 2021-02-12
change light color config to 0-255

liuxilu:


sk7725:
good, should be consistent with draw color r g b a

Anuken:
Honestly, I'm not sure if making `draw color` 0-255 was a good idea. 0-1 is easier to work with.

liuxilu:
We may add float literal.

liuxilu:
No, it already exist, just `.0`.

PR #4674, opened 2021-02-14
ClassEntry[] -> Seq<ClassEntry>

GlennFolker:
This way mods can register their own way of synchronization


Anuken:
Use custom InvokePacket IDs and/or `netServer.addPacketHandler(...)` instead.

PR #4689, opened 2021-02-15
Unit cap

TranquillyUnpleasant:
Allows custom maps to be able to set unit caps, as well as enable or disable dynamic caps based on cores.

Mainly allows bigger maps and more cores without making the maps a lag-fest, and also allows for better difficulty control.

Negative values are only there to be able to offset core contributions in case someone wants dynamic caps but doesnt want to have them too high. Cap value resets to 0 if its negative.

Settings Dialog:
image

Cap in game:
image

Cap cancelling out core effect:
image


AnOreo1:
this would be amazing for maps, I like this

PR #4717, opened 2021-02-17
Add message for mods that have hidden = true

MEEPofFaith:
Multiplayer Compatable
(Message could probably be different but you get the point.)


Quezler:
mod.multiplayer is probably not the right bundle key to use, could be interpreted out of context that the mod is only for multiplayer, maybe mod.multiplayer.compatible or something that is a little more descriptive?

PR #4718, opened 2021-02-18
Use github build status

buthed010203:
Why does this still use travis? It was ditched months ago


Anuken:
> Why does this still use travis?

Mindustry's build is usually passing, so nobody noticed.

PR #4723, opened 2021-02-18
Make content parser's parsers public

sk7725:
- This allows Java mods (or JS mods...somehow) to provide custom parsers, so other mods with content dependency to a Java mods will be parsed with the additional java classes.

- ...Wording is horrible, here is an example:

Java mod sk7725/BetaMindy has `PayloadTurret.java`
Json mod bluewolf/BetaBeta has
`routerpayturret.hjson`, with the following content:
```
type: PayloadTurret
size: 1
(and so on)
```
(BetaBeta has dependency set to BetaMindy of course)

### DIFF
--- a/core/src/mindustry/mod/ContentParser.java
+++ b/core/src/mindustry/mod/ContentParser.java
@@ -125,7 +125,7 @@ public class ContentParser{
     LoadedMod currentMod;
     private Content currentContent;
 
-    private Json parser = new Json(){
+    public Json parser = new Json(){
         @Override
         public <T> T readValue(Class<T> type, Class elementType, JsonValue jsonData, Class keyType){
             T t = internalRead(type, elementType, jsonData, keyType);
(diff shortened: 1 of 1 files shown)


Anuken:
- If you want to parse non-Mindustry content, use the fully qualified class name as the type.
- There's no particularly good reason to change the parser internals, as that requires copy-pasting. If you must do this, use reflection and try-catch it to prevent failures.
- Opening up implementation details like this is a bad idea.

sk7725:
> How do you envision this example actually working? You don't just need a custom parser, you need an override of the old block parser that somehow loads your own block files _and_ falls back on the old logic. I can't see that working without a lot of copy-pasting and flaky code.
> 
> I'm not even sure if mods can load classes from other mods' class loaders.

Add the java mod's packages (betamindy.world.xxxx) to the Block.class key of the object mapper, although it might not have worcked, and you already fixed it so it would work (thank you).

> use the fully qualified class name as the type.

*what?*

Anuken:
> Add the java mod's packages (betamindy.world.xxxx) to the Block.class key of the object mapper

You can't just add it as a key - the block types were resolved using code *inside the parser lambda*, which means adding new types would imply copy-pasting the entire lambda.

> what?

`type: "betamindy.world.PayloadTurret"`

sk7725:
`type: "betamindy.world.PayloadTurret"` and `type: "betamindy.world.blocks.defense.turrets.PayloadTurret"` both do not work, according to @BlueWolf3682

PR #4728, opened 2021-02-18
Added servers to list

thedimas3007:


CxZxRainzz:
The frfr1.quackhost.uk:20752 andfrfr1.quackhost.uk:20745 has no name or desc but other than that they are good, they have a discord it shows up when you join the game.

RebornTrack970:
Hi, RTOmega here,
I have some opinions about the server(s).

- First of all, QuackHost isnt a very, lets say, useable host system... They dont offer enough ram nor cpu power to host a community server and i highly recommand using another service if you dont get more than 2gb ram per server.

- Second of all, other than the Hub, all 4 servers are basicaly Vanilla, they dont have any uniqueness from servers of pl, io and others. Puting 4 servers in the list will just flood the list even more and devide the already limited playerbase. I recomand just adding the Hub as it would be enough to join the others.

- Thirdly, get some good maps for survival, its default map; get some moderators and have atleast the jseval plugin so they can do stuff if the owner isnt on; name the servers nicely and have desc and Motd.


Also, i dont know about you but i have never even heard the name of the owner + he didnt provide any information in this pull.

Other than these, servers apear to be working and they seem 24/7 .

**I Do NOT approve this pull for now as the list already has many vanilla servers.**

However, this is 100% my opinion and the server is pretty new so we can give it a chance to do smth that will make everyone say: *”Woah, this server is unqiue and it deserves to be on the list!”*

thedimas3007:
@RebornTrack970 Okay, i will search some good maps and plugins

thedimas3007:
It has hex PvP now. Tomorrow i will add something new also

PR #4738, opened 2021-02-20
unit boosting sensor

liuxilu:
Anuken/Mindustry-Suggestions#1802


Anuken:
What's the use case for this?

liuxilu:
Neither do I

PR #4741, opened 2021-02-20
Units should not be able to shoot when reloadMultiplier = 0

MEEPofFaith:
If a unit is fully reloaded, then it hit with a reloadMultiplier = 0 status effect, you can still shoot once, which doesn't really make much sense.

### DIFF
--- a/core/src/mindustry/entities/comp/WeaponsComp.java
+++ b/core/src/mindustry/entities/comp/WeaponsComp.java
@@ -81,7 +81,7 @@ abstract class WeaponsComp implements Teamc, Posc, Rotc, Velc, Statusc{
     }
 
     boolean canShoot(){
-        return true;
+        return reloadMultiplier > 0f;
     }
 
     @Override


Quezler:
it makes perfect sense, if it has already reloaded it has a bullet to shoot, reloading happens after that 👀

MEEPofFaith:
Well my logic is reloadMultiplier 0 is basically just dont shoot

Quezler:
units should probably just spawn without a bullet ready, that way it always needs to reload first 🤔

core/src/mindustry/entities/comp/WeaponsComp.java
```diff
@@ -81,7 +81,7 @@ void aim(float x, float y){
     }
 
     boolean canShoot(){
-        return true;
+        return reloadMultiplier > 0f;
```
Quezler, line 84: ps, this doesn't even seem to trigger at all, the method gets replaced with the `canShoot` found in `UnitComp` 😗

PR #4754, opened 2021-02-22
Prevent tick-perfect factory chains from halting impact reactors

sk7725:
Cultivators and impact reactors produce / consume items once every 140 ticks. 
While this suggests a perpetual, perfect flow - deltatime comes in and ruins it all, grinding the impact to a halt. 
This PR buffs cultivators ever so slightly so the 140 - 140 chain is no longer tick-perfect, allowing small interferences to not halt the chain.


sk7725:
* Edit: blast Mixer is actually 80, but this just proves that even with 2 blocks deltatime can come and ruin everything.

GreyMario:
You... "buffed" cultivators... by making them take longer to produce items?

sk7725:
> typo

Anuken:
Possible alternative: Mod progress instead of setting it to `0` on completion, so delta overflow doesn't result in lost progress. Test this setup with the latest BE.

PR #4757, opened 2021-02-22
DefenderAI that makes octs follow units.

MEEPofFaith:
https://user-images.githubusercontent.com/54301439/108773617-7bfdef80-7513-11eb-810a-9218a85aefcc.mp4

Foreshadow-like targetting
image

### DIFF
--- a/core/src/mindustry/ai/types/DefenderAI.java
+++ b/core/src/mindustry/ai/types/DefenderAI.java
@@ -1,0 +1,34 @@
+package mindustry.ai.types;
+
+import arc.math.*;
+import mindustry.entities.*;
+import mindustry.entities.Units.*;
+import mindustry.entities.units.*;
+import mindustry.gen.*;
+import mindustry.world.meta.*;
+
+public class DefenderAI extends AIController{
+
+    @Override
+    public void updateMovement(){
+        if(target != null){
+            moveTo(target, unit.range(), 5f);
+            unit.lookAt(target);
+        }else{
+            Teamc block = targetFlag(unit.x, unit.y, BlockFlag.rally, false);
+            if(block == null) block = unit.closestCore();
+            moveTo(block, 60f);
+        }
+    }
+
+    @Override
+    protected void updateTargeting(){
+        if(retarget()) target = findTarget(unit.x, unit.y, 0f, true, true);
+    }
+    
+    @Override
+    protected Teamc findTarget(float x, float y, float range, boolean air, boolean ground){
+        //Sort by max health and closer target.
+        return Units.closest(unit.team, x, y, u -> !u.dead() && u.type != unit.type, (u, tx, ty) -> -u.maxHealth + Mathf.dst2(u.x, u.y, tx, ty) / 800f);
+    }
+}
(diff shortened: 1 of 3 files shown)


core/src/mindustry/ai/types/DefenderAI.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Override
+    protected void updateTargeting(){
+        if(retarget()) target = findTarget(unit.x, unit.y, 0f, true, true);
+    }
+    
+    @Override
+    protected Teamc findTarget(float x, float y, float range, boolean air, boolean ground){
+        Sortf sort = (u, tx, ty) -> -u.maxHealth + Mathf.dst2(u.x, u.y, tx, ty) / 800f; //Sort by max health and closer target.
```
Anuken, line 31: This should not be defined as a variable - inline it.

core/src/mindustry/ai/types/DefenderAI.java
```diff
@@ -0,0 +1,34 @@
...
+import mindustry.entities.units.*;
+import mindustry.gen.*;
+import mindustry.world.meta.*;
+
+public class DefenderAI extends AIController{
+
+    @Override
+    public void updateMovement(){
+        if(target != null){
+            moveTo(target, -1f);
```
Anuken, line 15: It should not hug targets. Circle length should be greater than 0 and depend on target hitbox size - octs crashing into eclipses is a bad idea.

core/src/mindustry/ai/types/DefenderAI.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Override
+    protected void updateTargeting(){
+        if(retarget()) target = findTarget(unit.x, unit.y, 0f, true, true);
+    }
+    
+    @Override
+    protected Teamc findTarget(float x, float y, float range, boolean air, boolean ground){
+        Sortf sort = (u, tx, ty) -> -u.maxHealth + Mathf.dst2(u.x, u.y, tx, ty) / 800f; //Sort by max health and closer target.
+        return Units.closest(unit.team, x, y, u -> !u.dead() && u.type != unit.type && !u.isPlayer(), sort);
```
Anuken, line 32: Why not follow players?

Anuken (changes requested):

core/src/mindustry/ai/types/DefenderAI.java
MEEPofFaith: I already changed this back before you sent this message.

guncatt:
I would rather have oct target places where it can cover the most units but this is good

core/src/mindustry/ai/types/DefenderAI.java
Anuken: This still hugs targets, and will crash into Eclipse units.

MEEPofFaith:
> I would rather have oct target places where it can cover the most units but this is good

I'm not gonna bother trying to figure out how to math out how to target the "densest cluster"

MEEPofFaith:
I've resolved the changes why does it still say changes are requested.

Anuken:
I'm still not entirely happy with this implementation, will refine it after merge.

PR #4762, opened 2021-02-23
Add disarmed status effect that disarms

Quezler:
> resolves #4741 by adding a new status effect property (+ a status effect that uses it, though currently no particle effects)

ingame showcase with the parallax: (fire held down all the time)

Screen Shot 2021-02-23 at 10 05 54

https://user-images.githubusercontent.com/3179271/108822158-fafc3200-75be-11eb-84bf-f4e1584c8fce.mp4

> weapons stop rotating if you can't shoot, leaving that as-is for now, oh and the status color is Fx.plasticburn :)

### DIFF
--- a/core/src/mindustry/entities/comp/StatusComp.java
+++ b/core/src/mindustry/entities/comp/StatusComp.java
@@ -20,6 +20,7 @@ abstract class StatusComp implements Posc, Flyingc{
     private transient Bits applied = new Bits(content.getBy(ContentType.status).size);
 
     @ReadOnly transient float speedMultiplier = 1, damageMultiplier = 1, healthMultiplier = 1, reloadMultiplier = 1;
+    @ReadOnly transient boolean disarmed = false;
 
     @Import UnitType type;
 
@@ -111,6 +112,7 @@ abstract class StatusComp implements Posc, Flyingc{
 
         applied.clear();
         speedMultiplier = damageMultiplier = healthMultiplier = reloadMultiplier = 1f;
+        disarmed = false;
 
         if(statuses.isEmpty()) return;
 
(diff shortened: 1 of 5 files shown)


XEN0PHIL:
Will multiple parallaxes re-target to different unit if it comes in their range?
To stop focusing a single unit and spread the disarming effect.

Quezler:
this pull doesn't change parallaxes at all, i just used it as a simple way to visualize & test how it works.

anyways, their normal behavior is to target the closest unit to itself, not to divide and conquer.

(which should be a suggestion/pull of its own, which i would probably support.)

Quezler:
some extra things to consider for @Anuken:

yes this effect & boolean aren't used in the base game (currently), but that also seems the case with 1 or more other status effects, but having this would also open up new possibilities for server plugins that need to prevent units from shooting for whatever reason, as well as opening it up to modders like @MEEPofFaith with minimal overhead on their side, seems nice 🙂

core/src/mindustry/entities/comp/StatusComp.java
```diff
@@ -132,6 +134,9 @@ public void update(){
                 healthMultiplier *= entry.effect.healthMultiplier;
                 damageMultiplier *= entry.effect.damageMultiplier;
                 reloadMultiplier *= entry.effect.reloadMultiplier;
+
+                if(entry.effect.disarms) disarmed = true;
```
Anuken, line 138:
```suggestion
                disarmed |= entry.effect.disarms;
```

core/src/mindustry/type/StatusEffect.java
```diff
@@ -17,8 +17,10 @@
     public float healthMultiplier = 1f;
     /** Unit speed multiplier */
     public float speedMultiplier = 1f;
-    /** Unit speed multiplier */
+    /** Unit reload multiplier. */
     public float reloadMultiplier = 1f;
+    /** Unit weapon(s) disabled. */
+    public boolean disarms = false;
```
Anuken, line 23: `disarm` could be a better name for this.

joshuaptfan:
What if `disarm` disabled building as well? Could be useful, and they're similar enough that a separate status isn't necessary.

PR #4763, opened 2021-02-23
Add IPv6 of routerchain

BasedUser:
transition™


Quezler:
why not an AAAA dns record?

BasedUser:
> why not an AAAA dns record?

i'm planning to move away from ddns.net

XEN0PHIL:
Happy transition

PR #4767, opened 2021-02-23
Add Power Laser Build

Voz-Duh:
I wondered why the power node is not showing the lines of lasers that will connect to it?

image
image
image
image


joshuaptfan:
Might be too visually distracting when you're building manually?

Voz-Duh:
> Might obstruct your view when you're building manually?

Maybe this can clutter your vision

But on this screenshot this help me
image

Anuken:
Implemented manually with some differences:
- No setting
- Opacity set to 50% of node opacity
- White instead of yellow nodes
- Don't display nodes that involve going over the link limit

PR #4769, opened 2021-02-23
Don't heal target's that don't need healing.

MEEPofFaith:
Why is the heal outside of the check for if the unit is damaged?


Anuken:
It doesn't make a difference. Health is clamped.

Quezler:
close?

MEEPofFaith:
Time to make my invincibility button also change max health YEET

PR #4771, opened 2021-02-23
fix up some pump things

BlueWolf3682:
i decided to change these because at least to me, they were too basic and missed some points. Accept if you wish.


Anuken:
This doesn't seem necessary. The user can already see the rate at which liquids are produced by glancing at the stats.

BlueWolf3682:
sounds good, though i just found the copypaste feeling of the descriptions annoying.

PR #4775, opened 2021-02-24
schematic uses camera position

Minxyzgo:
This is useful on mobile phones

### DIFF
--- a/core/src/mindustry/input/InputHandler.java
+++ b/core/src/mindustry/input/InputHandler.java
@@ -544,7 +544,7 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
     }
 
     public void useSchematic(Schematic schem){
-        selectRequests.addAll(schematics.toRequests(schem, Core.camera.position.x, Core.camera.position.y));
+        selectRequests.addAll(schematics.toRequests(schem, (int)Core.camera.position.x, (int)Core.camera.position.y));
     }
 
     protected void showSchematicSave(){
--- a/core/src/mindustry/input/InputHandler.java
+++ b/core/src/mindustry/input/InputHandler.java
@@ -544,7 +544,7 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
     }
 
     public void useSchematic(Schematic schem){
-        selectRequests.addAll(schematics.toRequests(schem, player.tileX(), player.tileY()));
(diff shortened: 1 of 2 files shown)


Quezler:
have you tested this ingame? it doesn't look like camera x & y would be a drop in replacement for tile coordinates 🤔

Minxyzgo:
> 你测试过这个游戏了吗？它看起来像照相机x和y不会替代瓷砖

> have you tested this ingame? it doesn't look like camera x & y would be a drop in replacement for tile coordinates

usable

buthed010203:
The thing you edited is in InputHandler but is overridden in MobileInput, and you changed this from being a tile coordinate to a world coordinate (world coordinates being 8x larger than tile coordinates).

core/src/mindustry/input/InputHandler.java
```diff
@@ -544,7 +544,7 @@ public void drawBreaking(int x, int y){
     }
 
     public void useSchematic(Schematic schem){
-        selectRequests.addAll(schematics.toRequests(schem, player.tileX(), player.tileY()));
+        selectRequests.addAll(schematics.toRequests(schem, (int)Core.camera.position.x, (int)Core.camera.position.y));
```
buthed010203, line 547: This should be in MobileInput, not input handler. Instead of casting to an int, use `World.toTile` to properly convert from world to tile coordinates

buthed010203 (changes requested):

Minxyzgo:
It's ok now

core/src/mindustry/input/MobileInput.java
```diff
@@ -424,7 +424,7 @@ public boolean isBreaking(){
     @Override
     public void useSchematic(Schematic schem){
         selectRequests.clear();
-        selectRequests.addAll(schematics.toRequests(schem, player.tileX(), player.tileY()));
+        selectRequests.addAll(schematics.toRequests(schem, (int)(Core.camera.position.x / tilesize), (int)(Core.camera.position.y / tilesize)));
```
buthed010203, line 427: This still isnt using the appropriate method to convert world coords to tile coords, use `World.toTile` instead of dividing and then casting to an int

buthed010203 (changes requested):

liuxilu:
Tip: convert to darft at the right-top

liuxilu:
I just wonder why a machine translation there.

buthed010203 (approved):

Issue #4778, opened 2021-02-24
Loadout not always working


catsz:
**Platform**: *Android/iOS/Mac/Windows/Linux*
windows

**Build**: *The build number under the title in the main menu. Required. "LATEST" IS NOT A VERSION, I NEED THE EXACT BUILD NUMBER OF YOUR GAME.*
125.1

**Issue**: *Explain your issue in detail.*
I haven't personally experienced this but it has affected some plugins I have made. On PlayEvent if you loop through the active teams cores it won't get every single one but right below PlayEvent's fire is: 

`if(!state.isCampaign()){
            for(TeamData team : state.teams.getActive()){
                if(team.hasCore()){
                    Building entity = team.core();
                    entity.items.clear();
                    for(ItemStack stack : state.rules.loadout){
                        entity.items.add(stack.item, stack.amount);
                    }
                }
            }
        }`

this is normally fine for loadouts because the team data is almost guaranteed to have at least one core. The problem is the TeamData isn't working as intended.

this can be fixed a couple ways.

1. loop through the cores using Groups.build
2. fix Teams wherever your assigning it on world load

---

*Place an X (no spaces) between the brackets to confirm that you have read the line below.*  
- [x] **I have updated to the latest release (https://github.com/Anuken/Mindustry/releases) to make sure my issue has not been fixed.**
- [x] **I have searched the closed and open issues to make sure that this problem has not already been reported.**

Anuken:
No save, no exact code, no description of when exactly this fails or which teams don't have cores. Submit a pull request.

PR #4782, opened 2021-02-24
Fix Loadout Issue

catsz:
The loadout could sometimes not get applied this is unfortunately less efficient but more consistent.


Catchears:
(more info in #4778)

Anuken:
I still don't understand what issue this fixes, and the code proposed here is unusable.

PR #4787, opened 2021-02-25
Make recoil work similarly to heat

MEEPofFaith:


MEEPofFaith:
image

GlennFolker:
`restitutionTime` -> `restitution`, in my opinion. Not sure if Anuke agrees.

MEEPofFaith:
> `restitutionTime` -> `restitution`, in my opinion. Not sure if Anuke agrees.

It uses `cooldownTime` instead of `cooldown` so I did the same for restitution.

PR #4788, opened 2021-02-25
Enable mining of sand when double click to mine is enabled

buthed010203:
Since the only reason sand is un-mineable in the first place is because accidentally mining it is annoying, it should be enabled for double click to mine.


MEEPofFaith:
> Since the only reason sand is un-mineable in the first place is because accidentally mining it is annoying, it should be enabled for double click to mine.

Have you actually confirmed if that's the reason?

buthed010203:
> Have you actually confirmed if that's the reason?

I havent asked anuke but based on the javadoc it would seem so yes, I have had double click mineable sand in my client for quite a while and it works with no issues.

Thankyou69:
yay its getting up there

catatagun:
Yes

PR #4798, opened 2021-02-26
Fix a few issues w/ double tap mining

buthed010203:
Currently, double tap mining consumes clicks so it prevents the player from configuring any block if they click twice within 500ms. The double tap mining was also using the `selected` var which could have been null causing a crash. This is a very messy solution but it gets the job done and achieves the intended functionality,


Anuken:
@joshuaptfan Thoughts?

joshuaptfan:
I'll check out the fix once I get home.

joshuaptfan:
Finally had time to look at the fix; this breaks single-click on mined tile to stop mining. I will PR a fix.

PR #4801, opened 2021-02-26
Fixed #4791 since it is 100% a bug

### LINKED ISSUE #4791: Auto bridge doesn't respect conveyor rotation
**Platform**: Windows

**Build**: BE build 20820

**Issue**: The newly added auto bridge doesn't rotate the bridge links when the conveyor is rotated with the scroll wheel.

**Steps to reproduce**: Just try placing a conveyor and rotating it, the bridges will link backwards: https://aethex.is-a.fail/Dw8vEZ.webm

**Link(s) to mod(s) used**: NA

**Save file**: [338.zip](https://github.com/Anuken/Mindustry/files/6046280/338.zip)

**(Crash) logs**: NA

---

- [X] **I have updated to the latest release (https://github.com/Anuken/Mindustry/releases) to make sure my issue has not been fixed.**
- [X] **I have searched the closed and open issues to make sure that this problem has not already been reported.**

buthed010203:
This fix is somewhat janky as it checks if the rotation matches the exact original rotation but the likelihood of someone needing bridges along a belt rotated 1 or -1 times is super low. https://aethex.is-a.fail/t13ETd.webm

### DIFF
--- a/core/src/mindustry/input/Placement.java
+++ b/core/src/mindustry/input/Placement.java
@@ -119,6 +119,7 @@ public class Placement{
 
         var result = plans1.clear();
         var team = player.team();
+        var rotated = plans.first().tile().absoluteRelativeTo(plans.peek().x, plans.peek().y) != plans.first().rotation;
 
         outer:
         for(int i = 0; i < plans.size;){
@@ -144,7 +145,11 @@ public class Placement{
                         //found a link, assign bridges
                         cur.block = bridge;
                         other.block = bridge;
-                        cur.config = new Point2(other.x - cur.x, other.y - cur.y);
+                        if(rotated){
+                            other.config = new Point2(cur.x - other.x,  cur.y - other.y);
+                        } else {
+                            cur.config = new Point2(other.x - cur.x, other.y - cur.y);
+                        }
 
                         i = j;
                         continue outer;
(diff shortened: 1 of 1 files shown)


buthed010203:
Ok I changed it so it now only rotates the bridges when the belt is rotated twice

Anuken:
1. This doesn't look like it actually works. Rotation happens after placement lines are calculated.
2. Doesn't match style guidelines.
3. Doesn't check for nulls (use an IDE, it warns about `@Nullable`) or empty lists. May crash.

> it is 100% a bug

It is not. Bridges **do not rotate.**

buthed010203:
> It is not. Bridges do not rotate.

@Anuken I understand that but it just makes sense for the links themselves to be in reverse order to match the direction in which items are flowing.

Always forget about style guidelines, intellij says absolutely nothing about any of this being nullable, I don't see how my code could cause any crashes that the existing code couldn't.

Anuken:
> intellij says absolutely nothing about any of this being nullable

You likely need to configure Mindustry's `Nullable` annotation to show up in the settings.

> I don't see how my code could cause any crashes that the existing code couldn't.

`tile()` can be null.

buthed010203:
> `tile()` can be null.

Ah yeah that was totally an oversight on my part.

@Anuken The rotation happens after the line is calculated but that shouldn't really be a problem considering that the line is recalculated every time the line is rotated anyways

buthed010203:
Ok that should be everything

PR #4804, opened 2021-02-27
Change to overdrive sprite to help color blind people

bloool:
This includes changes both to the overdrive projector and overdrive dome, overdrive main sprite change is due to it being just a recolor of mender, and many people have red to green colorblindness which can make differentiate from both very hard. Overdrive dome also gets a sprite change to match the new small one.

Even if these don't make it into the game, getting anything similar would still be nice, even if OD has a little animation can still be hard to distinguish them quickly.

There is also another alternative made by Newguy#7763 on discord, that covers more of the middle part.
image


liuxilu:
Better a layer on the final image for everything, or just OS build-in color-blind filter.

sk7725:
I would do a different approach, mainly tweak the square effect drawn on top. There are striking UI differences when hovered over(the overdrive has an extra bar), and both projectors are normally almost always in a powered state constantly.
Also the 3x3 looks bad, and you don't need to change the 3x3 anyways

bloool:
> I would do a different approach, mainly tweak the square effect drawn on top. There are striking UI differences when hovered over(the overdrive has an extra bar), and both projectors are normally almost always in a powered state constantly.
> Also the 3x3 looks bad, and you don't need to change the 3x3 anyways



> I would do a different approach, mainly tweak the square effect drawn on top. There are striking UI differences when hovered over(the overdrive has an extra bar), and both projectors are normally almost always in a powered state constantly.
> Also the 3x3 looks bad, and you don't need to change the 3x3 anyways

Yeah but think about it in a image or scheme, you wouldn't be able to see a difference if you are colorblind, and I just changed the 3x3 just so it doesn't look way too different from the 2x2, but anyway, even if those exact sprites doesn't get in the game, anything would be better than a recolor

sk7725:
> image or schem

Why didn't I even think of them, you're right.
But still I don't think sprite change is the best approach here...hmm...

bloool:
The only thing I can think about atm is a sprite change, because effects are out of the equation in schems or just some quick look at a defense

MEEPofFaith:
Why not sprite and effect changes?

bloool:
> Why not sprite and effect changes?

That would work wonders, but idk if anuke would want to change the sprites completely, since its very iconic by now, but who knows, if plastanium can get a change after so much time

Quezler:
@bloool is this still needed, or was #4807 enough?

bloool:
> @bloool is this still needed, or was #4807 enough?

That would still not solve the problem when using schems, but it's really good already

PR #4808, opened 2021-02-27
Avoid drawing potential links for contacting blocks

Quezler:
something that has become a little more noticeable with the ghost cables is that it also previews in places where it transfers via proximity, this adds a small check that prevents drawing any visuals if it on direct contact on any of the 4 axis:
(doesn't draw the purple thing either, but why should it? no wire would spawn there anyways)

Screen Shot 2021-02-27 at 14 45 03
Screen Shot 2021-02-27 at 14 56 13

> though maybe this has to go in the potentiallinks part of the code (that i don't fully understand) since it could take up 1 of the drawn links that could have happened if one wasn't reserved just for the adjecency ones*

### DIFF
--- a/core/src/mindustry/world/Tile.java
+++ b/core/src/mindustry/world/Tile.java
@@ -58,7 +58,7 @@ public class Tile implements Position, QuadTreeObject, Displayable{
         return Point2.pack(x, y);
     }
 
-    /** Returns true if these tiles touch on any axis. */
+    /** Returns true if these tiles are right next to eachother. */
     public boolean adjacentTo(Tile tile){
         return relativeTo(tile) != -1;
     }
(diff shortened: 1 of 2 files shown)


sk7725:
It should not draw a link but still draw the purple square, intuitively introducing the power link system and adjacent conductivity to new players.

Anuken:
I don't think it should be part of `potentialLinks`, as you mentioned. Power nodes should not consider adjacent blocks as a link, even if it conducts power - no *connection* is made.

What makes this complicated is that 2x2 blocks can be next to other 2x2 blocks, and the algorithm also needs to take adjacency into account so conducting power graphs don't get linked twice.

Quezler:
> What makes this complicated is that 2x2 blocks can be next to other 2x2 blocks, and the algorithm also needs to take adjacency into account so conducting power graphs don't get linked twice.

allright let me think 🤔

Anuken:
I'll give a shot at implementing this myself.

Quezler:
as you wish ❤️

PR #4814, opened 2021-02-28
Make other things use modulo

MEEPofFaith:
Since genericCrafter now `%` instead of setting to 0, I figured other things should also do the same.

### DIFF
--- a/core/src/mindustry/world/blocks/power/ItemLiquidGenerator.java
+++ b/core/src/mindustry/world/blocks/power/ItemLiquidGenerator.java
@@ -134,7 +134,7 @@ public class ItemLiquidGenerator extends PowerGenerator{
                     Item item = items.take();
                     productionEfficiency = getItemEfficiency(item);
                     explosiveness = item.explosiveness;
-                    generateTime = 1f;
+                    generateTime %= 1f;
                 }
 
                 if(generateTime > 0f){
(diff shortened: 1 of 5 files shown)


core/src/mindustry/world/blocks/power/ItemLiquidGenerator.java
```diff
@@ -134,7 +134,7 @@ public void updateTile(){
                     Item item = items.take();
                     productionEfficiency = getItemEfficiency(item);
                     explosiveness = item.explosiveness;
-                    generateTime = 1f;
+                    generateTime %= 1f;
```
Anuken, line 137: yikes

MEEPofFaith: -0.3 % 1 should = 0.7.
Or should I convert to `Mathf.mod` instead?

MEEPofFaith:
oh dear

MEEPofFaith:
https://user-images.githubusercontent.com/54301439/109405507-2d42c200-7926-11eb-9667-dd8bd46e4093.mp4

Somehow I can compile but not Travis.
Also, this is very much not right.

core/src/mindustry/world/blocks/power/ItemLiquidGenerator.java
Anuken: > -0.3 % 1 should = 0.7.

No, that's not how mod works, and there's no reason to use mod here at all.

sk7725:
Modulo anly makes sense when a variable(n) is:
from 0 to a large float, += delta() each frame

some blocks have n go from 0 to 1, with a small float added each frame. Don not the modulo there.

MEEPofFaith:
I'm lazy just gonna not bother.

PR #4816, opened 2021-02-28
Add buildSpeedMultiplier status effect

joshuaptfan:
~~This PR makes the disarmed status effect into a general "stops units from doing things" effect, by also preventing building. This makes disarmed effective against builder units like polys and megas.~~

This PR adds the `buildSpeedMultiplier` status effect. `canBuild()` returns false when this effect is ≤ 0.

Related: #4762

### DIFF
--- a/core/src/mindustry/type/StatusEffect.java
+++ b/core/src/mindustry/type/StatusEffect.java
@@ -19,7 +19,9 @@ public class StatusEffect extends MappableContent{
     public float speedMultiplier = 1f;
     /** Unit reload multiplier. */
     public float reloadMultiplier = 1f;
-    /** Unit weapon(s) and building disabled. */
+    /** Unit build speed multiplier. */
+    public float buildSpeedMultiplier = 1f;
+    /** Unit weapon(s) disabled. */
     public boolean disarm = false;
     /** Damage per frame. */
     public float damage;
--- a/core/src/mindustry/type/StatusEffect.java
+++ b/core/src/mindustry/type/StatusEffect.java
@@ -15,11 +15,11 @@ public class StatusEffect extends MappableContent{
     public float damageMultiplier = 1f;
     /** Unit health multiplier. */
     public float healthMultiplier = 1f;
-    /** Unit speed multiplier */
+    /** Unit speed multiplier. */
     public float speedMultiplier = 1f;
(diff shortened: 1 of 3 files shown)


Quezler:
imo building should be blocked by a different status effect, i forsee usecases where people would want to block shooting or building, but not both at the same time, for example in https://discord.com/channels/391020510269669376/807692120160337981/814861007624339457 & https://github.com/unjown/playerDam/blob/master/src/example/ExamplePlugin.java it would not make sense to block building, just shooting 🤔

joshuaptfan:
Ok, I'll make it a separate status effect once I come up with a good name for this effect.

sk7725:
> Ok, I'll make it a separate status effect once I come up with a good name for this effect.

no, please do not. Rather add a field: buildSpeedMultiplier

it would be 1 by default, and if it is 0 or below it would prevent building at all.

Quezler:
note that `canBuild` being false also prevents placing/using schematics, might be intentional or not based on the duration 🤔

Quezler:
@joshuaptfan build is failing since `buildSpeedMultiplier` needs to be defined in `StatusComp.java` as well 😗

Quezler:
> > Ok, I'll make it a separate status effect once I come up with a good name for this effect.
> 
> no, please do not. Rather add a field: buildSpeedMultiplier
> 
> it would be 1 by default, and if it is 0 or below it would prevent building at all.

yea we already discussed this together on the discord 😋

core/src/mindustry/type/StatusEffect.java
```diff
@@ -15,10 +15,12 @@
     public float damageMultiplier = 1f;
     /** Unit health multiplier. */
     public float healthMultiplier = 1f;
-    /** Unit speed multiplier */
+    /** Unit speed multiplier. */
     public float speedMultiplier = 1f;
```
Quezler, line 19: hmm, maybe rename it to `moveSpeedMultiplier` to make it a little more descriptive, but then again it might break modding compatibility, and rules.java has borked speed rules anyways, i don't know tbh 🤔

joshuaptfan: I'm not qualified to make that call, I'll leave it to Anuke.

PR #4818, opened 2021-02-28
Draw to-be affected blocks when placing projectors

Quezler:
previously overdrive & mend projectors did not display which blocks in range would be affected during placement mode, this pull request makes that happen, as well as change the placement mode circle color to the one you see when its selected:

Screen Shot 2021-02-28 at 15 35 12
Screen Shot 2021-02-28 at 15 35 19


XEN0PHIL:
Can't believe it's not in the game yet

sk7725 (approved):

Issue #2012, opened 2020-05-09
Graphic glitch/bug


sk7725:
**Platform**: *Android*

**Build**: *104.6*

**Issue**: *Black bars appearing after a unit dies.*

**Steps to reproduce**: *Played around with the mod sk7725/Commandblocks to fix stuff, and suddenly i noteced when a unit dies, these black bars are left. Even after disabling the mod or restarting the app, these bars are still there.*

**Link to mod(s) used, if applicable**: *sk7725/Commandblocks was present, but the bug persisted even after disabling that mod.*
---

*Place an X (no spaces) between the brackets to confirm that you have read the line below.*  
- [X] **I have searched the closed and open issues to make sure that this problem has not already been reported.**

sk7725:
![Screenshot_20200509-173157_Mindustry](https://user-images.githubusercontent.com/58885089/81468758-7c472b80-921c-11ea-89c3-af03176e602d.jpg)

sk7725:
Also, I dunno what happened with the format but cannot un-uglify it for some reason

sk7725:
I found the reason, my mod overused print(i forgot to remove that after debugging)and made the last_log.txt a whopping 6 MB which resulted in performance issues, it was fixed when I deleted last_log.txt. However, I still believe this is not working as intended and should be fixed(malicious mods?).

Anuken:
The sprite that's becoming black here doesn't exist in 6.0, so this isn't relevant anymore.

> malicious mods

There's nothing I can do about mods allocating too much memory.

sk7725:
Correction: this problem recurred, independant of the size of the last_log.txt.

sk7725:
It is quite annoying, any idea how to fix it?

Anuken:
Swipe out of the app completely when restarting (as opposed to pressing 'quit').

PR #4820, opened 2021-02-28
Color power lasers depending on battery drain

joshuaptfan:
This PR resolves [suggestion #2012](https://github.com/Anuken/Mindustry-Suggestions/issues/2012).

Currently, while batteries are being drained, power lasers remain full bright because battery power is fulfilling power needs. That means that when batteries are finally depleted, players are suddenly and rudely awakened to total power loss with no warning.

This PR colors power lasers according to the percent of total stored power being drained each tick. This solution should scale well to power grids both small and large, early game and late game. The color used is the same as [that used to indicate battery fullness](https://github.com/Anuken/Mindustry/blob/master/core/src/mindustry/world/blocks/power/Battery.java#L17).

The color change is designed to be sudden and immediately noticeable upon transitioning to negative power production. From there, as the total stored power depletes, the color becomes increasingly redder until total power loss, at which there is another sudden color change to indicate that.

These images show power laser color at 1%, 33%, 66%, and 99% battery depletion. Notice that the 1% laser color varies according to drain.

Power laser battery color

How it looks at 50% power laser opacity:

Power laser battery color 50%

### DIFF
--- a/core/src/mindustry/world/blocks/power/PowerGraph.java
+++ b/core/src/mindustry/world/blocks/power/PowerGraph.java
@@ -19,7 +19,7 @@ public class PowerGraph{
     private final Seq<Building> all = new Seq<>(false);
 
     private final WindowedMean powerBalance = new WindowedMean(60);
-    private float lastPowerProduced, lastPowerNeeded, lastPowerStored;
+    private float lastPowerProduced, lastPowerNeeded, lastPowerStored, lastBatteryUsed;
     private float lastScaledPowerIn, lastScaledPowerOut, lastCapacity;
     //diodes workaround for correct energy production info
     private float energyDelta = 0f;
@@ -80,6 +80,10 @@ public class PowerGraph{
             return 1f;
         }
         return Mathf.clamp(lastPowerProduced / lastPowerNeeded);
+    }
+
+    public float getBatteryUsedFraction(){
+        return Mathf.zero(lastPowerStored) ? 0f : lastBatteryUsed / lastPowerStored;
     }
 
     /** @return multiplier of speed at which resources should be consumed for power generation. */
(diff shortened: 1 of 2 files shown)


Quezler:
the colors shown in the picture feel a bit out of place 🤔

joshuaptfan:
It should be some shade of orange-red to fit in with the current power/battery color scheme, and this color is pulled directly from that palette.

Kyllingene:
I like this, however it feels odd that the "completely out of power" color seems less negative than the "66% depleted" color.

joshuaptfan:
I suppose it would be most sensible to make "batteries depleting" be orange, and "no power" reddish.

However, the problem I need to solve to get there is, there are actually 2 different states between full power and no power:

* insufficient power (power needed < power produced)
* batteries depleting (power needed < power produced, but batteries are making up for the deficit)

So, I need to come up with 2 separate color schemes to show that. Currently, I have not altered the "insufficient power" color scheme, that's why the "no power" color is orange just as it has always been before.

So how about this:

* red for "no power"
  * for insufficient power, blend between white and red, until power lasers turn white at full power
* orange for "batteries depleting"
  * as batteries deplete, blend between white and orange
  * sharp jump in color from "full power → depleting"
  * at the "depleting → no power" transition, the color will change from orange to red, so that transition will be clear too

Kyllingene:
I like the new suggested color scheme, and think it would work well. Also, I assume those less-than signs are supposed to be greater-than signs?

joshuaptfan:
> Also, this should be explained in hints or something.

I think with this more intuitive color scheme, hints won't be necessary. Here it is in action:

Power laser battery color

At 50% power laser opacity:

Power laser battery color 50%

joshuaptfan:
Current color scheme (less red):

Power laser battery color

Jammo2000:
The transition b/w sufficient power and insufficient power isn't clear enough. Battery draining looks good though.

joshuaptfan:
I applied a [parabolic curve](https://en.wikipedia.org/wiki/Parabola) to the insufficient power gradient, such that the transition from insufficient power to full power is at the vertex of the curve. Blue line is the old curve (linear), red is the new curve. On the Y axis, 0 is white and 1 is orange-red.

Parabolic curve

Current color scheme:

Power laser battery color

PR #4836, opened 2021-03-02
Processors should have priority

MEEPofFaith:
Example: You place a reactor schem that uses logic to shut itself off. Unfortunately, the reactor is placed before the processor then the entire thing blows up.
If processors are placed before the reactor they can disable the reactor pretty much as soon as it's placed.

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -2067,6 +2067,8 @@ public class Blocks implements ContentList{
             instructionsPerTick = 2;
 
             size = 1;
+            
+            schematicPriority = 10;
         }};
 
         logicProcessor = new LogicBlock("logic-processor"){{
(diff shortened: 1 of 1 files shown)


GreyMario:
(How is it blowing itself up without a connected item input? Connect inputs to the schematic _after_ placing it. 🍿🐬)

the-red-soul:
Maybe by expanding currently built thor gens with a schematic. I approve of this.

MEEPofFaith:
> (How is it blowing itself up without a connected item input? Connect inputs to the schematic _after_ placing it. 🍿🐬)

what if im lazy and don't want to disconnect anything

MEEPofFaith:
I figured that since plastanium walls have special use and they already have schematic priority, I decided that processors, which likely need to start controlling things when they're built should also have priority.

Quezler:
it should probably not be on the same schematic priority tier as insulated blocks

joshuaptfan:
I guess it would look aesthetically nicer for insulator walls and diodes to be allowed to finish building first by setting `schematicPriority` to 9, but practically speaking it doesn't matter as power nodes don't connect to processors.

Anuken:
This should be defined in `LogicBlock`.

PR #4837, opened 2021-03-02
Remove duplicates and mild formatting.

MEEPofFaith:
Removes duplicate definitions
- Pulsar `shootY`
- Quasar `hitSize`
- Corvus `mineTier`
- Tsunami `recoilAmount`

Indent `ammo` on `ItemTurret`s and `results` on `Separator`s
Looks nicer, and matches with Tsunami.


MEEPofFaith:
(If people can find value definitions that are just the defaults tell me so I delete them as well.)

PR #4839, opened 2021-03-02
Payload units spawn units on crash, but the spawned units will have reduced health.

MEEPofFaith:
It doesn't really make sense for the unit to crash but then the units inside just cease to exist.

https://user-images.githubusercontent.com/54301439/109629963-3e8ff800-7af9-11eb-9d34-415d73e69343.mp4

### DIFF
--- a/core/src/mindustry/entities/comp/PayloadComp.java
+++ b/core/src/mindustry/entities/comp/PayloadComp.java
@@ -162,4 +162,34 @@ abstract class PayloadComp implements Posc, Rotc, Hitboxc, Unitc{
             table.image(p.icon(Cicon.small)).size(itemSize).padRight(pad);
         }
     }
+
+    @Override
+    public void destroy(){
+        //if this unit was carrying other units, spawn them but at reduced health
+        payloads.each(payload -> {
+            if(payload instanceof UnitPayload p){
+                Unit u = p.unit;
+
+                //drop all units, let ground units die on walls
+                Fx.unitDrop.at(this);
+
+                if(Vars.net.client()){
+                    //clear removed state of unit so it can be synced
+                    Vars.netClient.clearRemovedEntity(p.unit.id);
+                }else{
+                    //clients do not drop payloads
+                    u.set(this);
+                    u.trns(Tmp.v1.rnd(Mathf.random(2f)));
+                    u.rotation(rotation);
+                    //reset the ID to a new value to make sure it's synced
+                    u.id = EntityGroup.nextId();
+                    //decrement count to prevent double increment
+                    if(!u.isAdded()) u.team.data().updateCount(u.type, -1);
+                    //take damage from the landing
+                    u.health = u.maxHealth * u.type.crashDurability;
+                    u.add();
+                }
+            }
+        });
+    }
 }
--- a/core/src/mindustry/entities/comp/PayloadComp.java
+++ b/core/src/mindustry/entities/comp/PayloadComp.java
@@ -186,7 +186,7 @@ abstract class PayloadComp implements Posc, Rotc, Hitboxc, Unitc{
                     //decrement count to prevent double increment
                     if(!u.isAdded()) u.team.data().updateCount(u.type, -1);
                     //take damage from the landing
-                    u.health -= u.type.payloadCrashDamage + u.maxHealth * u.type.payloadCrashDamagePercent;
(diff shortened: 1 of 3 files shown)


core/src/mindustry/entities/comp/PayloadComp.java
```diff
@@ -162,4 +162,34 @@ void contentInfo(Table table, float itemSize, float width){
...
+                    //clients do not drop payloads
+                    u.set(this);
+                    u.trns(Tmp.v1.rnd(Mathf.random(2f)));
+                    u.rotation(rotation);
+                    //reset the ID to a new value to make sure it's synced
+                    u.id = EntityGroup.nextId();
+                    //decrement count to prevent double increment
+                    if(!u.isAdded()) u.team.data().updateCount(u.type, -1);
+                    //take damage from the landing
+                    u.health = u.maxHealth * u.type.crashDurability;
```
Quezler, line 189: doesn't take into account units that already had some damage, they damaged to or healed up to 10% health 🤔

MEEPofFaith: Then let's make it take damage instead.

core/src/mindustry/entities/comp/PayloadComp.java
```diff
@@ -162,4 +162,34 @@ void contentInfo(Table table, float itemSize, float width){
...
+                    //clients do not drop payloads
+                    u.set(this);
+                    u.trns(Tmp.v1.rnd(Mathf.random(2f)));
+                    u.rotation(rotation);
+                    //reset the ID to a new value to make sure it's synced
+                    u.id = EntityGroup.nextId();
+                    //decrement count to prevent double increment
+                    if(!u.isAdded()) u.team.data().updateCount(u.type, -1);
+                    //take damage from the landing
+                    u.health -= u.maxHealth * (1 - u.type.crashDurability);
```
Jammo2000, line 189:
```suggestion
                    u.health -= 200 + u.maxHealth * (1 - u.type.crashDurability)
```
Instead of the 0 crash durability, there should be some damage that is independent of maxHealth. The 200 should probably be set in a constant somewhere, but I dont' know where.

MEEPofFaith: Maybe rename `crashDurability` and instead make it just "damage taken on crash?"

Jammo2000: Idea - Upon crash, units take crashDurability% damage, plus the damage value for the payload unit crashing onto enemies, plus unit explosion damage.

MEEPofFaith: I feel like it should have a different name so that it doesn't get confused with the `crashDamageMultiplier` for just flying units dying.

core/src/mindustry/entities/comp/PayloadComp.java
```diff
@@ -162,4 +162,34 @@ void contentInfo(Table table, float itemSize, float width){
...
+                    //clients do not drop payloads
+                    u.set(this);
+                    u.trns(Tmp.v1.rnd(Mathf.random(2f)));
+                    u.rotation(rotation);
+                    //reset the ID to a new value to make sure it's synced
+                    u.id = EntityGroup.nextId();
+                    //decrement count to prevent double increment
+                    if(!u.isAdded()) u.team.data().updateCount(u.type, -1);
+                    //take damage from the landing
+                    u.damagePierce(u.type.payloadCrashDamage + u.maxHealth * u.type.payloadCrashDamagePercent);
```
Jammo2000, line 189:
```suggestion
                    u.damagePierce(this.type.payloadCrashDamage + u.maxHealth * this.type.payloadCrashDamagePercent);
```

core/src/mindustry/entities/comp/PayloadComp.java
Jammo2000: Sorry, I thought something was different. When I suggested, I meant that it should look up the damage for the carrying unit dying/crashing/exploding, and apply that to the dropped units.

Anuken:
This just encourages suicide bombing.

> It doesn't really make sense for the unit to crash but then the units inside just cease to exist.

You would expect a massive ship exploding as it falls from the sky to destroy all payloads completely, no? If the unit is *inside* the cargo hold, there's no way it could survive. If it's strapped *under* the ship, then it's essentially getting flattened by a massive chunk of metal - either way, letting payload units live doesn't make any sense.

PR #4840, opened 2021-03-02
Nuclear reactor schem priority of -5

buthed010203:
This will ensure that reactors are built after everything else (except nodes), this should reduce the likelihood of accidentally blowing them up due to lack of cryo (the change to logic block priority already fixes this in most cases but its better to be safe than sorry)

### DIFF
--- a/core/src/mindustry/world/blocks/power/NuclearReactor.java
+++ b/core/src/mindustry/world/blocks/power/NuclearReactor.java
@@ -54,7 +54,7 @@ public class NuclearReactor extends PowerGenerator{
         hasLiquids = true;
         rebuildable = false;
         flags = EnumSet.of(BlockFlag.reactor, BlockFlag.generator);
-        schematicPriority = -15;
+        schematicPriority = -5;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


Quezler:
i am against this being lower than the priority for power nodes

core/src/mindustry/world/blocks/power/NuclearReactor.java
```diff
@@ -54,6 +54,7 @@ public NuclearReactor(String name){
         hasLiquids = true;
         rebuildable = false;
         flags = EnumSet.of(BlockFlag.reactor, BlockFlag.generator);
+        schematicPriority = -15;
```
Quezler, line 57:
```suggestion
        schematicPriority = -5;
```

GreyMario:
I'm against people jumping on the block priority bandwagon for every little thing. Logic blocks might have had a reason. This? If you're that scared, build RTG schematics instead.

buthed010203:
> I'm against people jumping on the block priority bandwagon for every little thing. Logic blocks might have had a reason. This? If you're that scared, build RTG schematics instead.

Its not that I'm scared its just a minor change that improves qol ever so slightly

Quezler (approved):

PR #4842, opened 2021-03-02
Build and test gradle in parallel, uses multiple cores so it goes significantly faster

buthed010203:


Anuken:
Doesn't make a difference on my machine, breaks compilation.
Non-parallel: ~58 sec
Parallel: Compiles for 45 sec, then throws an error. This is 45 seconds **without** the desktop module being compiled or any tests being run.

buthed010203:
@Anuken I noticed the exact same thing when running `gradlew clean cleanTest test --parallel`, it crashed on my machine however it somehow works fine on CI. `gradlew test desktop:dist --parallel` works fine for me and it saves like 10s on my machine, I don't see why it wouldn't work on the CI though gradle tends to be weird

PR #4848, opened 2021-03-04
Add `setupRequirements`, which calls `requirements` so that it can be used in rhino js.

MEEPofFaith:
This'll allow it to be used in rhino js without rhino having a stroke about it.

I chose to change the method instead of the property because changing the method only breaks Java mods while changing the property would break every non-java mods. Java mods would be easier to fix because you could just do a `ctrl + h` to replace every `requirements` with `setRequirements` (assuming all blocks are in one file like Mindustry source code) while in non-java mods you would have to go through every file individually and fix it.

(I'm sorry sk)

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -625,6 +625,14 @@ public class Block extends UnlockableContent{
         return cacheLayer == CacheLayer.walls;
     }
 
+    public void setupRequirements(Category cat, ItemStack[] stacks){
+        requirements(cat, stacks);
+    }
+
+    public void setupRequirements(Category cat, BuildVisibility visible, ItemStack[] stacks){
+        requirements(cat, visible, stacks);
+    }
+
     public void requirements(Category cat, ItemStack[] stacks, boolean unlocked){
         requirements(cat, BuildVisibility.shown, stacks);
         this.alwaysUnlocked = unlocked;


GlennFolker:
_*distant sk noises_

sk7725:
no. just no.

just do not this PR; keeping it as is will only break *js mods that attempt to use the method*.
And, *js mods that attempt to use the method* < *java mods* < *json/js mods*, and according to your logic of majority (break java mods instead of non-java mods) this PR should not exist.

sk7725:
if you *need* to do this, since rhino prioritizes fields, you can just add 3 lines:

public void setupRequirements(Whatever whatever){
    requirements(Whatever);
}

and you can break no mods while achieving exactly this.

MEEPofFaith:
I don't know much ok? Maybe I should just wait for Anuke to appear and let him handle this.

MEEPofFaith:
> if you _need_ to do this, since rhino prioritizes fields, you can just add 3 lines:
> 
> public void setupRequirements(Whatever whatever){
> requirements(Whatever);
> }
> 
> and you can break no mods while achieving exactly this.

I hope I did exactly has you said.

sk7725 (approved):

MEEPofFaith:
(Sounds like a yes)

PR #4849, opened 2021-03-04
Disable unitLand Effect For Metal Floors

Voz-Duh:
I think what `isHard` bad name for this boolean but I don't have more ideas for this /-/
[video](https://user-images.githubusercontent.com/61590217/109904220-ef3ef880-7cad-11eb-8c94-9cb877102b60.mp4)

### DIFF
--- a/core/assets/contributors
+++ b/core/assets/contributors
@@ -114,3 +114,4 @@ simba-fs
 RedRadiation
 Marko Zajc
 CPX MC
+VozDuh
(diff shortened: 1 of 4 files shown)


joshuaptfan:
How about `isFirm`?

Voz-Duh:
> How about `isFirm`?

Hm... I think what this name better what `isHard`

Anuken:
This doesn't seem necessary. Metal can still have dust on it; things don't look any better without the effect.

sk7725:
Better suggestion: add
```
public Effect landEffect = Fx.none;
```
and make the metal floors have "sparks" (like Fx.despawn) so it looks like metal hitting metal.

joshuaptfan:
Sparks would be too distracting, it would look like there's something combat-related going on or the unit is taking damage, when it's just walking.

PR #4851, opened 2021-03-04
Enable client to use a different data dir through env MINDUSTRY_DATA_DIR

txomon:
I tried to use as many `Core` calls as I could but didn't find anything in regards to absolute vs relative paths, so I enforced absolutes to avoid a bigger fingerprint of the code

### DIFF
--- a/core/src/mindustry/ClientLauncher.java
+++ b/core/src/mindustry/ClientLauncher.java
@@ -34,6 +34,11 @@ public abstract class ClientLauncher extends ApplicationCore implements Platform
 
     @Override
     public void setup(){
+        String dataDir = OS.env("MINDUSTRY_DATA_DIR");
+        if(dataDir != null){
+            Core.settings.setDataDirectory(files.absolute(dataDir));
+        }
+
         checkLaunch();
         loadLogger();
 


core/src/mindustry/ClientLauncher.java
```diff
@@ -182,6 +185,14 @@ public void exit(){
 
     @Override
     public void init(){
+        String dataDir = OS.env("MINDUSTRY_DATA_DIR");
+        if(dataDir != null){
+            if(Paths.get(dataDir).isAbsolute()){
+                Core.settings.setDataDirectory(Core.files.absolute(dataDir));
+            }else{
+                throw new IllegalArgumentException("MINDUSTRY_DATA_DIR must be an absolute file path");
```
Quezler, line 193:
```suggestion
                throw new IllegalArgumentException("MINDUSTRY_DATA_DIR must be an absolute file path.");
```

core/src/mindustry/ClientLauncher.java
```diff
@@ -182,6 +185,14 @@ public void exit(){
 
     @Override
     public void init(){
+        String dataDir = OS.env("MINDUSTRY_DATA_DIR");
+        if(dataDir != null){
+            if(Paths.get(dataDir).isAbsolute()){
```
Anuken, line 190: Do not use `Paths` or any other Java file API here. See the guidelines.

core/src/mindustry/ClientLauncher.java
```diff
@@ -182,6 +185,14 @@ public void exit(){
 
     @Override
     public void init(){
```
Anuken, line 187: This shouldn't be in `init()` - all initialization logic is in `setup()`.

Anuken (changes requested):

core/src/mindustry/ClientLauncher.java
txomon: Done, I have realised that `Files.absolute` only means `verbatim`, so that it doesn't suffer any modifications, not related to it being an absolute path. I will try to update the docstring

core/src/mindustry/ClientLauncher.java
txomon: Done

core/src/mindustry/ClientLauncher.java
txomon: Just in case, this change has been removed because there is no need to enforce paths being absolute, thanks for the contribution!

Erisfiregamer1:
YES! Finally, I can make a portable mindustry version!!

Erisfiregamer1 (approved):
Hopefully this becomes a setting- Having a way to set the directory that Mindustry uses would be AWESOME for portable mindustry copies.

txomon:
> Hopefully this becomes a setting- Having a way to set the directory that Mindustry uses would be AWESOME for portable mindustry copies.

That would be difficult, settings by definition need to exist somewhere, making it a chicken and egg problem in your usecase. Your best bet would be to have a bash script that starts the server and do the setting of the env var there

PR #4857, opened 2021-03-05
Fix typo to preserve sanity

Quezler:
also puts a period in between to categorize it slightly


joshuaptfan:
Good job nipping that before it became permanent.

PR #4858, opened 2021-03-05
Remove dead servers from v5 server list

buthed010203:
From what I can see, atnanner is the only network that still hosts any v5 servers, all the others are either on v6 or are offline.

### DIFF
--- a/servers.json
+++ b/servers.json
@@ -1,35 +1,5 @@
 [
   {
-    "address": "mindustry.indielm.com:1101"
-  },
-  {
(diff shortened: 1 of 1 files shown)


Quezler:
objection, nydus v5 isn't dead (yet), i just forgot to point the new ip to the right place yet 😗

Anuken:
@Quezler Once that's done, make a PR to re-add it to the list and I'll merge it.

Quezler:
nah, its probably just a sign it has had its final sunset, ill close it down completely soon 🤔

PR #4859, opened 2021-03-05
Add a server config option

TranquillyUnpleasant:
For servers without plugins

### DIFF
--- a/core/src/mindustry/net/Administration.java
+++ b/core/src/mindustry/net/Administration.java
@@ -363,7 +363,6 @@ public class Administration{
         return getCreateInfo(uuid).banned;
     }
 
-
     public boolean isAdmin(String id, String usid){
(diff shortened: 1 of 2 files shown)


TranquillyUnpleasant:
Also check admins for ban on invokePacket, unsure if this is redundant.

core/src/mindustry/net/Administration.java
```diff
@@ -359,6 +363,7 @@ public boolean isIDBanned(String uuid){
         return getCreateInfo(uuid).banned;
     }
 
+
```
joshuaptfan, line 366: Accidental line break.
```suggestion
```

Anuken:
This *should* be redundant now, as I made kicked connections skip any connection/packet logic yesterday.

PR #4865, opened 2021-03-05
deleting of rally point assets

AnOreo1:
they aren't used or mentioned so why not delete them


Voz-Duh:
image
?

Voz-Duh:
> they aren't used

Really?
image

AnOreo1:
ok

PR #4870, opened 2021-03-06
dragMultiplier for status effects

sk7725:
+ A very fun property to mess around with.
+ Some new status effects may be possible with this; for example a status effect that freezes you in a chunk of ice, making you unable to control your movement.
+ As my usual PR, this is a feature that will lead to more creative mod features (with an added bonus; benefits json modders too!).

- [X] StatusComp
- [x] UnitComp
- [x] StatusEffect


core/src/mindustry/type/StatusEffect.java
```diff
@@ -21,6 +21,8 @@
     public float reloadMultiplier = 1f;
     /** Unit build speed multiplier. */
     public float buildSpeedMultiplier = 1f;
+    /** Unit drag multiplier */
```
joshuaptfan, line 24: Javadocs should have punctuation, look at the surrounding javadocs.
```suggestion
    /** Unit drag multiplier. */
```

joshuaptfan:
Maybe put all `dragMultiplier` stuff after `speedMultiplier` so the 2 movement-speed-affecting status effects are grouped together?

core/src/mindustry/entities/comp/StatusComp.java
```diff
@@ -135,6 +135,7 @@ public void update(){
                 damageMultiplier *= entry.effect.damageMultiplier;
                 reloadMultiplier *= entry.effect.reloadMultiplier;
                 buildSpeedMultiplier *= entry.effect.buildSpeedMultiplier;
+                dragMultiplier *= entry.effect.dragMultiplier;
```
Anuken, line 138: This needs to be reset to 1 every update above the loop.

sk7725: right. I forgot

PR #4874, opened 2021-03-06
light block placement

liuxilu:
https://github.com/Anuken/Mindustry-Suggestions/issues/1410

For the two magic numbers, I just place explanations here for now:
- I just measured "circle-shadow.png" in photoshop by my brightness recognizability and got 0.8.
- Multiplied 3 is the square of the distance between centers of overlapping regular-triangular tessellated unit circles, to have lights just contiguous.

x0.8:

image
x0.8

x0.75 (I wrongly multiplied `brightness` before):
x0.75

### DIFF
--- a/core/src/mindustry/world/blocks/power/LightBlock.java
+++ b/core/src/mindustry/world/blocks/power/LightBlock.java
@@ -3,12 +3,15 @@ package mindustry.world.blocks.power;
 import arc.graphics.*;
 import arc.graphics.g2d.*;
 import arc.math.*;
+import arc.math.geom.*;
 import arc.scene.ui.layout.*;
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/blocks/power/LightBlock.java
```diff
@@ -3,12 +3,15 @@
 import arc.graphics.*;
 import arc.graphics.g2d.*;
 import arc.math.*;
+import arc.math.geom.Point2;
```
genNAowl, line 6: Always use wildcard imports (use import arc.math.geom.*;)

Anuken:
Implemented, with some constant tweaks.

PR #4880, opened 2021-03-06
MassDriverBolt shouldn't collide with ground

buthed010203:
It seems odd to me that it is able to go over buildings but can be intercepted by ground units


Anuken:
Eh, the only reason it doesn't collide with tiles is so that people don't try to attack enemy drivers by putting random conveyors between them. Placing ground units between drivers is harder to pull off, and more interesting as tactic.

PR #4900, opened 2021-03-08
Translate the entire bundle into Danish.

nilq:
There were close to no existing translations of Mindustry in Danish, and the ones there were had concord issues, spelling issues or contained a mix of Danish and English. These existing translations have been improved.

Furthermore, this pull-request provides a thourough translation; covering all names - unit names have remained mostly the same, - as well as tutorials and descriptions.

A collaborative effort with @AsgerHB.


Catchears:
To have complete coverage, you could some of the newer entries: 
[https://github.com/Anuken/Mindustry/blob/master/core/assets/bundles/bundle.properties](https://github.com/Anuken/Mindustry/blob/master/core/assets/bundles/bundle.properties)

nilq:
Definitely could. This is merely an overwrite of the existing danish bundle. Will get back with those additions.

PR #4909, opened 2021-03-09
event for UnitSpawnAbility creating a unit

catsz:


Quezler:
why not reuse one of the existing unit creation events?

Anuken:
No longer necessary, merged into the unit creation event.

PR #4912, opened 2021-03-10
buff shale

unjown:
[suggestion](https://github.com/Anuken/Mindustry-Suggestions/issues/2129)


BlueWolf3682:
it will just be darksand but you cant drill sand from it then

genNAowl:
True; It needs to be better than dark sand or else people will still keep using dark sand instead.

Bohaska (approved):
> good good better better best best

Bohaska

Issue #2127, opened 2020-05-30
Sometimes, ressources stop being stored on multiplayer server


MasterBroki:
**Platform**: Windows

**Build**: 104.10

**Issue**: I played a game with a friend and even though the conveyer belts were still rolling and the thorium coming in the core, the counter wasn't going up. 

**Steps to reproduce**: Still working on reproducing the issue

**Crash report, if applicable**: *The contents of relevant crash report files.*

---

*Place an X (no spaces) between the brackets to confirm that you have read the line below.*  
- [x ] **I have searched the closed and open issues to make sure that this problem has not already been reported.**

LeoVen:
Not sure if related... but:

I've had similar issues with (what I believe) synchronization. There is this one sand that somehow got into a titanium conveyor going into the core. Only one of my friends saw this problem, but for everyone else, the conveyor was just fine with all the titanium going in.

Also, I've had occasions where someone couldn't delete a conveyor or a conduit. In this case, there was also the problem of certain resources not matching in each player's screen.

I wonder if these issues have been already fixed for 6.0. Or if there is any discussion about it. I haven't opened an issue on this because I couldn't reproduce it and it seems to be such a rare event that doesn't bother the gameplay at all.

zethnest:
Have you tried doing `/sync`? Is the wave counter stuck at 0? Are the players moving? Are the enemies moving? Or there are no players in the player list but you and others can still chat with each other?

Pota2s:
This happened to us when I was playing with my brother, the surge alloy goes in, and is immediately deleted upon reaching the core, it's fine when it's on containers not connected to the core but it makes us unable to build anything using surge alloys

DeltaNedas:
use /sync

Anuken:
Duplicate.

PR #4915, opened 2021-03-10
Desktop: Boost toggle setting

joshuaptfan:
This PR resolves [suggestion #2127](https://github.com/Anuken/Mindustry-Suggestions/issues/2127).

It adds a game setting (default: off) to change the desktop boost key from a hold to a toggle, to lessen strain on players' pinkie fingers.

Mechanically, boost behavior is unchanged; the player unit will continue to only boost while moving, and will remain boosted if boost is turned off while the player unit is over an obstruction.

## Needs localization
* "Boost Toggle" game setting


Quezler:
boost while capslock is on 😄

joshuaptfan:
Players are free to modify their keybinds as such. However, the boost key will only toggle boost while you are controlling a boostable unit, so Caps Lock may get out of sync with the state of boost.

Bohaska (approved):

sk7725:
Should be left as a control option setting, I myself want the boost to be held and not toggle, and I know that some others would too.

joshuaptfan:
I have been fishing for specific situations when boost hold is better, and I'm sure this PR would be more palatable to Anuke if it were optional. Can you explain yours?

sk7725:
> I have been fishing for **specific situations** where boost hold is better, and I'm sure this PR would be more palatable to Anuke if it were optional. However, I feel game settings are getting a bit bloated, and an option isn't worth adding if one of the choices is worse in every way.
> 
> Can you explain yours? "I want" is not enough; _why_ do you want?

Because 99% of the time, you *don't* want to be boosting; boosting makes the mech unable to react and use weapons. Most of the time, if you are boosting to jump over solid blocks, you dont need to hold it down anyways, so it isnt too much of a strain on the pinky (wait, you press shift with your pinky?)

sk7725:
In PvP/PvE/PvT situations, you need to react quickly, and forgetting to press shift again, or failing to do so quickly, will cost you an unit. Not to mention that the increased urgent presses will actually stress your pinky more, thus negating the pros of this change, if any.

sk7725:
uh, hello? Why are you not addressing my points at all? I addressed yours.

sk7725 (changes requested):
for above reasons, this must be an option if implemented at all.

joshuaptfan:
Let's refactor your points to make them impartial.

* Not boosting is the ideal state to be in, because you can't shoot while boosting.

Moving while boosting is significantly faster than not boosting, so when traveling long distances in boostable units, you will be holding down the boost key for a long time. This is where pinkie strain comes into play.

* Forgetting to toggle off boost can be dangerous.

Any movement while boost is toggled on will cause your unit to visibly rise. It's practically impossible to forget that boost is toggled on.

Thus, considering the hypothetical situation where you are commanding a group of novas in an attack map, against a mix of anti-air and anti-ground defenses, and accidentally boosting within range of anti-air can cause your units to be destroyed: Even with boost as a toggle, it would be impossible to accidentally boost without you manually pressing the boost key. It is impossible to "leave it toggled on by accident" because you can't move without your movement reflecting your current boost state.

sk7725:
+ Moving while boosting is significantly faster than not boosting

Nope! Units have a *different*, not *faster* speed while boosting. Modded units, as well as (maybe) future planned units may have boosting speed slower than normal speed.

+ Any movement while boost is toggled on will cause your unit to visibly rise. It's practically impossible to forget that boost is toggled

You are vastly overestimating the capacity of the human brain, or underestimating its stupidity. Static (the engine does grow and shrink but only too subtley) visory cues are commonly disregarded in the human brain, and its a gleden rule that for one "A cannot be forgotten" presumption, there are 10 instances of A forgotten. (don't trust me? go to any irl-topic website/media/subreddit etc.)

joshuaptfan:
I'm not persuaded by your memory arguments, as they are getting rather contrived. However, I'm adding the setting anyway, because Anuke would otherwise be uncomfortable with a PR that irreversibly alters the current behavior of controls.

Bohaska (approved):
hmm looks good ok i approve this

Anuken:
This is a year late, but with the Erekir update, boosting becomes even less relevant, and regardless of usefulness, I just don't want to maintain another keybind or add another setting for something that very few people will use.

joshuaptfan:
Ok, as long as it's not an issue.

PR #4925, opened 2021-03-12
setting zoom key for editor

liuxilu:
Implements https://github.com/Anuken/Mindustry-Suggestions/issues/2138


Quezler:
have you tested this ingame? on first glance it looks like the button based zoom would only work one way 🤔

liuxilu:
Yes. And I'm using the same input function as in-game zoom.

joshuaptfan:
I support changing from a hardcoded binding to reusing an existing configurable binding.

PR #4931, opened 2021-03-13
Leg mass for spider units.

MEEPofFaith:
Adds `legPayloadMass`, which is the mass taken up per leg.

Reason: It makes sense for the legs to take up payload space, but also since some of the vanilla spiders have smaller `hitSize`s because most of their size is leg (example: Toxopid, which is about the same size as Zenith)

### DIFF
--- a/core/src/mindustry/content/UnitTypes.java
+++ b/core/src/mindustry/content/UnitTypes.java
@@ -589,6 +589,7 @@ public class UnitTypes implements ContentList{
             legLength = 9f;
             legTrns = 0.6f;
             legMoveSpace = 1.4f;
+            legPayloadMass = 9f;
             hovering = true;
             armor = 3f;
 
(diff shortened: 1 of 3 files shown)


Quezler:
does this change the balance of certain units being able to get picked up by any if the 3 payload air tiers at all?

MEEPofFaith:
> does this change the balance of certain units being able to get picked up by any if the 3 payload air tiers at all?

Oct can only carry 2 Toxopids now ;)

MEEPofFaith:
though all the other tiers were kinda random because I don't know how to balance.
I'm sure Anuke would be better at balancing that.

joshuaptfan:
Payload capacity is not measured in mass, but size. `legPayloadMass` should be `legPayloadSize`, and `legMass` should be `legSize`.

BlueWolf3682:
This would change the somewhat not well known fact that a quad can currently pick up a toxopid. That's about it.

Quezler:
i suppose that would be an acceptable balance change ^

Anuken:
Leg unit hitboxes were already small for their body size; with the 7.0 adjustments this isn't necessary anymore.

PR #4938, opened 2021-03-15
chargeMoveFract

MEEPofFaith:
Instead of locking the turret's rotation, it applies a rotation speed modifier similarly to `firingMoveFract` on `LaserTurret`s when charging.


Anuken:
This doesn't work properly. Turrets usually create a charge effect, which can't rotate with the turret.

PR #4945, opened 2021-03-17
Swap map name and mode params in the host command.

TranquillyUnpleasant:
Currently if you want to host a specific mode you have to specify the map, which should not be necessary.

If you pick a specific map you should also have to specify its game mode, since all maps aren't survival.

Ergo swap.

### DIFF
--- a/server/src/mindustry/server/ServerControl.java
+++ b/server/src/mindustry/server/ServerControl.java
@@ -285,7 +285,7 @@ public class ServerControl implements ApplicationListener{
             info("Stopped server.");
         });
 
-        handler.register("host", "[mapname] [mode]", "Open the server. Will default to survival and a random map if not specified.", arg -> {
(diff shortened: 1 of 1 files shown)


Anuken:
This will break far too many servers.

> Currently if you want to host a specific mode you have to specify the map, which should not be necessary.

I don't see the logic here. With the parameters swapped, it becomes *"If you want to host a specific map you have to specify the mode, which should not be necessary"* - which makes just as much (if not more) sense.

> If you pick a specific map you should also have to specify its game mode, since all maps aren't survival.

I disagree. Survival is the most common (vanilla) gamemode. It is not unreasonable to leave it as a default.

PolloUsami:
maybe he was trying to say that host can be able of changing gamemode ( survival ,atack, pvp )

honestly i believe that it can crash specialy with changing from survival to pvp

LeoSko:
Simply introducing pattern-like command would work, like
`host * pvp`
`host Veins *`

PR #4951, opened 2021-03-17
Power Source Power Generation Stat

MEEPofFaith:
Adds power source power generation to its stats.


joshuaptfan:
Why not reuse the "Base Power Generation" string instead of creating a new string?

MEEPofFaith:
Power Source power generation doesn't change dynamically, so the power generation is a base that changes, but is instead constant.

joshuaptfan:
Then at least name the string `stat.powergeneration` so that it can be reused for other constant power sources.

liuxilu:
But it's not a 'generator' but a 'magic source'.
And there're other generators with unbootable output.

MEEPofFaith:
🤔

Anuken:
I'd rather not add another stat for this, it just seems unnecessary.

PR #4955, opened 2021-03-18
Fix json weather parsing

QmelZ:


Joshwelyt:
Yessss

Anuken:
This doesn't fix anything; in fact, it only breaks the default type. How do you expect it to instantiate `Weather`?

Weather parsing already works fine.

BlueWolf3682:
(current weather oarsing doesnt work at all, as far as i know)

QmelZ:
As bluewolf already said, ParticleWeather does not parse at all.

I have done my tests, the changes i made actually work and i can create both types of weathers using hjson.

Anuken:
> As bluewolf already said, ParticleWeather does not parse at all.

Works fine for me. 126.1 is **not** equivalent to the master branch.

> I have done my tests, the changes i made actually work

Have you tested on master *before* your changes were applied, or did you just launch 126.1?
Your code **will** fail with a cryptic abstract instantiation error when no type is specified.

PR #4958, opened 2021-03-18
Update servers_v6.json

osp54:
Add to the global server list. More modes will be added soon. Now only Mini Pvp


joshuaptfan:
Do not close a PR just because there's a mistake that can be fixed. You can push additional commits to modify the PR after it has been submitted.

Do not spam PRs.

PR #4961, opened 2021-03-19
Added some thedimas' servers

thedimas3007:
MS:GO, Annexation, Hex PvP, Campaign


Darkness6030:
You have hub. You don't need any other servers in list

thedimas3007:
It would be faster join server from the list, than from hub

PR #4962, opened 2021-03-19
Added TSR Sandbox to serverlist

GaviTSRA:


Darkness6030:
We have a lot of Sandbox servers, for example RCR, Mindurka, TheDimas. We don't need another one

Issue #4772, opened 2021-02-24
bridges pointing down don't link when rotated


Quezler:
**Platform**: `Mac OS X x64`

**Build**: `bleeding-edge build 20812`

**Issue**: if you have a bridge pointing down in a schematic and rotate it (or not) in any way it doesn't link.

**Steps to reproduce**: 

- place a bridge block
- place another bridge block below it
- make a schematic of it
- paste it in any of the 4 axis
- notice none of them linking as intended

![Screen Shot 2021-02-24 at 11 21 51](https://user-images.githubusercontent.com/3179271/108986823-da051100-7692-11eb-814a-44dac01d16c0.png)

---

*Place an X (no spaces) between the brackets to confirm that you have read the line below.*
- [ ] **I have updated to the latest release (https://github.com/Anuken/Mindustry/releases) to make sure my issue has not been fixed.**
- [ ] **I have searched the closed and open issues to make sure that this problem has not already been reported.**

Quezler:
right to left also seems affected

joshuaptfan:
Doing a quick test on BE 20812, it is only the left-to-right facing bridge that doesn't link when rotated.

![Bridge bug](https://user-images.githubusercontent.com/15149002/109016048-9eeee780-766a-11eb-933b-a962ced34675.png)

PR #4964, opened 2021-03-19
Fix bridges not connecting when clicked and dragged

buthed010203:
[Before](https://aethex.is-a.fail/S7CLob.webm)
[After](https://xenon.is-ne.at/XqPcZ9.webm)
This happens due to the fix for #4772, I have made sure that this still works as it should.

### DIFF
--- a/core/src/mindustry/world/blocks/distribution/ItemBridge.java
+++ b/core/src/mindustry/world/blocks/distribution/ItemBridge.java
@@ -176,11 +176,9 @@ public class ItemBridge extends Block{
         public void playerPlaced(Object config){
             super.playerPlaced(config);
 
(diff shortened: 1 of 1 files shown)


Anuken:
This is completely broken:

buthed010203:
Ah thats what i get for testing in sandbox

buthed010203:
@Anuken I can't reproduce the top one with a completely vanilla build (aside from the one change made in this pr). I don't see the second thing there being an actual problem either, that's how it used to function and generally when people place bridges they want them to auto link

Anuken:
> I can't reproduce the top one with a completely vanilla build

You need to be out of range and place the blocks out of order.

PR #4971, opened 2021-03-22
Split parts of `hit(b, x, y)` into separate methods

MEEPofFaith:
Pros:
- Easier to read. You can tell what each part does by the method name.
- Scripting/Java modding: Instead of overriding the entirety of `hit` to change something, you can just override one of the methods instead
- `despawned` and `hit` still exist, so pre-existing scripts are unaffected.
- `despawned` no longer calls `hit`, fixing
  - `hitSound` being played twice
  - `hitEffect` and `despawnEffect` displaying at the same time
Cons:
- None that I can think of

(Feel free to suggest different names for the methods. They're kinda h)

### DIFF
--- a/core/src/mindustry/content/Bullets.java
+++ b/core/src/mindustry/content/Bullets.java
@@ -79,7 +79,7 @@ public class Bullets implements ContentList{
         }};
 
         artilleryPlastic = new ArtilleryBulletType(3.4f, 20, "shell"){{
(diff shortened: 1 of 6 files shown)


Anuken:
This breaks code by requiring bullets to define a separate despawn effect, which previous called hit(). It's not worth the changes.

PR #4976, opened 2021-03-23
Nerf block lighting, buff illuminator

MEEPofFaith:
After some discussion in balancing, we decided that illuminator is underused because everything else already creates enough light that there's no reason to build them.
image
This weakens light produced by other sources, and buffs illuminator light proudction.

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -221,7 +221,7 @@ public class Blocks implements ContentList{
             blendGroup = basalt;
 
             emitLight = true;
-            lightRadius = 30f;
+            lightRadius = 17f;
             lightColor = Color.orange.cpy().a(0.15f);
         }};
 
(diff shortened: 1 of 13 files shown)


MEEPofFaith:
Note: Still have no idea how to fix illuminator stopping producing light when off screen.

Anuken:
> Note: Still have no idea how to fix illuminator stopping producing light when off screen.

This is precisely why the illuminator doesn't have a bigger radius. There is no reasonable way to prevent it from clipping off the screen in 6.0.

> everything else already creates enough light

I'd rather not make everything darker. Nights are annoying enough as is.

MEEPofFaith:
Well, we need some way to give an incentive to use illuminator

Anuken:
I increased its radius in the 7.0 branch when better culling was implemented, but making *other* sources dimmer is not the right way to go.

PR #4977, opened 2021-03-23
Generate dialogue QoL

TranquillyUnpleasant:
Adds a clear button that, unsurprisingly, clears the generation filters.

This is because its very annoying to have to spam click if you want to clear the generation.

Also, adds the download icon to the "Reset to Default" button since it works with that i think and the other buttons have icons in the menu.

Also resized the buttons slightly to make the text less cramped looking, and so "Randomize" doesn't get cut on mobile.

Autogen:
image
Mapgen:
image


Quezler (approved):

Anuken:
Perhaps it should be hidden behind a submenu, with some other utilities? I remember wanting to implement a "copy/load from clipboard" button to save filters, but never got around to doing it.

TranquillyUnpleasant:
Made a few changes as per the suggestion:

The main buttons are now 4 buttons, randomize i feel is better in the menu as players will probably be using it a lot to get a good random distribution.
image

The edit dialogue is either this:
image
Or this:
image
Depending on the generation mode.

I also changed the styles of both dialogues to match the logic and schematic dialogues, here is the add dialogue for reference:

TranquillyUnpleasant:
Unfortunately couldn't find an easy way to add import/export functionality to generation as that needs some custom parsing from what i can see, and i don't know the filters well enough to add that.

Anuken:
Sorry, I already implemented something similar (but with an additional copy/load from clipboard submenu) shortly after my comment yesterday.

PR #4978, opened 2021-03-23
Colorize the inner corners of other conduit sprites

Quezler:
> class #4978 extends #4015 implements `morecorners` {

Screen Shot 2021-03-23 at 12 10 06
Screen Shot 2021-03-23 at 12 14 08

> }


Quezler:
originally i actually made this pull to prevent this issue with the conduit cap, but it ran into some issues with the cap near corners since they were inconsistent: (now that they're all equal i can extend the cap on both sides by a row of pixels) 🎨 

Screen Shot 2021-03-23 at 11 51 47
Screen Shot 2021-03-23 at 12 30 56

bvcbvc:
h

PR #4979, opened 2021-03-23
Move capping code from armored to normal conduits

Quezler:
this pull was mainly to get the check outside the draw function and into the proximity updater, but decided i might as well just move it into the base conduit class, capping behavior is separated from leaking and controlled by the sprite region existing:

Screen Shot 2021-03-23 at 13 21 14

> just a pic that shows its all still visibly working as intended ^ (the sprites were still packed from #4978)


Erisfiregamer1:
the name of the branch is ridiculous

Quezler:
https://user-images.githubusercontent.com/3179271/112450759-0032d500-8d55-11eb-87b7-74f68b306f32.mp4

PR #4983, opened 2021-03-24
Add assetsJar artifact for core

way-zer:
Add artifact output: core.assets.jar

This artifact can provider dependence for projects like coreBot, that don't run the game but need assets for some work.


Anuken:
Seems strange to package *assets* without music or sounds.

way-zer:
> Seems strange to package _assets_ without music or sounds.

May add another artifact. Because music and sounds are so large.

PR #4989, opened 2021-03-25
Make units not shoot when idle

MEEPofFaith:
https://user-images.githubusercontent.com/54301439/112408736-932c2a80-8cd5-11eb-9e5f-6e19f85c03b3.mp4

---
(Should this be with normal Idle or should a 4th command center option be added?)


Anuken:
Neither. I see no good reason for units not to shoot when idle.

Quezler:
@MEEPofFaith it can still be done by a server sided plugin or mod, just apply the disarmed status effect (keep duration & unit network sync time in mind to prevent coverage gaps)

PR #4995, opened 2021-03-25
Customizable timescale per team?

Quezler:
just a questionable toy for plugins, being able to speed up (or even slow down) certain teams, some potential for pvp when teams are unbalanced, "map wide" overdrive projector coverage without abusive tricks, but calling `rules.teams.get(team)` for each block each tick might be a performance hit? i'll just leave it here as a draft for others / anuke to comment on for now 🤔 

> doesn't get subtracted from how much overdrive projectors boost, but if the number is higher they are just decoration


joshuaptfan:
Probably a good example of "just because you can, doesn't mean you should".

Bohaska:
i don't think we need this... at least for now... just my opinion though

Kieaer:
I think this would be fun

PR #5033, opened 2021-03-31
`firstSpawnDelay` + fix `spawnX` being mirrored

MEEPofFaith:
Fixes `spawnX` being mirrored. (Positive x would be on the left instead of the right)
Code_KUdzHEexkM
javaw_JigaLsD0jv

---

Adds `firstSpawnDelay` to stagger the spawning of spawns of the same `spawnTime`.

https://user-images.githubusercontent.com/54301439/113111598-304aff80-91bd-11eb-921c-b9f77930ec05.mp4


Anuken:
On second thought, it doesn't look great when the spawns are staggered like that.

PR #5036, opened 2021-03-31
🌸

Quezler:
> Screen Shot 2021-03-31 at 17 12 21

in collaboration with sk, we bring you: **flowers** (kinda started out as a joke, but it turned into something neat)

Screen Shot 2021-03-31 at 17 06 01

- shadows? yea tried those, and they sucked (not to mention clipped with the cache edges a bit)
- rotation? yea i did try steps of 45 before, but full 360 rotation randomness looked better
- modding? no idea how moddable this implementation is, generally new content is registered
- rarity? yea 0.12 might be a bit much, should probably be toned down a little, but hey
- seed? using the tile pos as a seed might make the patterns annoyingly noticeable
- ores? it was a bit to situational to hide them if the block/overlay wasn't air, and this looks fine
- detail? sprites might be a little bit overly detailed (not to mention oversized with padding) 
- +? well i just like the look of that next to where - is used
- boop? owo


Anuken:
This is neat, but on closer examination seems like something that could just be implemented with a 'flowery' grass variant. With ~8 different textures, it'll be hard to spot any repeating patterns, and no additional code would be needed.

joshuaptfan:
Also, the flowers are kinda large. If you think about the scale of a tile, these are about the size of [Rafflesia arnoldii](https://en.wikipedia.org/wiki/Rafflesia_arnoldii).

Anuken:
~~mutant alien flowers?~~

TranquillyUnpleasant:
> This is neat, but on closer examination seems like something that could just be implemented with a 'flowery' grass variant. With ~8 different textures, it'll be hard to spot any repeating patterns, and no additional code would be needed.

An attempt to do such:
image
Might need a different variant scattering algorithm because it becomes a but much if you have a lot of variants. This is with 6 extra flower variants of grass

sk7725:
> With ~8 different textures

I should sprite them soon...

bvcbvc:
this looks really nice, i hope it finds its way into the game

Anuken:
This isn't really going anywhere, so I'll close it for now.

PR #5041, opened 2021-04-01
Can do Schematics select Edit such kind of operation after dead or while using unit can't build.

shugen002:
In some PvP server , player want to save schematics after dead or while observing competition. So i do some adjust for that.

### DIFF
--- a/core/src/mindustry/graphics/OverlayRenderer.java
+++ b/core/src/mindustry/graphics/OverlayRenderer.java
@@ -17,7 +17,7 @@ import static mindustry.Vars.*;
 
 public class OverlayRenderer{
     private static final float indicatorLength = 14f;
-    private static final float spawnerMargin = tilesize*11f;
+    private static final float spawnerMargin = tilesize * 11f;
     private static final Rect rect = new Rect();
 
     private float buildFade, unitFade;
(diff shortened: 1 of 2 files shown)


core/src/mindustry/graphics/OverlayRenderer.java
```diff
@@ -26,13 +26,15 @@
...
+        input.drawBottom();
+            
         if(player.dead()) return;
 
         if(player.isBuilder()){
             player.unit().drawBuildPlans();
         }
 
-        input.drawBottom();
+
```
joshuaptfan, line 37: Extraneous line break.

shugen002:
Not all files in this project seems formated correctly as the file in .github/Mindustry-CodeStyle-IJ.xml . 
And I think we need someone to make a format check action for this project.

Anuken:
The formatting config IntelliJ provides doesn't handle all the cases. Sometimes, I like to split statements into multiple indented lines, or (in very specific situations) omit spaces around operators.

shugen002:
> The formatting config IntelliJ provides doesn't handle all the cases. Sometimes, I like to split statements into multiple indented lines, or (in very specific situations) omit spaces around operators.

i agree that. but at least it can help us avoid some changes just some spaces add and have a nice code history.

shugen002:
Is there any other problem in this PR except the format action request?
The format action request can be discuss later in the issue or discord.

shugen002:
close due to too long time.

Anuken:
The issue with letting dead players poll input is that it has all sorts of potential side effects - the player may try to unload items or queue blocks to a non-existent unit.

PR #5049, opened 2021-04-03
Add Kotlin Support

blahblahbloopster:
This PR makes the annotations module use the kotlin annotation processor, kapt, instead of the java one.  This enables kotlin code to be put in the core module, as well as speeds up compilation time to only a few seconds if no code in core has been changed (so not often).
NOTE: This breaks tests because it resets entity revisions to 0.  I'm not sure how to fix this because I haven't worked with revisions.

### DIFF
--- a/annotations/src/main/java/mindustry/annotations/impl/AssetsProcess.java
+++ b/annotations/src/main/java/mindustry/annotations/impl/AssetsProcess.java
@@ -20,8 +20,8 @@ public class AssetsProcess extends BaseProcessor{
 
     @Override
     public void process(RoundEnvironment env) throws Exception{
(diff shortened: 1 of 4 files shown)


sk7725:
Would this bteak existing java mods?
(@GlennFolker please test pu and bm)

blahblahbloopster:
It shouldn't cause any mod issues

sk7725:
so, does it not actually cause issues?

blahblahbloopster:
I haven't tested it with mods but it doesn't really change anything accessible to mods

sk7725:
> I haven't tested it with mods but it doesn't really change anything accessible to mods

better safe than sorry

Anuken:
Merged into kotlin-support branch, will be merged once/if I fix the dex issues.

PR #5050, opened 2021-04-03
Bullets shouldn't home on what they've already collided with.

MEEPofFaith:
Why home in on something when you can't damage them anyways.


TheRealIvyX:
this should only apply to non-piercing homing bullets imo since piercing bullets can hit the same thing several times

MEEPofFaith:
They can, but they cannot damage the same thing twice.

MEEPofFaith:
`collision` will not be called on the same target twice

bvcbvc:
yea, the only reason this was done was because v5 scorch

PR #5060, opened 2021-04-05
Big unit light radius

BlueWolf3682:
Before: 
Screenshot_20210405-035822

After: 
Screenshot_20210405-034627

Really minor change, but ✨ polishing™ ✨ 
(Let's just hope it went well, I'm not a java fanatic)


BlueWolf3682:
This has been done in 7.0, this is now unnecessary.

PR #5062, opened 2021-04-05
Reduce fire particle count

buthed010203:
Currently, the particle count spirals out of control very easily, this should *somewhat* fix that


buthed010203:
This probably won't be merged considering it's already fixed in v7 but if another v6 update is ever pushed, this could actually get merged so I'm gonna leave it open

Anuken:
Nah, this will just cause conflicts in the future, so I'll close it.

PR #5068, opened 2021-04-06
Allow for block inventory as well as config

buthed010203:
Can now see items in mass drivers and bridges, items can be added or removed from them directly. Also added an extra button to the core in campaign that allows for quick selection of launchpad destinations (select a sector which well send items to the current one).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,6 +1,6 @@
 name: Pull Request Tests
 
-on: [pull_request]
+on: [pull_request, workflow_dispatch]
 
 jobs:
   buildJava14:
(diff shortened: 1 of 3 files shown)


joshuaptfan:
In-action screenshots?

buthed010203:
> In-action screenshots?


@joshuaptfan https://vexera.is-ne.at/UBGkAl.webm https://yagpdb.is-ne.at/E9TVMz.webm https://mee6.is-terrible.com/khvx7k.webm That should be more than enough heh, it has some flaws but it works fine overall

joshuaptfan:
Ok, my concern was (un)linking adjacent bridges on the right, but it seems easy to do at high zoom.

buthed010203:
> Ok, my concern was (un)linking adjacent bridges on the right, but it seems easy to do at high zoom.

Alternatively you can just double click the bridge and the inventory will disappear (the link will also reset but if that is what you were trying to do, its faster and if you were just trying to link it then it doesnt even matter that it was relinked)

PR #5077, opened 2021-04-07
Refactored duplicated tests into a parameterized one

eas5:
**Problem:**
Classes containing test methods with repeated test code steps, typically, parts that set up test fixtures, generate additional effort in case of test maintenance.

**Solution:**
Parameterized tests make it possible to run a test multiple times, with different arguments, as individual and independent tests. In this refactoring, no original assertion parameter was changed. A further refactoring can also be applied as soon as JUnit 5 library version is updated to at least version 5.7, which is indicated in test class.

### DIFF
--- a/tests/src/test/java/IOTests.java
+++ b/tests/src/test/java/IOTests.java
@@ -3,6 +3,8 @@ import arc.util.io.*;
 import mindustry.game.*;
 import mindustry.io.*;
 import org.junit.jupiter.api.*;
+import org.junit.jupiter.params.*;
+import org.junit.jupiter.params.provider.*;
 
 import java.nio.*;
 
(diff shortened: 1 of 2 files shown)


tests/src/test/java/IOTests.java
```diff
@@ -3,29 +3,32 @@
 import mindustry.game.*;
 import mindustry.io.*;
 import org.junit.jupiter.api.*;
+import org.junit.jupiter.params.ParameterizedTest;
```
Anuken, line 6: Use wildcard imports here ([CONTRIBUTING.MD](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md#follow-the-formatting-guidelines)).

tests/src/test/java/IOTests.java
```diff
@@ -3,29 +3,32 @@
...
+    void writeStringTest(String string){
         ByteBuffer buffer = ByteBuffer.allocate(500);
-        TypeIO.writeString(buffer, "这个服务器可以用自己的语言说话");
+        TypeIO.writeString(buffer, string);
         buffer.position(0);
-        assertEquals(TypeIO.readString(buffer), "这个服务器可以用自己的语言说话");
+        assertEquals(TypeIO.readString(buffer), string);
     }
 
+    /**
```
Anuken, line 27: Any particular reason why you can't just update the libraries to 5.7 instead?

tests/src/test/java/IOTests.java
```diff
@@ -3,29 +3,32 @@
...
-        ByteBuffer buffer = ByteBuffer.allocate(500);
-        TypeIO.writeString(buffer, "asd asd asd asd asdagagasasjakbgeah;jwrej 23424234");
-        buffer.position(0);
-        assertEquals(TypeIO.readString(buffer), "asd asd asd asd asdagagasasjakbgeah;jwrej 23424234");
-    }
-
-    @Test
-    void writeChinese(){
+    @ParameterizedTest
+    @ValueSource(strings ={
```
Anuken, line 16:
```suggestion
    @ValueSource(strings = {
```

Anuken (changes requested):

tests/src/test/java/IOTests.java
eas5: Done

tests/src/test/java/IOTests.java
eas5: No particular reason. I have updated the libraries

PR #5080, opened 2021-04-08
Copy ItemStack functions over to LiquidStack

MEEPofFaith:


Anuken:
so I went to use some of these, and it turns out half of them are completely broken, you're treating liquid amounts as ints when they should be floats

MEEPofFaith:
lmao

PR #5082, opened 2021-04-08
Boss locator with spore color

DoctorFuchs:
This update it means if there is a boss, the boss has a extra line that stroke to it. The color is the spore color.
(Sorry for my bad English)

This pull request solves this [issue](https://github.com/Anuken/Mindustry-Suggestions/issues/2268) But it have no settings, because a boss/guardian is very rare.

### DIFF
--- a/core/src/mindustry/graphics/OverlayRenderer.java
+++ b/core/src/mindustry/graphics/OverlayRenderer.java
@@ -14,6 +14,7 @@ import mindustry.ui.*;
 import mindustry.world.*;
 
 import static mindustry.Vars.*;
+import static mindustry.graphics.Pal.spore;
 
 public class OverlayRenderer{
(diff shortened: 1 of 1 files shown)


Quezler:
the color should probably remain the same, but maybe the shape/width/length/outline can be played around with 🤔

DoctorFuchs:
yes but Lines.stroke() takes no arguments for outline, ... . But I can look in the code for some others (like the shoot of one of the monsters) But why not. Any Ideas?

DoctorFuchs:
> the color should probably remain the same, but maybe the shape/width/length/outline can be played around with 🤔

okay I can change it maybe as a second circle further outside. I will update this.

Anuken:
If this is really necessary, I can implement it myself later - this PR has awful style&formatting.

PR #5086, opened 2021-04-09
Increase Surge Swarmer Ammo Multipler

HTMLPhoton:
Increase Surge alloy Ammo Multiplier from x2 to x5
Increase Pyratite Ammo Multiplier from x2 to x3
x2 Is far too low for a small and fast firing turret like swarmer, especially for surge alloy, which is very expensive to make. This is also different from the blast multiplier, which is at x4.
This low multiplier is actively hindering the viability of swarmer, as currently blast compound is the only viable ammunition.
This cosensus has been reach over #balancing on the Mindustry Discord and over issue #1180. Although the surge multiplier was the heat of the discussion the Pyratite multiplier is also too low, considering the multiplier for metaglass on cyclones.

PR #5088, opened 2021-04-09
Cultivator adjustments (buff)

PickAPlayerNameFirst:
Spores are pretty underrated, all the time. And so, this could be the fix for the spore pods to catch up with other items.

This issue has been discussed in issue https://github.com/Anuken/Mindustry-Suggestions/issues/2248. Production time should be decreased to 1.33/s while the water consumption is still proportional to the original one.
The power consumption however is not completely proportional (should be 96/s) but if you compare it to the oil extractors, it still consumes a lot of power. So as a slight buff, 
from 96 units/s to 84 units/s.

Effects: 
It could change the oil industries and schematics, reducing its space consumption (from 7 to 4 cultivators) and power consumption. 
The blast compound will also get a fair buff, the blast mixer ratio to cultivators will lessen (from 2:1 to 1;1)
And lastly, powercubes, but they should be built near a water source, like a hydro generators.

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1299,14 +1299,14 @@ public class Blocks implements ContentList{
         cultivator = new Cultivator("cultivator"){{
             requirements(Category.production, with(Items.copper, 25, Items.lead, 25, Items.silicon, 10));
             outputItem = new ItemStack(Items.sporePod, 1);
-            craftTime = 140;
+            craftTime = 80;
             size = 2;
             hasLiquids = true;
             hasPower = true;
             hasItems = true;
 
-            consumes.power(0.9f);
-            consumes.liquid(Liquids.water, 0.2f);
+            consumes.power(1.4f);
+            consumes.liquid(Liquids.water, 0.35f);
         }};
 
         oilExtractor = new Fracker("oil-extractor"){{


genNAowl:
Oh no another balance PR

RusLeonid:
You see, this would allow powercubes even like this 
Screenshot_20210410-031106_Mindustry to not only self-sustainable, but also generate reasonable power (25 power in this example), not even including actual well-ratioed oal centrifuge powercubes

Anuken:
Closing for the same reason as https://github.com/Anuken/Mindustry/pull/5086#issuecomment-817038833

PickAPlayerNameFirst:
> Oh no another balance PR

These suggestions came from #balancing in discord, including the surge ammo multiplier buff. 

So that the gameplay will be more effective, like a buff and nerf section.

genNAowl:
Fair enough; I guess balancing PRs are less of something to be merged and more of a way to communicate a wanted change to Anuke

PickAPlayerNameFirst:
> Fair enough; I guess balancing PRs are less of something to be merged and more of a way to communicate a wanted change to Anuke 

Nope, we wanted it to be merged. Cuz these spore losers are a crap of nonsense. And we wanted it to change.

PR #5102, opened 2021-04-13
[Net] Add support for ipForward

way-zer:


core/src/mindustry/net/Administration.java
```diff
@@ -479,6 +479,7 @@ private void load(){
         autosave("Whether the periodically save the map when playing.", false),
         autosaveAmount("The maximum amount of autosaves. Older ones get replaced.", 10),
         autosaveSpacing("Spacing between autosaves in seconds.", 60 * 5),
+        ipForward("Support ip forward behind proxy(don't enable if no proxy).", false),
```
joshuaptfan, line 482:
```suggestion
        ipForward("Support IP forward behind proxy. Don't enable if no proxy.", false),
```

core/src/mindustry/net/ArcNetProvider.java
```diff
@@ -81,6 +82,10 @@ public void received(Connection connection, Object object){
             @Override
             public void connected(Connection connection){
                 String ip = connection.getRemoteAddressTCP().getAddress().getHostAddress();
+                if(Config.ipForward.bool()){
+                    Log.debug("&bWait For IpForward: @", ip);
```
joshuaptfan, line 86:
```suggestion
                    Log.debug("&bWait For IP Forward: @", ip);
```

joshuaptfan (changes requested):
Spelling/grammar.

Darkness6030 (approved):
Yes, it is a good feature

bvcbvc:
this is going to make port fowarding easier i assume?

way-zer:
> this is going to make port fowarding easier i assume?

not port forward, It is ip forward. 


> What does this actually do? Why are these changes necessary? 

There is a proxy like bungeecord but for mindustry. The proxy has the real ip address of client and send to tell the server behind proxy.
Reuse ConnectPacket is the simplest way to support ip forward in server.
To implement this in proxy, you need to send Connect Packet(with current address) after RegisterUDP.

> How would I test this?

May custom client or simulate receive Connect Packet.

Anuken:
> There is a proxy like bungeecord but for mindustry. The proxy has the real ip address of client and send to tell the server behind proxy.
Reuse ConnectPacket is the simplest way to support ip forward in server.
To implement this in proxy, you need to send Connect Packet(with current address) after RegisterUDP.

I'm not familiar with bungeecord - what is the purpose of doing this?

> May custom client or simulate receive Connect Packet.

I don't understand what you mean by this.

way-zer:
> I'm not familiar with bungeecord - what is the purpose of doing this?

like nginx?

> May custom client or simulate receive Connect Packet.
For ipForward disable, it has no change to cureent code.
For ipForward enable, you need to send Connect Packet (mock or from custom client)

PR #5112, opened 2021-04-16
Turn MiningRange into a UnitType variable.

MEEPofFaith:
Allows for custom mining range.

### DIFF
--- a/core/src/mindustry/ai/types/MinerAI.java
+++ b/core/src/mindustry/ai/types/MinerAI.java
@@ -63,7 +63,7 @@ public class MinerAI extends AIController{
                 return;
             }
 
-            if(unit.within(core, unit.type.miningRange)){
+            if(unit.within(core, unit.type.range)){
                 if(core.acceptStack(unit.stack.item, unit.stack.amount, unit) > 0){
                     Call.transferItemTo(unit, unit.stack.item, unit.stack.amount, unit.x, unit.y, core);
                 }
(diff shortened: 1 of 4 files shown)


core/src/mindustry/ai/types/MinerAI.java
```diff
@@ -63,7 +63,7 @@ protected void updateMovement(){
                 return;
             }
 
-            if(unit.within(core, unit.type.range)){
+            if(unit.within(core, unit.type.miningRange)){
```
Anuken, line 66: The changes on this line and the one below don't seem necessary - mining range should not affect core deposit & circling radius.

core/src/mindustry/entities/comp/MinerComp.java
```diff
@@ -36,7 +36,7 @@ boolean mining(){
     }
 
     public boolean validMine(Tile tile, boolean checkDst){
-        return !(tile == null || tile.block() != Blocks.air || (!within(tile.worldx(), tile.worldy(), miningRange) && checkDst)
+        return !(tile == null || tile.block() != Blocks.air || (!within(tile.worldx(), tile.worldy(), this.<Unit>self().type().miningRange) && checkDst)
```
Anuken, line 39: `@Import UnitType type;` instead of `this.<Unit>self().type()`

Anuken (changes requested):

PR #5115, opened 2021-04-16
Make some limits public.

ThebestkillerTBK:


Anuken:
Values outside of these limits break the game.

bvcbvc:
yea i dont think we need negative unit cap anytime soon

PR #5120, opened 2021-04-17
Make MenuRenderer units use Vars.content rather than hardcoded values

NiChrosia:
Adds a `calculateUnit` function to randomize the units based on Vars.content, hitSize, flying, and whether the region is error.png


AzhurniyBaklan (approved):

Drullkus (approved):

Anuken:
This idea is fine, but the selection can be implemented in one line, so I'd rather do it myself.

PR #5125, opened 2021-04-18
Convert some editor keys to keybinds

joshuaptfan:
This PR changes some editor keys to their equivalent default keybinds. No keys are changed. Resolves [suggestion #2241](https://github.com/Anuken/Mindustry-Suggestions/issues/2241), resolves [suggestion #2138](https://github.com/Anuken/Mindustry-Suggestions/issues/2138).

* Mouse left → "Select" binding
* Mouse middle → "Pick Block" binding
* Mouse right → "Break Block" binding
* Mouse scroll → "Zoom" binding

I did not want to rock the boat, but since game controls have evolved a little since editor keys were first mapped, I would prefer that editor keys were changed as follows, for consistency with game controls:

* Mouse middle (used for panning) → "Pan" binding (mouse forward)
* L-Shift (used for picking tiles) → "Pick Block" binding (mouse middle)
  * Behavior should be changed so key does not need to be held and combined with left-click to perform the action

If desired, I can amend this PR to make these changes.

### DIFF
--- a/core/src/mindustry/editor/MapView.java
+++ b/core/src/mindustry/editor/MapView.java
@@ -70,16 +70,16 @@ public class MapView extends Element implements GestureListener{
                     return false;
                 }
 
-                if(!mobile && button != KeyCode.mouseLeft && button != KeyCode.mouseMiddle && button != KeyCode.mouseRight){
+                if(!mobile && button != Core.keybinds.get(Binding.select).key && button != Core.keybinds.get(Binding.pick).key && button != Core.keybinds.get(Binding.break_block).key){
                     return true;
                 }
-                
-                if(button == KeyCode.mouseRight){
+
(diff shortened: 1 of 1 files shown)


Anuken:
> Mouse middle (used for panning) → "Pan" binding (mouse forward)
L-Shift (used for picking blocks) → "Pick Block" binding (mouse middle)

Aseprite and GIMP use the mouse middle button for panning, and alt/ctrl for picking colors. As the editor is similar to these programs in function, changing these bindings would make navigation very inconvenient and unfamiliar.

joshuaptfan:
Ok, the changes I suggested are not in the PR, so it should be acceptable.

Anuken:
Looking at these bindings again, I don't agree with the choices of controls. Picking a block is very different from panning or moving the camera. If this is ever to be re-opened, I think all of these should be entirely separate bindings.

PR #5141, opened 2021-04-22
Block Fire Damage Variable

MEEPofFaith:
In case you want some blocks to take more or less damage from fire than usual.

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -109,6 +109,8 @@ public class Block extends UnlockableContent{
     public int health = -1;
     /** base block explosiveness */
     public float baseExplosiveness = 0f;
+    /** how much damage fire does per hit */
+    public float fireDamage = 1.6f;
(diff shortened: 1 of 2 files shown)


Quezler:
boop

core/src/mindustry/world/Block.java
```diff
@@ -109,6 +109,8 @@
     public int health = -1;
     /** base block explosiveness */
     public float baseExplosiveness = 0f;
+    /** how much damage fire does per hit */
```
Quezler, line 112:
```suggestion
    /** how much damage fire does per hit. */
```

Anuken:
This will cause conflicts with 7.0 fire code, so I can't merge this.

Legomonster33:
stuff should take more fire damage if it has flammable in it, and less if it has coolants

MEEPofFaith:
I can recreate a working non-conflicting version after 7.0 is merged.
Or you could do it yourself.

PR #5178, opened 2021-05-01
Better Item Consuption Display

MEEPofFaith:
Overhauler from Exotic Mod:
many
Vanilla is still ok:
single

### DIFF
--- a/core/src/mindustry/world/consumers/ConsumeItems.java
+++ b/core/src/mindustry/world/consumers/ConsumeItems.java
@@ -39,7 +39,10 @@ public class ConsumeItems extends Consume{
             for(ItemStack stack : items){
                 c.add(new ReqImage(new ItemImage(stack.item.icon(Cicon.medium), stack.amount),
                 () -> tile.items != null && tile.items.has(stack.item, stack.amount))).padRight(8);
-                if(++i % 4 == 0) table.row();
+
+                if(i++ % 5 == 4){
+                    c.row();
+                }
             }
         }).left();
     }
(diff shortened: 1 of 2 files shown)


MEEPofFaith:
consuption :flushed:

oinkymoo:
what seems to be the issue? only unit buildings normally have build time and the unit count?

core/src/mindustry/world/consumers/ConsumeItems.java
```diff
@@ -39,7 +39,10 @@ public void build(Building tile, Table table){
             for(ItemStack stack : items){
                 c.add(new ReqImage(new ItemImage(stack.item.icon(Cicon.medium), stack.amount),
                 () -> tile.items != null && tile.items.has(stack.item, stack.amount))).padRight(8);
-                if(++i % 4 == 0) table.row();
+
+                if(i++ % 5 == 4){
```
Anuken, line 43: Why is the math here different?

MEEPofFaith: the previous math did not create new rows

Anuken: There's nothing wrong with the math (`++i % 4 == 0`), only the table needs to be changed.

MEEPofFaith: If there's nothing wrong with the math, why doesn't it create new rows?
image

I checked the v1.26.2 tag, it has that math on it.

It could be because the `++` was before the variable?

Whatever the case, the equation is like that of the one in ItemSelection.

Anuken: > If there's nothing wrong with the math, why doesn't it create new rows?

Because there's a typo, it creates rows for the **parent** table, not the table that actually has the icons in it.

Anuken:
*will implement myself due to potential conflicts*

MEEPofFaith:
👍

PR #5182, opened 2021-05-02
Add missed scarlet color to vote command response

skykatik:
This color is used in all responses :thinking:


Darkness6030 (approved):
Oh yes

PR #5186, opened 2021-05-03
Fix a ui bug on desktop

ThebestkillerTBK:
The buttons on menu will be weird if you try to put more than 8 buttons.

Before
java_vEGWaGlPhv

After
java_NXB18kjEAk

### DIFF
--- a/core/src/mindustry/ui/fragments/MenuFragment.java
+++ b/core/src/mindustry/ui/fragments/MenuFragment.java
@@ -144,20 +144,28 @@ public class MenuFragment extends Fragment{
             }).colspan(2);
         }
     }
-
+	private float getMenuTableHeight(){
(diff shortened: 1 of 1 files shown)


Quezler:
this is probably a horrible solution, and anuke would likely want to do it himself if he ever wants more buttons on the menu 🤔

ThebestkillerTBK:
This can be a temporary solution before anuke done the ui.

Quezler:
your code currently looks like a mess, so i doubt he will use it as a temporary solution 😅

sk7725:
.row()

Anuken:
> This can be a temporary solution before anuke done the ui.

There is no need for any solution at all. This is not a bug; the game doesn't **have** more buttons in the menu. If you make a mod that adds more, it is your responsibility to get it working properly *in the mod*.

PR #5188, opened 2021-05-03
Update Contributors

123e443:
Added my name because translating to greek.


WilloIzCitron:
mention a pull which you contributed?

no pull no proof

PR #5192, opened 2021-05-04
Fix small linecast issue

MEEPofFaith:
With the current code, if unit and building aren't null, and the unit isn't closer than the building, the `return` is skipped, then it'll return unit anyways. The change now checks for building instead of unit since the default return is unit.


MEEPofFaith:
I fix an error in my fix.

PR #5201, opened 2021-05-05
Update thedimas hub ip

thedimas3007:
Sry for doing this again, but server was migrated to new, more powerfull host


thedimas3007:
thx <3

PR #5207, opened 2021-05-06
4 new Application Tests added

andrew3737:
liquidRouterOutputAll verifies that liquid routers output form the other three sides that the input came from.
sorterOutputAll verifies that wanted item goes straight through the sorter whereas the unwanted items are sorted out the sides of the sorter.
routerOutputAll verifies that routers output form the other three sides that the input came from.
junctionOutputAll verifies that items can be passed on conveyors under and over each other through a junction without being mixed.


tests/src/test/java/ApplicationTests.java
```diff
@@ -290,6 +290,120 @@ void liquidJunctionOutput(){
...
+        router.setBlock(Blocks.liquidRouter, Team.sharded);
+        conduitLeft.setBlock(Blocks.conduit, Team.sharded,2);
+        conduitUp2.setBlock(Blocks.conduit, Team.sharded, 1);
+        conduitRight.setBlock(Blocks.conduit, Team.sharded, 0);
+        leftTank.setBlock(Blocks.liquidTank, Team.sharded);
+        topTank.setBlock(Blocks.liquidTank, Team.sharded);
+        rightTank.setBlock(Blocks.liquidTank, Team.sharded);
+
+        updateBlocks(200);
+        assertEquals(true, rightTank.build.liquids.currentAmount() > 0);
```
LeoSko, line 313: Why not simply `assertTrue`?

andrew3737: Anuken requested to use assert equals for comparisons

andrew3737: assertTrue is more refined in this case though, I'm going to make an adjustment here.

andrew3737: Let me know what you think of the adjustment.

LeoSko: I'd just wait for Anuken to see. The change is too minor to think much about.

tests/src/test/java/ApplicationTests.java
```diff
@@ -290,6 +290,119 @@ void liquidJunctionOutput(){
...
+        junction.setBlock(Blocks.junction, Team.sharded);
+
+
+        vault1.setBlock(Blocks.vault, Team.sharded);
+        vault2.setBlock(Blocks.vault, Team.sharded);
+
+        updateBlocks(200);
+        assertEquals(Items.coal, vault1.build.items.first());
+        assertEquals(Items.copper, vault2.build.items.first());
+        assertNotEquals(Items.copper, vault1.build.items.first());
```
Anuken, line 402: These two `assertNotEquals` are redundant. If the items are already known to be coal and copper from the above assertions, they can't possibly be anything else.

tests/src/test/java/ApplicationTests.java
```diff
@@ -290,6 +290,119 @@ void liquidJunctionOutput(){
...
+        source1.build.configureAny(Items.coal);
+        source2.setBlock(Blocks.itemSource, Team.sharded);
+        source2.build.configureAny(Items.copper);
+        conveyor1.setBlock(Blocks.conveyor, Team.sharded, 1);
+        conveyor2.setBlock(Blocks.conveyor, Team.sharded, 2);
+        conveyor3.setBlock(Blocks.conveyor, Team.sharded, 1);
+        conveyor4.setBlock(Blocks.conveyor, Team.sharded, 2);
+
+        junction.setBlock(Blocks.junction, Team.sharded);
+
```
Anuken, line 394: unnecessary newline?

tests/src/test/java/ApplicationTests.java
andrew3737: I will make some updates here

tests/src/test/java/ApplicationTests.java
andrew3737: I'll clear that up for you!

andrew3737: I'll clear that up for you!

tests/src/test/java/ApplicationTests.java
andrew3737: I see what you mean, I will update the tests.

andrew3737:
I updated the files in this most recent commit. I change the names of two of the tests to better reflect what their intention is, as well as removed the extra newline and redundant assertEquals statements.

PR #5208, opened 2021-05-06
Launchpad progress bar

pyokken:
Small QOL improvement that adds a progress bar for launchpads. Based on the unit factory and reconstructor progress bars.

launchpad-progress-bar


Anuken:
This is unnecessary, launch pads already have progress indicated with lights on the sides.

pyokken:
Yes, but the progress lights are hard to see when zoomed out and not as easy to read at a glance. Maybe I'm alone in this, but as a new player especially I wanted a progress bar to easily see what the launchpad was doing.

Anuken:
> Maybe I'm alone in this, but as a new player especially I wanted a progress bar to easily see what the launchpad was doing.

Why? It launches at regular intervals, and *when exactly the launch happens* isn't all that important. Another progress bar just feels like bloat.

oinkymoo:
i didn't even know that was a thing

oinkymoo:
> * when rushing to launch resources to a defending sector, you can easily see how long you have to wait for the next launch (and do other things in the meantime)

it's 20 seconds per launch if its always filled, that's really not enough to do things in the meantime

PR #5210, opened 2021-05-06
Logic sensor for unit item/payload timer

BenMan95:
Adds the logic sensor values `@itemTimer` and `@payTimer`, which sense the amount of time until a unit can take or drop items/payloads with logic.
Based on this suggestion I made earlier: [https://github.com/Anuken/Mindustry-Suggestions/issues/2394](https://github.com/Anuken/Mindustry-Suggestions/issues/2394)

### DIFF
--- a/core/src/mindustry/entities/comp/UnitComp.java
+++ b/core/src/mindustry/entities/comp/UnitComp.java
@@ -146,8 +146,8 @@ abstract class UnitComp implements Healthc, Physicsc, Hitboxc, Statusc, Teamc, I
                     controller instanceof FormationAI ? ctrlFormation :
                     0;
             case commanded -> controller instanceof FormationAI && isValid() ? 1 : 0;
-            case itemTimer -> controller instanceof LogicAI && isValid() ? ((LogicAI)controller).itemTimer : 0;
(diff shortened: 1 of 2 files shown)


core/src/mindustry/entities/comp/UnitComp.java
```diff
@@ -146,6 +146,8 @@ public double sense(LAccess sensor){
                     controller instanceof FormationAI ? ctrlFormation :
                     0;
             case commanded -> controller instanceof FormationAI && isValid() ? 1 : 0;
+            case itemTimer -> controller instanceof LogicAI && isValid() ? ((LogicAI)controller).itemTimer : 0;
+            case payloadTimer -> controller instanceof LogicAI && isValid() ? ((LogicAI)controller).payTimer : 0;
```
summetdev, line 150:
```suggestion
            case itemTimer -> controller instanceof LogicAI ai && isValid() ? ai.itemTimer : 0;
            case payloadTimer -> controller instanceof LogicAI ai && isValid() ? ai.payTimer : 0;
```

Anuken:
The logic/item timer is mostly an implementation detail to prevent spam, I don't want sensors for it.

BenMan95:
Still. it would be nice to have at least some way to check if a unit can move items/payloads. Maybe instead, there could be a boolean check to see if the timer has expired?

PR #5220, opened 2021-05-07
Accurate Lancers

MEEPofFaith:
~~This can't possibly be unbalanced.~~

https://user-images.githubusercontent.com/54301439/117496248-a0bb0e00-af2b-11eb-9ae5-6917273a6e0f.mp4

(Daggers made speedy for demonstration purposes.)

It can still be inaccurate if it aims at something, then that something dies ~~which happens a lot in the video.~~

Hehe I'ma just sit here and watch the comments go chaotic.

image

### DIFF
--- a/core/src/mindustry/entities/Predict.java
+++ b/core/src/mindustry/entities/Predict.java
@@ -64,6 +64,16 @@ public class Predict{
         return intercept(src.getX(), src.getY(), dst.getX(), dst.getY(), ddx, ddy, v);
     }
 
+    public static Vec2 instant(Position t, float delay){
+        float ddx = 0, ddy = 0;
+        if(t instanceof Hitboxc h){
+            float mul = Math.max(delay, 0f);
+            ddx += h.deltaX() * mul;
+            ddy += h.deltaY() * mul;
+        }
+        return vresult.set(t).add(ddx, ddy);
+    }
+
     /**
      * See {@link #intercept(float, float, float, float, float, float, float)}.
      */
--- a/core/src/mindustry/world/blocks/defense/turrets/Turret.java
+++ b/core/src/mindustry/world/blocks/defense/turrets/Turret.java
@@ -213,9 +213,11 @@ public class Turret extends ReloadTurret{
             BulletType bullet = peekAmmo();
             float speed = bullet.speed;
             //slow bullets never intersect
-            if(speed < 0.1f) speed = 9999999f;
-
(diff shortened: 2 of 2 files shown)


OnyxFinger:
actually good lancers..
lancers were usable already though, I prefer to see them more as a situational gun, it's just better in general : learning what to use in what situations is a better way of doing things, especially since knowledge in this game is informational for most.
also it would totally not be unbalanced INDDDEEEEEEEEEEEEEEEEEEEDDD SIR

Anuken:
While this seems nice in theory, I don't like the balance implications.

OnyxFinger:
balancing is hard and lancers are pretty much find as they are right now in my opinion, no need to touch them.

MEEPofFaith:
¯\_(ツ)_/¯

Anarchy.

Anuken:
at the same time, balancing something by making it stupid is not good design

OnyxFinger:
> Well, lancer doesn't really need balance. I created this more of to see how it would be, knowing there will be people who like it and people who don't.

and you placed it in a place where you're supposed to report bugs / send suggestions or talk about balance :/

MEEPofFaith:
eh, let the opinions roll in

WilloIzCitron:
but lancer cant aim crawler because crawler is **TOO FAST**

sk7725:
I think that there should be a way to utilize this behavior in logic.

notrealn:
Considering that this could possibly be implemented with logic, I think merging this would make it more accessible for everyone.

LeoSko:
Please yes.

Logic is better to support strategy decisions (e.g. choosing to attack enemy with highest HP) and not tactics where you make something inaccurate - accurate, something with low range - high range or something stupid - smart.

As a potential balance impact - there is practically none, since you previously could place lancers in such a way that would still make it hit fast units, but now you just won't have this headache of finding correct place/angle.

OnyxFinger:
Well yes. It would mean lancers need a damage nerf however, their innacuracy is their only issue (they can't hit air but you can just place scatters or another AA gun and woo, now you can hit air, so it's not an issue)

MEEPofFaith:
~~Dumb people friendly~~

Markty07:
Please no

Anuken:
Implemented as an optional parameter, in case it ever becomes necessary, or a mod decides to change it.

MEEPofFaith:
ight

PR #5224, opened 2021-05-08
Sector import statistics

pyokken:
It's not super efficient, but it's only computed when sector statistics are displayed. If efficiency becomes an issue, this can be fixed later.

https://user-images.githubusercontent.com/83549256/117556380-48246780-b036-11eb-8195-9dda495c0317.mp4

### DIFF
--- a/core/src/mindustry/game/SectorInfo.java
+++ b/core/src/mindustry/game/SectorInfo.java
@@ -27,8 +27,6 @@ public class SectorInfo{
     public ObjectMap<Item, ExportStat> rawProduction = new ObjectMap<>();
     /** Export statistics. */
     public ObjectMap<Item, ExportStat> export = new ObjectMap<>();
-    /** Import statistics. */
-    public ObjectMap<Item, ExportStat> imports = new ObjectMap<>();
     /** Items stored in all cores. */
     public ItemSeq items = new ItemSeq();
     /** The best available core type. */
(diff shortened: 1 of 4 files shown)


core/src/mindustry/game/SectorInfo.java
```diff
@@ -27,6 +27,8 @@
     public ObjectMap<Item, ExportStat> rawProduction = new ObjectMap<>();
     /** Export statistics. */
     public ObjectMap<Item, ExportStat> export = new ObjectMap<>();
+    /** Import statistics. */
+    public ObjectMap<Item, ExportStat> imports = new ObjectMap<>();
```
Anuken, line 31: Imports should be calculated on-demand when the import dialog is opened - they aren't state.

core/src/mindustry/game/SectorInfo.java
```diff
@@ -116,7 +118,7 @@ public void handleItemExport(Item item, int amount){
 
     /** Subtracts from export statistics. */
     public void handleItemImport(Item item, int amount){
-        export.get(item, ExportStat::new).counter -= amount;
+        //import statistics are handled separately
```
Anuken, line 121: Why is this here?

core/src/mindustry/game/SectorInfo.java
pyokken: Ah yeah I just did that so that we don't have to rebuild and gc the object every time. Should it be transient or just local to the method?

core/src/mindustry/game/SectorInfo.java
pyokken: As far as I'm aware, `handleItemImport` isn't used anywhere at the moment. I removed that line so that any future calls won't accidentally change export statistics when it should change import statistics. But maybe I misunderstood something here

core/src/mindustry/game/SectorInfo.java
Anuken: Local to the method. Allocating a ObjectMap when a very specific button is pressed outside of the main game is inconsequential.

Anuken:
Implemented in linked commit.

PR #5227, opened 2021-05-09
Power node update time

pyokken:
Power bar (on power nodes) doesn't take a full second to reflect the current state of the power graph. I also reduced the WindowedMean size so that the displayed power is more current (shows the past 0.25s mean rather than the past 1.0s).


Anuken:
`rawMean` is even more misleading than displaying zero, as its value tends toward zero and gradually ramps up. I don't see a good reason to reduce the average value window.

PR #5228, opened 2021-05-09
Rotate slider in the mirror filter.

TranquillyUnpleasant:
Allows applying 180 degree rotation within the mirror filter.

With mirror:
image

With rotate:
image

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -477,6 +477,7 @@ filter.option.circle-scale = Circle Scale
 filter.option.octaves = Octaves
 filter.option.falloff = Falloff
 filter.option.angle = Angle
+filter.option.rotate = Rotate
 filter.option.amount = Amount
 filter.option.block = Block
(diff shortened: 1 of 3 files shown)


TranquillyUnpleasant:
Map filter options dont have a checkbox atm

TranquillyUnpleasant:
I added a button on my local branch instead of a slider, can push if thats preferable
image

LeoSko:
@TranquillyUnpleasant, all I mean that the effect that you get with this flag is more that you *flip* the mirrored part rather than *rotate*. *Rotate* in terms of *mirroring* makes no sense to me (but I'm not native speaker so I might be wrong).

TranquillyUnpleasant:
A 180 degree rotation is literally a combination of two flips... unsure how that makes it not a rotation?

TranquillyUnpleasant:
Converted it to a checkbox, new Filter Option called ToggleOption for this purpose (name subject to change of course)

PR #5229, opened 2021-05-10
Smoother boundaries between light sources

pyokken:
Fixes the apparent dark lines between light sources. While the lines aren't actually darker, I think the lack of smoothing between light sources made them stick out.

Before:
https://user-images.githubusercontent.com/83549256/117671817-35657c00-b177-11eb-9fad-c73c92871034.mp4

After:
https://user-images.githubusercontent.com/83549256/117671849-3e564d80-b177-11eb-82ef-c23cd33034ac.mp4


pyokken:
Ahhh my mistake, I should have tested more before submitting.

PR #5232, opened 2021-05-11
`fragChance`

MEEPofFaith:
Someone in #modding-general asked if there was a `fragChance`.
I said no, then went ahead and implemented it.

0-1 chance that a bullet will create frag.

original mesage

### DIFF
--- a/core/src/mindustry/entities/bullet/BulletType.java
+++ b/core/src/mindustry/entities/bullet/BulletType.java
@@ -90,6 +90,7 @@ public abstract class BulletType extends Content{
 
     //additional effects
 
+    public float fragChance = 1f;
     public float fragCone = 360f;
     public float fragAngle = 0f;
     public int fragBullets = 9;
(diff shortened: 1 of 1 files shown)


sk7725:
if a bullet has 2 frag bullets, and .5 fragChance, it should spawn 1 bullet 50% of the time, none 25% ott, two 25% ott. It should follow a general distribution or something, i'm not a statistics man tho

MrDuck557:
I have no idea what this would be used for lmao

Anuken:
What's the point?

1ARC1:
itd make more sense tbh, having the same amount of frags every single bullet seems off when you really think about it

MEEPofFaith:
Someone asked if it was possible so I did it because it wasn't.

I guess at least one person out there would like their bullet have a chance to not create frag.

stuffyAI:
I'd much rather the frag is balanced by its type and not by a chance, like instead of giving it a 50% chance to frag just make the frag deal 50% damage...
I don't see how this is not unnecessary inconsistency

Anuken:
Eh.

MEEPofFaith:
Eh.

PR #5244, opened 2021-05-14
Add heat and elevation from Turret to PointDefence and TractorBeam

MEEPofFaith:


genNAowl:
At this rate, why not make heat/elevation fields in BaseTurret?

MEEPofFaith:
good point

Anuken:
Too stale.

PR #5252, opened 2021-05-16
Do not reset mineTile of units to null if client

LeoSko:
That allows server to put any tile into `mineTile` while client still draws it perfectly fine. Without that, client resets it regardless of server's data.
This way some ~~crazy~~ stuff can be allowed by servers if they want to, like
Анимация


oinkymoo:
can you explain what this means in layman's terms?

LeoSko:
@oinkymoo, gif pretty much explains that. Note that it can apply to custom servers, not to vanilla game.

oinkymoo:
basically, lets servers tell units to mine tiles they shouldn't

LeoSko:
Yes.

oinkymoo:
cool thank

PR #5269, opened 2021-05-18
Unit Command Radius

MEEPofFaith:
Allows for custom commanding radius.

### DIFF
--- a/core/src/mindustry/content/Fx.java
+++ b/core/src/mindustry/content/Fx.java
@@ -159,7 +159,7 @@ public class Fx{
     commandSend = new Effect(28, e -> {
         color(Pal.command);
         stroke(e.fout() * 2f);
-        Lines.circle(e.x, e.y, 4f + e.finpow() * 120f);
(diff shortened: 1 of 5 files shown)


core/src/mindustry/content/Fx.java
```diff
@@ -159,7 +159,7 @@
     commandSend = new Effect(28, e -> {
         color(Pal.command);
         stroke(e.fout() * 2f);
-        Lines.circle(e.x, e.y, 4f + e.finpow() * 120f);
+        Lines.circle(e.x, e.y, 4f + e.finpow() * e.rotation * 0.8f);
```
Anuken, line 162: This breaks the command center effect.

Anuken, line 162: It seems misleading to make the displayed radius smaller than the actual radius - why multiply?

MEEPofFaith: Well, the command radius was 150 before, but the effect would only go up to 4 + 120.

PR #5280, opened 2021-05-21
Remote Menu with callback via MenuOptionChooseEvent

LeoSko:
Please note https://github.com/Anuken/Arc/pull/62 is required for this to build.
This PR provides several things:
- Ability to get feedback from client's button presses for current `Call.infoMessage` from server, including "back/esc"
- Ability to create new menu on client from server, with feedback
- Allows to use `Call.infoMessage` with `title`

### Motivation
- Better UI for choosing, voting, answering online.
- I hate typing on mobile and I suspect some players don't even bother voting to kick griefer if they are from mobile because it takes typing out `/vote y` instead of 1 tap on screen.
- I have no idea if user *really* spent time reading my welcome screen or he is just standing afk

### Here is how it looks

General idea of menu with feedback:
menu_1

Test with different layout width:
image

### DIFF
--- a/core/src/mindustry/game/EventType.java
+++ b/core/src/mindustry/game/EventType.java
@@ -206,6 +206,17 @@ public class EventType{
         }
     }
 
+    /** Called when a player chooses an option in menu. */
+    public static class MenuOptionChooseEvent{
+        public final Player player;
+        public final int option;
+
+        public MenuOptionChooseEvent(Player player, int option){
+            this.option = option;
+            this.player = player;
+        }
+    }
+
     public static class PickupEvent{
         public final Unit carrier;
         public final @Nullable Unit unit;
--- a/core/src/mindustry/io/TypeIO.java
+++ b/core/src/mindustry/io/TypeIO.java
@@ -95,6 +95,9 @@ public class TypeIO{
         }else if(object instanceof BuildingBox b){
             write.b(12);
             write.i(b.pos);
+        }else if(object instanceof String[][] strings){
+            write.b(16);
+            writeStrings(write, strings);
         }else{
             throw new IllegalArgumentException("Unknown object type: " + object.getClass());
         }
@@ -126,6 +129,7 @@ public class TypeIO{
             case 13: return LAccess.all[read.s()];
             case 14: int blen = read.i(); byte[] bytes = new byte[blen]; read.b(bytes); return bytes;
             case 15: return UnitCommand.all[read.b()];
+            case 16: return readStrings(read);
             default: throw new IllegalArgumentException("Unknown object type: " + type);
         }
     }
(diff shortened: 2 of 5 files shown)


core/src/mindustry/game/EventType.java
```diff
@@ -206,6 +206,17 @@ public TapEvent(Player player, Tile tile){
         }
     }
 
+    /** Called when a player chooses an option in menu. */
```
Anuken, line 209: Probably should mention that this menu is the one shown in `Call.menu`

core/src/mindustry/game/EventType.java
```diff
@@ -206,6 +206,17 @@ public TapEvent(Player player, Tile tile){
...
+    public static class MenuOptionChooseEvent{
+        public final Player player;
+        public final int option;
+
+        public MenuOptionChooseEvent(Player player, int option){
```
Anuken, line 214: There's some potential ambiguity here. As far as I can tell, there's no way to identify *which menu* was used, only which option was chosen. A server may have multiple menus.

`Call.menu` should accept a unique menu ID that is returned by the player once they choose something, to make sure options are handled for the right menu.

core/src/mindustry/io/TypeIO.java
```diff
@@ -610,6 +614,35 @@ public static String readStringData(DataInput buffer) throws IOException{
...
+            write.b(strings[i].length);
+            for(int j = 0; j < strings[i].length; j++){
+                writeString(write, strings[i][j]);
+            }
+        }
+    }
+
+    public static String[][] readStrings(Reads read){
+        byte rows = read.b();
+        if(rows != -1){
```
Anuken, line 629: When would `rows == -1`?

Anuken (changes requested):

core/src/mindustry/game/EventType.java
LeoSko: I wouldn't agree with that.

Despite the fact that player can have multiple menus open at once, he can only interact with *one* and I think it is server's job to track which menu was last shown to player. It is also would be a bad UI design to show multiple menus to player from server's perspective, since they have a little transparency and hover over each other.

Anuken: Yes, multiple menus being shown at once aren't a good idea. However:

> I think it is server's job to track which menu was last shown to player

This seems like a lot of work, especially when a server may have any number of menus in it. You would have to write to a `ObjectMap<Player, Menu>` (or similar) every time a menu is opened.

In the event that a server *does* open multiple menus, everything would break in funny ways once the user started pressing buttons. Assuming that menus will always be singular doesn't seem very future-proof, and may lead to flaky behavior.

Anuken:
Implemented manually, see linked commit.

LeoSko:
@Anuken thanks for finishing this as I started to overengineer it. I might want to add ability to hide menu later *(and other legacy entities)* as well instead of timeouts, or adding a single clickable button, but that is subject of another PR*s*.

PR #5298, opened 2021-05-26
Shorten Mass Driver code by pulling some things from `Turret`

MEEPofFaith:


MEEPofFaith:
What did a do wrong... Lets see...

PR #5299, opened 2021-05-26
Fix turret shooting priority when there are multiple teams

notrealn:
before:
image

after:
image


LeoSko:
Wouldn't it make more sense to target team which is active (aka has core)?

notrealn:
im *pretty sure* im already doing that
image
this just loops over all active teams, right?

LeoSko:
@notrealn, no - activeTeams here contains all teams which have some *tile*, not some *core*.

notrealn:
this is just the logic thats already there
haha i have no idea how to do that

is there some sort of team.hascore()

notrealn:
ok i think it shouldnt matter whether or not a team has a core.
would you rather shoot a random copper wall or a neutral ripple shooting at you?

Anuken:
If you're going to check all teams, might as well check the distance too. With the current implementation, a turret further away that happens to have a lower team index will be targeted first.

notrealn:
image
woohoo it works (i think)

PR #5301, opened 2021-05-27
`Drawf.Laser` with different ends

MEEPofFaith:
`Drawf.laser`, but you can have each end be a different sprite.

Also made `TracatorBeamTurret` use this so that you can have different ended tractor lasers.

### DIFF
--- a/core/src/mindustry/world/blocks/defense/turrets/TractorBeamTurret.java
+++ b/core/src/mindustry/world/blocks/defense/turrets/TractorBeamTurret.java
@@ -20,10 +20,6 @@ import static mindustry.Vars.*;
 public class TractorBeamTurret extends BaseTurret{
     public final int timerTarget = timers++;
     public float retargetTime = 5f;
-
-    public @Load("block-@size") TextureRegion baseRegion;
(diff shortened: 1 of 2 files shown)


core/src/mindustry/world/blocks/defense/turrets/TractorBeamTurret.java
```diff
@@ -39,6 +35,11 @@
     public Sound shootSound = Sounds.tractorbeam;
     public float shootSoundVolume = 0.9f;
 
+    public @Load("block-@size") TextureRegion baseRegion;
+    public @Load("@-laser") TextureRegion laser;
+    public @Load(value = "@-laser-start", fallback = "@-laser-end") TextureRegion laserStart;
```
Anuken, line 40: What's the use case for this?

MEEPofFaith: image

Anuken: that's a very strange use case

PR #5303, opened 2021-05-27
Added Unit test

kdabiria:
Added three test cases testing the logic of the game.

### DIFF
--- a/tests/src/test/java/ApplicationTests.java
+++ b/tests/src/test/java/ApplicationTests.java
@@ -141,11 +141,20 @@ public class ApplicationTests{
         world.loadMap(testMap);
         logic.play();
         for(Teams.TeamData team : state.teams.getActive()){
-            if (team.hasCore()){
+            if(team.hasCore()){
                 assertTrue(team.team.items().has(Items.copper), "Has copper");
                 assertEquals(200, team.team.items().get(Items.copper.id),"Copper total number equal to initial items number");
             }
         }
+    }
+
+    @Test
+    void TeamsStatusTest() {
+        world.loadMap(testMap);
+        logic.play();
+        assertTrue(state.teams.isActive(state.rules.defaultTeam));
+        assertTrue(state.teams.areEnemies(state.rules.defaultTeam, state.rules.waveTeam));
+        assertFalse(state.teams.canInteract(state.rules.defaultTeam, state.rules.waveTeam));
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


tests/src/test/java/ApplicationTests.java
```diff
@@ -21,7 +21,7 @@
 import mindustry.type.*;
 import mindustry.world.*;
 import org.junit.jupiter.api.*;
-
+import mindustry.editor.*;
```
Anuken, line 24: What is this import used for?

tests/src/test/java/ApplicationTests.java
```diff
@@ -136,6 +136,40 @@ void spawnWaves(){
         assertFalse(Groups.unit.isEmpty(), "No enemies spawned.");
     }
 
+    @Test
+    void checkInitialItems() {
```
Anuken, line 140: I don't really see the purpose of this test. It may not always be the case that every map has 200 copper as a starting item, and some maps may not have any starting items at all. If I ever increase the starting copper amount to *300* copper, this test will break.

tests/src/test/java/ApplicationTests.java
```diff
@@ -136,6 +136,40 @@ void spawnWaves(){
...
+        for(Teams.TeamData team : state.teams.getActive()){
+            if(team.hasCore()){
+                assertTrue(team.team.items().has(Items.copper), "Has copper");
+                assertEquals(200, team.team.items().get(Items.copper.id),"Copper total number equal to initial items number");
+            }
+        }
+    }
+
+    @Test
+    void TeamsStatusTest() {
```
Anuken, line 152: spurious 
```suggestion
    void TeamsStatusTest(){
```

tests/src/test/java/ApplicationTests.java
```diff
@@ -235,6 +269,8 @@ void load(){
         assertTrue(state.teams.playerCores().size > 0);
     }
 
+
```
Anuken, line 272: unnecessary extra lines

tests/src/test/java/ApplicationTests.java
```diff
@@ -268,6 +304,8 @@ void liquidOutput(){
         assertTrue(world.tile(2, 1).build.liquids.current() == Liquids.water);
     }
 
+
```
Anuken, line 307: unnecessary extra lines

Anuken (changes requested):

kdabiria:
we fixed the issues you mentioned and send a new commit

PR #5306, opened 2021-05-28
FPS cap begone

MrDuck557:
Removed item and turret fps caps.
1000x boosts will work now, but do not
1.5 reload turrets will now shoot 4 times every 6 ticks, not 3


Anuken:
I won't merge this for many reasons:
- This makes significant changes to core code, there are almost certainly major bugs here.
- Spiral of death: If FPS gets low, every block will loop *more* to compensate, which will lower FPS even further, and so on.
- Very high potential for abuse or game freezes: setting certain stats to very low or very high values will freeze the game **with no obvious cause**.
- Bad formatting.
- Causes a performance decrease even for vanilla blocks at 60fps due to additional loops and calculations.
- Solves a problem that doesn't really exist in vanilla.

Alidonis:
> * Spiral of death: If FPS gets low, every block will loop _more_ to compensate, which will lower FPS even further, and so on.

why compensate with even slower methods ? the real thing to do here would be less loops. like do a stop on thinking to let fps go up again, or do less computation. im just saying.

Anuken:
> do less computation

*that's exactly what the game already does.*

PR #5307, opened 2021-05-29
Display Mass Driver range and reload stats

MEEPofFaith:
image

### DIFF
--- a/core/src/mindustry/world/blocks/distribution/MassDriver.java
+++ b/core/src/mindustry/world/blocks/distribution/MassDriver.java
@@ -16,6 +16,7 @@ import mindustry.gen.*;
 import mindustry.graphics.*;
 import mindustry.type.*;
 import mindustry.world.*;
+import mindustry.world.meta.*;
 
 import static mindustry.Vars.*;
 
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/blocks/distribution/MassDriver.java
```diff
@@ -50,6 +51,14 @@ public MassDriver(String name){
         config(Integer.class, (MassDriverBuild tile, Integer point) -> tile.link = point);
     }
 
+    @Override
+    public void setStats(){
+        super.setStats();
+
+        stats.add(Stat.shootRange, range / tilesize, StatUnit.blocks);
+        stats.add(Stat.reload, 60f / (reloadTime + 1), StatUnit.none);
```
Anuken, line 59: The +1 shouldn't be necessary, that was a turret-specific quirk.

MEEPofFaith: Interesting. I wonder why.

SMOLKEYS:
PMJava features but in vanilla mindustry

PR #5309, opened 2021-05-29
99.9% Culti

MrDuck557:
Cultivators operate at 99.9% efficiency, but it gets annoying not seeing them at 100%


oh-no-sphere:
he is right it's not the big deal but why it's just not 100 %

Anuken:
Fixed in 7.0 by making crafter warmup time linear.

PR #5311, opened 2021-05-29
Fix turret spread with even shots.

MEEPofFaith:
image

(Also casually gave charge turrets the ability to have multiple shots.)

### DIFF
--- a/core/src/mindustry/world/blocks/defense/turrets/Turret.java
+++ b/core/src/mindustry/world/blocks/defense/turrets/Turret.java
@@ -389,7 +389,9 @@ public class Turret extends ReloadTurret{
                     tr.trns(rotation, shootLength);
                     recoil = recoilAmount;
                     heat = 1f;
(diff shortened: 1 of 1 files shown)


MEEPofFaith:
`xRand` affects where the base of the shots come from.
`shotRnd` makes it so that if there's multiple shots each individual shot has their own slight randomness.

MrDuck557:
~~just so everyone knows, the mod where everything is a shotgun is [here](https://github.com/MrDuck557/everything-shotgun)~~

genNAowl:
Hmm that might of been caused by my PR here: https://github.com/Anuken/Mindustry/pull/3174

Although I guess the same problem would probably occur with the original calculation too.

MEEPofFaith:
_Angles.shotgun is better than a for loop_

sk7725:
> `xRand` affects where the base of the shots come from.
> `shotRnd` makes it so that if there's multiple shots each individual shot has their own slight randomness.

shotRand & xRand

PR #5313, opened 2021-05-30
Knockback following bullet direction

MEEPofFaith:
if `impact` is true, then the knockback direction will be the bullet rotation instead of the direction from the bullet to the unit


MEEPofFaith:
~~For some reason i didn't expect this to actually get merged~~

PR #5314, opened 2021-05-30
Fix BlockUnits saving to the wrong tile

sk7725:
The tile of the unit should be saved, not the position; while it makes little difference in vanilla, it is quite crucial for mods/plugins.

### DIFF
--- a/core/src/mindustry/io/TypeIO.java
+++ b/core/src/mindustry/io/TypeIO.java
@@ -179,7 +179,7 @@ public class TypeIO{
 
         //block units are special
         if(unit instanceof BlockUnitc){
-            write.i(((BlockUnitc)unit).tile().pos());
+            write.i(((BlockUnitc)unit).tile.pos());
         }else if(unit == null){
             write.i(0);
         }else{


sk7725:
okay what?!

Anuken:
This doesn't compile, and even if it did the result would be exactly the same. `BlockUnitc` is an interface.

> The tile of the unit should be saved, not the position

Your code still saves the tile position...?

sk7725:
> This doesn't compile, and even if it did the result would be exactly the same. `BlockUnitc` is an interface.
> 
> > The tile of the unit should be saved, not the position
> 
> Your code still saves the tile position...?

Screenshot_20210530-220452_Samsung Internet

I assume BlockUnitComp is responsible for BlockUnitc, is it not?

sk7725:
I want it to save *that* tile field, as it breaks for some blocks (for example, a "binocular" block that scrolls the player (can be achieved with routers & plugins too btw) via player.unit().set(), or some other stuff that needs to scroll the camera, like clear pipes)

Anuken:
> I want it to save *that* tile field

I don't see what you mean. There is only one.

PR #5318, opened 2021-05-31
Update servers_v6.json

uncaught1:
Changed server name in the server list


Quezler:
was moving it _up_ in the list really necessary?

uncaught1:
Hello, yes, there is a very small asset on the server

Quezler:
wut?

PR #5323, opened 2021-06-01
Add server to global list

KotMilkMeoW:
The server on which so far only survival. Then I'll add more modes.


KotMilkMeoW:
Now it's a hub

PR #5339, opened 2021-06-03
Update servers_v6.json

ulwepo:
This is a PVP server for Japanese.


Phinner:
👀 french server when ? 👀

PR #5372, opened 2021-06-05
Allow any block to have variants

MEEPofFaith:
Requested by @BlueWolf3682 in modding-general.

It compiles, and it does not crash and I have not found any issues, and looking at the code it should work.


AzhurniyBaklan (approved):
Fair change lmao

PR #5374, opened 2021-06-06
Turkish Translation 2

RebornTrack970:
Added a few V7 things and added missed translates.


RebornTrack970:
Fixed until Line 122

RebornTrack970:
Translated everything that has been missed on V6 and new things in V7!
Also, this took me a really long time.

PR #5382, opened 2021-06-06
segment is a turret, so is parallax

buthed010203:
sure it doesnt deal damage but its still a turret


buthed010203:
apparently these blocks actually have the flags now

PR #5386, opened 2021-06-06
remove stray newline

buthed010203:


Voz-Duh:
@buthed010203 if you can't make good PR don't doing PR spam

Quezler (approved):

buthed010203:
> @buthed010203 if you can't make good PR don't doing PR spam

???? What is that supposed to mean, they are two different PRs that fix two different problems it's not "pr spam"

Quezler:
anuke would prefer if you combine cleanups in the same pr if one is still unmerged ^

buthed010203:
> anuke would prefer if you combine cleanups in the same pr if one is still unmerged ^

Fair enough, will do next time

PR #5387, opened 2021-06-06
chat prefix/suffix prototype

Quezler:
throughout versions 5.0 and 6.0 my nydus server(s) have used a custom chat format, but their implementation is imperfect:

in v5 we used `Call.sendMessage(message, null, null);`
this has as a side effect that the chat message does not appear above the name of the player

in v6 we used `Call.sendMessage(message, null, player);`
this has as a side effect the entire message is duplicated above the player:

photo_2021-06-06 19 05 58

ideally there should be a way to introduce a mechanic to allow for a prefix/suffix/format without duplicate text.

in this prototype i have added a configurable prefix & suffix that currently gets sent with each chat call.
(server sided and not client sided, to allow for per player/message chat formatting, fairly important)

Screen Shot 2021-06-06 at 18 57 43

Screen Shot 2021-06-06 at 18 56 49

now this is far from perfect, i don't like the code i wrote for this _that_ much, and it feels like i am over complicating it.

for mods & plugins i have kept in some deprecated backwards compatibility, but that is currently not very ideal:
(not to mention the default is now duplicate, and while typing this i found `Config.<key>.defaultValue`)

Screen Shot 2021-06-06 at 19 11 56

for now i'll be leaving this here as a draft to gather some reactions & suggestions while i think about what i'll end up doing 🤔

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -160,11 +160,24 @@ public class NetClient implements ApplicationListener{
         clientPacketReliable(type, contents);
     }
 
-    //called on all clients
+    @Deprecated
     @Remote(targets = Loc.server, variants = Variant.both)
     public static void sendMessage(String message, String sender, Player playersender){
         if(Vars.ui != null){
             Vars.ui.chatfrag.addMessage(message, sender);
+        }
+
+        if(playersender != null){
+            playersender.lastText(message);
+            playersender.textFadeTime(1f);
+        }
+    }
+
+    //called on all clients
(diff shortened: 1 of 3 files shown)


sk7725:
I am very much against servers giving """special""" perks to """special players""", and this just seems to make such stuff so much easier and more wide-spread.

Quezler:
the vanilla code will have no support to set a per player prefix & suffix, its identical for everyone.
it only makes it _slightly_ cleaner to do it via plugins, which was always already possible.

L0615T1C5-216AC-9437:
this would be nice

PR #5388, opened 2021-06-06
Add RCR BE Server

SachaTending:


Quezler:
~~added near the top of the list 👀~~

SachaTending:
added

J-VdS:
I you want your server to be added do it at the bottom of the list! No need to bump it to the top.

PR #5402, opened 2021-06-09
allow javascript access to all java mods

catsz:
the title explains it all but basically this just makes the ModClassLoader the loader the scripts context uses


Anuken:
This **doesn't compile**, and will not work for JS in other threads. See commit.

PR #5405, opened 2021-06-09
Renamed the category and added it easyplay.su

romanin-rf:
We decided to rename the category in order to add to it EasyPlay.Su (we have an agreement)


EasyPlaySu:
Yes.

EasyPlaySu (approved):

PR #5421, opened 2021-06-13
Freedom for javascript mods

QmelZ:
Removes the script blacklist completely.

### DIFF
--- a/core/src/mindustry/core/Platform.java
+++ b/core/src/mindustry/core/Platform.java
@@ -79,8 +79,6 @@ public interface Platform{
     }
 
     default Context getScriptContext(){
-        ContextFactory.getGlobalSetter().setContextFactoryGlobal(new ContextFactory());
-
         Context c = Context.enter();
         c.setOptimizationLevel(9);
         return c;
--- a/core/src/mindustry/core/Platform.java
+++ b/core/src/mindustry/core/Platform.java
@@ -79,14 +79,7 @@ public interface Platform{
     }
 
     default Context getScriptContext(){
-        ContextFactory.getGlobalSetter().setContextFactoryGlobal(new ContextFactory(){
-            @Override
-            protected Context makeContext(){
-                Context ctx = super.makeContext();
-                ctx.setClassShutter(Scripts::allowClass);
-                return ctx;
-            }
-        });
+        ContextFactory.getGlobalSetter().setContextFactoryGlobal(new ContextFactory());
 
         Context c = Context.enter();
         c.setOptimizationLevel(9);
--- a/android/src/mindustry/android/AndroidRhinoContext.java
+++ b/android/src/mindustry/android/AndroidRhinoContext.java
@@ -62,13 +62,6 @@ public class AndroidRhinoContext{
             initApplicationClassLoader(createClassLoader(AndroidContextFactory.class.getClassLoader()));
         }
 
-        @Override
-        protected Context makeContext(){
-            Context ctx = super.makeContext();
-            ctx.setClassShutter(Scripts::allowClass);
-            return ctx;
-        }
-
         /**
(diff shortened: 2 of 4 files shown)


oinkymoo:
i would assume it exists for good reason, so i wouldn't be surprised if this gets bonked

QmelZ:
it's pointless, java can't be limited and there is an option to actually make java mods so i don't see the point of js having limits

core/src/mindustry/core/Platform.java
```diff
@@ -79,14 +79,7 @@ default Scripts createScripts(){
...
     default Context getScriptContext(){
-        ContextFactory.getGlobalSetter().setContextFactoryGlobal(new ContextFactory(){
-            @Override
-            protected Context makeContext(){
-                Context ctx = super.makeContext();
-                ctx.setClassShutter(Scripts::allowClass);
-                return ctx;
-            }
-        });
+        ContextFactory.getGlobalSetter().setContextFactoryGlobal(new ContextFactory());
```
Anuken, line 82: You don't need to set a context factory at all if there's no shutter.

Anuken:
> it's pointless, java can't be limited

Java mods give you a warning when you import them, JS mods do not.

---

Some questions that need to be answered before this is resolved:

1. Why is the sandbox a problem for you?
2. What do other mod loaders do? Forge/Fabric for MC don't do any sandboxing, but neither are official. Does TModLoader do sandboxing?
3. How easy is it to bypass the Rhino sandbox with the context bug fixed? I'm positive that there are more methods of doing so. If the security "holes" are too large, it might not be worth the maintenance effort.

QmelZ:
> Java mods give you a warning when you import them, JS mods do not.

I actually tried to somewhat add a warning but that will probably require a 2-part pull request with changes to the MindustryMods repository, i might try adding a warning for that.

> Why is the sandbox a problem for you?

I don't like the jvm languages that can be used to write mods, but i want/need to use some of the blacklisted classes in certain cases.

> What do other mod loaders do? Forge/Fabric for MC don't do any sandboxing, but neither are official. Does TModLoader do sandboxing?

TModLoader isn't official either, and it doesn't seem like there is any sandboxing.

> How easy is it to bypass the Rhino sandbox with the context bug fixed? I'm positive that there are more methods of doing so.

I haven't found any sandbox bypasses after the fix.

Anuken:
> i want/need to use some of the blacklisted classes in certain cases.

Obviously, you want to use the blacklisted classes, but *why*, and which ones?

> TModLoader isn't official either

Yes, but it's available on Steam & was officially endorsed by the developers as a "free DLC".

> it doesn't seem like there is any sandboxing.

Do you have a source for this?

---

In terms of games with built-in mod support, I know that Noita and Factorio both have "sandboxed" environments for Lua scripting; you are not allowed to access files or the network. Noita has a special "unsafe mods" setting for scripting.

QmelZ:
> but why, and which ones?

Mostly reflection and files, though i do not have specific "why" examples for those.

> Do you have a source for this?

I just looked through the source code, https://github.com/tModLoader/tModLoader, and didn't see anything related.

> Noita has a special "unsafe mods" setting for scripting.

That makes sense but is hard to implement in rhino without large issues.

l-Luna:
iirc TML just exposes regular MonoMod methods to mods, no point attempting to sandbox code that can (and is supposed to) modify other code

Anuken:
Historically, I don't think there's ever been a point where the class shutter "sandbox" has been able to stop a determined individual... even now, after the threaded shutter bug was fixed, I can think of several escape-related exploits that could be used to run arbitrary shell scripts.

It's also worth noting that the sandbox lets you use the network and make HTTP requests anyway, so tracking & remote access was always possible.

PR #5428, opened 2021-06-16
Make Impact Reactor explosion effects not hardcoded

MEEPofFaith:


MEEPofFaith:
Seems like anuke did this himself so I'll close this.

PR #5450, opened 2021-06-23
json moment

BlueWolf3682:
I'm not sure why you used json5 in json - though it *does* parse, it looks horrible.

### DIFF
--- a/core/assets-raw/sprites/pack.json
+++ b/core/assets-raw/sprites/pack.json
@@ -1,8 +1,8 @@
 {
-  duplicatePadding: true,
-  combineSubdirectories: true,
-  flattenPaths: true,
-  maxWidth: 4096,
-  maxHeight: 4096,
-  fast: true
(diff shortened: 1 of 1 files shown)


Quezler:
here is the rest of your nightmare: https://github.com/Anuken/Mindustry/tree/master/annotations/src/main/resources/revisions

BlueWolf3682:
oh come on seriously

Anuken:
That is not JSON5, it is libGDX's standard "minimal JSON" format, which is used throughout the game. The official libGDX wiki [uses this format](https://github.com/libgdx/libgdx/wiki/Texture-packer#configuration) for examples. 

I don't see any good reason to change it.

BlueWolf3682:
It just looks terrible on GitHub due to all of it erroring. On top of that, modders do this a lot and I really don't like the way it looks. But if you don't want to change it I guess I'll have to leave it be.

PR #5454, opened 2021-06-23
Allow shock mines to spawn bullets

MEEPofFaith:
https://github.com/Anuken/Mindustry-Suggestions/issues/2591

I would rename it to just `Mine` like in the suggestion but that seems unnecessary and would break stuff so I didn't.

### DIFF
--- a/core/src/mindustry/world/blocks/defense/ShockMine.java
+++ b/core/src/mindustry/world/blocks/defense/ShockMine.java
@@ -3,6 +3,7 @@ package mindustry.world.blocks.defense;
 import arc.graphics.*;
 import arc.graphics.g2d.*;
 import arc.math.*;
+import arc.util.*;
 import mindustry.annotations.Annotations.*;
 import mindustry.entities.*;
 import mindustry.entities.bullet.*;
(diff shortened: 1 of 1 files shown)


AzhurniyBaklan (approved):
wow, this is just pure genius right there

core/src/mindustry/world/blocks/defense/ShockMine.java
```diff
@@ -46,17 +49,26 @@ public void draw(){
...
             }
         }
+
+        public void triggered(){
+            for(int i = 0; i < tendrils; i++){
+                Lightning.create(team, lightningColor, damage, x, y, Mathf.random(360f), length);
+            }
+            if(bullet != null){
+                for(int i = 0; i < shots; i++){
+                    bullet.create(this, x, y, Mathf.random(360f));
```
Quezler, line 69: would dividing 360 by the amount of bullets not make more sense? 
(equal bullets in all direction, possibly angled towards the initial unit in case of just 1 bullet)

core/src/mindustry/world/blocks/defense/ShockMine.java
```diff
@@ -18,6 +19,8 @@
     public int length = 10;
     public int tendrils = 6;
     public Color lightningColor = Pal.lancerLaser;
+    public int shots = 6;
+    public BulletType bullet;
```
Anuken, line 23: must be marked as `@Nullable`

MrDuck557:
Thank you

PR #5459, opened 2021-06-25
V7 mod api limit

ThebestkillerTBK:


Quezler:
where did you get the number `127` from?

ThebestkillerTBK:
> where did you get the number `127` from?

After v126 is naturally v127

Anuken:
No. Many v6 mods should be compatible with v7.

PR #5475, opened 2021-06-25
Swap mode and mapname

Xasmedy:
I swapped the arguments because more "effective":

Now we can have a map randomized with a specific mode, but we can still randomize the map without arguments;
Now spaces are accepted inside the map argument.

As a Server manager, this change will be helpful. (Or at least more useful than before, where I could just randomize a map in survival)

### DIFF
--- a/server/src/mindustry/server/ServerControl.java
+++ b/server/src/mindustry/server/ServerControl.java
@@ -285,7 +285,7 @@ public class ServerControl implements ApplicationListener{
             info("Stopped server.");
         });
 
-        handler.register("host", "[mapname] [mode]", "Open the server. Will default to survival and a random map if not specified.", arg -> {
+        handler.register("host", "[mode] [mapname...]", "Open the server. Will default to survival and a random map if not specified.", arg -> {
             if(state.is(State.playing)){
                 err("Already hosting. Type 'stop' to stop hosting first.");
                 return;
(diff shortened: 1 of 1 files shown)


LeoSko:
https://github.com/Anuken/Mindustry/pull/4945

Anuken:
Adding [wildcard support for maps](https://github.com/Anuken/Mindustry/pull/4945#issuecomment-802673539) is a better solution that won't break existing server configs.

Xasmedy:
> Adding [wildcard support for maps](https://github.com/Anuken/Mindustry/pull/4945#issuecomment-802673539) is a better solution that won't break existing server configs.

Months passed and using wildcard still doesn't work (I haven't checked the v127).
For the server config is an easy fix for the server who needs it, and honesty who is going to make a program using `host`? As long as you don't have multiple servers instance that starts and closes automaticaly or dont need to restart the server after a crash, it won't affect much, and as I said is easy to fix. (And about breaking servers didn't you in the past did updates that destroyed a lot of plugin? Expecially with the java threads? Far more complicated to find a fix.)

Then I don't understand your logic at https://github.com/Anuken/Mindustry/pull/4945#issuecomment-800735341

I feel like you want the map to be more important than the mode, or just caring about survival as only mode.
Where's the logic in saying that without the swap as more sense? If the survival is the only mode of course it has! But it is like ignoring the others modes. And at the question "Why I need to specify a mode for a map?" As https://github.com/Anuken/Mindustry/pull/4945#issue-594343327 says, not all maps are made for survival, and don't all servers run survival, and as long as the wildcards are not added it doesn't have sense, and honesty with this easy swap you even save time in implementing the wildcards. And if it will crash after the swap? Of course if not implemented correctly, make a check if the map supports the gamemode, if no map inside the folder support it, show the user the error of not finding map with that mode, you can still use the `map.rules().mode();`. But then is your game, do what you want.

Anuken:
> Months passed and using wildcard still doesn't work (I haven't checked the v127).

Of course it doesn't work, nobody ever implemented it. That was a link to an alternative suggestion, an implementation was never mentioned.

> who is going to make a program using host

Me, for every server I've ever hosted. `config startCommand`

> Where's the logic in saying that without the swap as more sense?

I already explained the logic.

> not all maps are made for survival

But most are, as I've already pointed out. Therefore, it is a reasonable default. Nobody is claiming that *all* maps are survival.

> as long as the wildcards are not added it doesn't have sense

And why is that? Repeating your point doesn't make it true.

PR #5491, opened 2021-06-27
[TH] android and steam translation

JniTrRny:
Translate google play and steam fastlane/metadata folder. ~~`also translate main TH bundle as a bonus _(I don't want to create separate PR just for some little changes)`~~ separated it

~~`Also, I didn't translate the android changelogs, that would be a very **HUGE** workload just for some outdated changelogs that not everyone reads it.`~~ **changelogs got removed, see Anuke's commit on changelogs**

### _NOTE:_ Yes, I know. this will change in V7 and I shouldn't translate this, But v7 is just in Pre-Alpha stage, and most content aren't finished yet. so it is safe to translate, And as a base to the overhaul when v7 is finished.


JrTRinny:
WTF drunk Github

JrTRinny:
will reopen soon

PR #5493, opened 2021-06-27
thedimas v7 server

thedimas3007:


Anuken:
should be merged into [baseduser's list](https://github.com/Anuken/Mindustry/blob/master/servers_v7.json#L11)

PR #5516, opened 2021-06-30
outlaw separator desycning

Quezler:
an attempt to solve the age old problem of separators desyncing so hard that each time it happens a child in africa dies:

Screen Shot 2021-06-30 at 17 44 35

(having the `0` be `MIN_VALUE` causes issues, as well as not `- 1` ing the `MAX_VALUE`, but for the seed itself its seems fine)

### DIFF
--- a/core/src/mindustry/world/blocks/production/Separator.java
+++ b/core/src/mindustry/world/blocks/production/Separator.java
@@ -46,7 +46,12 @@ public class Separator extends Block{
         public float progress;
         public float totalProgress;
         public float warmup;
-        public int seed = Mathf.random(0, Integer.MAX_VALUE - 1);
+        public int seed;
+
+        @Override
+        public void created(){
+            seed = Mathf.randomSeed(tile.pos(), 0, Integer.MAX_VALUE - 1);
+        }
 
         @Override
         public boolean shouldAmbientSound(){
--- a/core/src/mindustry/world/blocks/production/Separator.java
+++ b/core/src/mindustry/world/blocks/production/Separator.java
@@ -45,6 +45,7 @@ public class Separator extends Block{
         public float progress;
         public float totalProgress;
         public float warmup;
+        public int seed = Mathf.random(0, Integer.MAX_VALUE - 1);
 
         @Override
         public boolean shouldAmbientSound(){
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/blocks/production/Separator.java
```diff
@@ -91,7 +93,7 @@ public void updateTile(){
                 int sum = 0;
                 for(ItemStack stack : results) sum += stack.amount;
 
-                int i = Mathf.random(sum);
+                int i = Mathf.randomSeed(seed++, 0, sum);
```
Quezler, line 96: this overflowing is fine

core/src/mindustry/world/blocks/production/Separator.java
```diff
@@ -30,6 +30,7 @@ public Separator(String name){
         solid = true;
         hasItems = true;
         hasLiquids = true;
+        sync = true;
```
Quezler, line 33: doesn't sync the seed when initially placed

core/src/mindustry/world/blocks/production/Separator.java
```diff
@@ -45,6 +46,7 @@ public void setStats(){
         public float progress;
         public float totalProgress;
         public float warmup;
+        public int seed = Mathf.random(0, Integer.MAX_VALUE - 1);
```
Quezler, line 49: see comment under screenshot

USMP-lancer:
yes please

core/src/mindustry/world/blocks/production/Separator.java
Anuken: How about making the seed depend on the tile position, then? Initialize it in `created`, with something like `seed = Mathf.randomSeed(tile.pos(), 0, Integer.MAX_VALUE - 1)`

Quezler: @Anuken your code sample looks fine & implemented as-is

Quezler, line 33: just push the implement suggestion button if you want to get rid of the syncing boolean:
```suggestion
```

Anuken: All GenericCrafters are synced, so it's probably fine with `sync = true`.

Quezler:
anything still left to discuss / test before this gets merged?

Anuken:
Looks about as deterministic as it can be, so this should be fine to merge.

PR #5520, opened 2021-07-01
Spore Rebalance

EclipsePlays:
So the current state of spore tech became worse since the initial PR was released. 

The water to spore pod ratio rigged it's production since it takes more water to produce a spore pod. 

The only way to regain the buff back is to change the values to the original ones proposed in #5088,  but with an extra change; flammability. 

If we change the flammability to 90% (100% is already good, but feels similar to coal) it should affect the power production of steam generators and can (possibly) remove the existence of direct spore pod - steam generator powercubes. 

Speaking of powercubes, adding the extra commit to Items.java is considered unnecessary since it won't bring back land powercubes, but will possibly nerf the water limited powercubes.

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1306,7 +1306,7 @@ public class Blocks implements ContentList{
         cultivator = new AttributeCrafter("cultivator"){{
             requirements(Category.production, with(Items.copper, 25, Items.lead, 25, Items.silicon, 10));
             outputItem = new ItemStack(Items.sporePod, 1);
-            craftTime = 100;
+            craftTime = 80;
             size = 2;
             hasLiquids = true;
             hasPower = true;
@@ -1320,8 +1320,8 @@ public class Blocks implements ContentList{
             drawer = new DrawCultivator();
             maxBoost = 2f;
 
-            consumes.power(80f / 60f);
-            consumes.liquid(Liquids.water, 20f / 60f);
+            consumes.power(85f / 60f);
+            consumes.liquid(Liquids.water, 24f / 60f);
         }};
 
         oilExtractor = new Fracker("oil-extractor"){{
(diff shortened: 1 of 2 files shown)


EclipsePlays:
Yes, a balance PR. Already discussed in #balancing and in https://github.com/Anuken/Mindustry-Suggestions/issues/2547.

MrDuck557:
~~while you're at it remove extra water consumption on moss tiles~~
~~nobody likes that, moss is almost useless~~

MostForgettablePerson:
Tile affinity, make sure you fix that

EclipsePlays:
> ~~while you're at it remove extra water consumption on moss tiles~~
> ~~nobody likes that, moss is almost useless~~

Welp spore mosses' only use is to make those cultivators OP for a sec, removing the extra affinity would make it absolutely useless, ~~and more of a decorative block instead~~

EclipsePlays:
> Tile affinity, make sure you fix that

Its rare to see cultivators placed on tiles that give it affinities tho
Yes it makes the cultivator itself OP, but the water extractor ratio will make it ugly.

EclipsePlays:
>but the water extractor ratio will make it ugly. 
>

Unless if there's a water source directly beside it, another rare occurrence.

itcannotbe:
https://github.com/Anuken/Mindustry/pull/5086#issuecomment-817038833

genNAowl:
Oh no your gonna screw up all my schems with cultivators

EclipsePlays:
> https://github.com/Anuken/Mindustry/pull/5086#issuecomment-817038833

It was my alt account who made that PR soo

AMindustryPlayer:
Yes, the cultivator is nerfed and the affinity is not found. Anuke should fix the affinity and he can slightly buffed a little bit if the cultivator nerfed too much

EclipsePlays:
The affinity tile will be shown when there is one in your map.

EclipsePlays:
Seems that cultivator was buffed.

AMindustryPlayer:
> Seems that cultivator was buffed.

So stop asking for buffing cultivator now and go do other tasks or any suggestions here won't help

PR #5534, opened 2021-07-03
Server qol

TranquillyUnpleasant:
Some QoL changes to the server terminal interface. Changes:
- `maps` displays custom maps only by default.
- Remove colors from map names so the name is actually readable for some maps.
The command has an additional `all` optional parameter to show default maps as well. 
The command description has been updated to reflect this.

- Color codes are removed from various places that show map names in terminal.
- Map finding function now strips both input and map name equally before comparing, making it compatible with all input/map name combinations. This change is reflected in both `host` and `nextmap` commands.

- The `help` command has an additional optional command parameter, to get help on a specific command.
The command description for help has also been updated to reflect this change.

- Look ma im a contributor!

### DIFF
--- a/server/src/mindustry/server/ServerControl.java
+++ b/server/src/mindustry/server/ServerControl.java
@@ -344,14 +344,17 @@ public class ServerControl implements ApplicationListener{
             }
         });
 
-        handler.register("maps", "Display all available maps.", arg -> {
+        handler.register("maps", "[all] [full]", "Display available maps. 'all' to include default maps. 'full' to show full map names.", arg -> {
+            boolean all = Arrays.asList(arg).contains("all");
+            boolean full = Arrays.asList(arg).contains("full");
             if(!maps.all().isEmpty()){
                 info("Maps:");
-                for(Map map : maps.all()){
+                for(Map map : all ? maps.all() : maps.customMaps()){
+                    String mapName = (full ? map.name() : Strings.stripColors(map.name())).replace(' ', '_');
                     if(map.custom){
-                        info("  @ (@): &fiCustom / @x@", map.name().replace(' ', '_'), map.file.name(), map.width, map.height);
+                        info("  @ (@): &fiCustom / @x@", mapName, map.file.name(), map.width, map.height);
                     }else{
-                        info("  @: &fiDefault / @x@", map.name().replace(' ', '_'), map.width, map.height);
+                        info("  @: &fiDefault / @x@", mapName, map.width, map.height);
                     }
                 }
             }else{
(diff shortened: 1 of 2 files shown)


TranquillyUnpleasant:
P.S. might want to look at the individual commits for less headache.

server/src/mindustry/server/ServerControl.java
```diff
@@ -259,7 +259,17 @@ protected void setup(String[] args){
     }
 
     protected void registerCommands(){
-        handler.register("help", "Displays this command list.", arg -> {
+        handler.register("help", "[command]", "Display the command list, or get help for a specific command.", arg -> {
+            if (arg.length > 0){
```
Anuken, line 263: incorrect formatting similar to this in several places, see guide

server/src/mindustry/server/ServerControl.java
```diff
@@ -259,7 +259,17 @@ protected void setup(String[] args){
...
+                Command command = handler.getCommandList().find(c -> c.text.equalsIgnoreCase(arg[0]));
+                if (command == null) {
+                    err("Command " + arg[0] + " not found!");
+                }else{
+                    info(command.text + ":");
+                    info("  &b&lb " + command.text + (command.paramText.isEmpty() ? "" : " &lc&fi") + command.paramText + "&fr - &lw" + command.description);
+                }
+                return;
```
Anuken, line 271: use an else here instead, a return would be confusing

server/src/mindustry/server/ServerControl.java
```diff
@@ -335,14 +345,17 @@ protected void registerCommands(){
             }
         });
 
-        handler.register("maps", "Display all available maps.", arg -> {
+        handler.register("maps", "[all] [full]", "Display available maps. 'all' to include default maps. 'full' to show full map names.", arg -> {
+            boolean all = Arrays.asList(arg).contains("all");
+            boolean full = Arrays.asList(arg).contains("full");
```
Anuken, line 350: I see no reason to display full map names in the console.

server/src/mindustry/server/ServerControl.java
```diff
@@ -335,14 +345,17 @@ protected void registerCommands(){
             }
         });
 
-        handler.register("maps", "Display all available maps.", arg -> {
+        handler.register("maps", "[all] [full]", "Display available maps. 'all' to include default maps. 'full' to show full map names.", arg -> {
+            boolean all = Arrays.asList(arg).contains("all");
```
Anuken, line 349: This is a bizarre way to handle arguments, `all` should be a boolean argument with false as a default, true if there's an argument and `arg[0].equals("true")`

Anuken (changes requested):

server/src/mindustry/server/ServerControl.java
TranquillyUnpleasant: Fixed

TranquillyUnpleasant:
Requested changes have been made.

PR #5544, opened 2021-07-05
Remove deconstructThreshold

Quezler:
Not needed anymore for its original purpose since boulders now deconstruct instantly, and this would return an incomplete refund if the block had building materials anyways.

I did search github for mods that seemingly used this property in any way, and so far i only came across @sk7725’s betamindy payload deconstructor.

Though i didn’t go through all the github results since like 99% are just results of mindustry forks, the damage (if any) caused by removing this value will probably be minimal.

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -223,8 +223,6 @@ public class Block extends UnlockableContent{
     public BuildVisibility buildVisibility = BuildVisibility.hidden;
     /** Multiplier for speed of building this block. */
     public float buildCostMultiplier = 1f;
-    /** Build completion at which deconstruction finishes. */
-    public float deconstructThreshold = 0f;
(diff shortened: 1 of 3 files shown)


Anuken:
I'm fine with removing the field from Prop and ConstructBlock, but if it's used even in one mod I'd rather `@Deprecate` than remove outright.

Quezler:
@sk7725 would it cost you more than 10 seconds to patch your mod?

Anuken:
The problem is that most people don't download updates for mods, so patches don't matter.

sk7725:
Please don't remove/deprecate this, I need it for block that cannot be harvested with right-click alone

sk7725:
And besides, a field that us not used ingame but one modders can mess with is always a good thing.

Anuken:
I'll remove it from Prop, but as mentioned above there isn't a good reason to remove it everywhere.

PR #5546, opened 2021-07-05
Compress PNGs, saving ~6.8MB

AeolianOwl:
Done with zopflipng. After cloning repo, I ran:
yes | find -name "*.png" -exec zopflipng -m {} {} \\;


Anuken:
Unnecessary. This pollutes git history, increases repo size due to binary diffs and makes little difference in the end - only a few large images are actually packaged with the game; most are packed into an atlas, which is not included in this repo.

AeolianOwl:
Thanks for the feedback!

Anuken:
I *did* end up optimizing one very specific set of large skybox sprites, which saved about 1mb of space. There aren't many other images worth optimizing, though.

PR #5554, opened 2021-07-08
Lets plastanium conveyor keep its highest speed

Quezler:
roombas currently bump into eachother a lot when going from an overdriven tile from one that is not, hardly looks pretty.

this prototype lets the crater/stack/roomba keep its highest boost for the rest of its life, allowing it to race to its destination.

if they bump into the back of another one the speed boost resets, this avoids "are we there yet?" bumping extravaganza.

it just feels like this makes (more) sense since its not the tile that is moving the item, but rather the entity on it.

not the most perfect solution to streamline their visuals, but its at least a step up from how they behave now. 🤔 

https://user-images.githubusercontent.com/3179271/124968331-bf21b080-e025-11eb-9b7f-de1fe2633000.mp4

### DIFF
--- a/core/src/mindustry/world/blocks/distribution/StackConveyor.java
+++ b/core/src/mindustry/world/blocks/distribution/StackConveyor.java
@@ -104,6 +104,7 @@ public class StackConveyor extends Block implements Autotiler{
         public int link = -1;
         public float cooldown;
         public Item lastItem;
+        public float velocityScl;
 
         boolean proxUpdating = false;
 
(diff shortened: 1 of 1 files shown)


Humanoid-X:
yes

LeoSko:
But doesn't overdrive actually overdrives **tile/building** and not **item** for the rest of the game? What proposed here is quite an unusual approach, I doubt someone would *expect that*.

Quezler:
@LeoSko the **tile/building** does **not** move an **item** since unlike conveyor belts there are roombas that drive on top of it, **they** will be getting the boost, the tile itself has nothing to do with it. (what good would boosting asphalt do?)

Anuken:
My take: If the *roombas* are not in range of the dome itself, they shouldn't get boosted. It's consistent with behavior of other blocks.

PR #5563, opened 2021-07-11
Multi output crafter

MEEPofFaith:
Adds `outputItems` to `GenericCrafter`. Input an array of ItemStacks and upon craft will produce everything in the array.

`outputItem` still exists so as not to break stuff.

### DIFF
--- a/core/src/mindustry/world/blocks/production/GenericCrafter.java
+++ b/core/src/mindustry/world/blocks/production/GenericCrafter.java
@@ -15,7 +15,8 @@ import mindustry.world.draw.*;
 import mindustry.world.meta.*;
 
 public class GenericCrafter extends Block{
-    public @Nullable ItemStack outputItem; //Keep for backwards compatibility
+    /** @deprecated use outputItems instead, outputItem would likely be removed eventually **/
+    public @Nullable ItemStack outputItem;
     public @Nullable ItemStack[] outputItems;
     public @Nullable LiquidStack outputLiquid;
 
--- a/core/src/mindustry/world/blocks/production/GenericCrafter.java
+++ b/core/src/mindustry/world/blocks/production/GenericCrafter.java
@@ -15,7 +15,6 @@ import mindustry.world.draw.*;
 import mindustry.world.meta.*;
 
 public class GenericCrafter extends Block{
-    /** @deprecated use outputItems instead, outputItem would likely be removed eventually **/
     public @Nullable ItemStack outputItem;
     public @Nullable ItemStack[] outputItems;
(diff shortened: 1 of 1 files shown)


BlueWolf3682:
comically large sized thank

MrDuck557:
deprecate `outputItem` to make it clear that you shouldn't use it
This opens up possibilities

MEEPofFaith:
> deprecate `outputItem` to make it clear that you shouldn't use it
> This opens up possibilities

~~I don't know how lol~~

notMEEPoffaith:
heh so uh i started this with a few pains in making a box unpacker have multiple outputs

core/src/mindustry/world/blocks/production/GenericCrafter.java
```diff
@@ -15,7 +15,9 @@
 import mindustry.world.meta.*;
 
 public class GenericCrafter extends Block{
+    /** @deprecated use outputItems instead, outputItem would likely be removed eventually **/
     public @Nullable ItemStack outputItem;
```
Anuken, line 19: `outputItem` is too convenient and too widely used to deprecated. Instead, use **only** `outputItems` in the update code, and make `outputItems` a single-item array if it is null and `outputItems != null` in `init()`.

notMEEPoffaith:
un-deprecate output item

notMEEPoffaith:
and the cat will merge basically

core/src/mindustry/world/blocks/production/GenericCrafter.java
Anuken: Also worth noting that this implementation doesn't handle `shouldConsume` properly.

murtaza64: Isn't `outputsItems` also not handled correctly?

Anuken: yes

MEEPofFaith:
Done

core/src/mindustry/world/blocks/production/GenericCrafter.java
```diff
@@ -133,15 +141,26 @@ public void updateTile(){
...
+                    }
+                }
             }
 
             if(outputLiquid != null){
                 dumpLiquid(outputLiquid.liquid);
             }
         }
 
+        public void produceItem(ItemStack item){
```
Anuken, line 158: This should be inlined, since it's only used in one location now.

Anuken:
Merged manually.

MEEPofFaith:
k

PR #5564, opened 2021-07-11
Move slider values to the right of the slider

buthed010203:
This is a minor change but prevents the slider from shifting around constantly which can be rather annoying.
Before: https://discord-cdn.is-terrible.com/53W3H4pMT.webm
After: https://vexera.is-ne.at/53W2dq2ev.webm

### DIFF
--- a/core/assets/contributors
+++ b/core/assets/contributors
@@ -127,3 +127,4 @@ WilloIzCitron
 SAMBUYYA
 genNAowl
 TranquillyUnpleasant
+buthed010203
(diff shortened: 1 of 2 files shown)


Anuken:
A better way to display slider text (and save space) would be to display the text *on top* of the slider, likely with a different style:

image

buthed010203:
I agree, ~~the minecraft way~~ that way is significantly better

buthed010203:
I tried with 2 rows in a table and it looks horrible, guess a stack is better here

Anuken:
I'm trying to make the bars "boxy" first, but the margin isn't being calculated properly.
image

buthed010203:
Fair enough

buthed010203:
It doesnt look *terrible* in its current state with them stacked, the text is pretty hard to read though. The text isnt centered either here but thats a really easy fix. https://extremely.questionable.link/543NPKUrb.webm

Anuken:
New implementation:
image

Phinner:
My god, that's beautiful.

PR #5572, opened 2021-07-13
Update bundle_pl.properties

Ksawi999:
Added translation of statuses and processor functions. Changed few translations.


Ksawi999:
Could you please give me the reason for your negative reaction @TheRealIvyX ? Is there still something I should change?

TheRealIvyX:
the status effect translations were kinda wrong when i disliked this but you fixed them now so keeping the negative reaction is pointless now

PR #5574, opened 2021-07-13
Change numeral units to support various numbering systems

minacle:
- Deprecate `unit.thousands`, `unit.millions` and `unit.billions`.
- Introduce `unit.numeric`.

# `unit.numeric`

This is comma-separated list of numeral units. Each part is meaning 1, 10, 100, 1000 and so on.

For example: the position of `unit.billions` is 10th, because of the value 1,000,000,000 has 10 digits. So `unit.numeric` is `,,,,,,,,b`.

The positions of `unit.thousands` and `unit.millions` are 4th and 7th. So combined `unit.numeric` becomes `,,,k,,,mil,,,b`.


Anuken:
> are available to use units.1e4, units.1e5 and others

Why would anyone want to do that? Looking up <20 keys *in a loop* for every single format call looks very expensive in terms of both CPU time and garbage generation.

> These are never inherited by other language bundles or custom translation mods.

Why? How is this enforced?

minacle:
> Why would anyone want to do that? Looking up <20 keys _in a loop_ for every single format call looks very expensive in terms of both CPU time and garbage generation.

Some languages does not have units for 1,000 or 1,000,000 and/or others.

For example, east asians use a myriad based numbering system like 1,0000(万), and 1,0000,0000(億) and so on.

They should use abnormal units for 1,000,000(100万) to display.

The value 123,456,789 is displayed like “123.6 100万” in east asian languages, but they want to see “1.2億” for it. You can think the value “123万” in east asian is displayed like “123 10 thousands” in English.

This problem is not confined in east asian languages.

Indian numbering system is more complex than ease asian.

They use 100,000 based unit and 10,000,000 based units for large numbers.

They want to see the value 123,456,789 as like “12.3 करोड”, not “123.6 10 लाख”.

Due to this variety, 20 loop is expensive, but I cannot think out the solution better than it.

…or we may need some hard coding for each languages.

---

> Why? How is this enforced?

Some languages does not have specific units like million and/or billion.

So, we need to block inheritance parents’ bundle in this scenario, or they will see “1.2 k”, “3.4万” and “56.7 mil” simultaneously in one language, one game screen.

Mnkai:
+1 for this feature.

As a Korean, seeing 170k (170,000) and converting that into our numbering system (17 * 10k-> 17만) within my head every time is not very intuitive. It would be better if this game can support multiple numbering systems by merging this PR.

liuxilu:
This breaks key consistence of .properties files and struggling finding the keys, please consider something like `digits = 4,万;8,亿;12,兆`/`,,,,万,,,,亿,,,,兆`.  The formatting performance seems bad. I got a pseudocode:
```
foreach e in [ [3, K], [6, M], [9, B] ]
    Word.push(e[1]);
    Num.push(10 ^ e[0]);

for i = Num.length To 0
    if number > Num[i]
        print number / Num[i] + Word[i];
```
____
TPEZY is not used originally, they are redundant.

`兆` is not so clear, since:
> 下数者，十十变之。若言十万曰亿，十亿曰兆，十兆曰京也。
中数者，万万变之。若言万万曰亿，万万亿曰兆，万万兆曰京也。
上数者，数穷则变。若言万万曰亿，亿亿曰兆、兆兆曰京也。

I think you'd better give decisions to players of different languages, other than change them directly there.

minacle:
I have applied poor man’s log10 algorithm. Now the digit calculation routine does not matter for CPU time.

__[[See Comparison]](https://gist.github.com/minacle/ed24f79df5d1da9ace0d1900a1872f2a)__

minacle:
> `兆` is not so clear

I have changed it to 万亿.

---

> I think you'd better give decisions to players of different languages, other than change them directly there.

I also do think so. We need an option that choose to use SI or language matched.

liuxilu:
Great performance. The keys are still problematic, different keys across translations are hard for translation tools to work.
The composed '万亿' reminds me that every digit have a word in small amount of chars for east asian.

core/src/mindustry/core/UI.java
```diff
@@ -579,22 +609,64 @@ public static String formatTime(float ticks){
...
+        if(value >= 100L){
+            unit += 2;
+            value /= 100L;
+        }
+        if(value >= 10L){
+            unit += 1;
+        }
+        while(unit > 0){
+            if(si){
+                if(numericSIUnits.containsKey(unit)){
```
liuxilu, line 654: - Move the `while` into the `if`.
- Array`[]!=null` should be faster.

minacle: > * Move the `while` into the `if`.

Can you explain me the reason?

minacle:
Now users can turn it on and off.

minacle:
I think I am done. Please review.

core/src/mindustry/core/UI.java
liuxilu: I was thinking of the performance at first, but it may be insignificant. And the logic order: "I'm looping finding units, seeing if I want SI for each." vs "Determine to use SI or not, then loop to find the units."

liuxilu: The `unit > 19` check while you changed to arrays.

core/src/mindustry/ui/dialogs/SettingsMenuDialog.java
```diff
@@ -450,6 +450,10 @@ void addSettings(){
         }
 
         graphics.checkPref("flow", true);
+
+        graphics.checkPref("siunits", false, value -> {
+            UI.useSIUnits = value;
```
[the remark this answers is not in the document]
minacle: > Don't add the option when `unit.numeric` is absent

I think we do not need to do for it. `unit.numeric` cannot be absent until set `unit.numeric = ` explicitly in bundle.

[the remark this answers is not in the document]
liuxilu: Oh, english bundle fallback, right?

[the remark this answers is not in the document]
minacle: Right. They will use (American) English units as fallback.

core/src/mindustry/core/UI.java
```diff
@@ -32,8 +32,29 @@
 import static arc.scene.actions.Actions.*;
 import static mindustry.Vars.*;
 
+import java.util.Arrays;
+
 public class UI implements ApplicationListener, Loadable{
-    public static String billions, millions, thousands;
+    private static final long[] power10s = {
```
Anuken, line 38: With this many values and this much logic, I think it would be better idea to move *all* the formatting logic into a separate class, say, `NumFormat`, and have `UI` delegate `formatAmount` to a static method in that class for compatibility.

core/src/mindustry/ui/fragments/BlockInventoryFragment.java
```diff
@@ -198,17 +198,6 @@ public void touchUp(InputEvent event, float x, float y, int pointer, KeyCode but
         }
     }
 
-    private String round(float f){
```
Anuken, line 201: This is not equivalent to the standard UI formatting; it doesn't display decimal thousands, for example. Have you tested how this look in inventories with large amounts of items?

Anuken:
Stale PR; review comments don't seem to have been addressed.

PR #5597, opened 2021-07-17
Remove spaces

GamerMoment69:
I'm surprised they were still here, unlike the feature request template in the suggestions repository.


Quezler:
you might want to change this file as well then:
https://github.com/Anuken/Mindustry/blob/master/core/src/mindustry/ui/Links.java#L76

Anuken:
This is wrong. I intentionally added a space in the box to make it show up *as an empty checkbox* when not edited. Users are expected to replace the space with an X.

PR #5602, opened 2021-07-18
Add settings hint

minacle:
Long description for setting items. It might be useful to clarify each settings.

Screenshot

---

If #5601 applied:

Screenshot if #5601 applied


Anuken:
Inline descriptions disrupt the UI layout, so I implemented these as tooltips instead:

PR #5618, opened 2021-07-19
Allow hiding of steam servers

buthed010203:
Theres no way to hide them while keeping them public in the steam api so instead this jank is needed. This wont do anything on a vanilla install but will allow people hosting headless steam servers to hide them if they only use the server to verify if players are on steam or not.

### DIFF
--- a/desktop/src/mindustry/desktop/steam/SNet.java
+++ b/desktop/src/mindustry/desktop/steam/SNet.java
@@ -312,6 +312,7 @@ public class SNet implements SteamNetworkingCallback, SteamMatchmakingCallback, 
             for(int i = 0; i < matches; i++){
                 try{
                     SteamID lobby = smat.getLobbyByIndex(i);
+                    if(!smat.getLobbyData(lobby, "hidden").equals("false")) continue;
                     String mode = smat.getLobbyData(lobby, "gamemode");
                     //make sure versions are equal, don't list incompatible lobbies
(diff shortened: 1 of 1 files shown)


Anuken:
Shouldn't lobbies be shown by default instead? i.e., check for the existence of a hidden property and only hide it *if the property is true*?

buthed010203:
@Anuken thats what this pr does

Anuken:
No, you're not. If the property isn't set, or it's "gsdsdgs", it'll still skip. In the former case, it may even crash.

Existing lobbies should not need to change any of their properties.

buthed010203:
Ah I see what you mean

buthed010203:
@Anuken fixed, tested and it doesnt crash

Anuken:
> !equals("")

Checked the Steam API, and it looks like `""` is returned as an empty property - that won't work. `true` should be fine.

USMP-lancer:
si

buthed010203:
😡 necroposts

USMP-lancer:
si

USMP-lancer:
((remember buthed, the button labeled unsubscribe))

Anuken:
@USMP-lancer The issue tracker is not a place for random meme comments. Asking people to *unsubscribe from their own pull requests* is absurd. If you don't have anything worthwhile to say, add a reaction to the main post and stop filling my notifications with spam.

PR #5626, opened 2021-07-20
Add the ability for draw() on status effects be able to access the effect's time

bloool:
this would be a nice addition specially for mods that may want to create effects with custom draws that increase / decrease as time passes by


Anuken:
You did not test your changes. This does not compile.

sk7725:
add it to update() too, but to not forget 6.0 compatibility/deprecation!

PR #5634, opened 2021-07-23
Better plan skipping

buthed010203:
This prevents the plans being skipped when there is a large number of items entering the core yet the core is still starved of that item.


Anuken:
Seems fishy. What if the core is at a stable number of 10x of some item? What if there are few resources in the core in general?

buthed010203:
It wont swap plans then but I've been playing with this for about a month and find that its way better than it was before adding this, previously on multiplayer if you wanted to build a row of something it would build each block to like 5% completion. The reason the number is set to 15 is because that's what I found was the optimal number, anything less and the plan swapping still occurs, any more and you start to actually notice the problems you mentioned. I think it will actually matter even less once everyone has it. You can now actually manually swap plans if you want to where previously no matter how furiously you clicked on one block it would constantly swap to a new one.

Anuken:
We'll see how well it works for the general playerbase, then.

buthed010203:
Sounds good, if people don't like it then it can always just be reverted

PR #5660, opened 2021-07-28
Fixed desync on servers

buthed010203:
When you ctrl click a unit into a payload, the unit desyncs on the client and isnt removed correctly. Removing these 3 lines fixes it, this fix probably isn't ideal but whatever, its better than it desyncing.


Anuken:
If the unit gets ejected again, the player won't be able to see it. This isn't a good fix.

buthed010203:
Yeah, probably should have tested that first 😅

PR #5664, opened 2021-07-28
This bothers me

buthed010203:
absolutely horrible


Anuken:
I'm pretty sure that option is on by default, you can probably remove it altogether.

buthed010203:
You are correct @Anuken

Anuken:
image
what

buthed010203:
Thats... interesting. I like how the test is still running too

PR #5665, opened 2021-07-29
Disposable container behavior for mods

sk7725:
- Uses instantDeconstruct field, as it implies that this payload is a disposable.
- Disposable blocks (think those cardboard/plastic packaging that are one-use) don't deserve a payload deconstructor.
- Initially additional behavior for mods, but i think some new vanilla blocks can use this as well
- Only disposes of blocks when it actually took items from it


sk7725:
- I really want to make a new DespawnEffect field but it seems like a too specific use case...

Anuken:
I'm not accepting any pull requests relating to the experimental classes; they've been moved, renamed and refactored in the feature branch. This would only cause conflicts.

Anuken:
(if you *really* need new functionality, just copy-paste it)

PR #5704, opened 2021-08-06
Muddy water

TranquillyUnpleasant:


TranquillyUnpleasant:
This is mainly aesthetic but allows for low water environments using mud for where groundwater isn't available.

Generally pumps on mud are worse than water extractors so doesn't actually change any balancing.

Also the water might be gross but that's besides the point, machines don't care about grossness.

Anuken:
Pumps should only be used for actual *water*, mud is just... mud. Pumps don't come with "filters", this is strictly water extractor territory.

PR #5726, opened 2021-08-09
stone craters consistencies fix

JrTRinny:
renamed `crater-stone` to `stone-craters` so it matches with the names in the bundle.
for some reasons all recently renamed environmental blocks, including deep water, water, slag and also craters are missing, pretty sure it isn't caused by my PR (tested on other branches, like master, still missing) so, literal bruh moment

-----


Anuken:
I think it's better to change the bundles instead; fewer problems will arise.

PR #5743, opened 2021-08-11
derelict icon formatting

JrTRinny:
use `Core.bundle.format` to get derelict icon instead of inserting it directly in the bundle, would cause less pain for translators (which including me) when they have to copy a new emoji from the main bundle, which isn't always up-to-date, when the icon is changed
thinking about the simpler way, I could've just yeet `block.derelict` out and use `team.emoji + team.name` instead, but that'd cause a massive deprecation when most bundles already has `block.derelict`, hmm

---


Anuken:
The icon ID/character should never change, though.

JrTRinny:
seems like true, hmm, another bonk.

PR #5771, opened 2021-08-15
reuse <content.sector.name> for sector list

JrTRinny:
reuse ~~plastic~~ bundle string, save the earth
- deprecate `sectorlist`, and reuse the `content.sector.name` for sector list instead

---


summetdev:
In some languages, for instance russian, these lines are not same. (the meaning is slightly different)

JrTRinny:
hmm, I mean, in the English bundle, both `content.sector.name` and `sectorlist` is the same; "Sectors". I could close this but let's see.

Anuken:
They happen to use the same word in English, but "the name of the content type of sector presets" and "the list of sectors" have very different meaning.

PR #5780, opened 2021-08-16
Added the ability to use mod status effects in json / hjson mods(now …

Zelaux:
…it's with if(data.isString))

It is possible to create status effects in separate files, but it was not possible to use them.
Because only the fields of the StatusEffects class were checked and not all the effects.


Anuken:
Implemented differently.

PR #5788, opened 2021-08-16
remove stray newline

buthed010203:
its ugly


Anuken:
can't you consolidate all these single-character PRs into one big change?

buthed010203:
I fix them as I find them, I don't think I have any other prs like this open rn

joshuaptfan:
Catch all the trailing spaces while you're at it too.

You should automate this with a search-and-replace.

buthed010203:
~~sounds like too much work~~

PR #5793, opened 2021-08-18
More team-specific rules

JniTrRny:
Add unit cap, core contribute, cheating, infinite ammo, ammo consumption multiplier and also sort team rules into titles to reduce clutter and preserve everyone's sanity. Feedbacks are welcome :D
- deprecate `rules.enemyCheat` and replace it with `rules.cheating`, I don't see where it is used anyways

moarTeamRules

checks:
- [X] make it works (completely and correctly)
- [X] ~~`make team-specific core contribute syncs with the main core contribute so people don't get confused`~~ found a workaround for it
- [x] maybe change or improve some mechanics in the rules so it's worth putting it in.
---


JrTRinny:
Improved the infinite ammo rule a bit, now the ammo is not consumed, rather than let the unit still shooting when the ammo is 0, + also added a ammo consumption multiplier for units
(in the video, the multiplier is set to 10)

https://user-images.githubusercontent.com/85090668/129890195-711d3d25-aa27-433a-b7a7-cee1357824f1.mp4

itcannotbe:
> Improved the infinite ammo rule a bit, now the ammo is not consumed, rather than let the unit still shooting when the ammo is 0, + also added a ammo consumption multiplier for units

Why not add an ammo capacity multiplier as well for good measure.

JrTRinny:
I mean, you can lower the ammo consumption multiplier to (partially) simulate that.

core/src/mindustry/type/Weapon.java
```diff
@@ -293,9 +293,9 @@ public void update(Unit unit, WeaponMount mount){
 
             mount.reload = reload;
 
-            if(useAmmo){
-                unit.ammo--;
-                if(unit.ammo < 0) unit.ammo = 0;
+            if(useAmmo && !unit.team.rules().infiniteAmmo){
+                unit.ammo -= state.rules.ammoConsumption(unit.team);
+                Math.max(unit.ammo, 0);
```
Anuken, line 298: Why was this changed? You're ignoring Math.max's return value.

JrTRinny: oops, changed it back, if that's what you mean :0

JrTRinny:
significantly improved the cheating rules, so it's worth putting it in the rules.
now turrets will be automatically filled with the first/last ammo, and if you want to change the ammo, you can just use item source, this applies to liquid turrets too, and also coolant will not be consumed when cheating is on.
and turrets that uses >1 ammo per shot, like Foreshadow, will bypass the >1 ammo rule, meaning it can shoot even with less items than item per shot.
this commit make cheating rule more fun(?) and less buggy.
also fixed some bugs from the rule.

https://user-images.githubusercontent.com/85090668/130017231-8910e527-8b11-428e-b15a-e29d092a9b24.mp4

(don't mind me previewing new commit's features from the PR)

Anuken:
Sadly I don't think this PR can be salvaged anymore; unit ammo has since been removed, and there are too many conflicts.

PR #5806, opened 2021-08-19
Hidden blocks only in the Placement UI

sk7725:
- when the blocks are sus.
- Hides the block from the player. This is different from `BuildVisibility`; The player & the world can still interact with the block (e.g. payload sources, picking up blocks) but it will not show up in the placementFragment, thus preventing the player from placing down those blocks *manually*.
- Not intended to be perfect (can be easily bypassed), but to prevent giving overwelming the player with too many similar blocks or something, to be used in mods.
- [ ] ~~todo: prevent schematics as it is just a glorified middle click~~


sk7725:
My conclusion: its actually for the better if you don't prevent the schems, as it is not there to hinder the player significantly, just not overwhelm.

PR #5809, opened 2021-08-19
Avoid ambiguity

SoMall-dumpling:
When you see 'boost: 150%'
It can be 'increase the speed to 150%' or 'increase the speed to 1+150%, which is 250% in total'. Boost can either be verb or be noun.
Maybe you will say that the colon has already avoided this ambiguity.
But you know there's always someone who ignores those things, **and translating 'boost' is hard with nothing to avoid ambiguity.**
**Previous:**
With a 'x', it will show 'Boost: x150%', which will never be thought as 1+150%.
**Latest:**
With a '+', it will show 'Boost: +50%', which not only avoid ambiguity, but also can show 'Boost: +0%' when power is not enough to hold the overdrive using _Math.max_. **Also it makes traslation way more easier.**


Quezler:
well, the default boost is `100%`, having `Boost: x150%` kinda sounds like `100 * 150 = 15000`

core/assets/bundles/bundle.properties
```diff
@@ -776,7 +776,7 @@ bar.corereq = Core Base Required
 bar.drillspeed = Drill Speed: {0}/s
 bar.pumpspeed = Pump Speed: {0}/s
 bar.efficiency = Efficiency: {0}%
-bar.boost = Boost: {0}%
+bar.boost = Boost: x{0}%
```
Quezler, line 779: an alternative might be including the "default" boost and indicating the increase/change somehow, perhaps like this:

```suggestion
bar.boost = Boost: 100% -> {0}%
```

Hexrotor: No, if power is not enough, it will display as 100% -> 80% etc.

Hexrotor: A way is using "boost: +50%" instead of 150%

SoMall-dumpling:
> well, the default boost is `100%`, having `Boost: x150%` kinda sounds like `100 * 150 = 15000`

But 100%*150%=150%, not 15000.

core/assets/bundles/bundle.properties
SoMall-dumpling: Maybe there's something needed to change in 
Mindustry/core/src/mindustry/world/blocks/defense/OverdriveProjector.java /
like when the power is low, show 'low power' instead of '80%', and if it is enough, show '+50%'.

SoMall-dumpling:
ohno wait a minute

SoMall-dumpling:
No need to make bar.low-power, actually if power is not enough, just show 'Boost: +0%' and nothing will get wrong.
Nah. The code get wrong.

SoMall-dumpling:
Alright it works.
So now the bar will show
(overdrive projector) Boost: +50%
(with phase fabric) Boost: +125%
(overdrive dome) Boost: +150%
and when the power is too low that it drops to <100%
it will show Boost: +0%

SoMall-dumpling:
Every 'bar.boost' I can find in the bundles are changed.

SoMall-dumpling:
overdrive projector
overdrive dome
bar
low power
So now everything's done.

PR #5811, opened 2021-08-19
Allows logic to get floor types with getBlock

yetaGitacc:

With this feature, there may be some cool logic, including but not limited to:
a brief map on the display
detecting floor for smarter auto-building


core/src/mindustry/logic/LUnitControl.java
```diff
@@ -16,7 +16,7 @@
     mine("x", "y"),
     flag("value"),
     build("x", "y", "block", "rotation", "config"),
-    getBlock("x", "y", "type", "building"),
+    getBlock("x", "y", "floor", "type", "building"),
```
summetdev, line 19: maybe add `floor` to end to make it compatible with older versions?

Anuken:
> With this feature, there may be some cool logic, including but not limited to:
a brief map on the display
detecting floor for smarter auto-building

How? There's no way to get the color of a floor or even compare floor types, so this feature is completely useless as-is.

core/src/mindustry/logic/LUnitControl.java
yetaGitacc: OK, I'll also change edit another file to make it compatible with older logics

yetaGitacc:
> > With this feature, there may be some cool logic, including but not limited to:
> > a brief map on the display
> > detecting floor for smarter auto-building
> 
> How? There's no way to get the color of a floor or even compare floor types, so this feature is completely useless as-is.

Players can use grey to display stone-like floor, red to stand for hot rocks, etc. The output of type is just like "crater-stone".So players can find or avoid specific floor, what they can do depends how far they can imagine an go.In my opinion, logic features are made for possibilities, not for a specific tasks.

Anuken:
> Players can use grey to display stone-like floor

They *can't*. There is no constant for stone floor. It's not possible to detect whether something *is* a stone floor.

yetaGitacc:
> They _can't_. There is no constant for stone floor. It's not possible to detect whether something _is_ a stone floor.

That's enough, the types of stone floor is limited, we just need to add another memory bank and a processor to determine which color should be used to display this floor.It's what I learned about in my group some days ago.(However, they did nothing in the end because they can't detect what floor is at the position)
There is not only one bullet can be used for spectre, but we can determine if this item can be used for spectre.
Back again, displaying map is not the only usage of this feature.

yetaGitacc:
@Anuken 
We are getting stuck in the topic of drawing map. Units can find water or hot rocks then build pumps or thermal generator automatically. When they a on a offence, this feature allows player to create a very configurable pathfinding logic.For example, avoiding deep water or tar floors. Some very useful logic in theory will come true with the feature.

Hexrotor:
With this, my ground units won't fall into the water lol

Anuken:
Instead of trying to explain my point once more, please just try implementing any of the designs you mentioned with this PR's changes applied.

PR #5823, opened 2021-08-20
Make a method public

buthed010203:
I want to use this method in a mod as it's handy but reflection is horrible


Anuken:
This is an implementation detail utility method that doesn't affect any state of the class; I really don't think it should be public API. Why not copy paste?

buthed010203:
I mean that works but I wanted to avoid that

PR #5837, opened 2021-08-22
Wave UI

TranquillyUnpleasant:
A few quality of life UI changes to the wave editor to make it less of a pain to work with:
- Units group by type, and are sorted in order of increasing health. This roughly corresponds with their strength and is better than simply retaining insertion order.
- The unit editing panels are now hidden unless clicked on so you can switch between different units more easily, as opposed to having to scroll. Changing the spawn group unit type has a dedicated button now:
image
- There is a clear button in the edit menu added. It does a confirmation to prevent accidental wave clearing.
- There is a back button in the unit selection menu.
- The buttons for removing a spawn group and for changing the unit are clear because bordered double buttons looked very weird.

### DIFF
--- a/core/src/mindustry/editor/WaveInfoDialog.java
+++ b/core/src/mindustry/editor/WaveInfoDialog.java
@@ -142,7 +142,9 @@ public class WaveInfoDialog extends BaseDialog{
             main.row();
             main.button("@add", () -> {
                 if(groups == null) groups = new Seq<>();
-                groups.add(new SpawnGroup(lastType));
+                SpawnGroup newGroup = new SpawnGroup(lastType);
+                groups.add(newGroup);
+                expandedGroup = newGroup;
                 buildGroups();
             }).growX().height(70f);
         }), new Label("@waves.none"){{
--- a/core/src/mindustry/editor/WaveInfoDialog.java
+++ b/core/src/mindustry/editor/WaveInfoDialog.java
@@ -172,7 +172,10 @@ public class WaveInfoDialog extends BaseDialog{
                             t.remove();
                             updateWaves();
                         }).pad(-6).size(46f).padRight(-12f);
-                    }, () -> showUpdate(group)).height(46f).pad(-6f).padBottom(0f).row();
+                    }, () -> {
+                        expandedGroup = expandedGroup == group ? null : group;
+                        buildGroups();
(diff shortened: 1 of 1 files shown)


TranquillyUnpleasant:
I did attempt to store just the index of the expanded group, but in case of identical groups indexOf seems to get all of them, and then the expansion becomes weird.

TranquillyUnpleasant:
Added a couple new things:
- Auto expands new spawn groups when they are added
- Opens unit selection menu for the new added spawn group

Anuken:
> Units group by type, and are sorted in order of increasing health. This roughly corresponds with their strength and is better than simply retaining insertion order.

What about grouping by start wave instead? Scrolling down the menu would naturally display units that appear later on.

TranquillyUnpleasant:
Start wave isnt generally what map makers are thinking about when organizing their waves i think, Its more what effect does a specific unit have id say. Sorting by health is actually not that great but its better than just insertion order.

Ideally it should sort by unit class and just order it the way units are ordered in the selection menu, but im not sure if that ordering is feasable dynamically.

Anuken:
> Start wave isnt generally what map makers are thinking about when organizing their waves i think

That's what *I* always do, personally. The first entry I want to see is the initial spawn group.

> Also sorting by start wave would make groups jump around while editing

This wasn't really a problem for me, as the list is only rebuilt when you switch units or click another group anyway.

> Its more what effect does a specific unit have id say, so like thinking about where daggers show up, where flares show up etc.

This is usually closely linked to the begin wave, in my experience.

I've started merging this in; here's what it looks like, currently:

image

TranquillyUnpleasant:
> That's what _I_ always do, personally. The first entry I want to see is the initial spawn group.
This may need just a general discussion on discord then, since map making styles are (very) different for different people.

I do like the extra initial wave label though, makes it nicer to distinguish between groups of the same units, i was going to add the index of the unit itself but that was being a bit messy.

But yeah this layout for me feels a bit like i would have to look around a few times before finding the unit i want to edit, just feels less smooth for editing wave layout later on. But again still a massive improvement over insertion order so no major complaints.

Anuken:
> This may need just a general discussion on discord then

Posted this in `#development`, and a couple people seemed to agree with the begin wave, so I'll stick with that.

Merged, see commit.

PR #5886, opened 2021-08-28
add 'allowBridging' for distributor and liquid tank

JniTrRny:
when in placement, the bridge will automatically replace over the solid terrain, this will save your time and sanity, but sometimes you don't want it.

this PR adds `allowBridging` which allows the placement to bridge over this block, true by default.
if false, it'll think that this block is accessible, and will out bridge over it, well, if only the target block can replace your conveyors/conduits/ducts in placement.

also made Distributor and Liquid Tank set the `allowBridging` to false, why? because people complained about it, see the first paragraph. Also why Liquid Tank? well _technically_, liquid tank is a big Liquid Router, same as Distributor.

### __*Before*__
Screenshot (87)
### __*After*__
Screenshot (86)

Conduit will still bridge over the Distributor, even if `allowBridging` is set to false, well because Distributor cannot replace the conduits, basically not in the same category as the blocked (Distributor) block.
Screenshot (88)
this applies to Liquid Tank too.
Screenshot (89)

### __*Family Picture*__
Screenshot (90)

---

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1008,6 +1008,7 @@ public class Blocks implements ContentList{
         distributor = new Router("distributor"){{
             requirements(Category.distribution, with(Items.lead, 4, Items.copper, 4));
             size = 2;
+            allowBridging = false;
         }};
 
         overflowGate = new OverflowGate("overflow-gate"){{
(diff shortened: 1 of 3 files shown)


Anuken:
I'm not sure if this is a necessary change. What if you *want* to bridge over the distributor or tank? Why chain tanks like that?

JrTRinny:
yeah, putting it to Liquid Tank seems unfitting, but I've seen people complaining about the Distributor one though 🤔

Anuken:
Closing for now as it's controversial.

PR #5888, opened 2021-08-28
Make navals able to be wet by Liquid Turrets

JrTRinny:
this PR removes wet immunity from naval units and make floors not apply status effect on naval units if the floor status effect is wet. 
but why? because the bottom part of the ship is touching the water, but the top part isn't, and obviously, not designed to be in touch with water. Wave/Tsunami/any Liquid turret sprays through the top part of the ship, which obviously, would damage the system inside.

nonsense explaining aside, this would make arc + wave combo viable on naval units.

---

### DIFF
--- a/core/src/mindustry/entities/comp/StatusComp.java
+++ b/core/src/mindustry/entities/comp/StatusComp.java
@@ -109,7 +109,7 @@ abstract class StatusComp implements Posc, Flyingc{
     @Override
     public void update(){
         Floor floor = floorOn();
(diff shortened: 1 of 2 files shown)


Anuken:
> but the top part isn't, and obviously, not designed to be in touch with water.

Why is this obvious? Real ships have water hit the deck all the time. As long as they are not completely submerged, futuristic AI-controlled ships with no crew or interior should not be affected by water in any way.

PR #5898, opened 2021-08-29
Add a button to duplicate wave unit groups

TranquillyUnpleasant:
Add a "duplicate" button to create and add a copy of the selected spawn group. The button is disabled if no group is selected:
image
Basically makes it easier to add units based on similar groups without having to edit all of the settings.
(Might need a better name but copy might get confused with copy to clipboard).

Also added a utility copy method to SpawnGroup because that seems like it would be useful, currently only used by the duplicate button.

---


Anuken:
Doesn't seem like the right place for a copy button, in my opinion.

TranquillyUnpleasant:
I only put it there since it accomplishes the same thing as the add button, i.e. creating a new spawn group.

TranquillyUnpleasant:
resolved some conflicts from the other pr

PR #5926, opened 2021-09-03
Wave graph y axis refactor

TranquillyUnpleasant:
The main change is that the health and total graphs actually show the health and total amounts, as opposed to keeping the count amounts.

Preview:
graph
Full changelog:
- `nextStep` obtains the next step in a `1 -> 2 -> 5 -> 10 -> 20...` sequence for cleaner graph intervals.
- `maxY` determines the maximum amount the Y value should reach.
- `maxY` also determines the amount of padding for the axis labels.
- The number of vertical sections in the graph are clamped to 10, which should be reasonable, as opposed to the relatively large amounts it reaches if the graph size is used.
- There is a mostly transparent horizontal line on each graph mark, for reference. It does not interfere with the graph itself much, but provides a nicer reference point compared to trying to guess where it is.
- Minor formatting fixes because spacing was strange for some reason.

---

### DIFF
--- a/core/src/mindustry/editor/WaveGraph.java
+++ b/core/src/mindustry/editor/WaveGraph.java
@@ -38,8 +38,14 @@ public class WaveGraph extends Table{
 
             lay.setText(font, "1");
 
+            int maxY = switch(mode){
+                case counts -> nextStep(max);
+                case health -> nextStep((int) maxHealth);
+                case totals -> nextStep(maxTotal);
+            };
+
             float fh = lay.height;
-            float offsetX = Scl.scl(30f), offsetY = Scl.scl(22f) + fh + Scl.scl(5f);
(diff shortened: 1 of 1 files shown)


TranquillyUnpleasant:
Alternate code for `nextStep`:
```java
  int nextStep(float value){
      int order = Mathf.pow(10, Mathf.ceil(Mathf.log(10, value)));
      if (order / 5f > value) {
          return order / 5;
      }
      if (order / 2f > value) {
          return order / 2;
      }
      return order;
  }
  ```
  But this is probably worse since it has to call Mathf.pow and Mathf.log

File: core/src/mindustry/editor/WaveGraph.java [hunk 1]
```diff
@@ -200,6 +210,23 @@ Color color(UnitType type){
         return Tmp.c1.fromHsv(type.id / (float)Vars.content.units().size * 360f, 0.7f, 1f);
     }
 
+    int nextStep(float value){
+        int order = 1;
+        while (order < value){
```
Anuken, line 215: incorrect formatting, see style guidelines

PR #5934, opened 2021-09-04
Censoring swear words on multiplayers + PlayerSwearEvent

JrTRinny:
add a regex for censoring swear words in multiplayer (by default, you can turn censoring off, but why tho)
also add `PlayerSwearEvent` for future stats counting and plugin? hmm.
Feedback are welcome :D
if you want to add more censors, just add more words in the regex.

TODO & thoughts:
- [ ] numbers and symbols will be removed from the message if the message has swear words, gonna fix that.
- [ ] what about language/server-specific swear censor words? how would that work though
- [ ] censoring symbol should be based on swearing word's length

_**FILE CONTAINS EXPLICIT CONTENT, LOOK AT YOUR OWN RISK.**_

---

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -27,6 +27,7 @@ import mindustry.world.modules.*;
 
 import java.io.*;
 import java.util.*;
+import java.util.regex.*;
 import java.util.zip.*;
 
 import static mindustry.Vars.*;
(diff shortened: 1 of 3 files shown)


Anuken:
I think this should be a plugin, not a built-in feature.

JrTRinny:
hmm, then only event should be enough

Anuken:
Even the event is unnecessary - there shouldn't be any swear detection included in the game at all. Server owners can choose how to deal with vulgar language themselves.

JrTRinny:
oh then bonk

PR #5957, opened 2021-09-07
Add a copy button to spawn groups

TranquillyUnpleasant:
It does exactly what you expect it to:
image

Other changes:
- Made sorting use comps and have a secondary sorting parameter to prevent this weird group jumping when copying groups with similar sorting parameters:

https://user-images.githubusercontent.com/62061444/132419627-46db39e4-d32e-4dc3-95ec-85131ad04502.mp4

- `buildGroups()` is called instead of `updateWaves()` when removing a spawn group to prevent desyncs between the UI and the value of expanded group, otherwise you can run into situations where clicking a group doesn't expand it and you have to click again:

https://user-images.githubusercontent.com/62061444/132419773-f52a6e01-1df7-418d-8511-f8b4e7ef9609.mp4

(This was already an issue but it didn't matter before since identical groups were basically non-existant before)

If this is too many items for the bar i still have the other layout for copying groups i can make a pr for.

---

### DIFF
--- a/core/src/mindustry/game/SpawnGroup.java
+++ b/core/src/mindustry/game/SpawnGroup.java
@@ -157,9 +157,12 @@ public class SpawnGroup implements JsonSerializable, Cloneable{
         '}';
     }
 
-    @Override
-    public SpawnGroup clone() throws CloneNotSupportedException{
-        return (SpawnGroup)super.clone();
+    public SpawnGroup copy(){
+        try {
+            return (SpawnGroup)clone();
+        }catch(CloneNotSupportedException how){
+            throw new RuntimeException("If you see this, what did you even do?", how);
+        }
     }
 
     @Override
--- a/core/src/mindustry/game/SpawnGroup.java
+++ b/core/src/mindustry/game/SpawnGroup.java
@@ -157,6 +157,22 @@ public class SpawnGroup implements JsonSerializable{
         '}';
     }
 
+    public SpawnGroup copy(){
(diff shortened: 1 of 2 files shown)


core/src/mindustry/game/SpawnGroup.java
```diff
@@ -157,6 +157,22 @@ public String toString(){
         '}';
     }
 
+    public SpawnGroup copy(){
```
Anuken, line 160: Implement `Cloneable` and use `clone()` instead, it's faster and future-proof.

TranquillyUnpleasant:
This is the correct way to implement this right? From what i can see you cant use `clone()` unless you override the method.

core/src/mindustry/game/SpawnGroup.java
```diff
@@ -157,20 +157,9 @@ public String toString(){
...
-        newGroup.max = this.max;
-        newGroup.unitScaling = this.unitScaling;
-        newGroup.unitAmount = this.unitAmount;
-        newGroup.shields = this.shields;
-        newGroup.shieldScaling = this.shieldScaling;
-        newGroup.effect = this.effect;
-        newGroup.items = this.items;
-        return newGroup;
+    @Override
+    public SpawnGroup clone() throws CloneNotSupportedException{
```
Anuken, line 161: I meant `copy()` should call `clone()` and handle the exception, see how `Weapon` does it.

Barabara343434:
Güvenlikli

PR #5853, opened 2021-08-24
Wave UI 2: Electric Boogaloo

JrTRinny:
Unofficial Sequel of #5837, Improved wave UI again... for good(?), added a status effect chooser, which will applies the chosen status effect to that wave group in all waves.
also added filter button, for filtering units in _**Begin, Health, Totals and Type**_, which gives mapmakers a nice time-saving choice.

full
full pic

button1
status effect chooser, presented in the pic as logic icon.

button2
sorting filter, presented in the pic as filter icon.

Screenshot (79)
full unit bar. the crawler in the pic has **Overclock** status effect, that's why there's overclock icon next to the unit icon.

https://user-images.githubusercontent.com/85090668/130610523-4a842dee-3985-4c50-817b-09bb54a1ec04.mp4

NOTE: idk why in the video, the guardian icon is missing, might be my computer being weird.

obviously, the code is not the best, to be honest, hence why I made this PR as a draft.
might post updates here later when some progresses is done

TODO and ideas:
- [x] make the code more compact and reduce the amount of space-eating if/else if possible, abuse ternary ? : operator.
- [x] add the "**Overdrive**" status effect back, as it fits right in for this PR.
- [ ] try to make the filter icon stay next to or near the "**add...**" button if possible, I tried but it's very buggy smh.
- [ ] If possible, try to fix/improve the other mechanics in the wave dialog.
 _
- [x] maybe use big filter icon for easier identification?
- [x] maybe option for filter to reverse sorting?

---
If your pull request is **not** translation or serverlist-related, read the list of requirements below and check each box:

- [X] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [X] I have ensured that my code compiles, if applicable.
- [X] I have ensured that any new features in this PR function correctly in-game, if applicable.

PR #5977, opened 2021-09-11
Wave UI 3: Electric Boogalee

JniTrRny:
**UPDATES:** see comment below for updated contents.

probably the biggest Pull Request I've ever done, and ~~the most (somewhat) broken one too.~~ it's pretty good, no(?) hardcode and broken code.

It's all started from a little changes to fix and continue my previous pull request, but then it grew bigger and bigger into this.
Everything works fine, at least on my machine.

Improved Wave UI again, yes, it takes more than a month to finalize this, and the contents changed are too much to be called "Improved", what you'll see below is changes from the first commits + some updates, for full updated summary see comment below.
yes, I know, the percentage of Anuke merging this is (very) low, but this project has taught me many things, so it is worth it even if this got rejected.

# ***ADDED CONTENTS***
### **Wave Search**
a search field, allows you to search by waves (you can only type in digits)
limits to 8 digits because integer limit and why'd you want to see trillionth waves

when you enter a number more than 0 in the search field, the wave group on the left will show only the group that spawns in the specific wave you searched.
using math, the wave graph will also shift to the specific wave you searched, for further lookup. The associated math is extremely accurate, so the graph shifted to the exact wave you searched, very satisfying huh?

Math for graph shifting if you accidentally saw this pull and don't want to browse the messy files:
`Math.max((searchText - 1) - (displayed / 2) - (displayed % 2), 0)` displayed -> displayed amount of waves in graph

https://user-images.githubusercontent.com/85090668/132953895-f26ba2c8-10a0-45f4-b139-f924dff15ecd.mp4

---
### **Wave Filter**
**UPDATE:** pretty much fixed all the heresy and problem, with the cost of filter health mode being taken off to only two

~~`this part is the most broken one in my opinion, yeah, 100+ lines of code, for a dialog.`~~ somewhat broken, just not as worse as before.
basically filter wave group like wave search, but more specific
filter comes with **Health, Wave start-end, Unit Amount in x wave, Group Effect** and also **Strict Filter** check to make the filters super accurate, to much accuracy isn't good either though so I made it as an option check.
Health filter is basically useless so I added modes for them, with 2 modes: **">"** and **"<"**, well, more than and less than.

image

you can access it from a filter icon button near wave search field.

---
### **Wave Group UI**
basically everything about the wave group that I can possibly fit, and it's still looks pretty.
moved wave duplicate and effect into one dialog, + with payloads and items (**UPDATE:** I also implemented working group items)
also with reset button, which resets the payloads, items and effect (but not wave begin, spacing, shields etc.)

image
(yes, upside-down pyramid shape, beautiful right?)

https://user-images.githubusercontent.com/85090668/132955846-de163079-588f-4bb3-a842-8a734acbb4b6.mp4

payloads, with a list of payloads on top, units that is bigger than the group type's payload capacity will be hidden, and the unit button will be disabled if that unit is bigger than group type's payload capacity minus the total payloads, and also with a big clear button for clear all.
image
Oct, with 28 crawlers as its payloads

image
items too, with amount on top.

image
**BONUS:** the payloads button will be disabled if the group type is not a payload unit (~~`currently using group.type.payloadCapacity <= 8 check because group.type instanceof Payloadc always returns false, for me.`~~ I use `constructor.get()` and it worked, finally.)

image
everything will be under a settings icon button.

---
### **Others**
as I promised in the [prequel](https://github.com/Anuken/Mindustry/pull/5853), I put the wave sort button next to the "**add...**" button.
image
also updated the sort enum to use Floatf secondary sort, rather than entire comparator, and added tertiary sort with oddly-specific Floatf, in case secondary sort isn't enough.

and since the effect button is replaced by wave group, I decided to put the effect icon near the unit icon, like in the prequel.
Q: Why not payloads/items too?
A: because it'd cause bloat, and people rarely adds items anyways, and payloads is a Seq. You can still see the items by clicking the Wave Group button to see its icon.
###### for payloads? good question, well.
image

I also make the "**edit..**" button dialog wider, because why not.
plus, make the `expandedGroup` null if the group you're removing is expanded, it's like some sort of a hidden bug found while testing I guess?

---
TODO and complaints, and thoughts:
- [x] ~~`Godzilla had a stroke trying to read the code and freaking died.`~~ Optimized the code significantly, or at least it's not broken anymore.
- [x] ~~`find a better check for Payload units, since group.type instanceof Payloadc always return false, for me.`~~ I use `constructor.get()` and it worked, finally.
- [ ] the `updateIcons()` is kinda wack with multitables.
- [x] ~~`I made between method specially for the filter statement, and it looks ugly as hell, there should be some alternate ways for this.`~~ simplified the filters, eliminating useless methods in the process
- [x] maybe some sort of zoom speed for wave graph? it'll be pretty fun, but useless, so I'll probably hide it under `Vars.experimental`.

[complaints part, I guess?](https://zajc.eu.org/mine.png)
- [x] ~~`**OH MY GOD 6 LINES `IF` STATEMENT!** should probably make it some sort of method(?) and remove between since we don't have to worry about the amount of lines anymore.`~~ I simplified the filters.
- [x] I should make the payloads clear button separate from the pane, because if the pane pop a new line, it'll pop too for cancel button, which looks ugly.
- [x] ~~`currently the group items doesn't work, the UI is fine, but the spawn group doesn't read/write the ItemStack, either fix it or hide the UI under Vars.experimental (well, that doesn't fix anything).`~~ finally implemented the working version it.
- [x] ~~`I probably reusing showUpdate() and showEffect() in the ugliest way possible; slapping booleans.`~~ well, fixed one, another one is impossible to fix so I'll leave it there.
- [x] ~~`the filter dialog is very space-eating, with around 100+ lines of code for a pretty useless dialog, should probably add a seq to filter group by units, that might make it not useless in a way, but it's still heresy`~~ I simplified it with less filters, and use `numField` method to reuse text fields.

with all it's heresy code, I'll make this as a draft, if there's something you don't like/don't want a UI of it, or the PR's content is useless/too much/generally bad/outright bad/delete this immediately, suggest me.

###### oh my god, how many hours it takes to write this PR.

---


JrTRinny:
# UPDATES
since this pull request is quite finalized, with all checks finished (except one; I tried to do it, but I think it'd be better to leave it there)
it is time to un-draft this pull request, so Anuke could merge it (which is *almost* impossible)

### CHANGES FROM THE *SECOND* COMMIT TO THE LATEST COMMIT

### Working group items
group items is working properly now, which is a very great news
and also improved the amount field, since when there's no way to test it, mistakes happen
image
now if you selected an item and the amount field is left empty, it will automatically set item amount to `itemCapacity` of the unit type.
Also removed `minWidth` from the item buttons since other bundle exists.

---

### Graph Speed
sometimes wave graph is too slow (or too quick?) and so I added graph speed button, default to 1, with x1, x2, x4, x8, x16 speed.
In the future, I may add x0.25, x0.5 and also x32, who knows.
image

Graph speed is hidden under experimental, since I don't think it'd be useful enough for people. perhaps it'd be if I add x0.25 and x0.5.

---

### Tertiary Sort
add tertiary sort value in case the first and second sort is not enough, it is oddly-specific sort though, and also get rid of ugly java comparators, use `Floatf` interface then turn into comparator later.
In the future, I may add none-able sort and make groups draggable and shuffleable (or custom combination of sorts?).

---

### Improved Edit Button
improved "**edit...**" button, changed the icon from `pencil` to `edit`, reordered and changed icons of the buttons in the edit dialog, and also increased the buttons (in the dialog) width a bit, for consistencies with the edit button (and dialog) in `MapGenerateDialog`
image
image
also make "Load from Clipboard" button more strict, with the condition of having "[" at the start of the clipboard.

---

### Merged all showContent pane dialog code into one
saves spaces, and also (kinda) not hardcoded, with cons and extra pane defined when using.
this might be heresy, but at least I tried.

---

### Improved Payloads Pane
reduced the amount of columns to ten, and add a visibility button in case of bloat in mod, this button will not be shown in vanilla (at least in Build 134) since Oct can't carry more than 30 of crawlers with no stats editing.
when clicked hidden, the pane will only show the first 30 units in the Seq, and when shown, it will shown all of the payloads Seq.
(using Exotic mod Flicker unit to cause bloat as much as possible)
image
image
also moved the clear button to a separate table with visibility button, and make the `payLeft` - variable for disabling the unit button, a float (which means, it is somewhat more accurate to payload units with floating-point `payloadCapacity`.

---

### `numField` for Filters and Spawns
since field with numbers is used everywhere in the file, why not making it a method to save some space?
yes, saved many file spaces, I am proud of this.

---

### Revised and Simplified Search System
now search wave is default to -1, so groups with start wave of 2 will not be included with wave 1 in results, and make the overall system less complicated.
And also make the shifting of the graph more accurate - now it should be extremely accurate with no mis-shift at all.

---

### Search Filters Simplification
removed `filterRange`, `filterShields` and `filterShieldsWave` and added `filterStrict` - should be more simple now
and made the clear button not exit the dialog.
image

In the future, I plan to add unit filters and 'search' filter, then remove other unnecessary filter and replace it with better ones.

---

### None Notification Text
basically text shown when there's no group that matches the filter/search, to not confuse players when they forgot they had search/filters on.
image

### Middle Click to Copy
now, you can middle click mouse on a group button in the pane to copy it - mobile will use the button in group settings instead.

### Button Tooltips
added tooltips for three buttons in the group button in the pane
image
image
image

also changed the word "Wave Group" to "Group Settings" in case you don't realize - should describe the dialog better now.

---

### Others
use `constructor.get()` for `Payloadc` detection - should be better for the eyes now.

Code Optimization and cleanups, gonna say I made so much disgusting mistakes in the first commit like `a != a` and many more.

---

My Future Plans (might do, or not):
- [ ] add unit filters and 'search' filter, then remove other unnecessary filter and replace it with better ones.
- [ ] add x0.25, x0.5 and x32(?) graph speed.
- [ ] draggable and shuffleable groups and none-able sort

---

lastly, I think it is the perfect time to un-draft this pull request, even when I know this pull request won't get merged easily, I might be wrong though, I will try to keep the quality of this pull request high - if possible, and thank you for everyone that supports this pull request, with love.

USMP-lancer:
neat

rylid-umil:
anuke pls add this but also, add an option for units to carry payload blocks and put items in the payload blocks carried by a unit, and items in on a unit being carried by a payload. each icon on a block/payload will have an x button and items button, x removes the payload, and the item assigns items to it. also, you can select a category for blocks or units, and can only add items that the block accepts.

rylid-umil:
also, add a filter that you can drag the units in the wave, just like in v6.

JniTrRny:
It was fun while it lasted :)

PR #2080, opened 2020-05-18
[QOL] Searchable Maps

zatnas:
screenshot200518195002537043020

Update 1:
Peek 2020-05-19 06-51

Update 2:
Peek 2020-05-19 11-27

Implements https://github.com/Anuken/Mindustry-Suggestions/issues/182

TODO:
- [x] add search icon
- [x] implement Gamemode filter

PR #6004, opened 2021-09-15
Map search

TranquillyUnpleasant:
Add map search to the map dialog. Also add a filter menu to filter maps by game mode, as well as to hide/show default maps in the map dialog. 

Figured i might as well add this since #2080 seems dormant. Since the entire search bar and filter button is a slim line it shouldn't cause too many issues with mobile landscape vertical space.

Demo:

https://user-images.githubusercontent.com/62061444/133461475-0ba7cea5-6bd7-4fed-9006-44ff5bb430fc.mp4

---


core/src/mindustry/ui/dialogs/MapsDialog.java
```diff
@@ -108,26 +118,53 @@ void setup(){
...
+        mapTable.clear();
+
+        mapTable.marginRight(24);
+
         int maxwidth = Math.max((int)(Core.graphics.getWidth() / Scl.scl(230)), 1);
         float mapsize = 200f;
 
         int i = 0;
-        for(Map map : Vars.maps.all()){
+        Seq<Map> copy = showAll ? Vars.maps.all().copy() : Vars.maps.customMaps().copy();
```
Anuken, line 153: This is very inefficient; unless I'm missing something, it should be faster and easier to iterate over `showAll ? Vars.maps.all() : Vars.maps.customMaps()` and `continue;` if maps don't match a criteria. A boolean flag can keep track of empty searches.

TranquillyUnpleasant:
I might as well ask this i guess, is this style better for the filter menu than the check mark? The check mark feels a bit out of place.

image

Anuken:
Either one is fine. For mode names, I would add the corresponding icon next to the text.

Editor should not be shown as a gamemode.

TranquillyUnpleasant:
Both sandbox and editor are hidden for filters, since they both don't have any conditions for the map, so the filter is pointless. They also don't have icons, so it would just look weird. Current layout.

image

Also changed the loop to use continue.

core/src/mindustry/ui/dialogs/MapsDialog.java
TranquillyUnpleasant: Fixed

core/src/mindustry/ui/dialogs/MapsDialog.java
```diff
@@ -213,4 +281,21 @@ void showMapInfo(Map map){
...
+        super.show();
+
+        if(Core.app.isDesktop()){
+            focusSearchField();
+        }
+
+        return this;
+    }
+
+    public void focusSearchField(){
```
Anuken, line 296: This should be inlined - the implementation in SchematicDialog was just dead code.

core/src/mindustry/ui/dialogs/MapsDialog.java
```diff
@@ -141,13 +181,41 @@ void setup(){
             i++;
         }
 
-        if(Vars.maps.all().size == 0){
-            maps.add("@maps.none");
+        if(mapList.size == 0){
```
Anuken, line 184: `mapList.size` can be non-zero, but the displayed list can still be empty. Add a boolean flag that is set to true after the `continue` in the loop, and check for that.

Anuken (changes requested):

core/src/mindustry/ui/dialogs/MapsDialog.java
TranquillyUnpleasant: Done

core/src/mindustry/ui/dialogs/MapsDialog.java
TranquillyUnpleasant: Done

PR #6017, opened 2021-09-18
Added unit index to radar statement

SingularityT3:
The radar statement now returns a unit based on the index after sorting. For example, when index is 0(and sort is distance and sortOrder is 1), the closest unit is returned. When the index is 1, the second closest unit is returned and so on. If the index is higher than the number of units nearby, the result is null.

Demo program:
logic
demo

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -739,6 +739,10 @@ public class LExecutor{
 
                     lastTarget = targeted = best;
                 }else{
+                    //return from cached units if possible
+                    if(idx < units.size){
+                        lastTarget = units.get(idx);
+                    }
                     targeted = lastTarget;
                 }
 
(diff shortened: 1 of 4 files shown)


core/src/mindustry/logic/LExecutor.java
```diff
@@ -747,7 +750,8 @@ public void run(LExecutor exec){
             }
         }
 
-        void find(Ranged b, float range, int sortDir, Team team){
+        void find(Ranged b, float range, int sortDir, Team team, int idx){
+            ObjectMap<Float, Unit> units = new ObjectMap<>();
```
Anuken, line 754: This violates the contribution guidelines in two ways:
1. Boxed `Float` used.
2. Large allocation in main loop.

SingularityT3: Should be fixed now.

PR #6027, opened 2021-09-20
Update servers_v7.json

jixishi:
Add two servers to the v7 list


Darkness6030:
You should add it to the bottom of list.
And also why 3 servers?
You said there are 2

Darkness6030:
And also you forgot a space after comma

PR #6029, opened 2021-09-20
Layer editor

Minxyzgo:

Implement simple layers in the map editor, It is useful in most cases


Anuken:
- Looks strange. I don't think this is very useful.
- Incurs an extra HashMap settings access every time any block is updated, which is significant.
- Implementation uses completely unrelated settings classes.

PR #6044, opened 2021-09-25
Vela Rebalance

SoMall-dumpling:
After getting enhanced in damage/speed/repair point, vela has become an ultimate unit.
3 velas can hold a cryofluided-alloy-cyclone, 9 velas can hold up to 5 foreshadows(with proper logic)
And the main cause is suggestion-flooding and bad-balancing. Vela shouldn’t be so strong.
My solution is to cut the repair efficiency in half, which means 42/s per repair point.
You know, the truth of the universe.
 

Ilya246:
it's not really an "ultimate unit", now it's just actually viable - the vision of it being "overpowered" most likely comes from several other t4s (antumbra, scepter) being weak in comparision to others - it's just that vela is just as good as quad and arkyid now

PR #6045, opened 2021-09-25
Make thruster cost some scrap

Darkness6030:
Now it costs some scrap and also i changed its position from distribution to defense


SoMall-dumpling:
So you can build *derelict* buildings.
Kinda doesn't make sense.

Anuken:
> So you can build derelict buildings.

You can already build the other scrap stuff in sandbox.

SoMall-dumpling:
> > So you can build derelict buildings.
> 
> You can already build the other scrap stuff in sandbox.

Ah, it’s sandbox mode. Then everything goes well.

PR #6046, opened 2021-09-25
Updated Indonesian Bundle V7

MissAmity:
V7 Indonesia Translation

Translated the new added wave ui, naval sectors, and other addition


Dioarya:
I think you should change the word "inti" to "pusat" so it's more accurate to the word "core" in the game. not forced

MissAmity:
~~will do consider later~~

PR #6064, opened 2021-09-28
Effect spawn delay

MEEPofFaith:
image

I should probably test this.

### DIFF
--- a/core/src/mindustry/entities/Effect.java
+++ b/core/src/mindustry/entities/Effect.java
@@ -30,6 +30,8 @@ public class Effect{
     public float lifetime = 50f;
     /** Clip size. */
     public float clip;
+    /** Time delay before the effect starts */
+    public float delay;
     /** Amount added to rotation */
     public float baseRotation;
     /** If true, parent unit is data are followed. */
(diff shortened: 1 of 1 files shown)


MEEPofFaith:
Yep, it works.

https://user-images.githubusercontent.com/54301439/135146362-2645a6dc-5962-4b99-9285-adbae45fbf25.mp4

Anuken:
Problem is that this adds a delay to *every single particle effect*, regardless of whether the delay is 0 or not. This means every particle effect will only spawn next frame, and has significant garbage collection / performance implications.

BlueWolf3682:
I'd recommend an \@Nullable field for this - that way it doesn't affect effects without the delay field.

MEEPofFaith:
> Problem is that this adds a delay to _every single effect_, regardless of whether the delay is 0 or not. This means every particle effect will only spawn next frame, and has significant garbage collection / performance implications.

~~This is the part where you do it yourself and it results in something much cleaner than my creation.~~

MEEPofFaith:
nice

PR #6075, opened 2021-09-29
Map Search by author and descriptions

JrTRinny:
This PR implements searching map by not only name, but also by author and descriptions, description would probably make the search too much, so I added a button for disabling search by it, true by default, and by only name would make the search too specific too, though, so without description search you can still search for names and author.

https://user-images.githubusercontent.com/85090668/135302521-819246ff-022d-45be-b76b-21f787830b25.mp4

also cleanup a bit and changed the "show default maps" button to a check, because it looks kinda... weird being a toggle button.

Forgot on video; this is the search description toggle button
image
when it's off
image

---

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -382,6 +382,7 @@ editor.center = Center
 editor.search = Search Maps...
 editor.filters = Filter Maps
 editor.showAll = Show Default Maps
+editor.searchDescription = Search Descriptions
 workshop = Workshop
 waves.title = Waves
 waves.remove = Remove
--- a/core/src/mindustry/ui/dialogs/MapsDialog.java
+++ b/core/src/mindustry/ui/dialogs/MapsDialog.java
@@ -20,6 +20,7 @@ import static mindustry.Vars.*;
 
 public class MapsDialog extends BaseDialog{
     private BaseDialog dialog;
+    private boolean searchDescription = true;
     private String searchString;
     private Seq<Gamemode> modes = new Seq<>();
     private Table mapTable = new Table();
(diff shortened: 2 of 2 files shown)


buthed010203:
Talk about fancy, wow

JrTRinny:
the original one didn't equipped the search bar to custom games dialog though, and custom games dialog already have more columns than maps, so I don't think adding it would be necessary, but it'd be great either.

buthed010203:
Ah, hadn't realized that a sear bar was added to vanilla prior to this pr lol

TranquillyUnpleasant:
> Description would probably make the search too much, so I added a button for disabling search by it.

The button seems a bit vague to be fair, and searching by author/description really seems more of an advanced setting that could be put into the settings menu instead of being a button in front view. That way it can also have a proper label. 

And it should probably also disabled by default, otherwise people who aren't aware of what it does would be confused why their search is being very weird. for example, people putting in "uk" for some map and because "anuke" has that, all the default maps still show in the search (hmm yes Ancient Cal*uk*dera).

> also cleanup a bit and changed the "show default maps" button to a check, because it looks kinda... weird being a toggle button.

That's a bit interesting to me, because i intentionally used a button instead of the check as you can check [here](https://github.com/Anuken/Mindustry/pull/6004#issuecomment-920292962). *The check* looked very out of place to me, since it would be the only part of the menu text just floating without a backing.

Of course the only other place with this type of menu is the wave editor settings selection so i guess the "standard" still needs to be set, we can probably have a discussion about this or something if you want.

buthed010203:
> > Description would probably make the search too much, so I added a button for disabling search by it.
> 
> The button seems a bit vague to be fair, and searching by author/description really seems more of an advanced setting that could be put into the settings menu instead of being a button in front view. That way it can also have a proper label.

It does have a tooltip which helps but it would probably make more sense to add it to the filter menu as that icon is self explanatory

JrTRinny:
#6194 is great

PR #6078, opened 2021-09-30
Update bundle_ru.properties

Kowkodivka:
б


Darkness6030 (approved):
All os correct. Approving

PR #6099, opened 2021-10-04
Slight sorter change

haltalt:
https://github.com/Anuken/Mindustry-Suggestions/issues/3060

- [ X ] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).

### DIFF
--- a/core/src/mindustry/world/blocks/distribution/OverflowGate.java
+++ b/core/src/mindustry/world/blocks/distribution/OverflowGate.java
@@ -11,7 +11,7 @@ import mindustry.world.meta.*;
 public class OverflowGate extends Block{
     public float speed = 1f;
     public boolean invert = false;
-
+    public boolean alter[] = {false, false, false, false};
     public OverflowGate(String name){
         super(name);
         hasItems = true;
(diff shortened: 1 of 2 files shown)


Anuken:
Sorters are a fundamental transportation block - there's no telling how many designs this will break, and I'd rather not change their behavior.

haltalt:
> Sorters are a fundamental transportation block - there's no telling how many designs this will break, and I'd rather not change their behavior.

See the linked up suggestion for explanation. This, as I believe, is actually unintended behaviour. Never saw this in use in any design whatsoever, and the only potential use i could see for it is belt logic, which is now not practised by anyone due to v6 introducing support for logic blocks. This is rather an example implementation, a proof of concept.

Darkness6030:
I think it is a good idea, but I didn't understand how that code works :-/

TranquillyUnpleasant:
The change does seem to make the actual behavior more consistent, even if the observed behavior would be something that happens in very rare circumstances, mostly in compact routing. 

Current behavior is framerate dependant and fairly inconsistent (the side that is outputting switches if items happen to flow even slightly differently), so i don't see how a design would be taking advantage of it, since its not even reliable in how it works.

PR #6100, opened 2021-10-04
Modded servers are not allowed

Darkness6030:


Anuken:
That's the wrong place to put such a notice.

USMP-lancer:
very much so

Anuken:
(added)

PR #6161, opened 2021-10-15
Add new (probably useless) method

Darkness6030:
Idk but why not


Anuken:
the best code is no code

PR #6231, opened 2021-10-25
Update servers_v6.json

maxutka99:


Master-Rus:
> Why adding not at the bottom? Anuke, maybe add to SERVERLIST.MD some info that the server must be added to the bottom?

Servers are sorted by ping, so location doesn't matter.

Master-Rus:
> It matters to some extent, as servers are pinged in order of listing, and there may be limits to how fast pings can happen.

I'm not sure if this is the problem. But it's up to you to decide.

Darkness6030:
> > It matters to some extent, as servers are pinged in order of listing, and there may be limits to how fast pings can happen.
> 
> I'm not sure if this is the problem. But it's up to you to decide.

*so why is easyplay always on the top even if using Korean VPN?*

PR #6266, opened 2021-10-30
Armour damage reduction changes.

Halt-alt:
This change makes armour damage reduction more fluent and less bashy, balancing armour for low damage events without the need to make sure the damage dealt is >= a certain value.

- [ X ] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).

### DIFF
--- a/core/src/mindustry/Vars.java
+++ b/core/src/mindustry/Vars.java
@@ -102,8 +102,6 @@ public class Vars implements Loadable{
     public static final float baseInvasionChance = 1f / 100f;
     /** how many minutes have to pass before invasions in a *captured* sector start */
     public static final float invasionGracePeriod = 20;
-    /** min armor fraction damage; e.g. 0.05 = at least 5% damage */
-    public static final float minArmorDamage = 0.1f;
     /** land/launch animation duration */
     public static final float coreLandDuration = 160f;
     /** size of tiles in units */
--- a/core/src/mindustry/entities/comp/ShieldComp.java
+++ b/core/src/mindustry/entities/comp/ShieldComp.java
@@ -25,7 +25,7 @@ abstract class ShieldComp implements Healthc, Posc{
     @Override
     public void damage(float amount){
         //apply armor
-        amount = Math.max(amount - armor, minArmorDamage * amount);
+        amount = amount - (1f/(1f/amount+0.6f/armor));
         amount /= healthMultiplier;
 
         rawDamage(amount);
--- a/core/src/mindustry/entities/comp/ShieldComp.java
+++ b/core/src/mindustry/entities/comp/ShieldComp.java
@@ -25,7 +25,7 @@ abstract class ShieldComp implements Healthc, Posc{
     @Override
     public void damage(float amount){
         //apply armor
-        amount = amount - (1f/(1f/amount+0.6f/armor));
+        amount = amount - (1f / (1f / amount + 0.6f / armor));
         amount /= healthMultiplier;
 
         rawDamage(amount);


Anuken:
This changes the entire balance of the game, and I don't see why armor needs to be changed. What does it mean for some numbers to be "bashy" or "fluent"?

haltalt:
> This changes the entire balance of the game, and I don't see why armor needs to be changed. What does it mean for some numbers to be "bashy" or "fluent"?
1. doesn't really change balance unit-wise. looking at the graph of both the old and the new formula, they are quite similiar
2. at the moment the formula isn't a fluent one, e.g. it uses a bashy solution of allowing damage reduction to only go up to 90% in order to ensure turrets like duos still do some *minimal* damage. however, this results in a graph which is just 2 straight lines (that's what i mean by non-fluent: the derivative of the damage graph isn't continous). this behaviour is quite nonlogical; and even though emphasy isn't usually put on the game to be realistic, non-fluent formulas like these tend to be worse than others.

Anuken:
> looking at the graph of both the old and the new formula, they are quite similiar

But they are not the *same*, so balance is affected.

> bashy solution

Again, I don't understand what you mean by "bashy". It doesn't seem to be a real word, I can't find any relevant definitions online.

> at the moment the formula isn't a fluent one

On the contrary, I would say it is *less* fluent. Yours still uses arbitrary constants (what is `0.6`?), and is much harder to explain to someone. How much real damage will a bullet with 20 damage do against a unit with 10 armor, with the current algorithm? Easy, 20 - 10. It can be succinctly summarized as "subtract armor from damage, but keep in mind that armor can only block 90% of damage." With your algorithm... the only option is to pull out a calculator.

> this behaviour is quite nonlogical

Why? The "real life damage" of bullets hitting armor playing really doesn't have to be a smooth curve, as some penetrate the plating, and others don't. After a certain amount of armor, bullets could simply be chipping off the plating itself, which is why further increases in the armor value wouldn't decrease the damage.

> non-fluent formulas like these tend to be worse than others.

Why?

haltalt:
> > looking at the graph of both the old and the new formula, they are quite similiar
> 
> But they are not the _same_, so balance is affected.
> 
> > bashy solution
> 
> Again, I don't understand what you mean by "bashy". It doesn't seem to be a real word, I can't find any relevant definitions online.
didn't explain the meaning of bashy... bashing is a brute force method of solving a problem that, while effective, may be impractical in the further scope and is just a quick patch to make stuff works
> 
> > at the moment the formula isn't a fluent one
> 
> On the contrary, I would say it is _less_ fluent. Yours still uses arbitrary constants (what is `0.6`?), and is much harder to explain to someone. How much real damage will a bullet with 20 damage do against a unit with 10 armor, with the current algorithm? Easy, 20 - 10. It can be succinctly summarized as "subtract armor from damage, but keep in mind that armor can only block 90% of damage." With your algorithm... the only option is to pull out a calculator.
This one is actually true.
> 
> > this behaviour is quite nonlogical
> 
> Why? The "real life damage" of bullets hitting armor playing really doesn't have to be a smooth curve, as some penetrate the plating, and others don't. After a certain amount of armor, bullets could simply be chipping off the plating itself, which is why further increases in the armor value wouldn't decrease the damage.
still, it's only *approximately* non-fluent... now that i think about it, i think the perfect solution would be using half a hyperbola whose asymptotes are the current two lines, because it doesn't change unit balance almost at all.
> 
> > non-fluent formulas like these tend to be worse than others.
> 
> Why?
In the general scope they tend to either kill any possible fun strategies, or breed a ton of broken ones. This doesn't happen in the case of vanilla armour. However, when it comes to modding, it could potentially change quite a lot. For example, if a mod siginificantly extends progression, then using a very high armour unit to take advantage of the fact that the enmey has rather spammed quite a lot low-damage earlier game turrets will be less viable due to how the formula is made at the moment. At some point more armour doesn't matter, because it is already enough. 

As said once again, both the term "bashy" and my aversion to bashy formulas is an effect of my own habits. I really do believe these formulas have a negative effect on any system they are present on, but i'd require some deeper explanation.

PR #6303, opened 2021-11-04
Added register of mod sounds (for Call.sound and Call.soundAt methods)

Zelaux:
Perhaps the implementation is not ideal, but Call methods with sounds will not work with sounds from mods.


Anuken:
I don't understand why this implementation uses `Prov<Sound>`. If it's because of headless support - passing empty sounds that have not been loaded should be perfectly valid.

PR #6307, opened 2021-11-06
Healing Buildings does not run `hit(b)`

MEEPofFaith:
Just watchin' Blue streaming and he mentions this.


sk7725:
I'd rather have them not run `hit(b)`, as it can cause harm to allied blocks, such as puddles or fires

(Also I think this change makes vela light ally buildings on fire)

Anuken:
It's not hitting `b` in the normal sense, it's healing them - it creates entirely different effects. This has **serious** consequences.

MEEPofFaith:
> It's not hitting `b` in the normal sense, it's healing them - it creates entirely different effects. This has **serious** consequences.

However, @BlueWolf3682 has mentioned to me that hitting a building does not produce frags, have hit effects, or anything else and just poof out of existence, which is why I made this pr in the first place.

Anuken:
> does not produce frags, have hit effects, or anything else and just poof out of existence

That's the point, it's not supposed to do any of those.

PR #6362, opened 2021-11-23
Allow to sensor building type through flag

radistmorse:
This could be useful for partial blueprints where processor is expected to be attached to a specific building, e.g. a turret. This change will allow to not hardcode the building name (which will require editing every time you place a blueprint) or assume that the link index will always be the same.

code
result

### DIFF
--- a/core/src/mindustry/entities/comp/BuildingComp.java
+++ b/core/src/mindustry/entities/comp/BuildingComp.java
@@ -1472,6 +1472,7 @@ abstract class BuildingComp implements Posc, Teamc, Healthc, Buildingc, Timerc, 
     public Object senseObject(LAccess sensor){
         return switch(sensor){
             case type -> block;
+            case flag -> block.flags.size() == 0 ? null : block.flags.iterator().next().toString();
             case firstItem -> items == null ? null : items.first();
             case config -> block.configSenseable() ? config() : null;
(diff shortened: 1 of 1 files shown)


Anuken:
I don't think this is a good idea, for three reasons:

1. Block flags are mostly an implementation details that is used for unit targeting. They can change at any time, which would break schematics.
2. It's not clear to users what the result is when blocks have multiple flags, or what flags blocks even *have*.
3. Returning a string is bad design, for type safety, performance and consistency reasons. No other sensor does this for "enum-like" values.

radistmorse:
> Block flags are mostly an implementation details that is used for unit targeting.

It's used in the logic already for ```unit locate```, so if it is to be changed, scripts will break anyway
> It's not clear to users what the result is

Yes, lack of documentation is the problem with the logic in general. For instance, how did I learn about ```@links```? I decided to implement it, went to the code, and found out that it's already there :) At least here the user can dump the result in message and see the possible values.
> Returning a string is bad design

I admire your memory discipline :) I wish unity developers were more like you. I wanted to return the enum value itself first, but that required some additional save/load code, so I decided that would be easier.

But never mind that, I come with the better idea. How about adding an additional filter for ```get link``` (again, same as ```unit locate```)? something like

link # ```i``` from group ```any/core/turret/etc```

That would be basically the same, only with less iterations. I could prepare a PR for that.

PR #6367, opened 2021-11-25
Allow to choose the building group for getlink

radistmorse:
Same idea as the previous PR, different way to implement it. This will allow to pick the needed linked building without hardcoding the name or index. Uses the same "group" list as unit locate, with the addition of the first element "any". "any" iterates through all the linked buildings, which is how this behaved before the change. It is chosen by default, so the change is backward-compatible with blueprints and saves.

code1
code2
result

TODO: The tooltip for "any" is from the RadarTarget enum, which means it is for "any unit", not for "any building".


radistmorse:
I decided to remove the tooltip for "any" altogether. If you don't understand what "any" means tooltip probably won't help.

PR #6386, opened 2021-11-29
Add filters only if necessary

Darkness6030:
We don't need to add default chatFilter and actionFilter if antiSpam is disabled.


LeoSko:
You can change config while running the server, and with such changes anti spam won't work.

Darkness6030:
🤔

Anuken:
This is pointless, doesn't match the style guidelines, and breaks configuration.

PR #6388, opened 2021-11-29
GameOverMessage?

MindustryIDK:
I think it will be a good idea because it can help to localise or change GameOver message.
I didn't test it, will test later.


MindustryIDK:
Can anyone approve running this workflow?

Anuken:
ServerControl is not an accessible class.

MindustryIDK:
🤔 so i need to move it?
For example to netServer

MindustryIDK:
🤔 how can i use roundExtraTime variable in netServer?
Maybe move it?

MindustryIDK:
Or there is a better idea?

PR #6397, opened 2021-12-02
add Call.copyText

Darkness6030:
A good feature for servers. I compiled it but didn't test in-game.


buthed010203:
This should really pop up a dialog with an option to copy or cancel, that would prevent servers from just copying random crap to people's clipboards constantly. Also when is this ever useful to send to everyone?

Anuken:
1. Servers should not have this power. The clipboard is something that should *never* be accessed remotely.
2. It is not useful for servers to copy text for *everyone.*

PR #6398, opened 2021-12-03
Configurable Power source

hortiSquash:

Following CleanupCrew's suggestion
There is an exponential slider (from 1 to 1M) to setup the power wanted but i'm kinda trash at UI so i'll let that here as a draft for now
Being able to type directly how much you need might be a better idea but i dont know how to do it
looks like this btw

image

(yes i dont know how to apply styles so here's a tex)

EDIT #1:

Remade the UI from message block. Ditched the slider for keyboard input. Now supporting negative power.
image


hortiSquash:
oh yeah and i couldnt change the normal power source because its config the same as power nodes, with links
if there is a way i couldnt find it

SMOLKEYS:
can you also add a TextField (for more better precision)

SMOLKEYS:
~~if this is too difficult dont make it~~

can you also make this but for item/liquid sources?

eg:

item source that produces <set amount> <item type> every <set time>

liquid source that produces <set amount> liquid units of <liquid type> every <set time>

MEEPofFaith:
what happens if you input something that isn't a number

Rex-Hm:
return 0?

hortiSquash:
> what happens if you input something that isn't a number

power get reset to 0
message displays "plz input a valid number"

SMOLKEYS:
~~im just gonna take the code and make a mod with it please dont kill me~~

buthed010203:
> what happens if you input something that isn't a number

> power get reset to 0
message displays "plz input a valid number"

The real question is why this is even allowed... just use the existing method (i think its called filter iirc?) to disable typing non numbers

PR #6401, opened 2021-12-05
Fix incorrect return value

buthed010203:
the javadoc is a lie

### DIFF
--- a/core/src/mindustry/net/Administration.java
+++ b/core/src/mindustry/net/Administration.java
@@ -302,11 +302,13 @@ public class Administration{
     public boolean adminPlayer(String id, String usid){
         PlayerInfo info = getCreateInfo(id);
 
+        if(info.admin) return true;
+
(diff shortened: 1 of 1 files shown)


core/src/mindustry/net/Administration.java
```diff
@@ -302,11 +302,13 @@ public boolean unbanPlayerID(String id){
     public boolean adminPlayer(String id, String usid){
         PlayerInfo info = getCreateInfo(id);
 
+        if(info.admin) return true;
```
Anuken, line 305: Might not want to return early here, so the usid is updated.

buthed010203: ah true

buthed010203:
fixed

buthed010203:
Idk why i wrapped it in newlines, i really shouldnt have but whatever

PR #6415, opened 2021-12-08
Splash Damage Patch

MrDuck557:
Before:
Hail failed to damage navanax even after visually hitting it
navanax
Scatter splash damage fails to damage antumbra properly, and the base damage is cut down by armor

https://user-images.githubusercontent.com/48808663/145219294-94bded66-6b1c-4e26-8c93-cc313e8e2cec.mp4

After:
This actually works
navanax
Antumbra dies properly, taking the 40 dmg splash

https://user-images.githubusercontent.com/48808663/145219541-f1d98241-af0e-43e3-8029-b11b14f3cdd6.mp4

### DIFF
--- a/core/src/mindustry/entities/Damage.java
+++ b/core/src/mindustry/entities/Damage.java
@@ -334,14 +334,14 @@ public class Damage{
     /** Applies a status effect to all enemy units in a range. */
     public static void status(Team team, float x, float y, float radius, StatusEffect effect, float duration, boolean air, boolean ground){
         Cons<Unit> cons = entity -> {
-            if(entity.team == team || !entity.within(x, y, radius) || (entity.isFlying() && !air) || (entity.isGrounded() && !ground)){
+            if(entity.team == team || !entity.within(x, y, radius + entity.type.hitSize) || (entity.isFlying() && !air) || (entity.isGrounded() && !ground)){
                 return;
             }
 
             entity.apply(effect, duration);
         };
 
-        rect.setSize(radius * 2).setCenter(x, y);
+        rect.setSize((radius + 100) * 2).setCenter(x, y);
         if(team != null){
             Units.nearbyEnemies(team, rect, cons);
         }else{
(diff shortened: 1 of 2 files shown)


MrDuck557:
I know that checks failed, this is the log
bm github thing

core/src/mindustry/entities/Damage.java
```diff
@@ -334,14 +334,14 @@ public static void damage(Team team, float x, float y, float radius, float damag
...
-            if(entity.team == team || !entity.within(x, y, radius) || (entity.isFlying() && !air) || (entity.isGrounded() && !ground)){
+            if(entity.team == team || !entity.within(x, y, radius + entity.type.hitSize/2) || (entity.isFlying() && !air) || (entity.isGrounded() && !ground)){
                 return;
             }
 
             entity.apply(effect, duration);
         };
 
-        rect.setSize(radius * 2).setCenter(x, y);
+        rect.setSize((radius + 100) * 2).setCenter(x, y);
```
Anuken, line 344: All queries take hitbox size into account when checking collisions; this shouldn't be necessary. Same goes for the other occurrence of this.

MrDuck557: fixing

LeoSko:
This patch makes sense to me, but it also significantly affects current balance. If it gets merged, can we also adjust healths/damages accordingly?

Anuken:
Yeah, now that I look at, it:

> Antumbra dies properly, taking the 40 dmg splash

...has very serious balance consequences. This is essentially a massive buff for flak against already-weak air T4s and T5s, which is not acceptable.

Anuken:
> can we also adjust healths/damages accordingly?

This isn't a viable strategy - there are far too many combinations of turrets/units, and basic value changes like "buff antumbra health to compensate" would indirectly nerf non-splash-damage weapons.

MrDuck557:
time to apply a different solution

LeoSko:
> This isn't a viable strategy - there are far too many combinations of turrets/units, and basic value changes like "buff antumbra health to compensate" would indirectly nerf non-splash-damage weapons.

Of course it doesn't. Balance is mainly done not via characteristics but abilities and tactics (~~hopefully we will get an obvious scheme what beats what some time huh, if any?~~). But it also doesn't mean we shouldn't do anything or not adjust balance at all after that particular buff.

Anuken:
> Balance is mainly done not via characteristics but abilities and tactics

Up to this point, I've addressed the overwhelming majority of unit/turret balancing problems by just changing some numbers - most problems *can* be solved this way.

I think the best solution here is to 1) do nothing (it's not a real problem, in my opinion) or 2) apply minor visual tweaks so that e.g. hails *do* deal damage to the navanax, but it's floored at 1, or some other tiny number that cannot possibly affect balance.

Anuken:
Won't be merging this regardless due to the balance implications.

PR #6423, opened 2021-12-10
update() move

MrDuck557:
Goober's comment

- [ ] I have ensured that my code compiles, if applicable. (it should)
- [ ] I have ensured that any new features in this PR function correctly in-game, if applicable. (probably should)

### DIFF
--- a/core/src/mindustry/entities/bullet/BulletType.java
+++ b/core/src/mindustry/entities/bullet/BulletType.java
@@ -352,6 +352,8 @@ public class BulletType extends Content implements Cloneable{
     }
 
     public void update(Bullet b){
+        updateTrail(b);
+
         if(homingPower > 0.0001f && b.time >= homingDelay){
             Teamc target;
             //home in on allies if possible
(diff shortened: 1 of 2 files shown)


SMOLKEYS:
the

Anuken:
What does this fix? You've just moved it past the targeting code, and unless I'm very much mistaken there *is* no movement there in the first place. In fact, there is no position change in `BulletType#update(Bullet)` at all, so nothing you do there matters.

image

Anuken:
@Goobrr Can you provide an example/screenshot of trails being buggy with the current code? I've created several trail-using bullets, and there aren't any significant issues with them relating to velocity. There **are** a number of bugs related to trail rendering in general that I have fixed recently, but they're not related to update order.

MEEPofFaith:
I think goob might be talking about how trails lag behind, which is completely unrelated to when it's updated.
*The lagging behind is intentional to make sure that delta time does not affect trail length, and is directly built into trail code.*

Anuken:
Yeah, @MEEPofFaith is correct. I fixed it in the dev branch by making the latest trail point always connect with the trail source, regardless of whether that point was actually added.

Also note that this change *still* doesn't change how trails would update, as velocity update() runs entirely before bullet update() in the generated code:

image

PR #6432, opened 2021-12-13
Fixed Omega Servers on V7

RebornTrack970:


Anuken:
this appears to be occurring quite frequently

RebornTrack970:
> this appears to be occurring quite frequently

yeah, it wont happen for abt a month again. I transferred servers from one node to another but then that node just died so back to the regular one. I hope it wont happen again.

PR #6439, opened 2021-12-17
Use `entity.consValid()` in consumer updates

MEEPofFaith:
`entity.consValid()` already returns `cons.valid(this)`, so it would have no change to vanilla things, however it would allow modded things to add some extra conditions to when the consumes should update through overriding `consValid()`.


MEEPofFaith:
*reads code harder*
actually gonna need as sec to think

MEEPofFaith:
oh yeah it's big brain time

PR #6448, opened 2021-12-21
Fix assertion message

buthed010203:
It previously said `exotic-mod must be loaded. ==> expected: <betamindy> but was: <exotic-mod>` when it failed which makes absolutely no sense.

On the subject of tests, how would you feel about running them in parallel? They finish a bit faster and the only code that needs tweaking is the line this PR changes.


buthed010203:
@Anuken probably didn't see this because I edited it at pretty much the exact same time as you merged the PR but I edited it with this question:
> On the subject of tests, how would you feel about running them in parallel? They finish a bit faster and the only code that needs tweaking is the line this PR changes.

Anuken:
Won't that make them very flaky? Each test uses the same mod directory, I don't see that ending well.

buthed010203:
I've run it a bunch of times with no issues myself, I suppose it could cause problems though..

PR #6454, opened 2021-12-27
Fix strange behavior of MinerAI

buthed010203:
See: https://discord.com/channels/391020510269669376/396416151032299521/925110542216073216

> Anuke, is the current behavior of miner ai where it wont mine anything if the item that the core has the fewest of has all of its ores covered?
The only place the hasOre code is used is miner ai so i assume thats a bug?

### DIFF
--- a/core/src/mindustry/ai/BlockIndexer.java
+++ b/core/src/mindustry/ai/BlockIndexer.java
@@ -78,8 +78,6 @@ public class BlockIndexer{
                 var drop = tile.drop();
 
                 if(drop != null){
-                    allOres.put(drop, 0);
-
                     int qx = (tile.x / quadrantSize);
                     int qy = (tile.y / quadrantSize);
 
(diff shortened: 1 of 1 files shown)


core/src/mindustry/ai/BlockIndexer.java
```diff
@@ -78,7 +78,7 @@ public BlockIndexer(){
                 var drop = tile.drop();
 
                 if(drop != null){
-                    allOres.add(drop);
+                    allOres.put(drop, 0);
```
Anuken, line 81: What's the point of setting it to 0? The default value is 0 anyway.

buthed010203: You could then check contains if needed

Anuken: > You could then check contains if needed for whatever reason

That's a pretty bad idea when you can just check `get(...) > 0`. I just don't see any reason for it to be here.

buthed010203: Fair enough, will fix

Anuken:
Now that I think about it, this could probably be made faster with a `int[]` instead of a map, but whatever.

PR #6457, opened 2021-12-30
Infection port change

RebornTrack970:


bendnuts:
can confirm that this change is necessary

nilederg:
agreed

PR #6459, opened 2021-12-30
made comment statement

Jenya705:

### DIFF
--- a/core/src/mindustry/logic/LParser.java
+++ b/core/src/mindustry/logic/LParser.java
@@ -31,9 +31,11 @@ public class LParser{
         return new LParser(text).parse();
     }
 
-    void comment(){
+    String comment(){
         //read until \n or eof
+        int from = pos;
         while(pos < chars.length && chars[pos++] != '\n');
+        return new String(chars, from, pos - from);
     }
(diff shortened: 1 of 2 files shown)


Darkness6030:
Did you test it?
If yes, please show some screenshots

Jenya705:
изображение
Comments are saving on rejoin

Darkness6030:
Oh

Darkness6030:
@Anuken review please :eyes:

Anuken:
This doesn't work properly:

1. Comments don't have an inherent length limit, which means logic block info can get arbitrarily large with comment spam. This has severe implications for both memory usage and network sync. I don't see a way around this.
2. Comments act as statements, meaning *having comments in your logic program will slow it down*. You can't just strip them out, because 1) you won't be able to view the comments afterwards then, and 2) jump offsets will be broken. Not an easy problem to fix.

Jenya705:
It will slow down only on code update, because it throw away all null instructions
```java
asm.instructions = st.map(l -> l.build(asm)).filter(l -> l != null).toArray(LInstruction.class);
```

PR #6466, opened 2022-01-03
Add serverControl listener

Darkness6030:
It will be very useful for some plugins

### DIFF
--- a/server/src/mindustry/server/ServerControl.java
+++ b/server/src/mindustry/server/ServerControl.java
@@ -35,8 +35,8 @@ import static arc.util.Log.*;
 import static mindustry.Vars.*;
 
 public class ServerControl implements ApplicationListener{
-    private static final int roundExtraTime = 12;
-    private static final int maxLogLength = 1024 * 512;
+    public static final int roundExtraTime = 12;
+    public static final int maxLogLength = 1024 * 512;
 
     protected static String[] tags = {"&lc&fb[D]&fr", "&lb&fb[I]&fr", "&ly&fb[W]&fr", "&lr&fb[E]", ""};
     protected static DateTimeFormatter dateTime = DateTimeFormatter.ofPattern("MM-dd-yyyy HH:mm:ss"),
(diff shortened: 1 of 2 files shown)


summetdev:
More global variables for the god of global variables!

Phinner:
This change is kinda nice, I used to use that hacky code to get the `ServerControl` instance.
```java
public static ServerControl getServer(){
    return (ServerControl)Core.app.getListeners().find(listener -> listener instanceof ServerControl);
}
```

Darkness6030:
Ok lets make global variables for EVERYTHING

Phinner:
wow, you sure about that ? You don't need to expose everything in this class.

Darkness6030:
I am tired of using Reflect :-/

Darkness6030:
@Anuken pls review :eyes:

Anuken:
These are mostly implementation details; they should either be set up with setters and validation code, or not exposed at all.

LeoSko:
Not sure about "implementation details" but I had to specifically rebuild server with ***all*** those properties to match my preferences, +`dateFormatter` because milliseconds matter. (~~who the hell keeps logs with sizes as low as 512kb, not configurable in any way?~~)

Anuken:
> who the hell keeps logs with sizes as low as 512kb

512kb is pretty large for a plain text file, some editors start lagging at those file sizes... and in any case, if someone wanted it configurable, making a *static final* field public isn't the answer, it's pointless and confusing - `Config` should be used instead.

Cryptic undocumented fields like `lastTask` or `suggested` should definitely not be exposed.

LeoSko:
Completely agree on `Config` point for that. (and surely they are not supposed to be final anymore) Also, `logFolder` might be a good idea to un-finalize as well.

> some editors start lagging at those file sizes

Out of curiosity: What software are you trying to use? I'm working with log files every day for the past several years with sizes anywhere between 50Mb to 1Gb and had no issues opening them in either notepad++, less or vi - if you have lags, try those options.

Anuken:
I have gedit, which doesn't handle large files well, although that seems to be more related to text-dense documents rather than long ones.

PR #6488, opened 2022-01-12
Unit based default controllers

LeoSko:
### Why

In a plugin, I want to write a more concise piece of code that relates to that specific unit AI, or to any property it has (team, for instance, but could also depend on type of weapons for ground units or anything else).

One special example is: I want base building team's AI's of daggers to make a formation before pathfinding to players' base, while this is unnessesary for players' daggers. Currenty it would take `if (unit.team == state.rules.waveTeam)` check on each update call and would only bloat the implementation of AI.

I have checked the implementation by replacing default controller for Monos and with NewHorizon mod which uses some custom AIs already. Any notes about `ContentParser.java` changes would be much appreciated.


Anuken:
Merged into private branch with some minor modifications - I ended up needing something similar.

PR #6505, opened 2022-01-19
Customizable gameOver listener and more

Darkness6030:
Why i did this?
Because i want a custom gameOver listener in my plugin. There are no ways to remove default one. But now you can change it to anything.

Why making public so many variables?
Because i am **very** tired of using Reflect.
Note that i made public only useful variables like nextMapOverride.

Also i moved 2 final variabled to Vars.java. Now you can use it anywhere.

### DIFF
--- a/core/src/mindustry/Vars.java
+++ b/core/src/mindustry/Vars.java
@@ -141,6 +141,10 @@ public class Vars implements Loadable{
     public static final int multicastPort = 20151;
     /** multicast group for discovery.*/
     public static final String multicastGroup = "227.2.7.7";
+    /** extra time after the end of the game. */
+    public static final int roundExtraTime = 12;
+    /** maximum log file size. */
+    public static final int maxLogLength = 1024 * 512;
     /** whether the graphical game client has loaded */
     public static boolean clientLoaded = false;
     /** max GL texture size */
--- a/server/src/mindustry/server/ServerControl.java
+++ b/server/src/mindustry/server/ServerControl.java
@@ -79,7 +79,7 @@ public class ServerControl implements ApplicationListener{
             state.set(State.menu);
             net.closeServer();
         }
-    }
+    };
 
     public boolean inExtraRound;
     public Task lastTask;
--- a/server/src/mindustry/server/ServerControl.java
+++ b/server/src/mindustry/server/ServerControl.java
@@ -51,7 +51,7 @@ public class ServerControl implements ApplicationListener{
         }
     };
 
-    public Cons<GameOverEvent> gameOver = () -> {
+    public Cons<GameOverEvent> gameOver = event -> {
         if(state.rules.waves){
             info("Game over! Reached wave @ with @ players online on map @.", state.wave, Groups.player.size(), Strings.capitalize(Strings.stripColors(state.map.name())));
         }else{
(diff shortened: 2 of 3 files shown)


LeoSko:
Another thought: events map from Events from arc also should be public, or provide less restricted interface of removing listeners.

core/src/mindustry/Vars.java
```diff
@@ -141,6 +141,10 @@
     public static final int multicastPort = 20151;
     /** multicast group for discovery.*/
     public static final String multicastGroup = "227.2.7.7";
+    /** extra time after the end of the game. */
+    public static final int roundExtraTime = 12;
```
Anuken, line 145: What is the point of placing either of these variables here? They're still static final, and they have nothing to do with "shared" variables. Either make them `Config` numbers, or don't move them.

server/src/mindustry/server/ServerControl.java
```diff
@@ -35,16 +36,18 @@
...
-
     protected static String[] tags = {"&lc&fb[D]&fr", "&lb&fb[I]&fr", "&ly&fb[W]&fr", "&lr&fb[E]", ""};
     protected static DateTimeFormatter dateTime = DateTimeFormatter.ofPattern("MM-dd-yyyy HH:mm:ss"),
         autosaveDate = DateTimeFormatter.ofPattern("MM-dd-yyyy_HH-mm-ss");
 
     public final CommandHandler handler = new CommandHandler("");
     public final Fi logFolder = Core.settings.getDataDirectory().child("logs/");
 
+    public boolean inExtraRound;
+    public Task lastTask;
```
Anuken, line 47: Implementation detail, confusing and should not be public. What is a "last task"? What does it refer to? Can it be null? What would canceling it do? When does accessing it change anything at all? None of these are clear from the variable alone.

server/src/mindustry/server/ServerControl.java
```diff
@@ -35,16 +36,18 @@
...
+    public Gamemode lastMode;
+    public @Nullable Map nextMapOverride;
```
Anuken, line 49: Should be private and set with a setter to support future changes.

server/src/mindustry/server/ServerControl.java
```diff
@@ -35,16 +36,18 @@
...
     protected static String[] tags = {"&lc&fb[D]&fr", "&lb&fb[I]&fr", "&ly&fb[W]&fr", "&lr&fb[E]", ""};
     protected static DateTimeFormatter dateTime = DateTimeFormatter.ofPattern("MM-dd-yyyy HH:mm:ss"),
         autosaveDate = DateTimeFormatter.ofPattern("MM-dd-yyyy_HH-mm-ss");
 
     public final CommandHandler handler = new CommandHandler("");
     public final Fi logFolder = Core.settings.getDataDirectory().child("logs/");
 
+    public boolean inExtraRound;
+    public Task lastTask;
+    public Gamemode lastMode;
```
Anuken, line 48: Should use a different name and/or documented getter to let users know what this really stores.

Anuken (changes requested):

way-zer:
I do this to disable default gameover logic
Snipaste_2022-03-03_10-19-25
https://github.com/way-zer/ScriptAgent4MindustryExt/blob/4fb49cc79b53c5915da205a43e36c19b88166f9f/scripts/wayzer/maps.kts#L86

PR #6509, opened 2022-01-21
Separate the Draw Z of Units and their Outlines

MEEPofFaith:
With a separation of layer between outlines and the rest of the unit, this allows for `layerOffset` on weapons to actually be able to put the weapon between the main body and the outline.

### DIFF
--- a/core/src/mindustry/type/UnitType.java
+++ b/core/src/mindustry/type/UnitType.java
@@ -605,7 +605,7 @@ public class UnitType extends UnlockableContent{
             drawShadow(unit);
         }
 
-        Draw.z(z - 0.02f);
+        Draw.z(z - 0.03f);
 
         if(mech != null){
             drawMech(mech);
@@ -623,7 +623,7 @@ public class UnitType extends UnlockableContent{
             drawLegs((Unit & Legsc)unit);
         }
 
-        Draw.z(Math.min(z - 0.01f, Layer.bullet - 1f));
+        Draw.z(Math.min(z - 0.02f, Layer.bullet - 1f));
 
         if(unit instanceof Payloadc){
             drawPayload((Unit & Payloadc)unit);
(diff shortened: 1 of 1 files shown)


Anuken:
> this allows for layerOffset on weapons to actually be able to put the weapon between the main body and the outline.

No, in the newer version, outline sprites are not generated unless weapons have `top = false` as an optimization, so there *is* no separate outline sprite in most cases.

MEEPofFaith:
well, `top = false` still draws the weapon sprite *above* the main body. You would still need a negative `layerOffset` to draw the weapon between the unit outline and the unit body.

Anuken:
I'm saying that:

> between the unit outline and the unit body

...won't be possible, because units draw only one sprite now, in most cases. The body isn't a separate region.

Anuken:
Look, I **understand** how the technique works - I'm saying that **THERE IS NO OUTLINE REGION NOW**, so **there is nothing to "sandwich" the weapon sprite between.** This PR will break as soon as I merge in the private branch.

MEEPofFaith:
ah
~~I probably should've understood earlier that "newer version" meant "private branch"~~

Anuken:
This *would* be nice to have, but I don't think a Z-offset is the way to do it, as it's nowhere near as obvious as `top = false` or `bottom = true`, messes with existing layer code, and doesn't provide a clear signal to the outline generator that a separate outline sprite is needed.

BlueWolf3682:
~~I was going to comment that not generating outlines in some cases may become an issue, but then i remembered that I manually add outlines and it will only affect Exogenesis~~

Anuken:
It might be a good idea to merge most of your outlines, if possible - since you have large units, and most of them probably aren't mechs, you're essentially doubling the amount of sprite packing & texture atlas memory usage per unit.

PR #6510, opened 2022-01-21
Remove my second server and a V6 server

Darkness6030:
n3.mindustry.me:5004 - 126.2 server, shouldn't be here

shizashizashiza.ml - my second server, doesn't work, domain lost


Darkness6030:
@RebornTrack970

PR #6522, opened 2022-01-23
Remove inactive servers from v6 list

Shawak:
This commit removes all servers which were offline for atleast three months according to https://mindservlist.de

### DIFF
--- a/servers_v6.json
+++ b/servers_v6.json
@@ -1,23 +1,15 @@
 [
   {
-    "name": "{AA}",
-    "address": ["aamindustry.play.ai", "aamindustry.play.ai:6571", "aamindustry.play.ai:6572", "aamindustry.play.ai:6573", "aamindustry.play.ai:6574"]
-  },
-  {
     "name": "Yeet Hosting",
     "address": ["n3.mindustry.me:5004"]
   },
   {
     "name": "RCM",
-    "address": ["185.104.248.61", "easyplay.su"]
-  },
-  {
-    "name": "Atanner",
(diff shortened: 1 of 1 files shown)


TranquillyUnpleasant:
Isn't the entire point of the server list the way it is now is that it doesn't require any maintenance. Servers that are offline aren't shown to anyone in any case.

LeoSko:
There are several reasons to maintain the list:
- If server is moved to another version - it is useless in the list and only takes space
- If server is down - ping request consumes resources
- If someone simply looks at file, they find a huge list which is not relevant

TranquillyUnpleasant:
> There are several reasons to maintain the list:
> 
>     * If server is moved to another version - it is useless in the list and only takes space
> 
>     * If server is down - ping request consumes resources
> 
>     * If someone simply looks at file, they find a huge list which is not relevant

The first reason is mostly irrelevant as when servers shift to v7 they remove the v6 entry, since they don't want a version incompatible server showing up on the list.

On another note anuke removing servers based on inactivity can result in extra effort if the server owners want to add themselves in again. Its just an extra unnecessary back and forth.

Also are we really going to argue about a handful of ping requests when you're playing a multiplayer game? The ping request checks if the server is down or not, its not wasting resources, manually editing the file periodically to make sure servers aren't down is a much bigger waste of resources.

And i don't see why people finding a giant list of servers in the file is relevant, people who want to change things in the file can do a quick ctrl+f, people who don't want to change things would not actually have any issues with the size of the file.

PR #6556, opened 2022-02-05
Save fires from being colored

sk7725:
Fires are critically endangered.
For some reason, fires soak up colors from bullets...?

### DIFF
--- a/core/src/mindustry/entities/comp/FireComp.java
+++ b/core/src/mindustry/entities/comp/FireComp.java
@@ -1,6 +1,7 @@
 package mindustry.entities.comp;
 
 import arc.*;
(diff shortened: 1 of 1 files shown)


Anuken:
> For some reason, fires soak up colors from bullets...?

Any bullets in particular? This means that a bullet is not resetting color after drawing, which may be a problem for other entities.

sk7725:
> > For some reason, fires soak up colors from bullets...?
> 
> Any bullets in particular? This means that a bullet is not resetting color after drawing, which may be a problem for other entities.

I for some reason thought that bullets are not guaranteed to do that (like effects), as opposed to blocks.

PR #6562, opened 2022-02-08
Allows for different Pixels/Area

hortiSquash:
Spent the entire day asking people if they knew of any mod that change would break, the answer is negative

modders, if you want the screen to take the entire space, remove both the -4

test : doesnt break current displays, automatically does borders, left one is 512x512 9x9 tiles
image

### DIFF
--- a/core/src/mindustry/world/blocks/logic/LogicDisplay.java
+++ b/core/src/mindustry/world/blocks/logic/LogicDisplay.java
@@ -29,7 +29,7 @@ public class LogicDisplay extends Block{
     public int maxSides = 25;
 
     public int displaySize = 64;
-    public float scaleFactor = 1f
+    public float scaleFactor = 1f;
 
     public LogicDisplay(String name){
         super(name);
--- a/core/src/mindustry/world/blocks/logic/LogicDisplay.java
+++ b/core/src/mindustry/world/blocks/logic/LogicDisplay.java
@@ -102,7 +102,7 @@ public class LogicDisplay extends Block{
             Draw.blend(Blending.disabled);
             Draw.draw(Draw.z(), () -> {
                 if(buffer != null){
-                    Draw.rect(Draw.wrap(buffer.getTexture()), x, y, buffer.getWidth() * Draw.scl, -buffer.getHeight() * Draw.scl);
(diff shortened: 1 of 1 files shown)


hortiSquash:
Since it can change the pixel density, you can have large but low resolution displays, perfect for pixel art, or power graphs

Anuken:
Aren't there mods that add borderless displays?

Anuken:
Why not have something like a `float scaleFactor = 1f` in the logic block that multiplies the buffer's rect size for scaling like this? It wouldn't affect any mods, and the functionality would remain the same.

hortiSquash:
> `float scaleFactor = 1f`

good idea, but what about `int borderSize = 4`?
so people dont have to do the maths to find what scale they have to put

Anuken:
The math for the scale is fairly easy, and adding a default border might still break mods.

If you wanted a version of the standard display (`80`px) at 4x magnification for pixel art and the same width border, you would just set the display size to 20 and the scale to 4.

hortiSquash:
misclick sorryy

core/src/mindustry/world/blocks/logic/LogicDisplay.java
```diff
@@ -102,7 +103,7 @@ public void draw(){
             Draw.blend(Blending.disabled);
             Draw.draw(Draw.z(), () -> {
                 if(buffer != null){
-                    Draw.rect(Draw.wrap(buffer.getTexture()), x, y, buffer.getWidth() * Draw.scl, -buffer.getHeight() * Draw.scl);
+                    Draw.rect(Draw.wrap(buffer.getTexture()), x, y, scaleFactor * Draw.scl, -scaleFactor * Draw.scl);
```
Anuken, line 106: This should multiply rather than replace. Don't remove `buffer.getWidth()` / height.

buthed010203:
> Aren't there mods that add borderless displays?
> 
> I say this because
> 
> > if you want the screen to take the entire space
> 
> implies that those would break.

~~Idk, no one has ever complained about it to me hehe `Draw.rect(Draw.wrap(buffer.getTexture()), x, y, (buffer.getWidth() + 16) * Draw.scl, -(buffer.getHeight() + 16) * Draw.scl);`~~

PR #6564, opened 2022-02-09
TR Bundle Fix

RebornTrack970:


Anuken:
This line won't be in the next release, so it's not really worth changing.

RebornTrack970:
> This line won't be in the next release, so it's not really worth changing.

yeah, just fixed a few lines added in the last Bundle update.
So, i dont have to deal with more after v7 branch merge.

PR #6568, opened 2022-02-11
Don't attempt to remove `itemCapacity` if there never was `itemCapacity` in the first place.

MEEPofFaith:
Might be related to [this](https://github.com/MEEPofFaith/prog-mats-java/issues/16).

### DIFF
--- a/core/src/mindustry/world/blocks/units/UnitFactory.java
+++ b/core/src/mindustry/world/blocks/units/UnitFactory.java
@@ -94,7 +94,7 @@ public class UnitFactory extends UnitBlock{
     public void setStats(){
         super.setStats();
 
-        stats.remove(Stat.itemCapacity);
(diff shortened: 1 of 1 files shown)


buthed010203:
~~why does this even throw an exception in the first place smh~~

Anuken:
Probably easier to remove the exception instead.

buthed010203:
Agreed, logging is probably a better option

Anuken:
Or just ignore it, I don't think anyone really cares. It never caught any real problems.

buthed010203:
Sure, you could do that but I think at least keeping logging might make it easier to debug stuff for modders? Especially considering that up until now it would just crash lol

Anuken:
> might make it easier to debug stuff for modders?

I don't really think removing a non-existent stat is something that modders need to care about in the first place.

Anuken:
Fixed in private branch by removing the check.

PR #6572, opened 2022-02-14
Add equals and hashCode for Content.java

way-zer:

When content like Blocks.arc reassign to new instance, references in containers will be wrong.

### DIFF
--- a/core/src/mindustry/ctype/Content.java
+++ b/core/src/mindustry/ctype/Content.java
@@ -5,8 +5,6 @@ import arc.util.*;
 import mindustry.*;
 import mindustry.mod.Mods.*;
 
-import java.util.*;
-
 /** Base class for a content type that is loaded in {@link mindustry.core.ContentLoader}. */
 public abstract class Content implements Comparable<Content>{
     public short id;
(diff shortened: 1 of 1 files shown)


core/src/mindustry/ctype/Content.java
```diff
@@ -49,6 +51,18 @@ public String toString(){
...
+    @Override
+    public boolean equals(Object o){
+        if(this == o) return true;
+        if(!(o instanceof Content content)) return false;
+        return id == content.id && getContentType() == content.getContentType();
+    }
+
+    @Override
+    public int hashCode(){
+        return Objects.hash(getContentType(), id);
```
Anuken, line 63: This is unsupported Java 8 API.

Anuken:
> When content like Blocks.arc reassign to new instance, references in containers will be wrong.

This sounds like a really bad idea in general.

way-zer:
> > When content like Blocks.arc reassign to new instance, references in containers will be wrong.
> 
> This sounds like a really bad idea in general.
> 
> Edit: It would also be misleading - you should use (possibly packed) `id | type << 16` keys, not objects. Your modified arc instance would not actually be equal in fields, it would just have the same ID.

As content type always single instance, the `equal` not bad. Just make new instance still be the same one as the old.

way-zer:
Wait, one problem, if mods set old instance to block. Need to add a helper function like `reGet`.

way-zer:
I try to resolve this in plugin internally,

PR #6582, opened 2022-02-17
Add Logic Lookup Id to ContentInfoDialog.java

Pietro303HD:

### DIFF
--- a/core/assets/contributors
+++ b/core/assets/contributors
@@ -132,3 +132,4 @@ hortiSquash
 King-BR
 citrusMarmelade
 Evolveye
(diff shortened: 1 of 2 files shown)


knzconnor:
Why not?

Because right now we have to check a spreadsheet to use lookup? Not either adding a reverse lookup/to_id, coercing to int I’d instead of bool, or adding it to the coredb seem unnecessarily obfuscated?  I wrote a logic processor in a sandbox world just to iterate the ids (before I found the spreadsheet) and that seems odd to have to do to find core functionality.

PR #6590, opened 2022-02-19
Update servers_v7.json

Omega596:
DM me if any problems appear with server (!-!Creator The Ransombyte!-!)


Anuken:
- Your server does not have a name or description.
- You are using a plain IP instead of a domain, which is prone to breaking when hosts change.

Please don't add servers unless they [meet the criteria](https://github.com/Anuken/Mindustry/blob/master/SERVERLIST.md), and you are prepared to invest time and effort into maintaining them.

PR #6593, opened 2022-02-20
Fix warnings

l-heggerz:

Just a teeny change to get rid of a bunch of the noise from the build project whilst I understand what is going on in the codebase


Anuken:
- The deprecation warnings were fixed in the private branch, removing them here would only cause conflicts.
- "Fixing" `<` and `>` arguably only makes things worse. They render fine in IntelliJ without warnings, and the only place where they *do* cause warnings is during javadoc duration - and why look at that? I don't want to see `&lt;` when I'm looking at code, it's just confusing.

PR #6612, opened 2022-03-02
Add `DamageSource` and `DamageEvent`

way-zer:
Add new interface `DamageSource`, Building/Unit/Bullet/StatusEffect extends it. And some special type.
Make all damage function accept `DamageSource`.
No BREAKCHANGE, all old functions using DamageSource.Unknown, and set `deprecated`

### DIFF
--- a/core/src/mindustry/ai/WaveSpawner.java
+++ b/core/src/mindustry/ai/WaveSpawner.java
@@ -98,7 +98,7 @@ public class WaveSpawner{
 
     public void doShockwave(float x, float y){
         Fx.spawnShockwave.at(x, y, state.rules.dropZoneRadius);
-        Damage.damage(state.rules.waveTeam, x, y, state.rules.dropZoneRadius, 99999999f, true);
+        Damage.damage(state.rules.waveTeam, x, y, state.rules.dropZoneRadius, 99999999f, true, DamageSource.SpawnShockwave);
     }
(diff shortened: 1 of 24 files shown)


Anuken:
I can't merge this in right now, as it conflicts with private branch changes.

MEEPofFaith:
*hissss*
image

way-zer:
> _hissss_ image

idea auto import....
wait anuke finish his private branch.

MEEPofFaith:
~~why are they in TitleCase and not camelCase~~
image

way-zer:
> ~why are they in TitleCase and not camelCase~ image

~may because i'm used to kotlin `object`, which is singleton class, using TitleCase.~

mse-k:
private branch conflict tag when, would probably make it easier to go and find all prs like these after v7 releases

Anuken:
No, that wouldn't help at all.

way-zer:
How about this pr?
Would it be merged if i adapt it to v136?

way-zer:
It will modify all function about damage, which will be a big change even no break-change(with deprecated).
So It may be harder to change when release beta/stable.

Anuken:
I've thought about this some more and I don't think it should be added at all. It's significantly more API breakage and maintenance burden for a feature most won't use.

PR #6623, opened 2022-03-06
Improved Bullet Velocity Inheritance

MEEPofFaith:
- Frag bullets now use the velocity of the parent bullet instead of the velocity of the unit that fired the original bullet
- The amount inherited is now scalable

https://cdn.discordapp.com/attachments/653293028869537843/949927139799609354/Improved_Bullet_Velocity_Inheritance.mp4

If your pull request is **not** translation or serverlist-related, read the list of requirements below and check each box:

- [X] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [X] I have ensured that my code compiles, if applicable.
- [X] I have ensured that any new features in this PR function correctly in-game, if applicable.

PR #6624, opened 2022-03-06
Unit frags shouldn't keep velocity

MEEPofFaith:
https://user-images.githubusercontent.com/54301439/156938030-5b74fd46-3c14-4475-803c-9003c67be30f.mp4

Alternative to #6623

### DIFF
--- a/core/src/mindustry/content/UnitTypes.java
+++ b/core/src/mindustry/content/UnitTypes.java
@@ -895,6 +895,7 @@ public class UnitTypes implements ContentList{
                     ammoMultiplier = 4f;
                     splashDamageRadius = 80f;
                     splashDamage = 75f;
+                    keepVelocity = false;
                     backColor = Pal.sapBulletBack;
                     frontColor = lightningColor = Pal.sapBullet;
                     lightning = 5;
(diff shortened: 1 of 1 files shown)


MEEPofFaith:
~~The velocity of the unit transferring into a bullet's frags makes no sense.~~

MEEPofFaith:
\*realizes I added to the wrong bullet\*

mse-k:
it makes sense kinda, but should be alot less powerful on the frags, if a bullet were to hit something and it was moving really fast in one direction it would make sense for the frags to also have a little of that velocity too

MEEPofFaith:
> it makes sense kinda, but should be alot less powerful on the frags, if a bullet were to hit something and it was moving really fast in one direction it would make sense for the frags to also have a little of that velocity too

But does this make sense?

https://user-images.githubusercontent.com/54301439/156938678-7797de4f-9e5b-442d-a4ae-a3b2c1cbf611.mp4

mse-k:
wait it takes velocity from the unit not the bullet??

MEEPofFaith:
always has been

mse-k:
maybe have the bullet save the velocity of the unit when it shot its bullet in the bullet and use that for the velocity of the fragments

Anuken:
If neither of these make sense, I can just make frag bullets get forced to keepVelocity = false everywhere as a default. As mentioned earlier, it's not a good idea to have broken behavior by default.

PR #6631, opened 2022-03-09
Removed redundant (and probably also broken) vars

buthed010203:
These shouldn't have existed in the first place.


buthed010203:
The real question is why the textField has an update() and why changed() is used instead of typed() but oh well

Anuken:
Now that I look at it, both of those classes are rather cursed.

buthed010203:
> Now that I look at it, both of those classes are rather cursed. The default value isn't even used.

Amazing

buthed010203:
FWIW, the default value isnt used in things such as checkpref either lmao

PR #6632, opened 2022-03-09
Change misleading comments

buthed010203:
The previous javadoc were more confusing than anything. I think this is better? Still dislike the naming of the targets() method considering those are the senders and not the receivers but oh well

### DIFF
--- a/annotations/src/main/java/mindustry/annotations/Annotations.java
+++ b/annotations/src/main/java/mindustry/annotations/Annotations.java
@@ -187,11 +187,11 @@ public class Annotations{
 
     /** A set of two booleans, one specifying server and one specifying client. */
     public enum Loc{
-        /** Method can only be invoked on the client from the server. */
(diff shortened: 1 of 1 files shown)


annotations/src/main/java/mindustry/annotations/Annotations.java
```diff
@@ -187,18 +187,16 @@
 
     /** A set of two booleans, one specifying server and one specifying client. */
     public enum Loc{
-        /** Method can only be invoked on the client from the server. */
+        /** Server only. */
```
Anuken, line 190: This is even less helpful, you're just re-stating the name of the enum. The old comment at least explains where it runs and who can invoke it.

buthed010203: Except that it was wrong half of the time since the meanings are completely different depending on which parameter its filling in for.

PR #6648, opened 2022-03-19
LiquidRouters - Loosen Throughput Reduction

hortiSquash:
temporary fix at best, but can improve the throughput by ~30% up to 4-5 routers in a row, and then go a tiny bit lower than with default scaling, but youre not supposed to have that many in a row anyways. 1.4-1.5 were good all rounders if you prefer. Didnt introduce bugs on my end, but i saw bugs when going lower like 1.01.

I mostly wanted to initiate the conversation about liquid routers. I want to fix them, and for real.
With no reduction (`scaling=1f`), its perfect, but backflow can make them hoard liquid that goes back and forth between two routers forever, and thats why you added that reduction by half, right?

I believe i can fix that if i modify `dumpLiquid()` to behave differently when outputting in the block it recieved liquid from the last. A solution i tried in here (using last block outputted into, so it keeps alternating and blinking cause its very bad), but the  speed is quite nice despite not being perfect (430-450/s compared to the full 500/s) (yes i dont play at 60fps) (and yes, it works less well at 60fps).
image

I am also proposing another solution: separate tanks/containers from routers, and remove all inventory from routers. And make them act like item gates or liquid junctions, just be a direct transfer. Which would have the side effect of trivializing cryo schematics for beginners, and reduce lag from anti-jam logic.

### DIFF
--- a/core/src/mindustry/world/blocks/liquid/LiquidRouter.java
+++ b/core/src/mindustry/world/blocks/liquid/LiquidRouter.java
@@ -16,7 +16,7 @@ public class LiquidRouter extends LiquidBlock{
         @Override
         public void updateTile(){
             if(liquids.total() > 0.01f){
-                dumpLiquid(liquids.current());
+                dumpLiquid(liquids.current(), 1.1f);
             }
         }
 


Anuken:
> I want to fix them, and for real.

The only true fix to the liquid system I can see is the completion of the fluid graph system (treating all connected liquid blocks as one entity, properly instant liquid transport...) but that's difficult to pull off properly. Transporting liquid based on full-ness is fundamentally flawed and I'm not sure if it's worth tweaking.

hortiSquash:
Well thats a whole another level of complexity
Did you start working on it yet, or is it Trello-only for the moment?

itcannotbe:
The more temporary fix is to make dump liquids ask the receiving block how much additional fluids it can store, so that it won't try to dump more fluids than the receiving block can store.

Might also help with removing that other bug.

Anuken:
> Did you start working on it yet

I implemented a basic prototype several months ago, but it was very flimsy - crashes, weird problems with junctions, broken throughput, etc. The "shared storage" part *did* seem to work, at least.

I should get back to it before final v7 release, but the code isn't fun to work with at all.

hortiSquash:
> I implemented a basic prototype several months ago, but it was very flimsy - crashes, weird problems with junctions, broken throughput, etc. The "shared storage" part _did_ seem to work, at least.
> 
> I should get back to it before final v7 release, but the code isn't fun to work with at all.

Do you have that somewhere? I'd like to see and maybe work on it, that would be such a huge upgrade to get that to work
Also the shared storage thing might be reusable for conveyors :D

Anuken:
> Also the shared storage thing might be reusable for conveyors 

No, unfortunately it is not even remotely similar. Liquid is based on a single floating-point number shared across blocks, conveyors have a list of item positions with IDs and such.

> Do you have that somewhere?

I do, but as mentioned earlier, it's very broken. I'm not sure if it's even salvage-able.

PR #6649, opened 2022-03-20
bind `usid` to hostname rather than ip

way-zer:

---
1. Cause some servers use DDNS, usid is useless and admin is easy to loss.
  Also, web uses host not ip address to identity server.
2. Don't provide backward compatibility for usid bind for ip, as that may cause present usid leak.
3. So, this pr may cause all admins need to be set again.

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -639,19 +639,28 @@ public class NetClient implements ApplicationListener{
     }
 
     String getUsid(String ip){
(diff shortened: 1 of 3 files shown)


way-zer:
Rewrite:
* provide back compatibility until 2023.
Other improve:
1. Support SRV for NetClient.connect
2. Support IPV6 address for `serverlist.json`
3. New Api `Net.currentServerAddress()`
4. No use `JoinDialog.safeConnect`

PR #6656, opened 2022-03-22
Update visibility of reset key in configurations

TonySantos9:
All the updates can be seen in " Rebind Keys" menu(Open Mindustry -> Settings -> Controls)

This PR update the showing of the reset key if it has been changed.
If the keybind is the same as the default key value, the reset button is not shown

There are two cases that can be implemented, but I choose the first because was making more sense.

The first: if all the keys are the default ones the reset button is not displayed and there is a gap in the actual missing button.
There is a blank space instead of Reset button.

The second: if all the keys are the default ones the reset button is not displayed and there is no gap in the actual missing button.


TonySantos9:
Original Version at the moment is in this template:
original_v1_2

This Pull Request main version:
fork_v1
fork_v1_2


or the alternative version(not in PR code):
fork_v2


If all keys are the default ones, the global reset button at the bottom will not appear in my both versions(not original mindustry version)

BalaM314 (approved):
Looks good

Anuken:
The keybind dialog _is_ in need of a refresh, but I think the buttons should be disabled as opposed to completely removed. It's something I will need to do myself [in the future.](https://trello.com/c/D79vicXU/996-keybind-dialog-refresh)

PR #6663, opened 2022-03-25
Refactor some power consumption

MEEPofFaith:
- For some reason, `ConditionalConsumePower` and `DynamicConsumePower` were in `/blocks/power/` instead of `/consumers/`
- Renamed them to `ConsumePowerConditional` and `ConsumePowerDynamic` to be consistent with other consumes.
- Overrode `display()` in `ConsumePowerDynamic`, as it should be handled by the block, similarly to `ConsumeItemDynamic`

I know renaming stuff breaks things, but v7 private branch already breaks everything so this is nothing compared to that.


SMOLKEYS:
~~Why not use a deprecated pointer?~~

Anuken:
I implemented 2/3 of those in the private branch already (as I mentioned earlier, the whole consumption system had major changes)

MEEPofFaith:
~~fair~~

PR #6670, opened 2022-03-28
Cease node dumbness

buthed010203:
I'm so sick of placing nodes in a schematic and having them link to some random buildings instead of the correct ones, this should half fix the problem (will still break if the node is placed before any of the other buildings but I really can't be bothered at this point).

### DIFF
--- a/core/src/mindustry/input/InputHandler.java
+++ b/core/src/mindustry/input/InputHandler.java
@@ -344,7 +344,14 @@ public abstract class InputHandler implements InputProcessor, GestureListener{
     public static void tileConfig(@Nullable Player player, Building build, @Nullable Object value){
         if(build == null) return;
         if(net.server() && (!Units.canInteract(player, build) ||
-            !netServer.admins.allowAction(player, ActionType.configure, build.tile, action -> action.config = value))) throw new ValidateException(player, "Player cannot configure a tile.");
+            !netServer.admins.allowAction(player, ActionType.configure, build.tile, action -> action.config = value))){
+            var packet = new TileConfigCallPacket(); //undo the config on the client
+            packet.player = player;
+            packet.build = build;
+            packet.value = build.config();
+            player.con.send(packet, true);
+            throw new ValidateException(player, "Player cannot configure a tile.");
+        }
         build.configured(player == null || player.dead() ? null : player.unit(), value);
         Core.app.post(() -> Events.fire(new ConfigEvent(build, player, value)));
     }
--- a/core/src/mindustry/world/blocks/power/PowerNode.java
+++ b/core/src/mindustry/world/blocks/power/PowerNode.java
@@ -355,7 +355,7 @@ public class PowerNode extends PowerBlock{
 
         @Override
         public void placed(){
-            if(net.client()) return;
+            if(net.client() || power.links.size > 0) return;
 
             getPotentialLinks(tile, team, other -> {
                 if(!power.links.contains(other.pos())){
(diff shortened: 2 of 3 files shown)


Anuken:
> I have ensured that any new features in this PR function correctly in-game, if applicable.

...but does it work?

buthed010203:
> ...but does it work?

You tell me :^)

buthed010203:
Actually.. I can make other improvements

buthed010203:
It works

bendnuts:
plz merge

Anuken:
What's the point of the other new changes?

buthed010203:
> What's the point of the other new changes?

The while loop change was just something I did, idk. The other thing allows you to hold shift when double clicking a node to add more links (clicking 4 times works but it often ends up exceeding ratelimits which is annoying)

Anuken:
> The while loop change was just something I did, idk.

It looks like more code, so unless that old loop is specifically broken, please don't change it.

> I feel like i've asked this before but what are your thoughts on changing tileConfig to use Variant.both and then sending players the correct config when they get ratelimited? I think it would be a lot better to have it be eventually consistent for the player than for it to just desync even if it does feel strange.

Yes, that would be better than what it does now.

buthed010203:
> It looks like more code, so unless that old loop is specifically broken, please don't change it.
Wouldn't the change I just proposed cause issues with the while loop running infinitely if the player has super low ping?

Anuken:
No, since the config will always happen locally immediately, even if it is reverted later by the server.

buthed010203:
Fair enough, now the only issue is that it still configs it on the server as well which is problematic

buthed010203:
> I don't really see a better solution - it's the only action that needs manual reverting on the client end, so you need similar code _somewhere_.

Fair enough, this is what ive settled with, unsure if I should be using the player or null for `packet.player`:
image

Anuken:
It's easiest to just merge one PR, so keep it as part of this one + the undone loop.

buthed010203:
Cool will do, just gonna test it to make sure it works first

Anuken:
Was fixed in the private branch, just ignore it for now.

buthed010203:
Ok cool it works, it will probably look really weird on servers that have the ratelimit warnings off (by setting the kick to the same as the limit itself) since the configs will just undo themselves then you will get kicked but thats not a thing any ~~good~~ server really does anyways.

Anuken:
That would be quite useful, but only after the private branch is merged - what needs to change?

...I should really hurry up with the private branch stuff after April 1st is over.

buthed010203:
> That would be quite useful, but only after the private branch is merged - what needs to change?

Should *in theory* be as simple as `- uses: gradle/gradle-build-action@v2`, see https://github.com/gradle/gradle-build-action#caching for info

Anuken:
Added it to the actions, will see if it explodes.

buthed010203:
I think the main reason its still slow is because its recompiling the whole module anyways as this branch doesn't have incremental compiling working maybe? idk

Anuken:
Yes, I know it doesn't, I'm just checking if it works at all.

buthed010203:
`kapt.classloaders.cache.size=1` In the `gradle.properties` seems to make the tiniest difference in speed though its hard to tell

buthed010203:
image
Its a lot faster when building the client with the caching stuff enabled, nice

buthed010203:
hmm, I'm tempted to actually just move the configuration on place from the server to the client so I can change it in a custom client...

buthed010203:
Or I could just do this server side
image

buthed010203:
~~I actually lied, I tested the other change and it works so im gonna make another pr lol~~

PR #6673, opened 2022-03-28
Better power nodes yet again

buthed010203:
It works

### DIFF
--- a/core/src/mindustry/world/blocks/power/PowerNode.java
+++ b/core/src/mindustry/world/blocks/power/PowerNode.java
@@ -351,7 +351,7 @@ public class PowerNode extends PowerBlock{
 
             getPotentialLinks(tile, team, other -> {
                 if(!power.links.contains(other.pos())){
-                    configureAny(other.pos());
+                    Call.tileConfig(player, this, other.pos());
                 }
             });
 
--- a/core/src/mindustry/world/blocks/power/PowerNode.java
+++ b/core/src/mindustry/world/blocks/power/PowerNode.java
@@ -26,6 +26,8 @@ public class PowerNode extends PowerBlock{
     protected static BuildPlan otherReq;
     protected static int returnInt = 0;
     protected final static ObjectSet<PowerGraph> graphs = new ObjectSet<>();
+    /** The maximum range of all power nodes on the map */
+    public static float maxRange;
 
     public @Load("laser") TextureRegion laser;
     public @Load("laser-end") TextureRegion laserEnd;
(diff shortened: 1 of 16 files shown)


core/src/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -357,9 +357,10 @@ public static boolean insulated(int x, int y, int x2, int y2){
         public void placed(){
             if(net.client() || power.links.size > 0) return;
 
+            var player = Groups.player.find(p -> p.name.equals(lastAccessed)); //player who built this block
```
Anuken, line 360: This is terrible, there must be a better way of fetching placement.

buthed010203: Why not just pass the builder to the placed function? Its only called in once place so it shouldn't be that hard (though it would probably break mods)

Anuken: That's fine, and breaking isn't an issue as the private branch breaks absolutely everything anyway.

buthed010203: > the private branch breaks absolutely everything anyway

Stop, you're making me scared. When I merge the average update, there are ~10 conflicts to merge manually, when the v7 pre alpha came out I had over 100 files with conflicts...

Anuken: > you're making me scared

~~maybe you should be~~

image

buthed010203: > ~maybe you should be~

Oh good, formations cause me so much pain

~i think this might actually crash intellij~

buthed010203:
@Anuken why do all the placed methods even have a `if(net.client()) return;`? They should never even be run on the client anyways

buthed010203:
What are your thoughts on something like this? It would allow the node auto linking when placing powered blocks to work with large nodes, surge towers and modded nodes while not sacrificing too much speed (if you use an IntFloatMap at least).
image
image

buthed010203:
Its pretty bad even with a map
image

buthed010203:
```java
        tempTileEnts.sort((a, b) -> {
            int type = -Boolean.compare(a.block instanceof PowerNode, b.block instanceof PowerNode);
            if(type != 0) return type;
            return Float.compare(a.dst2(tile), b.dst2(tile));
        });```
<img width="893" alt="image" src="https://user-images.githubusercontent.com/26940575/160869140-58b4032c-f0b4-462b-83f8-e2d01eebed06.png">

I guess when this was pasted into the `getNodeLinks` method, this sort wasn't changed... All of the blocks in the list are power nodes. Also wtf is up with the newline placement here

Anuken:
...this is getting pretty confusing and might even cause conflicts, so I'm just going to wait until the private branch merge until I can take a closer look.

Anuken:
> why do all the placed methods even have a if(net.client()) return;? They are never even be run on the client (at least in vanilla) anyways...

It's definitely redundant now, but I believe it *was* called at one point (before I refactored placement calls).

buthed010203:
The only issue now is that maxRange isnt set on world load... Should I add an Events.on(WorldLoadEvent) or just move the code to onproximityupdate?

buthed010203:
ok this is ready to merge now

PR #6676, opened 2022-03-30
Make closestEnemyCore find the actual closest core

Weathercold:
This method should return the closest core of all of `team`’s enemies, but it instead returns the closest core of the first enemy team that has a core

### DIFF
--- a/core/src/mindustry/game/Teams.java
+++ b/core/src/mindustry/game/Teams.java
@@ -36,11 +36,19 @@ public class Teams{
 
     @Nullable
     public CoreBuild closestEnemyCore(float x, float y, Team team){
-        Seq<CoreBuild> enemyCores = new Seq<>();
+        CoreBuild closest = null;
+        float closestDst = Float.MAX_VALUE;
+        
         for(Team enemy : team.data().coreEnemies){
-            enemyCores.addAll(enemy.cores());
+            for(CoreBuild core : enemy.cores()){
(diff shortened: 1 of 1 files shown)


core/src/mindustry/game/Teams.java
```diff
@@ -36,11 +36,19 @@ public Teams(){
...
         for(Team enemy : team.data().coreEnemies){
-            CoreBuild tile = Geometry.findClosest(x, y, enemy.cores());
-            if(tile != null) return tile;
+            for(CoreBuild core : enemy.cores()){
+                float dst = Mathf.dst2(x, y, core.getX(), core.getY());
+                if(closestDst > dst){
+                    closest = core;
+                    closestDst = dst;
+                }
+            }
```
MEEPofFaith, line 49:
```suggestion
           CoreBuild core = Geometry.findClosest(x, y, enemy.cores());
           float dst = Mathf.dst2(x, y, core.x, core.y);
           if(dst < closestDst){
               closest = core;
               closestDst = dst;
           }
```
Wouldn't this be simpler?

Weathercold: perhap (but it will raise NullPointerException since `core` might be null, then you have to add null check which makes it arguably as complicated)

MEEPofFaith: ~~right, I always forgor the null checks~~

PR #6678, opened 2022-03-30
Fix desync when two players control the same unit at the same time

buthed010203:
It seems to work fine though I havent extensively tested. Previously, one of the players would be banished to the shadow realm. The adminAction was seemingly ignored on the client before anyways so that probably desynced and this probably fixes that too..


Anuken:
How does setting the player unit twice fix it?

buthed010203:
I forgot to delete the other line 💀

buthed010203:
Causes a slight lag when swapping but well worth it imo

Anuken:
:skull:

PR #6714, opened 2022-04-17
S: Load icons,locales in server launcher

way-zer:
S: Add icons.properties,bundles,block_colors.png to server build
S: Load icons,locales in server launcher

PR from [MindusyX](https://github.com/TinyLake/MindustryX) project

### DIFF
--- a/core/src/mindustry/net/Administration.java
+++ b/core/src/mindustry/net/Administration.java
@@ -478,6 +478,7 @@ public class Administration{
         autosave("Whether the periodically save the map when playing.", false),
         autosaveAmount("The maximum amount of autosaves. Older ones get replaced.", 10),
         autosaveSpacing("Spacing between autosaves in seconds.", 60 * 5),
+        locale("Default locale in server", "default"),
         debug("Enable debug logging", false, () -> Log.level = debug() ? LogLevel.debug : LogLevel.info);
 
         public static final Config[] all = values();
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -17,7 +17,7 @@ task run(dependsOn: classes, type: JavaExec){
         args Eval.me(appArgs)
     }
     if(project.hasProperty("jvmArgs")){
-        jvmArgs((List<String>)Eval.me(project.getProperties()["jvmArgs"]))
+        jvmArgs((List<String>) Eval.me(project.getProperties()["jvmArgs"]))
     }
 }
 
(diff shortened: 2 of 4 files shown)


way-zer:
Problem: `block_color.png` not generate when build server. And it's hard to generate it. Shuild add it to git.

Anuken:
> Shuild add it to git.

No, it changes every block update, which leads to constant conflicts and confusing diffs on Github.

The only file I would agree to add to the server here is icons.properties - everything else seems like bloat and just increases startup time/size of the server.

core/src/mindustry/net/Administration.java
```diff
@@ -478,6 +478,7 @@ private void load(){
         autosave("Whether the periodically save the map when playing.", false),
         autosaveAmount("The maximum amount of autosaves. Older ones get replaced.", 10),
         autosaveSpacing("Spacing between autosaves in seconds.", 60 * 5),
+        locale("Default locale in server", "default"),
```
Anuken, line 481: This doesn't seem to do anything.

server/build.gradle
```diff
@@ -17,7 +17,7 @@ task run(dependsOn: classes, type: JavaExec){
         args Eval.me(appArgs)
     }
     if(project.hasProperty("jvmArgs")){
-        jvmArgs((List<String>)Eval.me(project.getProperties()["jvmArgs"]))
+        jvmArgs((List<String>) Eval.me(project.getProperties()["jvmArgs"]))
```
Anuken, line 20: Random unnecessary space

core/src/mindustry/net/Administration.java
way-zer: Use for server(headless). can't set locale in ui

server/src/mindustry/server/ServerLauncher.java
```diff
@@ -40,7 +41,7 @@ public static void main(String[] args){
     @Override
     public void init(){
         Core.settings.setDataDirectory(Core.files.local("config"));
-        loadLocales = false;
+//        loadLocales = false;
```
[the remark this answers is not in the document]
way-zer: Load Core.bundle. Make `xxx.localized` works

core/src/mindustry/net/Administration.java
Anuken: But it doesn't actually do anything, since the locale is never used anywhere in the vanilla server. `Config.locale` is *never read*. If you want a system for defining custom configs for plugins on the server, I can add that, but there's no sense adding a configuration option that does nothing.

way-zer: Snipaste_2022-04-17_22-08-02
`Vars.loadSettings` read this config

server/build.gradle
```diff
@@ -33,18 +33,19 @@ task debug(dependsOn: classes, type: JavaExec){
...
-    exclude("sounds/**")
-    exclude("fonts/**")
-    exclude("zones/**")
-    exclude("icons/**")
-    exclude("bundles/**")
-    exclude("cubemaps/**")
-    exclude("cursors/**")
-    exclude("shaders/**")
+    from{ configurations.runtimeClasspath.collect{ it.isDirectory() ? it : zipTree(it) } }
+    from(files(project.assetsDir)){
```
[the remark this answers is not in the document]
way-zer: Is `bloomshaders` useless?

core/src/mindustry/net/Administration.java
Anuken: That's... really indirect and hacky, you're relying on implementation details to get that, instead of the Config API. There's no validation or error handling on the locale string, either.

server/build.gradle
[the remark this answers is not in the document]
way-zer: I don't know how to exclude dir but one special file

way-zer:
Not include bundles. But should provide a way to load bundle? 
For example, user put bundle in config dir, then server will load it when start.

Darkness6030:
Uhhhh
How about gradle args to include assets?

```
./gradlew server:dist -include_assets
```

way-zer:
May add a new artifact for locale. Like assets.jar.

Block colors is still a problem.

PR #6723, opened 2022-04-19
filterable power node connections

KayyAyeAre:

### DIFF
--- a/core/src/mindustry/world/blocks/power/PowerNode.java
+++ b/core/src/mindustry/world/blocks/power/PowerNode.java
@@ -205,6 +205,7 @@ public class PowerNode extends PowerBlock{
             overlaps(tile.x * tilesize + offset, tile.y * tilesize + offset, other.tile(), laserRange * tilesize) && other.team == team &&
             !graphs.contains(other.power.graph) &&
             !PowerNode.insulated(tile, other.tile) &&
+            other.block.shouldConnect(other) &&
             !(other instanceof PowerNodeBuild obuild && obuild.power.links.size >= ((PowerNode)obuild.block).maxNodes) &&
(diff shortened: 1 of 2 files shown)


KayyAyeAre:
tested this by overriding `shouldConnect()` in the mend projector
image

GlennFolker:
I feel like, for the sake of the usefulness, other parameters should be passed as well. `boolean shouldConnect(Tile connector, Block connectorType, Building build)` would probably be sufficient.

> other.block.shouldConnect(other)

Turns to `other.block.shouldConnect(tile, this, other)`, and you should probably add to the document that `tile.build` might be `null` there.

MEEPofFaith:
(`@Nullable`)

bendnuts:
all this seems to do is change 2 lines of code... is it supposed to make it so mods can make content not automatically connect to power nodes?

core/src/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -205,6 +205,7 @@ protected void getPotentialLinks(Tile tile, Team team, Cons<Building> others){
             overlaps(tile.x * tilesize + offset, tile.y * tilesize + offset, other.tile(), laserRange * tilesize) && other.team == team &&
             !graphs.contains(other.power.graph) &&
             !PowerNode.insulated(tile, other.tile) &&
+            other.block.shouldConnect(other) &&
```
Mnemotechnician, line 208: I may not understand something , but why are you checking whether a block should connect to itself?

Anuken:
I agree with Glenn here; there aren't enough parameters. It's also in a strange file - if something requires its own building, it should be in BuildingComp, not Block. In case you still need this, re-open with the changes.

PR #6726, opened 2022-04-20
Name Saves on Custom Game Screen

Error-In-Code:
This thing has been bothering me for a while so I finally fixed it.
image
image

It currently uses the same strings as the save renaming screen but I can change that if needed.

### DIFF
--- a/core/assets/contributors
+++ b/core/assets/contributors
@@ -135,3 +135,4 @@ Evolveye
 Jerzy Paciorkiewicz
 YozoZChomutova
 Qendolin
+[Error_27]
--- a/core/src/mindustry/core/Control.java
+++ b/core/src/mindustry/core/Control.java
@@ -294,7 +294,7 @@ public class Control implements ApplicationListener, Loadable{
         }
     }
 
-    public void playMap(Map map, Rules rules){
+    public void playMap(Map map, Rules rules, String name){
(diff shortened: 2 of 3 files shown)


Anuken:
This takes up precious vertical space on mobile and can be trivially changed in the save screen after you begin. Why bother?

Error-In-Code:
The vertical space concern is a pretty good point, and that can probably be fixed by putting the button on the same row as the "Customize Rules" button. I'll see what I can do on this.

On your other point, it would be a lot easier, more convenient, and more obvious to put a rename feature on the "Custom Game" screen, rather than to leave it only in the "Save" and "Load" screens.

Error-In-Code:
> The vertical space concern is a pretty good point, and that can probably be fixed by putting the button on the same row as the "Customize Rules" button. I'll see what I can do on this.

I think this is better.
image

Anuken:
Sorry, but this still doesn't make sense to me. It's not clear at all what "rename" refers to in that dialog. Changing the save in the "load game dialog" doesn't seem difficult.

> a lot easier, more convenient

Not really, it's only two extra clicks/key presses.

> more obvious

I would expect saves to be edited in the dialog where saves are listed, not here. Maybe the "current" save should be highlighted for easier viewing or something, but that can be implemented independently of this PR.

PR #6790, opened 2022-05-08
Fix mods on mac, other misc items

buthed010203:
Opening a folder such as `mods/someMod` which only has one item (such as `/someModContent` (this type of thing happens w/ the mod updater)) breaks the line just below what I changed. This fixes that.

Another issue that I've noticed previously is that this line checks for `mod.json` and `mod.hjson` but not for the plugin equivalents. Is this intended? https://github.com/Anuken/Mindustry/blob/124cf5fa4e370e38550bfb326898188a05ef9c37/core/src/mindustry/mod/Mods.java#L405

Also yes, this breaks the style guidelines.

**EDIT:** Maybe I shouldn't have changed the url in the warning considering that the issue mentioned is specific to v5 -> v6+

### DIFF
--- a/core/src/mindustry/mod/Mods.java
+++ b/core/src/mindustry/mod/Mods.java
@@ -787,7 +787,7 @@ public class Mods implements Loadable{
 
             try{
                 Fi zip = file.isDirectory() ? file : new ZipFi(file);
-
+                if(OS.isMac) zip.child(".DS_Store").delete(); //macOS loves adding garbage files that break everything
                 if(zip.list().length == 1 && zip.list()[0].isDirectory()){
                     zip = zip.list()[0];
                 }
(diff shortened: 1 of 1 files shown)


buthed010203:
Ok now afaik tests are just failing because they dont initialize an atlas. I can't be bothered to fix this.

Anuken:
> Ok now afaik tests are just failing because they dont initialize an atlas

But that's *wrong*, the server doesn't have an atlas either! You've just introduced a major startup crash.

buthed010203:
> But that's _wrong_, the server doesn't have an atlas either! You've just introduced a major startup crash.

~~Eh servers are overrated anyways~~

core/src/mindustry/mod/Mods.java
```diff
@@ -787,7 +799,7 @@ public Seq<Fi> resolveDependencies(Seq<Fi> files){
 
             try{
                 Fi zip = file.isDirectory() ? file : new ZipFi(file);
-
+                if(OS.isMac) zip.child(".DS_Store").delete(); //macOS loves adding garbage files that break everything
```
Anuken, line 802: This will lead to very strange results if a zip file has a DS_Store entry in it... you can't delete those, it just closes the zip.

buthed010203: I mean... I've been doing it for months with no issues.

buthed010203: was a trivial fix either way

PR #6855, opened 2022-05-12
Adding Block: Heat Source

RebornTrack970:
A 1x1 Block that generates Infinite Heat. (For Sandbox)
Textures by: NullPointerException#7763


genNAowl:
Draft until sprite is made?

SMOLKEYS:
~~Better off making the value configurable, because there is a non-zero chance Anuke will add a melting point to blocks~~

RebornTrack970:
Compiled and Tested, the version works.

PR #6863, opened 2022-05-13
update bundle_zh_CN.properties

Lin-cat:
misc and \n


xem8k5:
very old?

Lin-cat:
> very old?

old, don't know why not merge

PR #6880, opened 2022-05-14
Misleading description fix - Spectre does not fire armor piercing bullets

MEEPofFaith:


MEEPofFaith:
~~why are there changes to a bunch of random other lines?~~

PR #6939, opened 2022-05-26
Achievements on non-steam versions

hortiSquash:
I added a desktop service if Steam is not launched that will register completed achievements.
I added an Achievement menu in the launch menu (based on the mod browser) that display all achievements and their completion.

image
Launch menu

image
Achievement menu:
completed achievements are being highlighted
you can search by name or description
textures are missing

image
Toast pop-up of a completed achievement (also missing textures)

Notes :
The Steam version keeps the exclusivity of achievement and player stats, including achievement progression. (rarity of achievements, etc)
image
This in-game achievement menu should work on Steam versions too (not the progression though)
Adding achievements through mods should also add them in the menu

In the future :
You can easily make a mobile version of this but i dont know how to use the GooglePlay nor AppleStore API.
I plan to add a lot more achievements (150+) to the game.

Special thanks to @buthed010203 and @JniTrRny for UI and cleanup


SMOLKEYS:
uwoooooooogh

VXsz:
> I don't really see the point of this when the achievements are supposed to be Steam-exclusive, and can be viewed much more comfortably in the Steam UI. I'm not going to put in extra effort into implementing a feature when it _actively de-values_ the Steam version when compared to the free one - it just doesn't make any sense.

This is just sad. Achievements should be available regardless of the platform & be viewable inside the game. 
In the case of Steam the UI elements could be disabled instead (like popup animation), or just be left in. A lot of games, like literally all Valve games, have it the same exact way as this PR; you can see it in the menu of the games, even see progress for the progression-based achievements & everything's fine. 

I think cloud saves & easy multiplayer are enough of an encouragement.
Also, Steam achievements happen in-platform, so the value isn't really lost? I can still showcase it on my Steam profile unlike people who didn't buy it where they only have it locally.

This PR looks really good as well :/
BTW I already have the game on Steam, I just don't think this is a good decision, nor does it "devalue" Steam version IMO. I mean look at valve games, c'mon man.

I hope this is reconsidered. I dislike having a degraded experience on my minimal laptop personally (like when I'm travelling etc).

PR #6944, opened 2022-05-28
Fixed loading mods with non-existent or disabled dependencies

Zelaux:
This error has existed since v6. But the first time I came across this, I prefer to make AnnotationProcessor for generate special class for cheching existing of dependencies.

[mods to check(version for v7)](https://github.com/Anuken/Mindustry/files/8792240/modsForCheching.zip)

### DIFF
--- a/core/src/mindustry/mod/Mods.java
+++ b/core/src/mindustry/mod/Mods.java
@@ -448,9 +448,8 @@ public class Mods implements Loadable{
     }
 
     private void resolveModState(){
-        mods.each(this::updateDependencies);
-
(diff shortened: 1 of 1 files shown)


Phinner:
Ma man should know I fixed this issue but I kept the loading of invalid mods because I did not know what to do with them, if you want an ez way to fix it, you can have a list of invalid mods here https://github.com/Anuken/Mindustry/blob/master/core/src/mindustry/mod/Mods.java#L819

Anuken:
Doesn't seem to be necessary anymore?

PR #6956, opened 2022-06-01
Merge functionality of getBlockI into unit getBlock

FlippingPotatoes:
Makes the old unit control getBlock have the same output options as the new getBlock for world logic (though of course with the same range restrictions as before) caveat being it outputs to one variable instead of two and uses an input of 0-3 to change its modes similarly to radar changing its order.

This lets players access floor and ore tile types and lets them read all wall tile types. It also makes "@ solid" redundant (you can still reach the same functionality of detecting "@ solid" by determining if it's a block without any related building, ie. all walls and cliffs will have a block type but no building)

Unlike in #5811, a similar pr, floor tiles and such have already been made into variables (for the new world processors) so we can actually compare things and make simple example pieces of logic like this:

getBlock Demo

(the logic basically just grabs the tiles in the copper square and draws them)

~~I basically just reused the cat's code so don't look at me if it's odd~~

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -121,7 +121,7 @@ committingchanges = Committing Changes
 done = Done
 feature.unsupported = Your device does not support this feature.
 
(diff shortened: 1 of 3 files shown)


FlippingPotatoes:
I also changed the description of getBlock to match the new functionality.

Anuken:
Useful as it may be, I can't merge this as it breaks older logic that uses this instruction.

PR #6959, opened 2022-06-02
Fix BeamDrill mining ore covered by buildings.

BlackDeluxeCat:
136p1
Beam drills can mine wall ores covered by any solid blocks, **including player buildings**. This looks funny but it is unreasonable. 

After fix:
fix1
Also fix crash by drawing null wall drop. It occurs if you break a building covering a wall ore while beam drill is mining it.


Anuken:
This has the right idea, but my fix is to make wallDrop return null for tiles with a building on them instead; it makes sense and leads to less code.

PR #6982, opened 2022-06-06
Add support for SRV records

Phinner:
Rejoice, no more need to create a PR for changing the IP of a failing server, SRV records will handle it!

Works wonderfully well, I tested it with 2 fallback servers on `play.xpdustry.fr`.

This PR should also be merged with a "domain over ip for usid" PR like #6649 or #4341, but without including the port due to the workaround I used to get the thing working.

Also, the used library works on Android but I can't get my sdk working so I can't add the necessary line to make it work, it's explained [here](https://github.com/dnsjava/dnsjava/blob/1c94536723eb58ce4140487443c0ad1eec39653f/src/main/java/org/xbill/DNS/config/AndroidResolverConfigProvider.java#L28).


buthed010203:
Is this really worth 600KB in deps tho?

Phinner:
Well, I did my research, and I couldn't find "raw" code for SRV record lookup so using this library was pretty much the only convenient option. But you are right, maybe I could make a stripped down version of it like rhino. Imma mark this as draft and work on it.

Phinner:
I finished the stripped down version, got a rid of at least 1/3 of the classes, I just need to switch the logging from slf4j to arc.

Phinner:
The PR is ready for review, here is some information :
- I managed to get IDE support working with android so I could add the dns resolver line but I still can't compile cauz Intellij dies.
- The stripped down version is available on my dnsjava fork on the mindustry branch, you can see the code [here](https://github.com/Phinner/dnsjava/tree/511fe15ca2a814ef9980be96c976e65cbb7b5bdb).
- For Windows, it is possible to include another dependency that can use its native dns resolver, adding it would look like : 
```gradle
configure([":server", ":desktop"].collect{project(it)}){
    dependencies{
        //Add jna for dns lookups on windows (dnsjava)
        implementation("net.java.dev.jna:jna:5.11.0")
        implementation("net.java.dev.jna:jna-platform:5.11.0")
    }
}
```
- Without any optional dependency, the stripped down version of the library is 300kb, the original library is 540kb, ain't bad innit ?

Phinner:
Well, its an optional dependency... No need to add that then :).

way-zer:
SRV query with pure UDP?
https://github.com/Anuken/Mindustry-Suggestions/issues/1616#issue-771807787

Phinner:
Well, you can steal the dns resolvers of dnsjava. But much tnaks for this much simpler code.

way-zer:
> dnsjava

seems not easy
Snipaste_2022-06-10_17-57-10

Phinner:
It's not that hard, I just finished doing it, you steal `ResolverConfig` and some `ResolverConfigProvider`, simplify them by removing ndots and searchPaths, and finally, you can get name servers by just doing `new ResolverConfig().servers()`.

Phinner:
Mmmmmh, imma rename them

Phinner:
Why Im drafting it again:
- Still have some exams to finish.
- Imma PR simple dns utilities to Arc net extension and then bring it in Mindustry.

Phinner:
How did it became a PR again ?

Phinner:
Mmmmh, it works :) You may test with `be.xpdustry.fr` that redirects to `n1.xpdustry.fr:8000`...

Phinner:
I should also mention I managed to compile on Android and it works there too :)

core/src/mindustry/net/ArcNetProvider.java
```diff
@@ -189,8 +192,27 @@ public void sendClient(Object object, boolean reliable){
     @Override
     public void pingHost(String address, int port, Cons<Host> valid, Cons<Exception> invalid){
         try{
-            DatagramSocket socket = new DatagramSocket();
+            var host = pingHostImpl(address, port);
+            Core.app.post(() -> valid.get(host));
+        }catch(IOException e){
+            if(port == Vars.port){
+                for (var record : ArcDns.getSrvRecords("_mindustry._tcp." + address)) {
```
Anuken, line 199: formatting, unnecessary spaces after for and before brace

core/src/mindustry/net/ArcNetProvider.java
```diff
@@ -189,8 +192,27 @@ public void sendClient(Object object, boolean reliable){
...
+                        return;
+                    }catch(IOException ignored){
+                    }
+                }
+            }
+            Core.app.post(() -> invalid.get(e));
+        }
+    }
+
+    private Host pingHostImpl(String address, int port) throws IOException {
```
Anuken, line 212: unnecessary space before {

Anuken (commented):
Code seems fine aside from minor formatting issues; I don't know how this system works in the first place, and I am not interested in utilizing it myself, so that part is up to you.

core/src/mindustry/net/ArcNetProvider.java
buthed010203: > unnecessary space before {

Anuke why do you not just...
```suggestion
    private Host pingHostImpl(String address, int port) throws IOException{
```

Anuken: g

Phinner:
Oh... https://github.com/Anuken/Arc/blob/master/extensions/arcnet/src/arc/net/dns/ArcDns.java#L123

Anuken:
Save as int, call readUnsignedShort() (or `& 0xffff`)

way-zer:
A new problem. `Call.connect` seems not support SRV @Phinner

Phinner:
Ok, well use `ArcDns` manually until I implement that.

Phinner:
@way-zer Are you sure it is necessary ? I mean after having looked at `Call.connect`, it is a direct connection to the remote server so I'm not sure if adding SRV lookup in there is a good idea.

way-zer:
> @way-zer Are you sure it is necessary ? I mean after having looked at `Call.connect`, it is a direct connection to the remote server so I'm not sure if adding SRV lookup in there is a good idea.

I think it necessary. As you having done, SRV provides fallback for multiple actual address, and `Call.connect` also needs this.

For my problem, I give a SRV address for player to connect, but I use it for `Call.connect`, it will fail.(Even it can resolve SRV in server, but I may prefer resolve it in client)

Anuken:
Does anyone still use this feature? Is it worth keeping around? I am re-implementing server pinging to avoid creating a new thread per ping, and I can't find a single server on the list to even *test* SRV records on.

way-zer:
> Does anyone still use this feature? Is it worth keeping around? I am re-implementing server pinging to avoid creating a new thread per ping, and I can't find a single server on the list to even _test_ SRV records on.
> 
> There was a server network (`cn1.meiqiumdy.top`) that _had_ SRV records, but they weren't actually being used for anything, as far as I could tell.

I perfer to keep SRV feature, and it just one UDP request that can be implemented with server pinging(Also UDP). If you need help, I can implement this feature again after you updating.

I removed SRV record from servers list as some dns don't support. And if I keep both SRV and CName record in list, it will show replicate two. More, one advantage of SRV is to hide port, that is useless for global servers list.

PR #6991, opened 2022-06-08
update pl server address, put into server category

Ilya246:
do not merge until @Somka000 and @BasedUser comment


BasedUser:
https://github.com/tgstation/tgstation/blob/master/sound/misc/highlander_delayed.ogg?raw=1

PR #7026, opened 2022-06-15
add config to setblock

BasedUser:
Makes world processors be able to build blocks with configs just like `ucontrol build`.


BasedUser:
Maybe it would instead be better to *force* placement of that block regardless if it's the same block on the same team with the same config.

Anuken:
Can't config be set as a separate instruction? This seems slower/bulkier and unnecessary.

BasedUser:
> Can't config be set as a separate instruction? This seems slower/bulkier and unnecessary.

It actually could be merged into `ControlI`. It seemed to me like using `setblock` for this purpose would make more sense seeing how builder AIs and players do exactly that.

core/src/mindustry/logic/LStatements.java
```diff
@@ -1170,10 +1175,9 @@ public boolean privileged(){
 
         @Override
         public LInstruction build(LAssembler builder){
-            return new SetBlockI(builder.var(x), builder.var(y), builder.var(block), builder.var(team), builder.var(rotation), layer);
+            return new SetBlockI(builder.var(x), builder.var(y), builder.var(block), builder.var(team), builder.var(rotation), builder.var(config), layer);
         }
 
-        @Override
```
code-explorer786, line 1176: what? this seems weird

BasedUser:
stale PR, and breaks with nulls :ranaisquish:

core/src/mindustry/logic/LStatements.java
BasedUser: when did that get rmoved, huh

PR #7049, opened 2022-06-20
ES Translation.

Volas171:


Angel-24:
There is already an existing PR for this translation bundle, where we already covered some of these changes. Please make your changes and suggestions [here](https://github.com/Anuken/Mindustry/pull/7034) to avoid confusion and favor a better organization for the other collaborators. Thanks.

PR #7112, opened 2022-07-03
revert the most vehemently OPPOSED zh-CN translation

Jackson11500:
In fact I hadnt seen anyone who accepted these two translation, they are too strange and ungrammatically, most of Chinese would not treat "毁伤"(scathe) as a noun. Just like some player said, translate like this would be even worst than English.
At the meantime, I would suggest anyone who submit new translation later should be made only after fully discussion, and in line with correct grammar, this should not be a problem for Chinese I believe.


Jackson11500:
If I may could add some suggestion, I would not support the translation for a 1.4 billions country should be DECIDED just by a small groups that contains even no more than 20 people, especially for a large release like V7. 
Take some server communities for a example, the best situation is the players are told that they have decided ** have translated to **, and the translation have pr&merge. Not need to say single player which access the mindustry by steam or github, which would even lose the chance to discuss. However, there are more times that most of players have been tolded nothing before the tranlations have made. Shouldnt translation for a large release draft pr and merge only after being fully discussed like v126 or v105 for a large country?

Prosta4okua:
@Jackson11500 the dev doesn't know Chinese (and other languages), so the whole translation process depends on translators. You can also collaborate with other translators to proofread the translation and improve it's quality.

Jackson11500:
> @Jackson11500 the dev doesn't know Chinese (and other languages), so the whole translation process depends on translators. You can also collaborate with other translators to proofread the translation and improve it's quality.

I know I know, but the current situation is different. In previous version, zh_CN translation are broadly discussed at most of time, at least hundreds of players participated and discussed across many groups, while translation pr draft could lasts for at least few days to see if there are any suggestions from some single player. Never had this before that some very few people determined the translation, and some translation quality is so poor that nearly no person would support this translation (Almost everyone who see these v7 translation would have "WTF" at the first few sentense). That's the reason I hope Anuke could suspend merging the zh_CN translation and revert some very badly translation.
(https://github.com/Anuken/Mindustry-Suggestions/issues/3696)

Anuken:
Can you look at the [blame/changelog](https://github.com/Anuken/Mindustry/commits/master/core/assets/bundles/bundle_zh_CN.properties) for these translations, and let me know who is responsible for the bad changes? I can revert these changes if necessary. I do agree that there should be more organization in regards to what gets merged.

Jackson11500:
> Can you look at the [blame/changelog](https://github.com/Anuken/Mindustry/commits/master/core/assets/bundles/bundle_zh_CN.properties) for these translations, and let me know who is responsible for the bad changes? I can revert these changes if necessary. I do agree that there should be more organization in regards to what gets merged.

While some translation covering tower or units have received boardly criticisms, the other translation on settings or hint are fine. But most of xem's translation commits contains both of them, and thus its hard to say which one or some commits should be reverted.Thats why I just remove the most opposed two translations in this pr.
As far as I know, some players are trying to rework the v7 units and buildings translation, hope that the future translation could be better

Anuken:
Okay, I will revert for now. How would you propose the "correct" translation be coordinated? Where should this be discussed? How will I know which translation should or should not be merged?

Jackson11500:
> Okay, I will revert for now. How would you propose the "correct" translation be coordinated? Where should this be discussed? How will I know which translation should or should not be merged?

Thanks a lot! Currently, the v7 translation has lead to a heated discussion in many CN groups, I have heard that lots of players are trying to give their own version. Usually, there would be some representatives on github on behalf of their groups' opinions and discussed on one or few PR. If most of players are agreed with the translations of one certain PR, there would be words like 'please merge this' or something after serveral days. :)

PR #7122, opened 2022-07-04
Logic Cutscene Shake

Mythril382:


Mythril382:
This requires testing, I have no idea if this works because I use a phone.

Catchears:
Code compiles, but there are a few potential issues.

The `shake` text ends up with a newline in the selection box:
java_x26R9HhL6R

The selection box for `cutscene` is 1x4 while all other selection boxes with >=4 options go 2x2, not sure how important this is though.

It's also unclear in which unit duration is.

If you use it without a `stop` command afterwards it locks your camera in the bottom left corner of the map, but this seems to be true for `cutscene zoom` as well.

PR #7126, opened 2022-07-05
Deprecate DrawFrames

Goobrr:
DrawFrames is a terrible legacy drawer that is *only* used by the spore press. Even then, the spore press' animation can be replicated with DrawPistons, which i did.

I did not remove the class yet as it may be used by mods, and last time a class was removed without warning everyone got mad for a few days.


MEEPofFaith:
> YEEEEEEESSSSSS

n't

PR #7182, opened 2022-07-16
distributorium rename to .pl

Ilya246:
this has already been discussed by us and does not need comment confirmation like the previous PRs


BasedUser:
ok

Somka000:
ko

PR #7200, opened 2022-07-19
Update bundle_th.properties

JniTrRny:


KayyAyeAre:
merui-chan?

Anuken:
~~pretty sure the bundle update script is going to do strange things to do the extra lines~~

PR #7230, opened 2022-07-23
update heat source description to match other sources

Catchears:

- [N/A] I have ensured that my code compiles, if applicable.
- [N/A] I have ensured that any new features in this PR function correctly in-game, if applicable.

Just noticed that the heat source description was quite different compared to the others while translating and thought I'd suggest this for uniformity.


yesC48 (commented):
ps: should be "Outputs large amounts of heat. Sandbox only." because it outputs 1000 heat units for free

Catchears:
Ah - saw the number but assumed it was a similar order of magnitude to the power source.

PR #7346, opened 2022-08-11
Less ResetEvent

way-zer:
Less ResetEvent emit twice.
In many case, Call `SaveIO.load` will emit `ResetEvent` twice.

Another reason:
I want to do something between `ResetEvent` and `WorldLoad`, but it seems very hard as all `SaveIO.load` begin with a `logic.reset`

### DIFF
--- a/core/src/mindustry/core/Control.java
+++ b/core/src/mindustry/core/Control.java
@@ -347,6 +347,7 @@ public class Control implements ApplicationListener, Loadable{
     public void playMap(Map map, Rules rules, boolean playtest){
         ui.loadAnd(() -> {
             logic.reset();
+            SaveIO.justReset = true;
             world.loadMap(map, rules);
             state.rules = rules;
             if(playtest) state.playtestingMap = map;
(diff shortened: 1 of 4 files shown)


Anuken:
> I need to do some patch before world load like Blocks.arc.health
And patch will reset in ResetEvent.

Can't you do something like store whether something is "patched" or not, then roll it back in ResetEvent, and do nothing if it's not patched?

First event -> `arc` is patched -> reset `arc`, set patched to false
Second event -> `arc` is **not** patched, nothing happens

way-zer:
the problem is that
i need to patch before world load, but will be this ResetEvent affect. make
 the patch also reset.

I don't want to do special check to pass that reset. It is strange that not reset when ResetEvent.

way-zer:
These patch is with map lifecycle. Load before a map, reset after a map.

This mod give larger freedom to map author and server, modify content properities.

Kieaer:
There are UI issues, but you can set block health to whatever value you want.
use ``Blocks.arc.health = <value>``

way-zer:
> There are UI issues, but you can set block health to whatever value you want. use `Blocks.arc.health = <value>`
I maybe want a event before WorldLoadEvent and after ResetEvent, which emit just in the start of `endMapLoad`

way-zer:
#7523

PR #7370, opened 2022-08-15
SetProp world processor instruction

gorodmi:

adds a new world proc instruction that can set properties
java_pOQOEeyTTz (sets carbide count of container1 to c1)
this would be very useful in world proc maps (example: consuming/producing items without setblock shenanigans)


code-explorer786:
he also uses intellij idea ultimate, respect

Anuken:
This is really, really broken. I don't mean to be rude but it's clear that this was not tested properly at all, because some of the property operations here screw up the game in **major** ways, or don't make any sense whatsoever. It's hard to even specify all the issues because *most* of the properties here are broken. 

Some examples:

- You let it set size of blocks. **Why on earth?** This is a global value that affects every block and persists until restart. Easy way to screw up an entire game or save.
- Setting position of buildings is nonsense.
- Many other global values are settable, maxHealth, power capacity, etc - **these are all globals**, never edit them.
- No checks for negative values for things like items or liquids.
- Pointless property edits like power status, heat or efficiency, that will be reset immediately (you can't just make a block "have power", that's broken)
- Buggy behavior (making a unit "not dead" kills it)
- Using `LAccess` when a new enum should be used; many properties are read-only (again, size, capacity, etc) and you are only causing confusion by re-using this enum. 

Closing as this is basically beyond fixing, needs a re-write. Maybe I'll try doing it myself later.

code-explorer786:
@gorodmi 😭

PR #7413, opened 2022-08-22
Update servers_v7.json

Volas171:
Ownership change of AAS, now Eradicationdustry.


baseduse:
you mean, half of it has changed ownership

PR #7442, opened 2022-08-25
More control over wave sending

MEEPofFaith:
- New rule: `waveSending`. If set to false, the send wave button is disabled.
- `Spawn Wave` command changed: Can now be set to spawn a natural wave (The same as the send wave button)

https://user-images.githubusercontent.com/54301439/186617519-1b83c8e8-cd7e-4e14-8dea-213e5d13d6a9.mp4

Can be useful for world procs sending some waves (with drop zone kill), then later giving the control to the player.

### DIFF
--- a/core/src/mindustry/logic/LStatements.java
+++ b/core/src/mindustry/logic/LStatements.java
@@ -1291,15 +1291,12 @@ public class LStatements{
 
     @RegisterStatement("spawnwave")
     public static class SpawnWaveStatement extends LStatement{
-        public String natural = "false";
-        public String x = "10", y = "10";
+        public String x = "10", y = "10", natural = "false";
 
         @Override
         public void build(Table table){
             table.add("natural ");
-            fields(table, natural, str -> {
-                natural = str;
-            });
+            fields(table, natural, str -> natural = str);
 
             table.add("x ").visible(() -> natural.equals("false"));
             fields(table, x, str -> x = str).visible(() -> natural.equals("false"));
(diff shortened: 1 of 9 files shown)


core/src/mindustry/logic/WaveType.java
```diff
@@ -0,0 +1,8 @@
+package mindustry.logic;
+
+public enum WaveType{
```
Anuken, line 3: An enum with two values is pointless, just use a boolean instead.

MEEPofFaith:
https://user-images.githubusercontent.com/54301439/186734375-bafea54a-49dc-4f2b-87b6-d5283136d0b3.mp4

~~You know, I feel like there should be a selector for booleans similar to enum selection~~

MEEPofFaith:
https://user-images.githubusercontent.com/54301439/186806347-2e5397af-61f2-4a1a-bbbb-cc4aef25b98c.mp4

Anuken:
I don't see the point of making these instructions accept hard-coded booleans instead of variables, and you've almost certainly broken a lot of existing world logic with this change without good reason.

MEEPofFaith:
I mean there are some parameters in logic commands that only accept booleans, but fair enough.
That could be left to a separate PR later if people get tired of manually typing "true" or "false."

Anuken:
> That could be left to a separate PR later if people get tired of manually typing "true" or "false."

That would be a very unreasonable thing to get tired of, I am forced to do it all the time in Java and it's fine.

core/src/mindustry/logic/LStatements.java
```diff
@@ -1291,16 +1291,21 @@ public LCategory category(){
 
     @RegisterStatement("spawnwave")
     public static class SpawnWaveStatement extends LStatement{
+        public String natural = "false";
```
Anuken, line 1294: Would be more backwards-compatible to have this as a last parameter so existing logic doesn't break.

PR #7453, opened 2022-08-26
shield breaker rename

yesC48:
you should

rename shield breaker to shield annihilator.


Anuken:
That sounds rather silly.

MEEPofFaith:
Shield Skibidi bop mm dada-er

PR #7518, opened 2022-09-05
Upload game data to custom cloud 

hhjxx:
1. Export and upload the game data to a http server.
2. Download and import the data from the http server.
Storage server sample: https://github.com/hhjxx/CloudSave


Darkness6030:
bro
thats amazing, but anuke usually doesnt accept PRs with new libraries
is it possible to make this without any new libraries, only with ARC?

hhjxx:
> bro thats amazing, but anuke usually doesnt accept PRs with new libraries is it possible to make this without any new libraries, only with ARC?

thanks for your advice, let me try replace it with ARC

Anuken:
As I understand it, this requires a user to host their own server for cloud saves. I think very, very few users would utilize such a feature. It's far easier to just set up folder sync with an existing cloud storage provider.

hhjxx:
> As I understand it, this requires a user to host their own server for cloud saves. I think very, very few users would utilize such a feature. It's far easier to just set up folder sync with an existing cloud storage provider.

How about using github as a cloud storage provider, upload the zip data to a repository and download from it then import it. @Anuken can you advise?

Anuken:
That sounds like a horrible idea still, and very few would use it regardless.

**I do not wish to add cloud storage to the game itself.** Just use folder sync.

PR #7520, opened 2022-09-05
Generation quick copy & paste

JniTrRny:
small QOL changes :)

https://user-images.githubusercontent.com/85090668/188493614-805233b2-6624-4bea-83f3-f40a334d37ba.mp4

Middle click to copy, Right click to paste.
won't paste on options that doesn't allow that block.

also add ignore option to clear filter, that will help me a lot in terrain designing.

that's... it?
oh, also made the load dialog more... fancy.

Notes:
- is using block ID as a paste a good decision? should it be name instead?
- how to tell player that this feature exists? well I guess you don't.

---


Anuken:
> is using block ID as a paste a good decision? should it be name instead?

Yes. Names rarely change, IDs do so frequently. It's also far more intuitive for the clipboard to contain, say, "stone" rather than some number.

JniTrRny:
Got it.

PR #7527, opened 2022-09-07
[Plugins] Added MessageHandler for handling messages & commands.

Darkness6030:

- [x] I have ensured that my code compiles, if applicable. [Compiles]
- [x] I have ensured that any new features in this PR function correctly in-game, if applicable. [Tested]

I need this for my plugin to work correctly so why shouldn't i add it ;-;

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -227,47 +227,8 @@ public class NetClient implements ApplicationListener{
         if(message.length() > maxTextLength){
             throw new ValidateException(player, "Player has sent a message above the text limit.");
         }
-
-        message = message.replace("\n", "");
-
-        Events.fire(new PlayerChatEvent(player, message));
-
-        //log commands before they are handled
-        if(message.startsWith(netServer.clientCommands.getPrefix())){
-            //log with brackets
-            Log.info("<&fi@: @&fr>", "&lk" + player.plainName(), "&lw" + message);
-        }
-
-        //check if it's a command
-        CommandResponse response = netServer.clientCommands.handleMessage(message, player);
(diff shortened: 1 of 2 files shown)


Darkness6030:
Wait why didn't you compile

Phinner:
I don't think this is a good idea.

I mean, when a message is sent, a regular user or a plugin developer expect several things :
- `PlayerChatEvent` to be triggered.
- The player messages to be logged.
- The chat filters to be applied.
- etc...

But your PR delegates all these things into this `MessageHandler`, meaning this class has multiple responsibilities towards the game code.

This is a problem because a bad implementation or a simple mistake can lead to unpredictable behavior not so easy to debug.

> Example: Let's say you make a plugin that just changes the default `MessageHandler` but you forget to trigger the `PlayerChatEvent` in it. After that, you add an anti-afk plugin that record player movements and player messages to detect whether they are active or not. So now, because `PlayerChatEvent` is never fired, the anti-afk will start to kick players that are just talking. And when this will occur, the player will think he has been kicked because he was not moving. He has a very low chance of reporting this unwanted behavior.

That's why we split different responsibilities into different units of code like `TeamAssigner` assign teams, `ChatFormatter` formats chat messages, etc...

way-zer:
In fact, there are many way to handle message.
Like some in my [plugin](https://github.com/way-zer/ScriptAgent4MindustryExt/blob/v3.0.0/scripts/coreMindustry/utilNext.kts) :
Snipaste_2022-09-11_00-21-40

Anuken:
This also breaks the API... so, no.

Darkness6030:
Ok, btw thanks @way-zer for showing an example, will use this

PR #7529, opened 2022-09-08
deconstructor name change (https://github.com/Anuken/Mindustry-Suggestions/issues/3942)

Catchears:
Implements https://github.com/Anuken/Mindustry-Suggestions/issues/3942.

I'm not sure if this is too significant a change to make, but having it consistent with the constructor would be nice, so I thought I'd make it easy to change if you wanted to.

I split this up into 2 commits to avoid it looking like the 'normal' deconstructor being shrunk - depending on how important this is and whether or not it gets merged squashing may or may not be worth it.

I also updated the bundle entries in all languages manually because `updateBundles` made a slight mess of it - I assume it's not made to deal with renames. The translations themselves would have to be updated by the translators though.

- [X] I really hope I didn't mess anything up.


Anuken:
This breaks saves, and as such it isn't acceptable. I've already made some changes locally that simply change the display name of the blocks, which should be enough for users.

Catchears:
Ah, makes sense. Thanks for the explanation!

Issue #69, opened 2018-01-09
Conveyor Tunnels 


ICrowleyI:
A Conveyor Tunnel leading directly into the base, will not add ressources to your stock.
As seen in the video, coal reaches the tunnel, but is not added to my resources.
![ezgif-3-fd44035844](https://user-images.githubusercontent.com/30113543/34748950-7770673c-f59f-11e7-9fff-79604c1f55b4.gif)

PR #7562, opened 2022-09-13
Why it uses name instead of id?

Darkness6030:

### DIFF
--- a/core/src/mindustry/ui/fragments/PlayerListFragment.java
+++ b/core/src/mindustry/ui/fragments/PlayerListFragment.java
@@ -200,7 +200,7 @@ public class PlayerListFragment{
 
                 button.button(Icon.hammer, ustyle,
                     () -> ui.showConfirm("@confirm", Core.bundle.format("confirmvotekick",  user.name()),
-                    () -> Call.sendChatMessage("/votekick " + user.name())))
+                    () -> Call.sendChatMessage("/votekick #" + user.id)))
                 .size(h);
             }
 


Anuken:
PRs must be tested.

Darkness6030:
Ok, fine, will test today

Darkness6030:
And yes, it works

core/src/mindustry/ui/fragments/PlayerListFragment.java
```diff
@@ -200,7 +200,7 @@ public void draw(){
 
                 button.button(Icon.hammer, ustyle,
                     () -> ui.showConfirm("@confirm", Core.bundle.format("confirmvotekick",  user.name()),
-                    () -> Call.sendChatMessage("/votekick " + user.name())))
+                    () -> Call.sendChatMessage("/votekick #" + user.id)))
```
MEEPofFaith, line 203: This is only telling the user to type in `/votekick #69`.

Does the command `/votekick #69` actually work though?

~~(I probably shouldn't be asking this, I don't do any sort of server stuff.)~~

Anuken: I would hope so, considering he said it was tested.

Darkness6030:
@MEEPofFaith just download last BE build, join any server and try to kick any user. It will work

Darkness6030:
I committed this because while player clicks the kick button and sends the packet, the server can theoretically change the target's name (for example rainbow nickname plugin) and the votekick won't happen

PR #7588, opened 2022-09-17
editing ip address!

GTMZOMEAGEAmazon:
The server ip address is wrong

### DIFF
--- a/servers_v7.json
+++ b/servers_v7.json
@@ -122,6 +122,6 @@
   },
   {
     "name": "LightDustry",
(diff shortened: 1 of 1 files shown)


GTMZOMEAGEAmazon:
my server ip was wrong i fixed it

Anuken:
Neither the old nor new addresses point to functional servers, and the change makes no sense.... why are only two of them "lightdusy"?

GTMZOMEAGEAmazon:
> Neither the old nor new addresses point to functional servers, and the change makes no sense.... why are only two of them "lightdusy"?

because the light industry ip belongs to a different server from the other ips (the lightdustry ip is the server I rent and the lightdusy ip is the self host)

PR #7593, opened 2022-09-18
Fancy custom map play dialog

Goobrr:
Layout for maps without description
image

Layout for maps with description
image

Mobile layout, description is always hidden.
image

I know the contribution guidelines advises against making single-use methods, but inlining the build methods just makes the already messy code look messier.


Mythril382:
Maybe a collapser for description in mobile?

Anuken:
While this looks nice, I think there are several issues with re-doing the UI like this at the moment:

- It's not consistent with the current UI in other dialogs
- It changes a very frequently used part of the game before a major release
- It may or may not need stylistic changes (are we absolutely sure this is the best style to go with? maybe it would be better to go full material with drop shadows and such)

This isn't to say that this PR is being rejected, I just can't merge it before v7 release as a UI rework will inevitably have bugs and issues that will slow down the release. I'll wait until after the release before further investigating it.

PR #7595, opened 2022-09-18
Support for custom menu buttons

Mnemotechnician:
Added support for custom main menu buttons, similarly to custom categories in SettingsMenuDialog. Mods can register new menu buttons by using one of the `MenuFragment.addButton` methods, e.g. `Vars.ui.menufrag.addButton("die", Icon.none, () -> throw Error())`. For convenience, the quit button is always kept in the bottom right / bottom part of the menu, so the custom buttons are inserted between it and other buttons.

MenuFragment.Buttoni was made public and is now used to store the registered "button definitions"... I'm not sure if this is the best approach, but it works.

For example, running the following code in the console
Miserable_situation_20220918_170317
Results in this:
Miserable_situation_20220918_174142

### DIFF
--- a/core/src/mindustry/ui/fragments/MenuFragment.java
+++ b/core/src/mindustry/ui/fragments/MenuFragment.java
@@ -122,7 +122,7 @@ public class MenuFragment{
         Seq<MobileButton> customs = customButtons.map(b -> new MobileButton(b.icon, b.text, b.runnable == null ? () -> {} : b.runnable));
 
         if(!Core.graphics.isPortrait()){
-            container.marginTop(60f);
+            container.marginTop(60f).left();
             container.add(play);
             container.add(join);
             container.add(custom);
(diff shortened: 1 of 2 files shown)


Anuken:
This can be problematic, because there is absolutely no mechanism for scrolling the buttons, and even ~2 new buttons can easily get cut off on certain screens.

Mnemotechnician:
Uh, maybe just make the container a scroll pane?

Mnemotechnician:
Well, idk, i tried to make the scroll pane look better but couldn't think of any way... However, normally the menu shouldn't be big enough to be scrollable, so i think it's ok


https://user-images.githubusercontent.com/69920617/190924594-6df49073-bfe7-4f57-b443-485b4b789687.mp4

SMOLKEYS:
what about vertical screens

Mnemotechnician:
> what about vertical screens

Same thing. ScrollPane supports scrolling in both directions by default, so it will be scrollable vertically (but i doubt anyone will ever need that)

PR #7606, opened 2022-09-22
Configurable Rates on Some Source Blocks

MEEPofFaith:
image

https://user-images.githubusercontent.com/54301439/191634370-aa31c31a-82af-47bb-a01b-7cdd1dfce9e2.mp4

Does not break saves, I've tested.

---


hortiSquash:
> one block having multiple configs

logic processor -> there is a pencil icon under the block to enter the other mode
if that can give you an idea, i already made it here, i suck at UI but its usable https://github.com/Anuken/Mindustry/pull/6398

MEEPofFaith:
Yep, that works.

https://user-images.githubusercontent.com/54301439/192026132-caa60b85-506d-4d66-a084-6d4df043c1cb.mp4

SMOLKEYS:
finally

after so long

it has been made

*shall we begin?*

hortiSquash:
does the power source support negative values? thats very useful for testing schematics

MEEPofFaith:
> does the power source support negative values? thats very useful for testing schematics

No it does not. I guess I could make power void configurable.

MEEPofFaith:
https://user-images.githubusercontent.com/54301439/192432883-57155526-7773-46a2-9fb5-cbbf53e04a80.mp4

SpaceEarthEX:
This is what i needed finally

sk7725:
Would there be a way to visualize how powerful the sources are?
Maybe make the center dot go lerp(Color.orange, lerp(Color.yellow, Color.green)) based on fractiont (keep in mind, you need to cap the input as absurdly high values will cause bugs at high deltatime, and may lead to float.Infinity in some operations)
Also, adding a value preview when mouse is hovered (like message blocks) would be nice.

MEEPofFaith:
> Also, adding a value preview when mouse is hovered (like message blocks) would be nice.

https://github.com/Anuken/Mindustry/assets/54301439/95ed1cb6-7b1b-496b-bf23-bca3d91536aa

Anuken:
After some consideration, I think it's better that I not merge this, for two reasons:

1. Minor technical problems: Power source's config() no longer being a `Point[]` means certain rotation/flipping code may not work correctly anymore. This would need to be carefully investigated and fixed.
2. Clutter: In *99%* of scenarios, I do not see myself using the configuration UI. I *could* theoretically see myself needing it for some obscure power bug that requires low consumption/output, but there are plenty of simpler ways of replicating that in the game already. I could also see designers or modders using it (maybe?) for testing schematics or blocks, but that seems like a very niche use case. I *don't* want the default power node/void having extra UI that blocks me from clicking other blocks when that is unnecessary in 99% of situations.

I think this would be a good addition to a dedicated testing mod (if it isn't already), but not vanilla.

MEEPofFaith:
> I think this would be a good addition to a dedicated testing mod (if it isn't already), but not vanilla.
:3

PR #7624, opened 2022-09-25
Added teamset command to ServerControl.java

Bumpkin-Pi:
Added a teamset command that ServerControl.java with the syntax:

teamset <username> [team]

Team can be either the in-code name for the team, or its corresponding color.


way-zer:
May plugin be a better choose?
In fact, there are always many plugins including this function. Like mine in https://github.com/way-zer/ScriptAgent4MindustryExt/blob/7.5/scripts/wayzer/map/betterTeam.kts

Anuken:
Honestly, I think this is approaching "too specific" territory, and the implementation here is awful (manual switch instead of a lookup by name). It's a one-liner with the JS command, and I don't think dedicated servers often need to manually move players between teams.

PR #7627, opened 2022-09-26
[Refactor] Server cleanup and more

Darkness6030:
I did a small refactoring, but didn't break the formatting.
Also i added the ability for mods to change the default GameOverEvent listener so you can add a custom map loader or custom GameOver message
Also i made two fields public. Why? Because using getters and setters in this class is pointless

- [x] I have ensured that my code compiles, if applicable. _It compiles_
- [x] I have ensured that any new features in this PR function correctly in-game, if applicable. _It works_

### DIFF
--- a/server/src/mindustry/server/ServerControl.java
+++ b/server/src/mindustry/server/ServerControl.java
@@ -2,6 +2,7 @@ package mindustry.server;
 
 import arc.*;
 import arc.files.*;
+import arc.func.Cons;
 import arc.struct.*;
 import arc.util.*;
 import arc.util.Timer;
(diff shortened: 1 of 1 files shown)


Darkness6030:
What about all the other changes?

Darkness6030:
_I'll try to replace "lastServerMode" with "Vars.state.rules.mode()" and test how it works_

Darkness6030:
And yes, it works correctly for any gamemode. Also, it will choose the next map based on the gamemode of the current one, which is better.

Darkness6030:
@Anuken now nextMapOverride is documented, lastServerMode is replaced and removed.
What about all the other changes?

PR #7650, opened 2022-10-02
fix player name color on minimap

Darkness6030:
So now players without color codes in their names look like this
image
image

But i fixed it ;-;

- [ ] I have ensured that any new features in this PR function correctly in-game, if applicable. (Didn't test, but should i?)


Anuken:
There's already a team color parameter. This 1) seems like an issue for PvP where colors are important, and 2) absolutely needs testing.

Darkness6030:
ok, will test in 1-2 hours 👀

Darkness6030:
Yes, it shows the right color in any situation, but yes, it doesn't show the team. 
BUT on PvP the player's unit is already colored in color of his team, so i don't think it's a problem.

Anuken:
Realized that playerName() allocates and is pointless since you can just supply the player color as a parameter.

PR #7652, opened 2022-10-03
Minor map fixes

Dexapnow:
Changes to maps are:
- all maps got their block hp maxed out so that there are no longer blocks that are relicts from past when they had less max hp. (done manually so I might have missed something)
- tarfields no longer has inaccessible air pockets and vissible map edge. (now it's 260x260 instead of 250x250)
- player base on saltflats got cleaned up, crux got their batteries connected.
- overgrowth no longer has drills on sand. (that were next to salvos)


Somka000:
Dex Sex

PR #7702, opened 2022-10-12
Expand getBlock functionality

FlippingPotatoes:
Let's the player use getBlock to detect water, floor, and ore tiles.

A reimagining of pr #6956, but now it doesn't break old logic, and still adds the majority of its function.

~~Simply adds a boolean input to getBlock (in the same fashion as unitLocate's enemy boolean) that lets you choose to search for floor tiles instead of block tiles.~~ When used and a non-air overlay tile is present (ie. ores) it will default 'floor' to that, otherwise it will return the floor tile in 'floor'.

The main goal here was being able to make cool logic-based maps, detect water to allow (shitty) pathfinding, and have a more precise method of detecting ores.

Tested to make sure old logics are unchanged on this branch (simply because you already have 5 variables for unitControl thanks to build, all the considered problems are already solved)

Also changed the bundle description of getBlock to reflect this.

~~clearly me waiting two months to try this again was optimal, just attempting to squeeze this into v7~~

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -510,20 +510,20 @@ public class LExecutor{
                         if(!unit.within(x1, y1, range)){
                             exec.setobj(p3, null);
                             exec.setobj(p4, null);
+                            exec.setobj(p5, null);
                         }else{
                             Tile tile = world.tileWorld(x1, y1);
                             if(tile == null){
                                 exec.setobj(p3, null);
+                                exec.setobj(p4, null);
+                                exec.setobj(p5, null);
                             }else{
-                                if(exec.bool(p5)){
-                                    //Allows reading of ore tiles if they are present (overlay is not air) otherwise returns the floor
-                                    exec.setobj(p3, tile.overlay() == Blocks.air ? tile.floor() : tile.overlay());
-                                }else{
-                                    //any environmental solid block is returned as StoneWall, aka "@solid"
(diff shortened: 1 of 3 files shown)


core/src/mindustry/logic/LExecutor.java
```diff
@@ -512,10 +512,19 @@ public void run(LExecutor exec){
...
                         }else{
                             Tile tile = world.tileWorld(x1, y1);
-                            //any environmental solid block is returned as StoneWall, aka "@solid"
-                            Block block = tile == null ? null : !tile.synthetic() ? (tile.solid() ? Blocks.stoneWall : Blocks.air) : tile.block();
-                            exec.setobj(p3, block);
-                            exec.setobj(p4, tile != null && tile.build != null ? tile.build : null);
+                            if(tile == null){
+                                exec.setobj(p3, null);
+                            }else{
+                                if(exec.bool(p5)){
```
Anuken, line 518: This is very confusing; documented or not, most people would expect the 'floor' parameter to actually contain the floor instead of a boolean that is read.

FlippingPotatoes: How do you feel about changing the 'floor' parameter to 'tryFloor' or even 'isFloor'?

Anuken: Why not just make it `floor` and write into it?

FlippingPotatoes: Sometimes, I am dumb lmao

FlippingPotatoes:
Now just has the extra 'floor' parameter that gets set to either the ore or the floor value. (re-tested with some simple logic, still forward compatible etc.)

PR #7766, opened 2022-10-25
Add logic operator "char" for get char x

noverd:
Add logic operator "char" for get char x


Anuken:
I will not allow string operations in logic, especially ones that allocate.

noverd:
why?

Anuken:
Logic is already laggy enough; string allocations and comparisons make it even worse. I do not want to encourage them.

I see no good reason to do character operations.

PR #7775, opened 2022-10-27
contrast on large paydriver is way too high

stacktrace-error:
neither of these solutions is great but I think this is slightly better


Anuken:
Yes... I tried low contrast at first, looked bad. I'm not happy with either.

stacktrace-error:
I could try making it use a slightly different version of the palette instead

stacktrace-error:
image
this should do

PR #7796, opened 2022-11-01
fix factory plan icon scaling, unhardcode node beam regions

stacktrace-error:
Currently the game just recolors the existing power beam regions when you change its color parameters, which tends to make getting the colors you want difficult.

Also fixes this
image

### DIFF
--- a/core/src/mindustry/world/blocks/power/PowerNode.java
+++ b/core/src/mindustry/world/blocks/power/PowerNode.java
@@ -29,8 +29,7 @@ public class PowerNode extends PowerBlock{
     /** The maximum range of all power nodes on the map */
     protected static float maxRange;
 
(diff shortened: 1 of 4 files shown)


core/src/mindustry/world/blocks/power/PowerNode.java
```diff
@@ -29,8 +29,7 @@ public class PowerNode extends PowerBlock{
     /** The maximum range of all power nodes on the map */
     protected static float maxRange;
 
-    public @Load("laser") TextureRegion laser;
```
Anuken, line 32: Don't remove the annotation. You can use `@-laser` and `laser` as a fallback.

stacktrace-error:
I didn't change the long node since it didn't have one in the first place, is it fine now?

PR #7809, opened 2022-11-03
Precision errors in float to string methods

matusm1:
Related to Anuken/Arc#124, "About the other autoFixed()" paragraph. Since the pull request got accepted, it is necessary to copy the changes from Arc's `autoFixed(value, maxPrecision)` to Mindustry's `fixValue(value)` to make sure a correct precision is used.

The two methods were practically the same, so I replaced all occurrences of `fixValue()` with `autoFixed()` and made `fixValue()` deprecated. As far as I know `fixValue()` was used only in the StatValues class and nowhere else (not even some mods I briefly checked). I was tempted to delete it altogether, but decided to deprecate it instead since it is a public method.

I managed to find only 2 stats that were not displayed properly before the fix: Precept's speed (4.80 -> 4.8) and healing from Overdrive status (0.60 -> 0.6).

----

### DIFF
--- a/core/src/mindustry/world/meta/StatValues.java
+++ b/core/src/mindustry/world/meta/StatValues.java
@@ -34,15 +34,13 @@ public class StatValues{
         return table ->  table.add(!value ? "@no" : "@yes");
     }
 
-    /** @deprecated Use Strings.autoFixed(value, 2) instead. */
-    @Deprecated
     public static String fixValue(float value){
         return Strings.autoFixed(value, 2);
     }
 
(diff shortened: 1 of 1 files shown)


Anuken:
Why deprecate/replace it when you can just change the implementation and keep everything the same?

matusm1:
Because it seemed redundant and was not used much. But I agree, changing the implementation is the safest way to go about it.

Anuken:
It's still useful enough, since it enforces a "default behavior" of fixing to two decimal points.

PR #7814, opened 2022-11-04
Show Build Cost Even When In Sandbox

MEEPofFaith:
javaw_3KctlmSfLF

I don't see a reason why for the cost to be replaced with *. Just saves the very short amount of time it takes to just open the block info and check the requirements from there.


MEEPofFaith:
~~now replace the * with infinity symbols~~

PR #7871, opened 2022-11-11
Remove dead servers

Kieaer:
Check date: Friday 2:23pm (UTC +0)

### DIFF
--- a/servers_v7.json
+++ b/servers_v7.json
@@ -1,9 +1,5 @@
 [
   {
-    "name": "Mindustry Español",
-    "address": ["n1.xpdustry.fr:6001"]
-  },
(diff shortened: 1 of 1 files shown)


Phinner:
🤨 How come Xpdustry servers get yeeted ? They are still up.

Kieaer:
> 🤨 How come Xpdustry servers get yeeted ? They are still up.

image

Phinner:
I use SRV records, that's why it does not work.

Anuken:
You should PR it back

way-zer:
Oh, Why my server get deleted without any notification??
That is annoying.

PR #7933, opened 2022-11-17
Selecting only 1 block as a schematic picks it instead

Garen7:
Selects the block so it can be used for dragging a line of the block and for the UI helpers such as the drill range one the Plasma Bores. Doesn't select the block instead if its something that you might want to save as a one block schematic such as a logic block. I coded this for mobile too originally but tbh it just felt easier to select the block the normal way instead of copying it so I didn't add that to the PR.

Note that unlike the Pick block hotkey this does not change the category on the block selector panel since afaik there isn't a nice way to do that without being in the PlacementFragment class, not sure if that's a deal breaker but I didn't notice any bugs as a result of that.

### DIFF
--- a/core/src/mindustry/input/DesktopInput.java
+++ b/core/src/mindustry/input/DesktopInput.java
@@ -21,7 +21,6 @@ import mindustry.game.Teams.*;
 import mindustry.gen.*;
 import mindustry.graphics.*;
 import mindustry.input.Placement.*;
-import mindustry.type.Category;
 import mindustry.ui.*;
 import mindustry.world.*;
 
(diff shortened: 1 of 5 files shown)


LeoSko:
Why would one need that if there is a pick block hotkey again?

Garen7:
Why not merge the two hotkeys and remove the pick block hotkey to keep things simple?

Anuken:
I don't see a good reason to change this. Block picking trigger immediately and cannot be dragged, schematic selections can. Making a one-block schematic can copy configuration that block picking may not.

PR #6328, opened 2021-11-11
Better Mod/Plugin dependencies handling

Phinner:
Currently, all plugins are loaded, instantiated and then sorted, this leads to `ClassNotFound` exceptions if you dare to use static classes of other Plugins outside the `init` or if you try to create a general purpose `Plugin` classes like a `AbstractGamemodePlugin` for example that may be optimized for game mode plugins.

The solution is simple, resolving dependencies before loading anything, it features:
- more complex dependency trees
- detecting missing dependencies
- preventing circular dependencies

The code is tested and it works fine, It's just that the structure of `Mods` is a bit hard to follow, so I need your opinion on how the missing/circular dependencies should be handled and what to do with the `dependencies` and `missingDependencies` fields of `LoadedMod`.

Thanks for reading.

If your pull request is **not** translation or serverlist-related, read the list of requirements below and check each box:

- [x] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [x] I have ensured that my code compiles, if applicable.
- [x] I have ensured that any new features in this PR function correctly in-game, if applicable.

PR #7972, opened 2022-11-21
Mod dependency resolution improvements

Phinner:
This pull request introduce some improvements and bugfixes for the dependency resolution mechanism introduced in #6328 with :

- Support for soft dependencies: This aims to make hooking into a soft-depended plugin safe by guaranteeing it's always initialized before the plugins that use it.
- A better resolution algorithm: Improved to give more information about invalid mods, avoiding to resolve them another time later in the code.
- Fixed a bug where the mod initialization order is not guaranteed to be kept after the initial resolution.


Phinner:
Tested on both server and client 👍

Phinner:
**;-;**

Anuken:
So unless I'm missing something, this is actually completely broken. `orderedMods()`, instead of returning only enabled mods, returns **all** of them, which means you get all sorts of issues with mods that you are not even using.

Anuken:
`orderedMods()` is also called many times instead of getting cached once, which looks rather expensive.

Phinner:
> So unless I'm missing something, this is actually completely broken. orderedMods(), instead of returning only enabled mods, returns all of them, which means you get all sorts of issues with mods that you are not even using.

😞 My bad, didn't notice this issue while testing.

PR #7980, opened 2022-11-22
Update v7 server list

TNT331:


Anuken:
offline, needs resubmission

PR #7993, opened 2022-11-25
Additional block build over options

Garen7:
This PR just assigns some blocks that didn't have a group to some existing block groups so that you can build them over their similar counterparts, except for payload conveyors/routers which was reassigned to the payload group instead of the transportation group. The use cases are below:

-Vent condensers can be built over turbine condensers and vice versa
-Duct unloaders and electrolyzers can be built over themselves for easier rotating
-Payload conveyors no longer assume you want to erase normal conveyors you're building over
-Payload blocks like loaders and unloaders can be built over Payload conveyors

From what I can tell the only other place block groups are used besides building over blocks is for the SuicideAI that crawlers have, however that doesn't matter for vent/turbine condensers and duct unloaders since they're on Erekir. Though payload conveyors can now be targeted but afaik that shouldn't be an issue.

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1180,6 +1180,7 @@ public class Blocks{
             craftTime = 10f;
             rotate = true;
             invertFlip = true;
+            group = BlockGroup.liquids;
 
             liquidCapacity = 50f;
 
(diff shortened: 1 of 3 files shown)


Slotterleet:
This actually implements my [suggestion](https://github.com/Anuken/Mindustry-Suggestions/issues/4127) about condenser replacement 🤔

Garen7:
Oh lmao I didn't even notice

Garen7:
Added a block group to electrolyzers and updated the PR description

Also I just realized it might be better to make it so blocks with quickRotate can build over themselves by default, let me know if you would rather me do that and I can add it to the PR

PR #8077, opened 2022-12-21
Hides canvases if logic displays are disabled in the settings

Garen7:
Canvases appear with their default screen if the setting is enabled, they can still be edited and viewed by trying to edit them however. 

Note that the actual setting is called 'Hide logic displays' and not hide display blocks so it might be confusing maybe but I honestly couldn't think of a better name for it anyways and canvases are in the logic block section anyways so I thought it was probably fine, let me know if its not.

### DIFF
--- a/core/src/mindustry/world/blocks/logic/CanvasBlock.java
+++ b/core/src/mindustry/world/blocks/logic/CanvasBlock.java
@@ -13,6 +13,7 @@ import arc.scene.ui.layout.*;
 import arc.struct.*;
 import arc.util.*;
 import arc.util.io.*;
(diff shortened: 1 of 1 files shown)


Anuken:
Implemented in a better way (just don't render the texture.)

Garen7:
@Anuken Hey I tried doing it as just not rendering the texture but it didnt work when I did that, either the block turned invisible, drew this weird static texture, or didnt draw the borders merging based on where I added the return statement. Not sure, maybe the last one is preferable for code readability reasons but yeah just wanted to give a heads up

PR #8081, opened 2022-12-22
Delist omega for spoofing player count

buthed010203:
The hub's player count is spoofed. It shows over double the number of players that are actually connected to any of their servers. They also count the players twice by counting them in both the hub and the other servers in the server list.

### DIFF
--- a/servers_v7.json
+++ b/servers_v7.json
@@ -32,10 +32,6 @@
     "address": ["hexpvp.ml", "kycb42148.ddns.net:6614"]
   },
   {
-    "name": "Omega Hub",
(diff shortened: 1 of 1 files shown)


TimaHappy:
Bruh, EradicationDustry is partner server and it's been in hub as classic gamemodes (attack, survival and etc)

TimaHappy:
And you really care about it?

Volas171:
> The hub's player count is spoofed. It shows over double the number of players that are actually connected to any of their servers. They also count the players twice by counting them in both the hub and the other servers in the server list.

this is untrue, do you have any proof of this, the hub may have desynced with the servers, but this should *not* happen, and I'm 99.9% certainly sure this is not the case.

Volas171:
*Update*

**As Tima Happy mentioned, EradDustry is added to the hub as they are a partnered server, it is also added to the player count respectively, we have talked with eraddustry owner (as well as you have) and they are fine with it, same as us, there's nothing else other than that causing this issue. And here i bring a question, what problems does this arise? Servers are listed based off ping, not players, the user goes to the one they like the most.**

buthed010203:
I cant be bothered to deal with server drama any longer

Volas171:
awesome, now retreat before I come back with 70 lines of proof

buthed010203:
🤡

PR #8130, opened 2023-01-08
erekir T4s' price rebalanced

nanmenyangde:
Mainly optimize the problem of overpowered and too cheap ships

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -5506,7 +5506,7 @@ public class Blocks{
             regionSuffix = "-dark";
             size = 5;
             plans.add(
-            new AssemblerUnitPlan(UnitTypes.quell, 60f * 60f, PayloadStack.list(UnitTypes.elude, 4, Blocks.berylliumWallLarge, 8)),
+            new AssemblerUnitPlan(UnitTypes.quell, 60f * 60f, PayloadStack.list(UnitTypes.elude, 6, Blocks.tungstenWallLarge, 14)),
             new AssemblerUnitPlan(UnitTypes.disrupt, 60f * 60f * 3f, PayloadStack.list(UnitTypes.avert, 6, Blocks.carbideWallLarge, 20))
             );
             areaSize = 13;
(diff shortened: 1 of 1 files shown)


nanmenyangde:
Also make its price similar to other t4s.The ships are more expensive than other t4s because they have good mobility and long range

Garen7:
tbh this probably isnt a large enough quell nerf all things considered

image

nanmenyangde:
yes i agree the quell needs to be nerfed.it is much stronger than other t4
but i think disrupt don't need big modification because it has its unique use among t5s and its strength is appropriate

Garen7:
What is its unique use?

nanmenyangde:
> What is its unique use?

it's flying. disrupt can attack over the wall without damaged by the collaris, and when you get 20+ t5s disrupt is almost the only option

Garen7:
After your proposed changes why not just go quells instead of disrupt though, the cost of going them is still smaller and you get more stats at almost no penalty

Anuken:
I am hesitant to change required materials entirely, as it breaks builds, and campaign maps. Maybe it should require more blocks instead?

Garen7:
I dont know if that will change much since people would just need to add 2 more small bores on beryllium and add 4 more payload conveyors since the buffer would need to be bigger(a payload conveyor buffer is needed since unlike the other refabricators it cant pre-store the materials for the next unit)

12\*24 berl/70 sec/.75 berl per sec = 5.49 bores (6 rounded up)
8*24 berl/70 sec/.75 berl per sec = 3.66 bores (4 rounded up)
(I use 70 sec instead of 60 because theres around a 10 second time period needed for the payload conveyors to add it to the assembler)

I think a stat change would matter more but any change is welcome

moocowswag:
Quell needs a stat change if anything and tecta is already decent for its cost and is likely to receive a massive indirect buff from any nerfs to locus. (I dont think tecta should be touched at all until after locus are changed)

moocowswag:
quell is the "weakest" because its not meant to be a ball of stats and its the "strongest" because it has the capability to provide you value without putting itself in harms way, every quell you make will easily provide you value until the end of the game.

nanmenyangde:
yes it is. so i think make quell more expansive than other t4 is a good way to solve the problem.

nanmenyangde:
@Anuken What do you think of this?

Garen7:
a different change was already added for this so probably best to just wait and see how that one pans out https://github.com/Anuken/Mindustry/commit/3ec2ae797ef87028abe30997b1b2d4d051229909

Also this PR has merge conflicts

Anuken:
> What do you think of this?

What is "this"? If you mean the exact changes in this PR, I have already explain my opinion above. I would rather not change unit block requirements, as it breaks break maps/builds.

nanmenyangde:
ok i'll wait for more advice. i think that change doesn't work for nerfing quells, because i just need to put one more constructor.

PR #8197, opened 2023-01-19
Adds control groups

Garen7:
Adds the following controls for doing control groups

- Create Control Group (Ctrl+Num)
- Add to Control Group (Unset)
- Create Auto Control Group (Alt+Num)
- Focus Cam on Control Group (Double Tap Num)

The add to control group button is unset by default since that would normally go on the left shift key however that spot is currently taken and I don't want to cause confusion. I have also added a setting that will cause control groups to automatically remove the units you add to them from the other groups as that is usually the preferred way of doing it but sometimes you see games do it the other way, or offer both as an option.

2 Problems I want to mention however:
1. ~~If you sync the non auto control groups will disappear, I could maybe fix this by using the units id instead of a reference to the unit(not sure if I can do this actually or if its worth it)~~
2. When you focus the camera on the control group it will not use the smooth camera if that setting is enabled(I couldn't use Core.camera.position.lerpDelta since the focus camera thing only runs on one update tick, I could instead make it follow the group and it would be fine but I wanted to verify if that's preferred before actually adding that since it would be more code and less standard for control group stuff)

But yeah please let me know if you want me to try doing the potential fixes I mentioned or if theres a smarter way of doing that instead


core/assets/bundles/bundle.properties
```diff
@@ -1180,6 +1181,19 @@ keybind.chat_scroll.name = Chat Scroll
...
+keybind.control_group_01.name = Control Group 1
+keybind.control_group_02.name = Control Group 2
+keybind.control_group_03.name = Control Group 3
+keybind.control_group_04.name = Control Group 4
+keybind.control_group_05.name = Control Group 5
+keybind.control_group_06.name = Control Group 6
+keybind.control_group_07.name = Control Group 7
+keybind.control_group_08.name = Control Group 8
+keybind.control_group_09.name = Control Group 9
+keybind.control_group_10.name = Control Group 10
```
Garen7, line 1196: also should I just reuse the block select hotkeys or is this fine

MEEPofFaith:
> I could maybe fix this by using the units id instead of a reference to the unit

`Groups.unit.getByID(id)` can be used to grab the unit itself.
You could store unit ids, then use getByID to get the units.

Garen7:
> > I could maybe fix this by using the units id instead of a reference to the unit
> 
> `Groups.unit.getByID(id)` can be used to grab the unit itself. You could store unit ids, then use getByID to get the units.

oh true thanks! ill fix that in a bit

Garen7:
nvm, I believe units *sometimes* get new ids after a sync since I started getting crashes saying the unit would be null with an implementation that uses unit ids instead of references. Im gunna look more into it later though but I dont think that way would work unfortunately

Garen7:
> nvm, I believe units _sometimes_ get new ids after a sync since I started getting crashes saying the unit would be null with an implementation that uses unit ids instead of references. Im gunna look more into it later though but I dont think that way would work unfortunately

correction to this, the IDs *don't* change between syncs but getByID() sometimes returns null right after a /sync so rather than filtering invalid units from the control groups on update I changed it to only checking when the player hits the control group button and added a null check so it should almost never drop units on a sync now except in maybe an edge case scenario where the player selects the control group immediately after a sync

nanmenyangde:
Great! i like this

Anuken:
So this doesn't come with any UI, yes? Should it? Do other games have that?

Garen7:
This doesn't have a UI, most games do. The usual UI for control groups is a picture of the most common unit in the control group plus a number counting the total units in the control group

Starcraft 2
image

Age of empires 2
image

I didn't add one because I don't know how to do UI in mindustry yet and people don't usually click on the control group UI to select them in RTSs but rather hit the keyboard shortcuts. And also since I wasn't sure where you would want that on the screen. If you want me to I can make a UI for it @Anuken

Anuken:
Implemented:

- Didn't use anything related to autoControlGroups, not sure what that's even for
- Many various changes to make it work correctly (didn't save, didn't reset properly, lots of boxing, etc)

PR #8199, opened 2023-01-19
Add hiddenMods in ConnectPacket event

Kieaer:
This event sends all mods (including hidden mods) to the server.

At the extreme, even if a modified client shows up and hides the list of mods being sent to the server, it's better than a server that has no idea what hidden mods the client is installing and making the game unfair.
It's common to use hidden mods to unlock blueprint bans and server rate limiters. (rip arc packet spam detector)

=============================================================

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -91,6 +91,7 @@ public class NetClient implements ApplicationListener{
             c.name = player.name;
             c.locale = locale;
             c.mods = mods.getModStrings();
+            c.allMods = mods.getAllModStrings();
             c.mobile = mobile;
             c.versionType = Version.type;
             c.color = player.color.rgba();
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -91,7 +91,7 @@ public class NetClient implements ApplicationListener{
             c.name = player.name;
             c.locale = locale;
             c.mods = mods.getModStrings();
-            c.allMods = mods.getAllModStrings();
+            c.hiddenMods = mods.getHiddenModStrings();
             c.mobile = mobile;
             c.versionType = Version.type;
             c.color = player.color.rgba();
--- a/core/src/mindustry/core/NetServer.java
+++ b/core/src/mindustry/core/NetServer.java
@@ -197,6 +197,8 @@ public class NetServer implements ApplicationListener{
                 con.kick(result.toString(), 0);
             }
 
+            Events.fire(new PlayerMods(player, packet.allMods.copy()));
+
             if(!admins.isWhitelisted(packet.uuid, packet.usid)){
                 info.adminUsid = packet.usid;
                 info.lastName = packet.name;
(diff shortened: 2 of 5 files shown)


Phinner:
Why adding a new event, when `ConnectPacketEvent` is already available ? And also isn't it better to just send the hidden mod list (something like `ConnectPacket#hiddenMods`) instead of the whole list since a list of not hidden mods is already sent (`ConnectPacket#mods`) ?

Kieaer:
> Why adding a new event, when `ConnectPacketEvent` is already available ? And also isn't it better to just send the hidden mod list (something like `ConnectPacket#hiddenMods`) instead of the whole list since a list of not hidden mods is already sent (`ConnectPacket#mods`) ?

Good idea. I'll edit it when I get home.

Phinner (approved):
Much better :)

Kieaer:
My bad not pushed packets file 🤪

Phinner (approved):

TheRadioactiveBanana:
Rather sure a smart person can just re-name their mods to something not suspicious. And for a person who makes their own mods, this is easy to bypass. I think you would literally have to upload the mods to the server, scan each and every file with some sort of anticheat, and then allow the player. There is no point in scanning mods in my opinion.

notAXeony:
This will fix only Anyuk's addition of a unique ID for mods, and the banning of all mods except with a specific ID

Phinner:
> This will fix only Anuken's addition of a unique ID for mods, and the banning of all mods except with a specific ID

If you mean just banning specific hidden mods, yes.

Kieaer:
1. The client uploads the mod it is currently using to the server.
2. The server checks the hash value of the mode received from the client on Github.
3. If a match is found, the server administrator takes the desired action.

If you add the upload function to the game itself, there may be a large delay when entering the server.
Therefore, this feature only works when requested by the server.
How is it?

Kieaer:
Even if the mod name is modified, the content of the mod being sent by this event remains the same, which can be useful in situations where all mods must be banned, such as in PvP.

Anuken:
I don't really see the point of any of this when someone can trivially modify their client to *not* send the hash or hidden mods. I have no control over *foo's* client, for example, and I highly doubt buthed is going to merge in this functionality into that client when all it does is prevent users from joining servers.

Kieaer:
Not all users use modified clients.
I can't help it if the client hides itself and comes in.

Anuken:
So what's the point? Get cheat mod -> can't join server -> get a mod that bypasses this restriction? It doesn't really solve anything.

Anuken:
There are no real alternatives. You can't stop people from using the mods they want in an open source game, period.

Kieaer:
It's going to be pierced anyway, but I don't think it's not like you're not even trying to block it at all.. please

Kieaer:
On PvP servers where schematic is banned, use schematic mode to give teams with mods an overwhelming advantage, use fog removal mod to locate enemy players or units or ignore world processor/world maker's intentions And because of the difference between users with and without any mods, i will continue to see fights every day.

This's not a modified client.
I downloaded it from the steam workshop and tested it.

So, i added 7 more admins to stop this, but I couldn't stop it.
All of the griefers are blocked by the plugin, but the **hidden mod** is out of my control.

PR #8203, opened 2023-01-22
Rename omnidustry server to ohiodustry

piskabo:
You will ask why? Because this server is shit that violates the rules of the servers. For example, there is practically no administration, and their sandbox server is constantly in shit (NSFW).


TheEt1234:
Instead of trolling anuke why not confact the server owners

Darkness6030 (approved):

Def-Try:
> Instead of trolling anuke why not contact the server owners

They are aware of it, but thinking that "nothing could go wrong, there cant be a person that will kick out server from global" or something like that

TheRadioactiveBanana:
I can confirm their sandbox server has a lot of stuff wrong.
-COPIOUS amounts of nsfw content, like im talking logic art, sorter art, canvas art, you name it.
(atleast 25% of the map is covered with it a lot of the time, unless someone sensible (usually not a moderator) comes and removes it.)
-No moderation whatsoever, i saw on the survival server, about 3 griefers who joined did their thing and left. people are absolutely swearing the heck out of these griefers and votekick spamming yet the griefers just leave when something bad is happening. No moderator ever came.

Anuken:
NSFW content with no moderation is not acceptable, and as such I'll just remove the sandbox server until the issue is resolved.

To the owners: If you can't moderate such content, please just ban displays and canvas blocks in the server config.

@TheTrueInfinity (pinging since that's the person that last edited the omnidustry listing)

SSTentacleSS:
@Anuken 

image

Anuken:
It should have the canvas block banned too. People use it for NSFW too.

code-explorer786:
(fyi during the time i'm writing canvas blocks are still not banned and can cause an issue too)

PR #8204, opened 2023-01-22
Banned logic displays on sandbox

SSTentacleSS:
image
image


TheEt1234:
You should ban canvas too since a lot of nsfw was shared through that

SSTentacleSS:
image

PR #8210, opened 2023-01-23
More useful methods

Darkness6030:
Because it says "do not access directly" and i don't wanna use reflection
image

### DIFF
--- a/core/src/mindustry/net/Administration.java
+++ b/core/src/mindustry/net/Administration.java
@@ -321,7 +321,7 @@ public class Administration{
     /**
      * Returns all kicked IPs. This does not include the IPs of ID-kicked players.
      */
-    public Seq<String> getKickedIPs(){
+    public ObjectMap<String, Long> getKickedIPs(){
         // Remove all expired kicks
         kickedIPs.each((ip, time) -> {
             if (time <= Time.millis())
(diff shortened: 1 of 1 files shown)


Darkness6030:
Also i should add "/kicks" command to servers. Maybe later...

Kieaer:
I needed this too!

core/src/mindustry/net/Administration.java
```diff
@@ -304,6 +304,32 @@ public Seq<PlayerInfo> getBanned(){
...
+                result.add(info);
+            }
+        }
+        return result;
+    }
+    
+    /**
+     * Returns all kicked IPs. This does not include the IPs of ID-kicked players.
+     */
+    public ObjectMap<String, Long> getKickedIPs(){
```
code-explorer786, line 324: There's already `kickedIPs` (Administration.java, line 25) for that.

code-explorer786 (commented):
This is not really useful at all, since `kickedIPs` is a thing.

* `getKicked` can be done with something along the lines of `kickedIPs.keys().toSeq().map(ip -> findByIP(ip))`.
* `getKickedIPs` can be done with accessing `kickedIPs`.

In the end, I would recommend making a method to "refresh" `kickedIPs` in case there's an anomaly somehow.

core/src/mindustry/net/Administration.java
Darkness6030: Ok, i will try

PR #8220, opened 2023-01-25
Fix server auto pause when gameover

Kieaer:
image

I found a problem where the server doesn't auto pause when the gameover even if no players when the server is an attack map with units.

==================================================================

### DIFF
--- a/server/src/mindustry/server/ServerControl.java
+++ b/server/src/mindustry/server/ServerControl.java
@@ -191,7 +191,12 @@ public class ServerControl implements ApplicationListener{
 
                 info("Selected next map to be @.", Strings.stripColors(map.name()));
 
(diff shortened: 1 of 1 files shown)


inflexibledark:
Did you test it? I don't wanna have more issues with pausing

Kieaer:
@inflexibledark Not tested. But it will work correctly.

Anuken:
Testing is required.

Kieaer:
https://user-images.githubusercontent.com/44261958/216462846-2d618fe8-12bc-4375-980e-b282754154bf.mp4

Anuken:
But if you don't set wasAutopaused to true, will it ever unpause..?

Kieaer:
In my tests and thoughts, autoPause is always true when this condition is triggered, so I didn't think it was necessary.
Because the autoPause value must be true when the last player leaves.

PR #8246, opened 2023-02-01
Random servers order + refactoring

Darkness6030:
image

### DIFF
--- a/core/src/mindustry/ui/dialogs/JoinDialog.java
+++ b/core/src/mindustry/ui/dialogs/JoinDialog.java
@@ -117,7 +117,7 @@ public class JoinDialog extends BaseDialog{
     }
 
     void refreshAll(){
-        refreshes ++;
+        refreshes++;
 
         refreshLocal();
         refreshRemote();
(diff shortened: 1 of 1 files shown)


Darkness6030:
@Anuken what do you think about this PR?
I tried to make the serverlist order fair .-.

Anuken:
This isn't a good idea. Random shuffling means a different order every time you refresh, and I'm not convinced it matters anyway.

As for refactoring, I do not accept such PRs - the code works fine, the only thing you're giving me by refactoring it is introducing the possibility of bugs. I don't want to review all those changes.

Darkness6030:
> I'm not convinced it matters anyway.

It really matters.
Look, here's 3 different refreshes:
image
image
image

Maybe a random order should apply once when you load the game?

Anuken:
That's no good either. The order should depend on ping and *only* ping.

Darkness6030:
> The order should depend on ping and _only_ ping.

*But at least it should not depend on a position in servers.json*

kotasusmogus:
> > I'm not convinced it matters anyway.
> 
> It really matters. Look, here's 3 different refreshes: image image image
> 
> Maybe a random order should apply once when you load the game?

💀

PR #8292, opened 2023-02-13
Changed Darkdustry domain to direct IP address

Darkness6030:
Why? Because darkdustry.net is not available in some countries (Ukraine, Israel and some more) 
Also improved formatting


SSTentacleSS:
image

VozdyxDev (approved):
oh no

PR #8303, opened 2023-02-15
Fixed issue with Blocks scaledHealth

FredyJabe:
The way it was, it would only use item healthScaling value if the Block had no scaledHealth value, but from what I understand, we want to scale scaledHealth WITH requirements' healthScaling.

Simply moved a brace to make sure the calculations we're made even if a scaledHealth value is given.

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -1125,15 +1125,15 @@ public class Block extends UnlockableContent implements Senseable{
             boolean round = false;
             if(scaledHealth < 0){
                 scaledHealth = 40;
-
-                float scaling = 1f;
-                for(var stack : requirements){
-                    scaling += stack.item.healthScaling;
-                }
-
-                scaledHealth *= scaling;
-                round = true;
             }
+            
+            float scaling = 1f;
+            for(var stack : requirements){
+                scaling += stack.item.healthScaling;
+            }
+
+            scaledHealth *= scaling;
+            round = true;
 
             health = round ?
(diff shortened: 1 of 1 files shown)


FredyJabe:
Created a bug Issue : #8304

Anuken:
This is not a bug. scaledHealth is **only** to be used "raw", it has no relation to items.

FredyJabe:
> This is not a bug. scaledHealth is **only** to be used with exact values for per-block-area health, it has no relation to items.

From what I understand, it is impacted by the building requirements?

Currently, if you change the default value, it doesn't scale anymore, but if you just ommit to add anything (so, leaving the value to -1), it sets the block health to 40*size*requirement scaling values.
```java
    //initialize default health based on size
        if(health == -1){
            boolean round = false;
            if(scaledHealth < 0){
                scaledHealth = 40;

                float scaling = 1f;
                for(var stack : requirements){
                    scaling += stack.item.healthScaling;
                }

                scaledHealth *= scaling;
                round = true;
            }

            health = round ?
                Mathf.round(size * size * scaledHealth, 5) :
                (int)(size * size * scaledHealth);
        }
```

Anuken:
> From what I understand, it is impacted by the building requirements?

It's used in some intermediate calculations *if you don't provide a default value*, but that's mostly irrelevant (I could *not* use it there, and the result would be the same). 

Assigning any value to scaledHealth makes it use that value as-is, which is the desired effect.

FredyJabe:
> 

It does make it use that value as-is, but stop scaling, which shouldn't if I read the comments well in `Block` class.

PR #8308, opened 2023-02-17
Add spaces between settings

JniTrRny:
Added some space to make it more categorized, and looks cleaner I guess?
The old settings just looks messy (especially Graphics) and that makes adding new settings harder.

~~added `addText()` which is a customized `Setting` for anyone that wants to add a raw text to settings table and `addRow()` which is just a blank version of `addText()` but will not show up in `getSettings()` (should it?)~~
Added `Separator` setting which is literally just text in settings (with customizable background)
but this PR only uses the one with no background and without text, just so if someone wants to actually add some text :3

image

image

image
###### hope I didn't mess up anything


MEEPofFaith:
What about underlined text dividers like the ones I use in my mod?
image

[My code for this.](https://github.com/MEEPofFaith/prog-mats-java/blob/74fff2dceffa8fa4c9228ff8e6e2ce7d142e2e63/src/progressed/ProgMats.java#L302)

core/src/mindustry/ui/dialogs/SettingsMenuDialog.java
```diff
@@ -676,6 +672,20 @@ public void areaTextPref(String name, String def, Cons<String> changed){
...
+                public void add(SettingsTable table){
+                    table.add(text).row();
+                }
+            });
+            rebuild();
+        }
+
+        public void addRow(){
+            addText("newrow", "");
+        }
```
MEEPofFaith, line 687: Why even have `name` in the constructor?
Why not just leave it null?

Darkness6030 (commented):
copy().removeAll() is bad, use select() instead. Also an idea with "newrow" is VERY bad

JniTrRny:
Yah that was the last minute decision to slap a `copy()` on top of it

JniTrRny:
> What about underlined text dividers like the ones I use in my mod? image
> 
> [My code for this.](https://github.com/MEEPofFaith/prog-mats-java/blob/74fff2dceffa8fa4c9228ff8e6e2ce7d142e2e63/src/progressed/ProgMats.java#L302)

Looks interesting, guess I'll make it looks like the `CustomRulesDialog` but without the striking color.

Darkness6030:
Amazing

Anuken:
I think this needs proper text on the categories to be useful.

PR #8310, opened 2023-02-18
Smoother Screenshake Reduction

MEEPofFaith:
Screen shake will now properly respect the input duration. No longer will it awkwardly cut off the shake or fade away before the set duration ends.

Before:

https://user-images.githubusercontent.com/54301439/219851805-b3bc9158-8b52-4d45-9447-d5355acada42.mp4

After:

https://user-images.githubusercontent.com/54301439/219851841-2ad84fad-2b4e-48ee-b50c-7818a46bb2db.mp4

---

### DIFF
--- a/core/src/mindustry/core/Renderer.java
+++ b/core/src/mindustry/core/Renderer.java
@@ -72,6 +72,8 @@ public class Renderer implements ApplicationListener{
     landPTimer,
     //intensity for screen shake
     shakeIntensity,
+    //reduction rate of screen shake
+    shakeReduction,
     //current duration of screen shake
     shakeTime;
     //for landTime > 0: if true, core is currently *launching*, otherwise landing.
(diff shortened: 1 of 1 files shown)


core/src/mindustry/core/Renderer.java
```diff
@@ -83,14 +85,15 @@ public Renderer(){
...
+            shakeTime = shakeIntensity = shakeReduction = 0f;
             camShakeOffset.setZero();
         });
     }
 
     public void shake(float intensity, float duration){
-        shakeIntensity = Math.max(shakeIntensity, intensity);
+        shakeIntensity = Math.max(shakeIntensity, Mathf.clamp(intensity, 0, 100));
         shakeTime = Math.max(shakeTime, duration);
+        shakeReduction = shakeIntensity / shakeTime;
```
Anuken, line 96: shouldn't this only update if the intensity and/or time updates?

MEEPofFaith: Since `update()` reduces both intensity and time, having `shakeIntensity / shakeTime` wouldn't reduce the amount accurately.
If I make the shake based on `shakeIntensity > 0` and then just not decrease shakeTime it would accurately decrease over time, however inputing a new shake time wouldn't work properly if the current time doesn't decrease.

The reduction rate needs to be calculated upon inputting shake parameters.

PR #8336, opened 2023-02-24
Update servers_v7.json

SSTentacleSS:
We have enabled DDoS protection . _.


StableRibbon:
I don't think this server should be in the public list. It is important to address the issue of a server violating the Mindustry serverlist guidelines for NSFW content. Allowing such a server to remain public not only goes against the established rules, but also creates an unsafe and uncomfortable environment for users who may accidentally come across this bad server.
Furthermore, the toxicity of the server's moderation only exacerbates the issue. Abuse of power and a lack of respect for community members is unacceptable and has no place in any community, let alone a public one. Toxicity can lead to a negative user experience. What is more, server is hated by huge part of Mindustry community.

SSTentacleSS:
ok

image
image

SSTentacleSS:
image

PR #8387, opened 2023-03-12
Fix armor piercing ignoring status effect health multiplier

MEEPofFaith:
Was originally just a pr to make `rawDamage` protected so for overriding reasons, but then I noticed the problem and fixed it in the meantime.


MEEPofFaith:
tfw sublimate destroys a unit with the invincible status effect

PR #8441, opened 2023-03-27
Update servers_v7.json

Suportik-GFX:
adding one more server under the Exploding Cowards.


Suportik-GFX:
I am still setting up the gamemode with plugins, so you might catch the server offline.

PR #8442, opened 2023-03-27
Return `Setting` value type for each `pref`methods.

SITUVNgcd:
Added return `Setting` type for these methods (instead of `void`):
- `CheckSetting checkPref()`.
- ~~`SliderSetting sliderPref()`.~~ Why these ones have return type and others not, before my PR?
- `TextSetting textPref()`.
- `AreaTextSetting areaTextPref()`.
- `Setting pref()`. Oh! For it, also!
- Optimized for calling overload methods.

---


MEEPofFaith:
I don't see how?
In most cases where it doesn't get used the returned value can just be ignored.

Unless you're saying that old mods that don't increase the version of mindustry used for compiling will break because of this?
~~I really could just test this myself.~~

Anuken:
> We can always just increase what version we use in compiling.

Yes, meanwhile every single Java mod a user has installed will crash the game on startup and have to be updated manually.

MEEPofFaith:
*ah*

SITUVNgcd:
@Anuken Ok, i can't change exsited method (signalture).
But how about add new method? Is this allowed?

Anuken:
That would bloat the API and I don't see a good reason to do that.

SITUVNgcd:
> That would bloat the API and I don't see a good reason to do that.

Ok! I will not do that.

PR #8445, opened 2023-03-30
Update Menus.java

SSTentacleSS:
The commit is aimed at the ability to remove listners that we do not need, which can potentially be a huge number and which can cause the GC to not work as expected

### DIFF
--- a/core/src/mindustry/ui/Menus.java
+++ b/core/src/mindustry/ui/Menus.java
@@ -19,6 +19,11 @@ public class Menus{
         menuListeners.add(listener);
         return menuListeners.size - 1;
     }
+    
+    /** Remove a *global* menu listener */
+    public static void removeMenu(MenuListener listener){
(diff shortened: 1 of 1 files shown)


Phinner:
This approach would break all menus since the ids will be shifted. Use an `IntMap` to store the menus instead with an incrementing id so it never happens.

Darkness6030:
Зачем это вообще нужно? Ты можешь регистрировать одно меню для всех игроков

Darkness6030:
В любом случае на сервере не будет больше 5-10 меню, ну в крайнем случае 20. Зачем может пригодиться удалять их, это же не сильно влияет на память

SSTentacleSS:
> This approach would break most menus since the ids will be shifted. Use an `IntMap` to store the menus instead with an incrementing id so it never happens.

yea, good

Anuken:
PR is broken as mentioned above

PR #8451, opened 2023-04-01
Fix hjson batteries without a drawer crashing the game

MEEPofFaith:
Is `init` not called before `load`?


MEEPofFaith:
https://canary.discord.com/channels/391020510269669376/653293028869537843/1091526882941472929

Batteries seem to crash the game due to a NPE when used with hjson.
If a drawer is not defined, then it ends up null and crashes with trying to load, which doesn't make sense because supposedly it's supposed to not be null due to a default drawer being created in `init()`.

PR #8455, opened 2023-04-01
Fixing Call.menu() not hiding on option selection.

Xasmedy:
*When 3 hours of testing gets thrown in the garbage by a single refactoring...*

I was calling `UI.hide()` instead of `Dialog.hide()`, now fixed.

**Tests ran before this PR**
- (Menu) if closeOnBack gets called.
- (Menu) if the buttonListener get called + if it has the correct option.
- (Menu) if hidden after the option selection.
- (FollowUpMenu) if closeOnBack gets called.
- (FollowUpMenu) if the button listener gets called + if it has the correct option.

I went with `Cons2<Integer, Dialog>` ([to replace](https://github.com/Anuken/Arc/pull/141)) instead of having a `boolean hideMenu` to allow easier maintenance (if ever needed) in the future.
(*Stupid example that gives the point: hiding the menu with no delay when `option = -2`.*)

### DIFF
--- a/core/src/mindustry/core/UI.java
+++ b/core/src/mindustry/core/UI.java
@@ -594,7 +594,8 @@ public class UI implements ApplicationListener, Loadable{
         dialog.show();
     }
 
-    public Dialog newMenuDialog(String title, String message, String[][] options, Intc buttonListener, Runnable closeOnBack){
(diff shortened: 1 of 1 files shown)


Anuken:
~~So much for testing things, now every client has a broken menu...~~

PR #8501, opened 2023-04-10
Search bar for wave UI

JniTrRny:
Heavily stripped down ~~and not heresy~~ version of #5977, so apparently I did the funny Ctrl-C Ctrl-V and branded it as a new Pull Request.

Added search bar for wave info UI, a filter next to it because it feels too empty without one, and probably fixed other editor UI things.

https://user-images.githubusercontent.com/85090668/230944729-a29ca15a-921e-412f-9bf3-1b4b989841bd.mp4

Type a number, and the graph on the right will shift to the number inputted, the wave group on the left will change to only show the unit that will spawn in the wave that is the number you inputted too.

image

The filter icon next to the search bar, when clicked, will show a selection of units to choose, and that unit will be the unit the wave group pane will only show.
###### Apparently in the old PR, I have overcomplicated it, it was a big hassle and no one would use it anyways so stick to only the useful ones for now.

image

Moving on to one of the smaller things, moved the filter button to be next to the add button ~~exactly like the old PR.~~

image

Reordered and resized this edit dialog to be consistent with the generation ones ~~and the next target is the logic one.~~

image

Okay and great tooltips everywhere in case it isn't obvious enough ~~and make use of the `<none>` now that there is `<none found>` replacing almost all of its uses~~

And probably several internal things I could not care less to mention about it because apparently people wants to only see interesting things and not uninteresting things

Oh and maybe I should mentions some editor UI things too small to be a separate Pull Request ~~and as well name it random UI shit~~, so I might as well use the old PR tactic of sticking smaller changes with a big UI changes.

image

Updated one of the most hidden dialog, to a new flat UI.

image

extended the search bar of the editor, so more words can be shown, and it looks better that way.

image

so added a search icon to search bar, so that they searchingly searches for the search that you have searched for correctly.

Okay that's it, I should write less in an otherwise self-explanatory PR sometimes, might as well be just a video and some texts next time, people in general are not so keen at reading more than eight sentences.

---


core/src/mindustry/ui/dialogs/CustomRulesDialog.java
```diff
@@ -280,7 +280,7 @@ void setup(){
 
             main.button("[#" + team.color +  "]" + team.localized() + (team.emoji.isEmpty() ? "" : "[] " + team.emoji), Icon.downOpen, Styles.togglet, () -> {
                 shown[0] = !shown[0];
-            }).marginLeft(14f).width(260f).height(55f).checked(a -> shown[0]).row();
+            }).marginLeft(14f).width(260f).height(55f).checked(a -> shown[0]).update(t -> ((Image)t.getChildren().get(1)).setDrawable(shown[0] ? Icon.upOpen : Icon.downOpen)).row();
```
Anuken, line 283: this shouldn't be here

core/src/mindustry/world/blocks/ItemSelection.java
```diff
@@ -85,9 +85,11 @@ public static <T extends UnlockableContent> void buildTable(@Nullable Block bloc
 
         Table main = new Table().background(Styles.black6);
         if(rowCount > rows * 1.5f){
-            search = main.field(null, text -> rebuild.run()).width(40 * columns).padBottom(4).left().growX().get();
-            search.setMessageText("@players.search");
-            main.row();
+            main.table(s -> {
```
Anuken, line 88: what is the purpose of this?

core/src/mindustry/ui/dialogs/CustomRulesDialog.java
JniTrRny: whoops

core/src/mindustry/world/blocks/ItemSelection.java
JniTrRny: It just looks oddly weird because every other search bar has that zoom icon.

PR #8517, opened 2023-04-19
New categories and placement

nekit508:
Added search lines for categories and blocks and the ability to create new categories of blocks (does not conflict with existing mods).

java_0uLDgXVAlX


Anuken:
There are a lot of problems here:

- Takes up a lot more UI space, which is very important on mobile. In fact, this probably completely breaks portrait mode.
- Looks rather ugly compared to the old compact layout. There's a whole lot of empty space and a search box nobody will ever need to use in vanilla (or even in most mods...)
- Destroys existing muscle memory by rearranging button order.
- Breaks any mods that use the category array - it's an API change.
- Doesn't really solve the category problem neatly - some future mods will use new categories, but others will not. Users will be left with 90% of mods having their stuff injected into existing categories, and the rest having their own, which is messy and suboptimal. Mods can't easily share categories, either.

nekit508:
still, does the idea have a future?

MEEPofFaith:
Keep it 2 columns wide and 4 rows tall and just have extra categories require scrolling down, and a search bar for categories is *probably* not needed.


Also, **neuron activation**.

xStaBUx:
To reduce spam in modified categories, I suggest the following options:

- Add ability to scroll to save available screen space.
- Make it so that no more than 10 categories are displayed at the same time and add special buttons "<", "[category section number]" and ">" to quickly switch between category sections.
- Add the ability to customize categories. For example, you can add the option of "selective display" of categories, where the user can choose which categories he wants to see on the main page, and which - to hide in the menu. However, this may cause some modified blocks to disappear, so you need to think about solving this problem.

nekit508:
updated...

Otamamori917:
Look nice but how does this work for mobile?

xStaBUx:
Since I don't think @Anuken will ever approve this, I suggest you make a library mod that does what your PR does.

nekit508:
The point was to replace the enumeration with a class, but the library will not be able to do this.

PR #8531, opened 2023-04-23
Lock rotation of blocks with `rotate = false` to 0

MEEPofFaith:
https://user-images.githubusercontent.com/54301439/233856676-12391a82-cae1-456d-9329-0dd6bb485963.mp4

It's on a boolean because apparently some mods use the fact that blocks that don't rotate still rotate.


core/src/mindustry/world/Block.java
```diff
@@ -106,6 +106,8 @@ public class Block extends UnlockableContent implements Senseable{
     public boolean underBullets;
     /** whether this is rotatable */
     public boolean rotate;
+    /** if rotate is false and this is true, building rotation is set to 0 regardless of player input */
+    public boolean lockRotation = true;
```
MEEPofFaith, line 110: Should it default to false?

BlueWolf3682: probably not, PRed values probably shouldnt force changes that may affect vanilla if intentionally avoidable

sk7725:
I don't see any use with this field. If you want to lock rotation that bad, you can just set the rotation to 0 on block initialization for individual blocks (via Java/JS).
Json mods have no use whatsoever of this block, either.

sk7725:
> It's on a boolean because apparently some mods use the fact that blocks that don't rotate still rotate.

afaik vanilla hortiloaders use this too.
To specify, mods are not using the fact that it "still rotates" - mods are using the fact the initial rotation tells you how (and in what direction) the block was placed, even when rotate = false.

MEEPofFaith:
> I don't see any use with this field. If you want to lock rotation that bad, you can just set the rotation to 0 on block initialization for individual blocks (via Java/JS). Json mods have no use whatsoever of this block, either.

Case 1: Directional liquid outputs on a generic crafter that has `rotate = false`. Json has no way to disable.
Case 2: Setting the rotation on placement will not visually show it in building highlights. `IntputHandler` drawing block previews will still rotate the block, leading to improper displays for rotation-based things such as generic crafter liquid outputs.

sk7725:
> > I don't see any use with this field. If you want to lock rotation that bad, you can just set the rotation to 0 on block initialization for individual blocks (via Java/JS). Json mods have no use whatsoever of this block, either.
> 
> Case 1: Directional liquid outputs on a generic crafter that has `rotate = false`. Json has no way to set rotation to 0 on placement. Case 2: Setting the rotation on placement will not visually show it in building highlights. `IntputHandler` drawing block previews will draw a rotated preview, leading to improper displays for rotation-based things such as generic crafter liquid outputs.

Case 1: set `rotate = true` for json.
Case 2: override the preview draw functions.

BlueWolf3682:
this is a lot of complaining about useless additions from the person that added placeablePlayer

Anuken:
The setRotation change seems too invasive, I don't want to use a getter everywhere.

MEEPofFaith:
god damnit

MEEPofFaith:
why are the power gen tests failing what

Anuken:
I broke it

sk7725:
planRotation() sounds better, I am still skeptical that this will have an use out of very niche scenarios

sk7725:
> this is a lot of complaining about useless additions from the person that added placeablePlayer

I only complain about new field additions that can be achievable by using existing code workarounds. I'm cool with a field that can introduce new behaviors.

MEEPofFaith:
> 

ive never even heard of `placablePlayer` does anyone even use it

MEEPofFaith:
> > this is a lot of complaining about useless additions from the person that added placeablePlayer
> 
> I only complain about new field additions that can be achievable by using existing code workarounds. I'm cool with a field that can introduce new behaviors.

ive never even heard of `placablePlayer` does anyone even use it

core/src/mindustry/world/blocks/production/GenericCrafter.java
```diff
@@ -118,6 +118,8 @@ public void init(){
 
         if(outputItems != null) hasItems = true;
         if(outputLiquids != null) hasLiquids = true;
+        
+        if(!rotate) lockRotation = true;
```
MEEPofFaith, line 122: Should this be included?

Anuken: No, that looks redundant.

MEEPofFaith:
See convo in #pulls.
If a block has `rotate = false`, it doesn't make sense for it to rotate anyway. That is a bug and unintended behavior. If you want to not show rotation, then just do `drawArrow = false` instead.

PR #8547, opened 2023-04-30
Arbitrary Block Sorting & De-Enum Categories

Goobrr:
Grants mods the ability to organize both their own blocks and vanilla blocks in the place menu without having to adjust load order. Also de-enumerates block categories, but does not add proper UI support for it.

Not a good example but notice how the hyper-processor has been moved in between the message block and the switch.
image

### DIFF
--- a/core/src/mindustry/type/Category.java
+++ b/core/src/mindustry/type/Category.java
@@ -1,34 +1,70 @@
 package mindustry.type;
 
-public enum Category{
-    /** Offensive turrets. */
-    turret,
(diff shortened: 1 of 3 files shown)


Anuken:
> but does not add proper UI support for it.

I don't think any PR like this should be implemented until there is a good demonstration of a system that handles it on the UI-level.

Maybe there should be a separate list of icons at the top of the placement UI for mod tabs. Maybe new categories should be added to the right of the current ones... or maybe there should be a scrollpane down instead? It's something that needs to be decided.

PR #8560, opened 2023-05-05
update german translation

Catchears:
Should be compatible with #8543. Adds new keys since https://github.com/Anuken/Mindustry/commit/d81500d99d8772fdff0a7a5af05d576c923baeac.


Shawak (changes requested):

Jordansichs:
As a german i can confirm that most of the text is correct, but "diese mod"
is false, it would be called "dieser mod", if we're talking about
modifikations or moderators.

Maximilian ***@***.***> schrieb am Fr., 5. Mai 2023, 15:05:

Catchears:
> As a german i can confirm that most of the text is correct, but "diese mod" is false, it would be called "dieser mod", if we're talking about modifikations or moderators.

I'd agree with "Der Moderator", but if it's from "Die Modifikation" it should stay feminine imo.

Shawak (approved):

Jordansichs:
Your right from an gramatically view, but the word "mod" is very well known
and used in the german language, given to the fact Minecraft-youtuber used
mods very often and always used the male Version "dieser mod", "der mod"
etc.

It just sound really weird for a german ear. That's why i asked if you're
talking about modifications. We germans don't even know about the
gramatical rules, we just know which word to use throught it's sound.

I'm really sorry for my bad english, hope you understand.

Catchears ***@***.***> schrieb am Fr., 5. Mai 2023, 16:44:

Phippe:
"Der Mod" sounds better to me (probably also due to the Minecraft videos), but just because it sounds better shouldn't mean we may use an incorrect translation. Just my two cents :)

PR #8577, opened 2023-05-10
Add BlockRotateEvent

Phinner:
Basically, this PR adds an event for block rotations since monitoring them with an Action listener is not very efficient.

It also comes with a `@Nullable` annotation to the `player` field of `ConfigEvent` since configuration changes can be triggered by a plugin (I learned that the hard way through NPEs :smiling_face_with_tear:).

> Since for some reasons I can't compile locally, I will use the CI jar of this PR to test the feature. I will add a comment when I'm done :)


Phinner:
I tested it and it works :+1: 

> I used `Events.on(EventType.BuildRotateEvent, event => Call.sendMessage("Rotated " + event.build + " as " + event.unit + " to " + event.build.rotation + " with previous " + event.previous))` to test.

PR #8584, opened 2023-05-12
fix NPE in SpawnGroup.write

way-zer:

After loading some mod map, the payload will get null. Then NPE when write.


Anuken:
`Objects::isNull` is Java 8 API, which is not available on iOS / some Android versions.

PR #8599, opened 2023-05-13
Fix crash

buthed010203:


Darkness6030:
Maybe it'll be better not to call this event if the player is null?

Phinner:
> Maybe it'll be better not to call this event if the player is null?

What about logic units that place blocks ? It uses Unit to handle this case.

> Fix crash

Don't you mean "possible crash" ? The only place `Call.rotateBlock` is called have [a player that is never null](https://github.com/Anuken/Mindustry/blob/307943c098b99bf215675df0bce96832b0cef194/core/src/mindustry/input/DesktopInput.java#L403), don't scare me like that :(

PR #8654, opened 2023-05-31
Update Vietnamese translation

SITUVNgcd:
- Add @SITUVNgcd as contributor.
- Update Vietnamese translation.


Anuken:
image

File encoding broken?

SITUVNgcd:
> image
> 
> File encoding broken?

What? Ok! I will check it. Something gone wrong with my text editor.

SITUVNgcd:
@Anuken I known the problem, `^M` is a control character CR `0x0D`.
This character is usually used in Windows beside with LF (`0x0A`) (`\r\n` - CR LF).
I forgot to config my Text Editor use `\n` instead of `\r\n`.



^ | C0 | Abbr | Name | Effect
-- | -- | -- | -- | --
^M | 0x0D | CR | Carriage Return | Moves the cursor to column zero.

PR #8655, opened 2023-05-31
Make PlayerInfo map public

Darkness6030:
Because I'm tired of using Reflect to access it (for example, to find a PlayerInfo by USID)

### DIFF
--- a/core/src/mindustry/net/Administration.java
+++ b/core/src/mindustry/net/Administration.java
@@ -26,8 +26,8 @@ public class Administration{
 
 
     private boolean modified, loaded;
-    /** All player info. Maps UUIDs to info. This persists throughout restarts. Do not access directly. */
-    private ObjectMap<String, PlayerInfo> playerInfo = new ObjectMap<>();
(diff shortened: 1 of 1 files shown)


Anuken:
Maybe add methods to do what you want instead of exposing implementation details?

Darkness6030:
> Maybe add methods to do what you want instead of exposing implementation details?

What is the best method name you can choose in your opinion? "getPlayerInfos" / "getPlayerInfoMap" / something else?

Anuken:
No, I mean, to do things like

> find a PlayerInfo by USID

> getPlayerInfos

Adding a getter doesn't solve anything, since you can easily just mutate the map from there, and it still exposes the same implementation details.

Darkness6030:
> exposes the same implementation details.

I don't think it's a problem. What really bad can happen if the plugin changes some entries ​​in this map? 👀

Anuken:
> What really bad can happen if the plugin changes some entries ​​in this map? 

It won't get saved, or it will break plugins if the storage data structure ever changes. But if you're really fine with being able to shoot yourself in the foot with this in the future, sure.

PR #8666, opened 2023-06-03
Adapt interface to display notch of mac laptops

alxgsv:
Hey! 

- Just bought a game, immediately liked it. 
- However on fullscreen of macbooks with notches on screen, HUDFragment appears below the notch (so, basically, invisible). 
- I found [another guy who struggles with it](https://steamcommunity.com/app/1127400/discussions/0/3843304884854109789/).
- Initial idea was to make a mod, but after inspecting game code, I think it's impossible to adjust interface with mods.
- Sorry, this is the first time I edit Java code.
- Height of notch is 32 points. 
- This PR works for me with one critical exception: I didn't find a way to re-initialize UI after settings changed. 

Please help. Or feel free to drop it, if it's completely wrong. I can play my custom build, just wanted to share with people.


core/assets/bundles/bundle.properties
```diff
@@ -1115,6 +1115,7 @@ setting.bridgeopacity.name = Bridge Opacity
 setting.playerchat.name = Display Player Bubble Chat
 setting.showweather.name = Show Weather Graphics
 setting.hidedisplays.name = Hide Logic Displays
+setting.macnotch.name = Adapt interface to display notch
```
Anuken, line 1118: If it's not a dynamic, add a tooltip `(provide a .tooltip entry)` that explains that it takes effect after restart.

core/src/mindustry/ui/fragments/HudFragment.java
```diff
@@ -293,6 +293,11 @@ public void build(Group parent){
         //core info
         parent.fill(t -> {
             t.top();
+
+            if(Core.settings.getBool("macnotch") ){
+                t.margin(32);
```
Anuken, line 298: Is `32` always the correct value for this? Does it vary by device and monitor settings?

alxgsv: AFAIK, all notches are of the same height _currently_. 
And I didn't find a way to get notch dimensions dynamically in Java app on mac. 
I know that magic number are better to store as constants and reuse (I use this one twice), but not sure how to do it in Java, and in Mindustry project style.

core/assets/bundles/bundle.properties
alxgsv: Thank you! Will do!

core/src/mindustry/ui/fragments/HudFragment.java
Anuken: Maybe add a `public static float macNotchHeight = 32f` in `Vars`

core/assets/bundles/bundle.properties
alxgsv: Added `setting.macnotch.name.tooltip.tooltip` to all languages, but it's not fetched automatically by renderer. I don't think it's worth to adapt settings renderer for such a small tweak. Maybe I'll just do it like this?

```
setting.macnotch.name = Adapt interface to display notch (restart required)
```

Anuken: My bad, it's supposed to be `setting.macnotch.description`

alxgsv:
Thank you! Pushed a commit with both improvements.

PR #8670, opened 2023-06-04
Weather World Processor

Mythril382:
Uhhh... don't mind the commit history.

- [ ] I have ensured that any new features in this PR function correctly in-game, if applicable. (Need testing.)


core/src/mindustry/logic/LExecutor.java
```diff
@@ -1727,6 +1727,32 @@ public void run(LExecutor exec){
             }
         }
     }
-
+    
+    public static class ApplyWeatherI implements LInstruction{
+        public String weather;
```
Anuken, line 1732: This should be a variable that points to a weather constant like every other instruction, not a hard-coded string

Mythril382: Isn't this what Apply Effect used?

PR #8674, opened 2023-06-04
Configurable unit movement speed

chris-nada:
Very straight forward:
Makes unit (x/y)-movement speed configurable, 1 is default of course:

image


MEEPofFaith:
what about the range of missiles?
those are units as well

don't want 0.5x movement speed cutting the range of scathe, quell, disrupt, and anthicus in half.

Anuken:
I'm also concerned about what happens when this number is set to 0 or some ridiculously high value.

MEEPofFaith:
~~also did you seriously manually copy the bundle entry to every one of the bundles? there's an action that does this for you after merging~~

chris-nada:
> I'm also concerned about what happens when this number is set to 0 or some ridiculously high value.

There are lots of settings you can set to ridiculous values. What happens?  Fun of course.

MEEPofFaith:
ok but what about the missile problem

Anuken:
> What happens? Fun of course.

No, bug reports, actually. Many of the settings are capped because the game explodes when you set them too high or low.

JasonP01:
@Anuken  there is a way to see what happens when unit speed gets WAY too high (yes ik this is a very old pullrequest) anyway if u set dragMultiplier to -1 the units speed up infinitely, even after respawning they keep moving, infact is goes so fast that on pc the camera goes to maxInt and beyond, i played with this making a map and its unplayable (the only way to stop urself from moving is teleporting or removing drag from ur unit)         

players did complain about the infinite speed so u were correct about the bug reports thing   
(btw since u might see this can u make 0 drag work correctly, currently the units still stop moving even when its set to 0 which is less than ideal when making a 0 gravity map)

PR #8683, opened 2023-06-07
Update README.md with download widget

mathijs81:
Via AppBrain it's now possible to have a shield widget that shows the current download count and I thought that would be great to show on the first page so people can see how popular Mindustry is.

It looks like this:

image

(no code changes so no tests done)


Anuken:
I don't think this is necessary.

PR #8699, opened 2023-06-10
Save map tags

Darkness6030:
So the map name/author/description will be saved 👀 


Anuken:
This completely breaks the game. The tags provided overwrite the *standard* meta tags, which means the game rules of the map will forcefully overwrite the actual rules you have in the save - this means your saves will get un-sandboxed, any custom rules you set will be wiped, etc.

PR #8740, opened 2023-06-22
Better BuildRotate and Config events

Darkness6030:
(really better)

### DIFF
--- a/core/src/mindustry/game/EventType.java
+++ b/core/src/mindustry/game/EventType.java
@@ -261,12 +261,13 @@ public class EventType{
     public static class ConfigEvent{
         public final Building tile;
         public final @Nullable Player player;
-        public final Object value;
+        public final Object value, previous;
 
-        public ConfigEvent(Building tile, @Nullable Player player, Object value){
+        public ConfigEvent(Building tile, @Nullable Player player, Object value, Object previous){
             this.tile = tile;
             this.player = player;
             this.value = value;
+            this.previous = previous;
         }
     }
 
@@ -476,11 +477,12 @@ public class EventType{
     public static class BuildRotateEvent{
         public final Building build;
         public final @Nullable Unit unit;
-        public final int previous;
+        public final int rotation, previous;
 
-        public BuildRotateEvent(Building build, @Nullable Unit unit, int previous){
+        public BuildRotateEvent(Building build, @Nullable Unit unit, int rotation, int previous){
             this.build = build;
             this.unit = unit;
+            this.rotation = rotation;
             this.previous = previous;
         }
     }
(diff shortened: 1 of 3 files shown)


Darkness6030:
Tested with every rotatable / configurable block, everything works

core/src/mindustry/input/InputHandler.java
```diff
@@ -468,8 +468,11 @@ public static void rotateBlock(@Nullable Player player, Building build, boolean
         build.rotation = Mathf.mod(build.rotation + Mathf.sign(direction), 4);
         build.updateProximity();
         build.noSleep();
-        Fx.rotateBlock.at(build.x, build.y, build.block.size);
-        Events.fire(new BuildRotateEvent(build, player == null ? null : player.unit(), previous));
+        
+        if (build.rotation != previous) {
```
Anuken, line 472: incorrect formatting everywhere

Darkness6030: bruh I forgor

core/src/mindustry/input/InputHandler.java
```diff
@@ -488,8 +491,13 @@ public static void tileConfig(@Nullable Player player, Building build, @Nullable
 
             throw new ValidateException(player, "Player cannot configure a tile.");
         }
+
+        var previous = build.config();
         build.configured(player == null || player.dead() ? null : player.unit(), value);
-        Core.app.post(() -> Events.fire(new ConfigEvent(build, player, value)));
+
+        if(build.config() != previous){
```
Anuken, line 498: Don't call config() twice again, this will cause logic blocks to compress all their code again. Just use `value`.

core/src/mindustry/game/EventType.java
```diff
@@ -476,11 +477,12 @@ public BlockBuildEndEvent(Tile tile, @Nullable Unit unit, Team team, boolean bre
...
         public final Building build;
         public final @Nullable Unit unit;
-        public final int previous;
+        public final int rotation, previous;
 
-        public BuildRotateEvent(Building build, @Nullable Unit unit, int previous){
+        public BuildRotateEvent(Building build, @Nullable Unit unit, int rotation, int previous){
             this.build = build;
             this.unit = unit;
+            this.rotation = rotation;
```
buthed010203, line 485: What is the purpose of this? Seems redundant since this is already stored in `build.rotation`. If anything should be added to this event, it's the direction considering that it would remove the need for hacks like these (only really useful for quick rotations but still): https://github.com/mindustry-antigrief/mindustry-client/blob/b14aa9c94e756915cc0089ab9105acd93f2a9aaf/core/src/mindustry/client/utils/ClientUtils.kt#L471

core/src/mindustry/input/InputHandler.java
buthed010203: Kind of strange that the building is always configured and rate limit is always consumed but the event is only fired when the config has actually changed

Darkness6030: The problem is that `value` isn't always identical to `config()` result

way-zer:
A less relevant idea, is it possible to treat the processor as a unit in some action events.

buthed010203:
But why

way-zer:
> But why

Why not?

Why only Player but no Processor?

Why building modified without event?

PR #8747, opened 2023-06-22
Use Base64 to Encode schematic filename

CaryCatZ:
Some "unsafe" strings would be replaced with the same form. For example, “啊” and 额" would be replaced with "_", this may cause some problems in opera schematic files manually. So, use Base64 to encode filename and this also avoid "finding file" for many times.

### DIFF
--- a/core/src/mindustry/game/Schematics.java
+++ b/core/src/mindustry/game/Schematics.java
@@ -36,6 +36,8 @@ import mindustry.world.blocks.storage.CoreBlock.*;
 import mindustry.world.meta.*;
 
 import java.io.*;
+import java.nio.charset.*;
+import java.util.*;
 import java.util.zip.*;
 
 import static mindustry.Vars.*;
@@ -136,6 +138,10 @@ public class Schematics implements Loadable{
 
         try{
             Schematic s = read(file);
+            String expectedFilename = Base64.getUrlEncoder().encodeToString(s.name().getBytes(StandardCharsets.UTF_8));
+            if(!file.file().getName().split("_")[0].equals(expectedFilename)){//check if filename is compliant
+                file.moveTo(findFile(expectedFilename));
+            }
             all.add(s);
             checkLoadout(s, true);
 


Anuken:
This has many issues:

- Uses StandardCharsets, which is not supported on the target Android API
- Uses the Java base64 API instead of the Arc one
- Doesn't solve the problem of files being written with underscores in the first place
- Doesn't actually add support for these characters; the filesystem likely supports writing files with UTF-8 names, but there is no straightforward way of testing what is a valid filename and what isn't (what about reserved windows filenames? which characters are illegal? whitelisting won't work when one needs to support cyrillic/japanese/korean/chinese/etc)

CaryCatZ:
Or directly use user input as filename, if invalid, report error to user?

Anuken:
No, that's even worse. Most people don't care about the filename, and now *everyone* has to worry about what is and isn't an invalid character. What should the dialog even suggest, "your schematic has invalid characters, remove random things until it works?"

CaryCatZ:
Or detecting the type of user's OS, it would mostly be Windows or a Unix-like system. This means most of users' systems support non-ASCII chars in filename.
That is, in most situations, we can hard-code invalid filename characters of each system, and verify filenames, give suggestions. Even if some other chars are not allowed, we are still able to use `Strings.sanitizeFilename` as a fallback.

Links:
  1. [Invalid Chars on Windows](https://learn.microsoft.com/windows/win32/fileio/naming-a-file)
  2. [Invalid Chars on Unix & Linux](https://unix.stackexchange.com/a/230299)
  3. [Invalid Chars for Windows in .Net](https://github.com/dotnet/runtime/blob/HEAD/src/libraries/System.Private.CoreLib/src/System/IO/Path.Windows.cs#L15)
  4. [Invalid Chars for Unix in .Net](https://github.com/dotnet/runtime/blob/HEAD/src/libraries/System.Private.CoreLib/src/System/IO/Path.Unix.cs#L12C4)

Anuken:
> give suggestions.

Most users *should not care* what the internal filename is, and at no point should schematic saving fail or give any sort of popup because we can't find a proper name for it.

> Links:

Well, that's a long list of invalid characters and invalid names, so as mentioned above, all of those need to be handled correctly by the sanitizer.

CaryCatZ:
Yes, users should not care it.
And it seems there is not any best solution for this problem, and the best way is to keep it.

PR #8783, opened 2023-07-01
atNotShoot option for StatusFieldAbillity

Otamamori917:
atNotShoot a setting for StatusFieldAbillity only activates the ability when the unit is not shooting
when atNotShoot and onShoot are both false the ability acts normally
when when one is true they do their corresponding attribute 
when both are true the ability does nothing 

you can test yourself here https://github.com/Otamamori917/Axthrix-Modded-Java/actions/runs/5429503561
install mod and spawn a - b - c - d 
- a has both false.
- b has onShoot true.
- c has atNotShoot true.
- d has both true.

Thanks to @MEEPofFaith he did most of it


Otamamori917:
Could use a better name tbh

Otamamori917:
was this ignored or unnoticed 
no idea but been up for too long

PR #8817, opened 2023-07-15
Expose NetServer currentlyKicking for plugin access

BalaM314:

### DIFF
--- a/core/src/mindustry/core/NetServer.java
+++ b/core/src/mindustry/core/NetServer.java
@@ -98,8 +98,15 @@ public class NetServer implements ApplicationListener{
     private boolean closing = false, pvpAutoPaused = true;
     private Interval timer = new Interval(10);
     private IntSet buildHealthChanged = new IntSet();
+    
     /** Current kick sessions. */
     public VoteSession[] currentlyKicking = {null};
+    /** Duration of a kick in seconds. */
+    public static final int kickDuration = 60 * 60;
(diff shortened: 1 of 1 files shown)


BalaM314:
image
what

Phinner:
Wouldn't it be better to just copy the existing code and implement your own votekick command ?

BalaM314:
I'm trying to copy the "cancel votekick" code, but can't because closure
If you mean copy the entirety of votekick, that is planned but will take a while

buthed010203:
If you're going to make the various kick related variables public, it may be a good idea to make them not final so that servers can change them without writing their own kick logic

core/src/mindustry/core/NetServer.java
```diff
@@ -98,6 +98,15 @@ public class NetServer implements ApplicationListener{
     private boolean closing = false, pvpAutoPaused = true;
     private Interval timer = new Interval(10);
     private IntSet buildHealthChanged = new IntSet();
+    
+    /** Current kick sessions. */
+    public VoteSession[] currentlyKicking = {null};
```
Anuken, line 103: There is not reason to make this an array if it's not a local variable. Also, mark it `@Nullable.`

BalaM314: Oops, didn't realize it was an array only to allow other code to modify it

PR #8835, opened 2023-07-24
Same USID for same IP

Darkness6030:
⚠️ Needs discussing
- Will it work for IPv6?
- Will it be safe?
- What will happen with old USIDs?
- It will reset all admin or whitelist statuses.

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -655,7 +655,7 @@ public class NetClient implements ApplicationListener{
             ip = ip.substring(ip.indexOf("/") + 1);
         }
 
-        int colonIndex = address.lastIndexOf(":");
+        int colonIndex = ip.lastIndexOf(":");
         if (colonIndex != -1) {
             ip = ip.substring(0, colonIndex);
         }


Anuken:
> It will reset all admin or whitelist statuses.

I don't think this is acceptable.

Darkness6030:
> I don't think this is acceptable.

I just don't see any good idea how to sync player data between servers without the risk of being hacked. 
UUIDs are stolen all the time, so this is not an option.

squi2rel:
Some different panel servers are using the same IP address. The only difference is the port. It's not so good to do this

BalaM314:
It would be possible to keep old USIDs, by looping through each settings key and converting them.
However, I don't think this is a good idea. USID should be unique to each server, not just IP address.
Playerdata sync problems can be solved by storing USID as a Record<ip, usid>, then have the server determine its ip on startup.

buthed010203:
The best solution I can think of to reliably link a player's account across servers would be to send them from one server to the next via a command or something and just look for a player with the same uuid joining on the other server. Not the greatest solution but would probably work decently

PR #8891, opened 2023-08-03
Improve SyncI

BasedUser:
# what

This PR increases the sync rate from 10 times per second (100ms) to 66.6 times per second (15ms).

# why

You might want to sync some things a lot faster for precision (or other networking-related) reasons. However, there's no real point to synchronize anything faster than once every tick, because the vast majority of Mindustry players run at 60 FPS, and synchronizing a variable faster than that is wasteful. It also isn't a multiple-orders-of-magnitude level increase, your code would have to intentionally be bad (from a mapper PoV) in order for problems to occur.

I chose 15ms because it allows some more leeway with frametimes.

# what's the worst that can happen

Bandwidth might increase up to 6x for badly written code. Not really within crashing limits, but still probably on the same importance as `setblock` and `explosion` spam (tested just now on a statistically average client-server pair)


way-zer:
Maybe just at most once pre tick?

BasedUser:
> Maybe just at most once pre tick?

what do you think this does

BasedUser:
stale

PR #8899, opened 2023-08-06
Aegires Nerf

Zerenta:
Nerfed healing (3% to 1.5% on non-aegires, 0.5% on aegires) because 3% resulted in large groups of aegires healing eachother faster than damage (from endgame, surge tier defenses (aegires is phase tier)) could be dealt, 1 omura (a T5) has large issues trying to kill 2 aegires (T4s) when the healing is 3%
Buffed segments (reload 6 frames to 4 frames, 25 damage to 30 damage)
This is because aegires are overpowered as hell when in groups

### DIFF
--- a/core/src/mindustry/entities/abilities/EnergyFieldAbility.java
+++ b/core/src/mindustry/entities/abilities/EnergyFieldAbility.java
@@ -29,6 +29,8 @@ public class EnergyFieldAbility extends Ability{
     public boolean targetGround = true, targetAir = true, hitBuildings = true, hitUnits = true;
     public int maxTargets = 25;
     public float healPercent = 3f;
+    /** Multiplies healing to units of the same type by this amount. */
+    public float sameTypeHealMult = 1f;
 
     public float layer = Layer.bullet - 0.001f, blinkScl = 20f, blinkSize = 0.1f;
     public float effectRadius = 5f, sectorRad = 0.14f, rotateSpeed = 0.5f;
@@ -129,7 +131,8 @@ public class EnergyFieldAbility extends Ability{
                 if(((Teamc)other).team() == unit.team){
                     if(other.damaged()){
                         anyNearby = true;
-                        other.heal(healPercent / 100f * other.maxHealth());
+                        float healMult = (other instanceof Unit u && u.type == unit.type) ? sameTypeHealMult : 1f;
+                        other.heal(healPercent / 100f * other.maxHealth() * healMult);
                         healEffect.at(other);
                         damageEffect.at(rx, ry, 0f, color, other);
                         hitEffect.at(rx, ry, unit.angleTo(other), color);
--- a/core/src/mindustry/entities/abilities/EnergyFieldAbility.java
+++ b/core/src/mindustry/entities/abilities/EnergyFieldAbility.java
@@ -28,7 +28,7 @@ public class EnergyFieldAbility extends Ability{
     public float x, y;
     public boolean targetGround = true, targetAir = true, hitBuildings = true, hitUnits = true;
     public int maxTargets = 25;
-    public float healPercent = 1f;
+    public float healPercent = 3f;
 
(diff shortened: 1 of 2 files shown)


UnionofSovietSocialistRepublics:
We demand aegries nerd 🗣️📣🔥🔥🔥

Lime712:
Nerb now! ⛈️🗣️🔊📢📣🔥🧙

Gwapowo:
Nerp yourself, now! 🗣️📣💥

core/src/mindustry/entities/abilities/EnergyFieldAbility.java
```diff
@@ -28,7 +28,7 @@ public class EnergyFieldAbility extends Ability{
     public float x, y;
     public boolean targetGround = true, targetAir = true, hitBuildings = true, hitUnits = true;
     public int maxTargets = 25;
-    public float healPercent = 3f;
+    public float healPercent = 1f;
```
BalaM314, line 31: This is not the right way to do this, you are changing the default setting, which will affect mods. Change it in UnitTypes::aegires instead. (add a line inside the init block `healPercent = 1f`)

BalaM314 (changes requested):

BalaM314:
Aegires is absolutely broken in groups and definitely needs a nerf, but 3% to 1% might be a bit much. Maybe consider nerfing healing between aegires? (aegires heals 1.5% on other units, but only 0.5% on other aegires) I think that could work, stops aegires from being unkillable while keeping them useful as a support unit)
We also need to consider campaign, where the unit cap is 24... would probably still be fine

core/src/mindustry/entities/abilities/EnergyFieldAbility.java
Zerenta: Tysm! (For others reading this has been fixed)

UnionofSovietSocialistRepublics:
Halai

Zerenta:
> 



> Aegires is absolutely broken in groups and definitely needs a nerf, but 3% to 1% might be a bit much. Maybe consider nerfing healing between aegires? (aegires heals 1.5% on other units, but only 0.5% on other aegires) I think that could work, stops aegires from being unkillable while keeping them useful as a support unit) We also need to consider campaign, where the unit cap is 24... would prob

Zerenta:
Shit accidental

Zerenta:
> Aegires is absolutely broken in groups and definitely needs a nerf, but 3% to 1% might be a bit much. Maybe consider nerfing healing between aegires? (aegires heals 1.5% on other units, but only 0.5% on other aegires) I think that could work, stops aegires from being unkillable while keeping them useful as a support unit) We also need to consider campaign, where the unit cap is 24... would probably still be fine

Ye a 1.5% heal was considered but with how op 3% was, it felt like 1% was the only way to make it actually reasonable
I also don't know how to do 0.5 to just other aegires

Lime712:
we want nerve!

BalaM314:
> > Aegires is absolutely broken in groups and definitely needs a nerf, but 3% to 1% might be a bit much. Maybe consider nerfing healing between aegires? (aegires heals 1.5% on other units, but only 0.5% on other aegires) I think that could work, stops aegires from being unkillable while keeping them useful as a support unit) We also need to consider campaign, where the unit cap is 24... would probably still be fine
> 
> Ye a 1.5% heal was considered but with how op 3% was, it felt like 1% was the only way to make it actually reasonable I also don't know how to do 0.5 to just other aegires

Will do (ofc it needs to be generic)

UnionofSovietSocialistRepublics:
Nerf aegries to the ground

core/src/mindustry/ai/UnitCommand.java
```diff
@@ -19,6 +19,10 @@ public class UnitCommand{
         drawTarget = true;
         resetTarget = false;
     }},
+    ramCommand = new UnitCommand("ram", "modeAttack", u -> null){{
```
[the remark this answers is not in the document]
Zerenta: yeah my bad

PR #8910, opened 2023-08-10
Update servers_v7.json

a-big-fish-fish:
Add a new server


camelStyleUser:
somka there is already a server called fish

PR #8920, opened 2023-08-12
Fixed: Units lose LogicAI after reading.

MinRi2:

### DIFF
--- a/core/src/mindustry/ai/types/LogicAI.java
+++ b/core/src/mindustry/ai/types/LogicAI.java
@@ -40,6 +40,12 @@ public class LogicAI extends AIController{
     private float lastMoveX, lastMoveY;
     private int lastPathId = 0;
 
+    // LogicAI state should not be reset after reading.
+    @Override
+    public boolean keepState(){
+        return true;
+    }
+
     @Override
     public void updateMovement(){
         if(control == LUnitControl.pathfind){


Anuken:
What bug does this fix?

MinRi2:
> What bug does this fix?

Units read their controller correctly but in the function `afterRead` the controller which don't keep state such as `LogicAI` will be lost. And some logic players are confused about that.

Functuon afterRead in UnitComp is as follows:

```java
@Override
public void afterRead(){
    afterSync();
    //reset controller state
    if(!(controller instanceof AIController ai && ai.keepState())){
        controller(type.createController(self()));
    }
}

Anuken:
Yes, but *what actual bug* does it fix? What changes does this make on the client end?

MinRi2:
> 

Some logic players try to use @controller to find back their unit after sync but units losing LogicAI makes it imposible...

MinRi2:
> Yes, but _what actual bug_ does it fix? What changes does this make on the client end? How does this appear differently to a player?

Some logic players try to use `@controller` to find back their units after sync. But units losing LogicAI makes it imposible...

kJNCwBUq:
> Yes, but _what actual bug_ does it fix? What changes does this make on the client end? How does this appear differently to a player?

我是一名来自中国的玩家，不是很熟悉英语。
所以下文的英语部分由翻译器处理，可能会有许多问题。

I'm a gamer from China and I'm not very familiar with English. 
So the English part of the following is handled by the translator, 
which can have many problems.


在我的逻辑程序中，
我使用@controller和@time让多个逻辑块同时
工作的时候不会控制同一单位而冲突。
一个最简单的程序如下：

In my logic program, 
I use @controller and @time so that multiple logical blocks work at the same time 
without controlling the same unit and conflicting.
and the simplest program is as follows:


ubind @flare
sensor uControlled @unit @controlled
jump 0 notEqual uControlled false
ucontrol within 0 0 0 0 0
set t @time
jump 5 equal @time t
set t @time
jump 7 equal @time t
sensor uController @unit @controller
jump 0 notEqual uController @this
sensor dead @unit @dead
ucontrol move @thisx @thisy 0 0 0
jump 10 equal dead false


这种获取单位的方法理论上完全可以被应用到更复杂的大规模单位控制程序中，
但是这种方式要求单位始终保持受到逻辑块控制才稳定。

This method of acquiring units can theoretically be applied to more 
complex large-scale unit control programs, but this method requires that the units remain controlled 
by logical blocks to be stable.

PR #8931, opened 2023-08-14
Update servers_v7.json

a-big-fish-fish:


a-big-fish-fish:
This server no longer belongs to me

PR #8943, opened 2023-08-17
Make heat conductors under bullets

nanmenyangde:
it can be built in quantities in a short time to absorb the bullets.


NazrinNya:
sounds fair since they got plenty of health, but i think we should do the same thing with heat routers shall we?

PR #8981, opened 2023-08-27
most ability displays

stacktrace-error:
surgically removed the format markers from the existing names with unsurgical precision, should fix for most latinish alphabet languages but may break others

extremely stupid usage of `waves.shields` in shieldregenfield, I don't know whether that has to be added as its own bundle nor how to name it

image

image

image

  - Tested with all vanilla units.


stacktrace-error:
I don't have a fix for the minor text offset on empty displays rn, don't know whether there's something to use besides pad for it

stacktrace-error:
should there be 2 abilities per line?

Darkness6030 (changes requested):
What is this comment?

core/src/mindustry/entities/abilities/ShieldRegenFieldAbility.java
```diff
@@ -1,9 +1,15 @@
 package mindustry.entities.abilities;
 
+import arc.Core;
+import arc.scene.ui.layout.Table;
 import arc.util.*;
 import mindustry.content.*;
 import mindustry.entities.*;
 import mindustry.gen.*;
+import mindustry.world.meta.Stat;
```
Anuken, line 9: use wildcard imports everywhere

PR #9018, opened 2023-09-04
add `@server` to logic

way-zer:
**@server**: means server, room owner, local

There are many logic operations only work in server, like `setblock`.  
And it's useful to sync variables with `syncI` with such pattern.  
```js
//init variables
while (true) {
    if (@server) {
        //read memory
        //operation
    }
    //sync
    //display and write memory
}
```

---


Anuken:
Very confusing name should be `@server`

way-zer:
> Very confusing name should be `@server`

OK

PR #9028, opened 2023-09-06
Update servers_v7.json

Redik2:
Idk if i need to write description, but i will. My server is about pvp in open world, that generates every round again. Also there are serpulo and erekir content, that balanced not bad.


Redstonneur1256:
There should be two D's on "address", the server won't work unless this is fixed.

Redik2:
> There should be two D's on "address", the server won't work unless this is fixed.

Oh... so cringe mistake) Sry

PR #9068, opened 2023-09-18
(draft) StatusEffect with level

way-zer:
StatusEffect with `level` like `duration`.  Like `Fast II` `QuickBuild IV` which act like 4*Status

Also statusEffect like `burning` can damage `damage*level`.
Allow procssor and mods to do more modify to single unit.

TODO:
* `TypeIO` and Backwards compatibility
* may provide some base status like `Attack` `QuickReload` `QuickBuild`.
---

### DIFF
--- a/core/src/mindustry/content/StatusEffects.java
+++ b/core/src/mindustry/content/StatusEffects.java
@@ -25,10 +25,10 @@ public class StatusEffects{
 
             init(() -> {
                 opposite(wet, freezing);
-                affinity(tarred, (unit, result, time) -> {
-                    unit.damagePierce(transitionDamage);
+                affinity(tarred, (unit, result, time, level) -> {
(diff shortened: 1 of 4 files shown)


Darkness6030:
Also, it'll be very good to add the "alwaysEffect" field to UnitType
So the unit will always have this effect

Can you please do this for me? 👀

way-zer:
> Also, it'll be very good to add the "alwaysEffect" field to UnitType So the unit will always have this effect
> 
> Can you please do this for me? 👀

Maybe you can use custom `Ability`?

Darkness6030:
> Maybe you can use custom `Ability`?

🤔

Darkness6030:
Why closing it? It was a good idea...

Anuken:
Maybe a better approach would be to have a `@Nullable` modifiers object in units that contains direct multipliers for health/damage/etc.

PR #9081, opened 2023-09-22
Minor buffs to Scathe

python-nerd-3:

Scathe is highly known for being a very weak tower that is only used as a big radar. I changed two of its attributes (health and damage) to make it slightly more viable. I did not highly increase the stats to avoid it from becoming overpowered while still making it slightly more viable. If you would like you can more drastically alter the stats.

### DIFF
--- a/core/assets/contributors
+++ b/core/assets/contributors
@@ -161,3 +161,4 @@ SITUVNgcd
 Gabriel "red" Fondato
 Yoru Kitsune
 OpalSoPL
+sc3d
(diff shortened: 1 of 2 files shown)


Anuken:
Scathe was given significant buffs (2x+ unit damage) since last release. Further buffs wouldn't make sense right now, especially one that only increases damage by ~3%.

python-nerd-3:
I really need to look at the code more when making PRs

PR #9086, opened 2023-09-23
Optimize `wait`/`message`

way-zer:
1. `wait 0.001` use as yield, to ensure later instructions in the same tick
2. ~~`message label` can let map maker do many interesting things.~~(Duplicates with `marks`)
3. fix `message` not clear `textBuffer`
4. `message` now return `success` instead of waiting for available. (useful for message loop to flush latest info)
---


way-zer:
Should implement back backward compatibility for `outSuceess`? (fallback to blocking when `outSuccess` is special value.)

PR #9087, opened 2023-09-23
Add marker instruction to world processors

ApsZoldat:
Code changes:
1. Add wproc-controlled markers sequence (`IntMap`) to `Rules` (each marker has its own identifier)
2. New `Marker` instruction for world processors
3. Add `hidden` boolean and some controlling methods to `ObjectiveMarker` for use in world processors
4. Add new `LineMarker` to objective markers
5. `ObjectiveMarker` writer/reader methods in `TypeIO`, new `@Remote setMarker()` method

P.S: New `TypeIO` methods read/write a single marker data, not the whole sequence. And the whole markers sequence isn't being synced except server's `Rules` sync. This is probably more optimized and creates an opportunity to desync markers between server and clients (for example, this is really useful for displaying different markers text depending on player's selected locale).
_This is my first PR ever, so i could have screwed up anywhere in the code, even though i tested these features on both client and local server as much as i could_

---


core/src/mindustry/game/MapObjectives.java
```diff
@@ -612,6 +616,17 @@ public void drawMinimap(MinimapRenderer minimap){}
         public void added(){}
         /** Remove any UI elements, if necessary. */
         public void removed(){}
+        /** Control marker with world processor code*/
+        public void control(LMarkerControl type, float... values){
```
Anuken, line 620: Don't use varargs, they allocate arrays! There should be N parameters, where N is the maximum amount of parameters that may ever be used.

core/src/mindustry/logic/LStatements.java
```diff
@@ -148,7 +149,8 @@ void rebuild(Table table){
                     }
                     rebuild(table);
                 }, 2, cell -> cell.size(100, 50)));
-            }, Styles.logict, () -> {}).size(90, 40).color(table.color).left().padLeft(2);
+            }, Styles.logict, () -> {
```
Anuken, line 152: Lots of redundant reformatting all over this file, revert it.

core/src/mindustry/logic/LMarkerControl.java
```diff
@@ -0,0 +1,39 @@
...
+public enum LMarkerControl{
+    create("type", "x", "y"),
+    remove,
+    hide,
+    show,
+    toggleVisibility,
+    setVisibility("true/false"),
+    setText("text"),
+    flushText,
+    setX("x"),
```
Anuken, line 12: I don't think any of these should have the set_ prefix, it's clear enough what they do anyway.

Anuken (changes requested):

TheEt1234:
oh my god this will be awesomee

ApsZoldat:
Commited a last one small sudden change from mobile before PR possibly gets merged
I hope this is the last one

core/src/mindustry/game/MapObjectives.java
```diff
@@ -685,6 +703,54 @@ public void draw(){
...
+            this.text = text;
+            if(fetch){
+                fetchedText = fetchText(this.text);
+            }else{
+                fetchedText = this.text;
+            }
+        }
+
+        @Override
+        public void setColor(double value){
```
Anuken, line 751: Why is setColor a separate method - couldn't it be part of control, if the parameters were doubles?

ApsZoldat: Actually, earlier (even before the first commit) it used to be double, but there were tons of float typecasts in control() method, so i changed 
control()'s args type to float and created another method for controlling color as it strictly requires double value.

It's a choice between having a separate method for that (less optimized i guess) or having these typecasts in control(). Can revert this back if needed.

PR #9103, opened 2023-09-28
Add sensor for player camera

Redstonneur1256:
As the title says, it adds `cameraX`, `cameraY`, `cameraWidth` and `cameraHeight` to logic sensor operation.


Anuken:
1. What is the use case for this?
2. This code doesn't work properly on servers; it returns ` player.con.viewWidth / 2` instead of` player.con.viewX` for some reason. Why?

Redstonneur1256:
Logic lets us control the camera position and zoom level. Be able to access it can have many uses such as adding some custom UI elements, etc... as said in the [logic channel](https://discord.com/channels/391020510269669376/742769933926269069/1156756897848901734).
For the `player.con.viewWidth / 2` it was because I had `player.con.viewX - player.con.viewWidth / 2` from past experiences and I then removed the wrong part of the code, this should be fixed in the latest commit.

Anuken:
> custom UI elements

But these can't be made individual on servers, it sounds like a bad idea. How do you envision this working?

PR #9107, opened 2023-09-28
update bundle_es properties

ArkSourcer:
description of Oncet:

EN = Commence the conquest of Erekir. Gather resources, produce units, and begin researching technology.

ES = Comienza la conquista de Erekir. Reúne recursos, produce unidades y comienza a investigar tecnología.

and other translations from the schematics and editor section.


ArkSourcer:
*Dies**

PR #9138, opened 2023-10-04
Add privileged desynced client global variables

ApsZoldat:
A bit controversial feature: added new global variables `@clientLocale`, `@clientUnit`, `@clientName`, `@clientTeam`

These variables are used for managing custom world processor desync, which is weird, but allows displaying different information to different players. The implementation also required making these variables available only to privileged processors, as regular processors shouldn't be able to get client unit wherever it is.

example

---


Darkness6030:
@clientLocale can change sometimes, maybe you should use Vars.player.locale() instead?
Also why making these variables "privelleged"? Like what bad can even happen if they are not?

PR #9151, opened 2023-10-10
Update servers_v7.json

TheTrueInfinity:
Hello Anuken. Sorry to bother you, but only me and my deputy can delete my servers, and these servers are also backup servers in case of a DDoS attack. They will soon be used for new game modes. Thank you.

### DIFF
--- a/servers_v7.json
+++ b/servers_v7.json
@@ -124,6 +124,10 @@
     "address": ["162.248.101.95", "162.248.100.98", "162.248.100.133", "162.248.102.204"]
   },
   {
+    "name": "OmniCorp Classic II"
+    "address": ["omnidustry.ru:6567", "omnidustry.ru:6568", "omnidustry.ru:6569", "omnidustry.ru:6570", "omnidustry.ru:6571", "omnidustry.ru:6572", "omnidustry.ru:6573", "omnidustry.ru:6574"]
+  },
+  {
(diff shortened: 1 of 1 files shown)


Anuken:
Why do they appear twice on the same server list?

Darkness6030:
> Also because of DDoS

If you're getting ddosed (about it I highly doubt), you don't need to use two different hosts, because both of them can be ddosed at the same time 💀

osp54:
> > Also because of DDoS
> 
> If you're getting ddosed (about it I highly doubt), you don't need to use two different hosts, because both of them can be ddosed at the same time 💀

At the same time, the IP to which the domain redirects can be changed at any time.

SSTentacleSS:
> Why do they appear twice on the same server list?

We have been subjected to DDoS attacks on our main server for a long time, so the second server is used only as a server if the first one does not work (Anuken, what’s wrong with the community?)

Anuken:
> Anuken, what’s wrong with your community?

Don't act like I have anything to do with it. I'm not the one who opened an issue opened demanding removal of this server.

> If you really do not allow it, then you can delete the server with the address omnidustry.ddns.net. But, omnidustry.ru is our main server

I don't allow the same server group to be in the list twice; it wastes space. If you wish to replace the existing one, make a different PR.

PR #9168, opened 2023-10-14
Cleanup Voronoi

FxMorin:
Makes the `Voronoi` class static, removing the need to instantiate it. Plus a bit of cleanup


Anuken:
This still seems to involve a lot of instantiation and new classes, so... I'd rather just leave it alone.

PR #9199, opened 2023-10-25
Map-specific locale bundles system

ApsZoldat:
Really messy PR adding map-specific locale bundles for map objectives and world processors. And i'm 80% sure there's something i forgot to fix or change.

Change list:
1. Added two new world processor instructions:
- `Format` (replacing placeholders in text buffer)
- `Print Locale (privileged)` (printing locale bundles properties to text buffer using current player's selected locale (prints ???propertyname??? on server)).
2. Added `MapLocales` class to `mindustry.type` and i'm still not sure if it should be there.
3. Made map objectives and objective markers text use locale bundles properties whenever possible.
4. Made new dialog in `Map Editor > Map Info` for setting map locale bundles. Supports copying and pasting bundles as text.
5. Made map locale bundles support displaying mobile tips (searches for properties with names ending in ".mobile." first if you are on mobile device), but still added `@clientMobile` global var, may be useful too

P.S: i promise to wipe older commits from github fork next time...


eeve-lyn:
oh my god

ApsZoldat:
At this point i'm not even sure this pull is going to be merged

But it seems like i've done everything i needed to do

TheRadioactiveBanana:
Oh god a giant PR

ApsZoldat:
> Oh god a giant PR

Yes, and with funny hotfixes after another funny hotfixes

And a living hell of commits

TheRadioactiveBanana:
> > Oh god a giant PR
> 
> Yes, and with funny hotfixes after another funny hotfixes
> 
> And a living hell of commits

Truly

ApsZoldat:
This one is surely the last commit in PR, ready for review.

I've tested anything i could, if i will need to fix something, i will do it in later PRs (unless it's an extremely minor fix that doesn't change lots of the code or it's a really critical problem)

core/src/mindustry/logic/LExecutor.java
```diff
@@ -1048,6 +1048,55 @@ public static String toString(Object obj){
...
+
+        FormatI(){}
+
+        @Override
+        public void run(LExecutor exec){
+
+            if(exec.textBuffer.length() >= maxTextBuffer) return;
+
+            int placeholderIndex = -1;
+            int placeholderNumber = 10;
```
Anuken, line 1066: What is the point of placeholderNumber existing if it is never read from? There is no functional difference between `{0} {1} {2}` and `{5} {6} {7}`, which is very confusing. Why not make it match the log formatting system, and simply replace the single character `@`?

I understand that the system in `bundle.properties` uses positional indexes (`{0}`), but that's because the index actually matters. Here, it doesn't. Most people won't be using this in a world logic context, and will have no idea why the number between the brackets is there.

core/src/mindustry/editor/MapLocalesDialog.java
```diff
@@ -0,0 +1,759 @@
...
+import mindustry.gen.*;
+import mindustry.graphics.*;
+import mindustry.io.*;
+import mindustry.type.*;
+import mindustry.ui.*;
+import mindustry.ui.dialogs.*;
+
+import static mindustry.Vars.*;
+
+public class MapLocalesDialog extends BaseDialog{
```
Anuken, line 24: There's way too much code here for me to review all at once, but here's an issue I ran into immediately: Why isn't there English by default?

When I open the dialog, there are no languages on the left. I add an entry:

Screenshot from 2023-11-09 10-19-00

If I then proceed to add English, the entry disappears. Where did it go? I can't select it or go back to it:

Screenshot from 2023-11-09 10-19-18

When I copy the text, it shows up like this:

```
default:
name = text
en:
```

Why is there a "default" and an "en" section? Shouldn't en be the default? If not, why isn't there a "default" entry on the left?

Anuken (changes requested):

core/src/mindustry/editor/MapLocalesDialog.java
ApsZoldat: "default" locale just didn't display on the left as it isn't presented in Vars.locales and things like language dialog. It should've been replaced by Locale.getDefault().getLanguage(), which is done now. The property you've added just left in this default locale (which you couldn't switch to) as they don't copy when you add a new locale bundle, there are separate edit buttons for doing so.

core/src/mindustry/logic/LExecutor.java
```diff
@@ -1113,7 +1109,9 @@ public void run(LExecutor exec){
             if(exec.building(target) instanceof MessageBuild d && (d.team == exec.team || exec.privileged)){
 
                 d.message.setLength(0);
-                d.message.append(exec.textBuffer, 0, Math.min(exec.textBuffer.length(), maxTextBuffer));
+                // Backslash is an escape character for format instruction placeholders
+                String result = exec.textBuffer.toString().replace("\\@", "@");
```
Anuken, line 1113: This will allocate every single print flush. Is there any particular good reason to combine non-formatted "@" with format instructions? (Do we really need escaping?)

PR #9205, opened 2023-10-28
Update bundle_ru attributions

BasedUser:
i don't think anyone calls me by my old alias anymore


Darkness6030:
Yeah, also change me (Даркнесс#3729) to `inflexibledarkness`

PR #9218, opened 2023-10-31
setprop speed i guess

camelStyleUser:

### DIFF
--- a/core/src/mindustry/entities/comp/UnitComp.java
+++ b/core/src/mindustry/entities/comp/UnitComp.java
@@ -292,7 +292,7 @@ abstract class UnitComp implements Healthc, Physicsc, Hitboxc, Statusc, Teamc, I
                     this.team = team;
                 }
             }
-            case speed -> baseSpeed = value / 60f * tilesize
+            case speed -> baseSpeed = value / 60f * tilesize;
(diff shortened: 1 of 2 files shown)


Anuken:
This is fundamentally broken. You made it transient. Implementing setProp of any stat won't save, and *making* it save will waste network and memory.

camelStyleUser:
~~do i just make it non-transient or what~~

Anuken:
> do i just make it non-transient or what

No, because as mentioned above, making it non-transient wastes network bandwidth/save space without good reason. I don't want units to have a bunch of properties that don't need to be synced in 99% of cases.

camelStyleUser:
> > do i just make it non-transient or what
> 
> No, because as mentioned above, making it non-transient wastes network bandwidth/save space without good reason. I don't want units to have a bunch of properties that don't need to be synced in 99% of cases.

so what if i just make it non setpropable but able to be set by mods

Anuken:
Doesn't change my conclusion, and mods can override/control speed themselves in custom units anyway.

Anuken:
See https://github.com/Anuken/Mindustry/commit/e5047d752d62a0353f1ead4619ee5d4653fe9517

PR #9228, opened 2023-11-03
Assignment 1

hmgtech:

- [✓ ] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [✓] I have ensured that my code compiles, if applicable.
- [✓] I have ensured that any new features in this PR function correctly in-game, if applicable.

I have added Test cased for TileGen class of package mindustry.world;
And also covered test cases for methods of the World class of package mindustry.core;

### DIFF
--- a/tests/src/test/java/TileGenTest.java
+++ b/tests/src/test/java/TileGenTest.java
@@ -1,0 +1,38 @@
+import static org.junit.jupiter.api.Assertions.*;
+import org.junit.jupiter.api.Test;
+import mindustry.content.*;
+import mindustry.world.*;
+
+public class TileGenTest {
+
+    @Test
+    public void testInitialization() {
+        //Arrange: Initializing the titleGen objec
+        TileGen tileGen = new TileGen();
+
+        //Act As Blocks is a static metiond, it is called without creating object in the assert phase.
+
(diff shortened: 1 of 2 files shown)


Anuken:
I don't think these are useful tests.

`TileGenTest` does no meaningful testing; every field in Blocks is null unless initialized. You're comparing `null` to `null`. Even if that were not the case, you're just testing that the defaults are... set to the defaults? That doesn't matter, and would never break unless I *changed the defaults*.

WorldTest: 

- `testUnconv`: You're testing that a function correctly multiplies a value by the `tilesize`, but it's fundamentally wrong. The only reason for it to ever break is if I changed the tile size, and in that case it *should* produce different values. The test doesn't accomplish anything.
- `testTileBuildingWithNullTile`: You're testing that an empty world is empty, but that doesn't matter at all. You're not supposed to even use un-initialized worlds.
- `testTileBuildingWithNullBuild`: You're creating another empty world (invalid), empty tiles (also invalid), and then... comparing null with null? It doesn't test anything meaningful either.


Meaningful tests should verify that game mechanics or blocks/units function correctly.

PR #9270, opened 2023-11-12
increase the cost of the walls

nanmenyangde:
the walls are too cheap and too easy to be built now.


classicdirium:
Erekir Assembler units' wall cost should be decreased as they are also affected by the changes

nanmenyangde:
you are right.
it may be a better idea to revert the changes in erekir.

Ilya246:
wall cost could maybe be increased but not by that much

PR #9314, opened 2023-11-25
buff all solar panels

Ilya246:
makes them *much* cheaper and buffs small ones 6->7.2 p/s

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -2401,12 +2401,12 @@ public class Blocks{
         }};
 
         solarPanel = new SolarGenerator("solar-panel"){{
-            requirements(Category.power, with(Items.lead, 10, Items.silicon, 15));
-            powerProduction = 0.1f;
+            requirements(Category.power, with(Items.lead, 6, Items.silicon, 6));
+            powerProduction = 0.12f;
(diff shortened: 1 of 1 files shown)


Anuken:
That seems ridiculously cheap to me, 10/10 instead at most?

Ilya246:
> That seems ridiculously cheap to me, 10/10 instead at most?

this would still be 1.2 power/silicon, whereas turbines are 12 and other generators have similar figures
actually gives solar panels an use case which is using them when you have spare space and a low income of fuel-usable materials
a 20x20 square of solar panels would make you 2880 power and cost 2400 lead and silicon, this can be matched by 8 turbines which would take considerably less space and are more easily scalable, if you're fine with the coal upkeep (which should be trivial)

Kieaer:
I agree buff this.
In most resource-limited modes, placing this block is considered trolling.
It's expensive, very inefficient, and better off place other high-end buildings than place solar panels.

Anuken:
>  if you're fine with the coal upkeep (which should be trivial)

...except coal requires *a coal vein*, as well as infrastructure to route it and keep things safe from flying units. In contrast, solar panels can be spammed anywhere at any time with zero thought.

I'd accept a minor buff, but not this.

PR #9344, opened 2023-12-03
Update bundle_ca.properties

J0anJosep:


J0anJosep:
I will open a new PR with an updated translation.

PR #9356, opened 2023-12-07
Markers update

ApsZoldat:
This branch required some force-pushing due to serious skill issues (so I couldn't reopen previous PR)

# Changes:
1. Texture markers (able to draw any single texture from the game, limitless)
2. Markers now support drawing on different z layers (marker rendering moved from `OverlayRenderer.drawTop()` to `Renderer.draw()`)
3. Added minimap mode to other markers (default minimap marker just ignores this value), added autoscale option for markers.
4. `null` values now make marker control methods use previous values instead of converting `null` to 0 (small changes in `LExecutor`, `MapObjectives`), removed redundant marker control operations.
5. Minor locale bundles dialog fixes.
6. Added color pickers to some logic instruction where it's useful (new `col()` method in `LStatement`).
7. Swap 1st and 2nd argument in `makemarker` for argument consistency (usually a dropdown option comes before all the other args, and this kind of consistency is important for editing mlog code as text).


Darkness6030:
Бро реально было некуда потратить свободное время

ApsZoldat:
> Бро реально было некуда потратить свободное время

Ты не понимаешь, это такое особо опасное заболевание - тратить время и силы на ненужную работу когда у тебя куча недоделанной обязательной

core/src/mindustry/logic/LExecutor.java
```diff
@@ -1980,8 +1992,15 @@ public void run(LExecutor exec){
                 }else if(type == LMarkerControl.flushText){
                     marker.setText(exec.textBuffer.toString(), true);
                     exec.textBuffer.setLength(0);
+                }else if(type == LMarkerControl.texture){
+                    if(exec.obj(p1) != null){
+                        StringBuilder res = new StringBuilder(exec.obj(p1).toString());
```
Anuken, line 1997: This implementation is pretty weird. I don't think it should add `-` between images - what if someone wants to draw `stone1, stone2, stone3`?

Additionally, it should check for `p2` && `p3` being null, and avoid allocating the StringBuilder in that case.

core/src/mindustry/game/MapObjectives.java
```diff
@@ -719,58 +731,92 @@ public ShapeTextMarker(){}
...
             // font size cannot be 0
-            if(Math.abs(fontSize) < 1e-5) return;
+            if(Mathf.equal(fontSize, 0f)) return;
 
-            WorldLabel.drawAt(fetchedText, pos.x, pos.y + radius + textHeight, Draw.z(), flags, fontSize);
+            WorldLabel.drawAt(fetchedText, pos.x, pos.y + radius * scl + textHeight * scl, drawLayer, flags, fontSize * scl);
+        }
+
+        @Override
+        public void drawMinimap(MinimapRenderer minimap){
```
Anuken, line 759: There is a massive amount of duplication here. You're essentially copy-pasting the entire draw method for the minimap - why do the markers even need minimap-specific code?

You should not call `draw` *at all* if `hidden` is true, and you should not call `drawMinimap` / `draw` if the booleans to disable them are set. The marker should not be the one handling it in every single override of drawMinimap/draw.

ApsZoldat: About the minimap mode for other markers - it's useful when you want to make a situation awareness system (if the attack system gets too complicated). Default minimap markers blend with terrain when you, for example, display a red marker over crux base, and even if you just set them to other color - it's still not as plain as some sort of unit icon.

I'll better try to cleanup code duplication by merging actual drawing code into one method while leaving position calculations/other stuff as separate functions, maybe even the superclass methods. Will send to re-review when it will be done.

ApsZoldat:
I've refactored markers draw code as much as i could. Some overriden methods code duplication is still there (due to differences in subclasses like not having position variable in the superclass), but they're way smaller now and should at least look cleaner.

core/src/mindustry/logic/LExecutor.java
```diff
@@ -1987,17 +1987,15 @@ public void run(LExecutor exec){
...
-                    if(exec.obj(p1) != null){
-                        StringBuilder res = new StringBuilder(exec.obj(p1).toString());
-                        if(exec.obj(p2) != null) res.append("-").append(exec.obj(p2).toString());
-                        if(exec.obj(p3) != null) res.append("-").append(exec.obj(p3).toString());
-                        marker.setTexture(res.toString());
+                    if(exec.bool(p1)){
+                        marker.setTexture(exec.textBuffer.toString());
+                        exec.textBuffer.setLength(0);
+                    }else{
+                        marker.setTexture((exec.obj(p2) != null ? exec.obj(p2).toString() : "null"));
```
Anuken, line 1998: The outputs for toString() are different than what logic outputs when you try to print an object, so this should use `PrintI#toString` instead.

core/src/mindustry/game/MapObjectives.java
```diff
@@ -998,33 +1081,123 @@ public LineMarker(String text, float x1, float y1, float x2, float y2){
...
+
+        @Override
+        public void drawWorld(){
+            baseDraw(pos.x, pos.y, autoscale ? 4f / renderer.getDisplayScale() : 1f);
+        }
 
-            Lines.stroke(stroke, color);
-            Lines.line(pos1.x, pos1.y, pos2.x, pos2.y);
+        @Override
+        public void drawMinimap(MinimapRenderer minimap){
```
Anuken, line 1192: You have this same exact implementation for drawMinimap/drawWorld copy-pasted across most of the markers, maybe that should be the default implementation?

ApsZoldat: > You have this same exact implementation for drawMinimap/drawWorld copy-pasted across most of the markers, maybe that should be the default implementation?

I've explained the problem about it:
> Some overriden methods code duplication is still there (due to differences in subclasses like not having position variable in the superclass)

And i don't really know how to solve this

Anuken: How about an abstract class `Vec2Marker` / `PointMarker` that have those position variables, and override the method?

ApsZoldat:
I've just moved `pos` to superclass as it has the same name in every marker type. `MinimapMarker` has its own pos variable without `@TilePos` annotation, `LineMarker` has slightly different draw methods (with `endPos`). I've tested this change for breaking json serialization consistency, and it seems like the only marker type affected by this change is `LineMarker` (in old BE versions its pos variables were named `pos1` and `pos2`), but it doesn't really matter as `LineMarker` was introduced in BE and didn't get released yet.

core/src/mindustry/logic/LExecutor.java
BasedUser: IMO, `stone1`, `stone2` and `stone3` are themselves inconsistent. What if you wanted to have an `@type` variant drawn, but all of your code is written for conveyor rotations, animation frames and (rarely) shading?

ApsZoldat: Check the new texture implementation, you can now flush print buffer as texture name in `setmarker`. This solves the problem, i guess? Or are you talking about something else?

PR #9363, opened 2023-12-09
Fix servers-v7.json

osp54:
We need a json validator in tests 💀

### DIFF
--- a/servers_v7.json
+++ b/servers_v7.json
@@ -9,7 +9,7 @@
   },
   {
     "name": "meiqiuMDT",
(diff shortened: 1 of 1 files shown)


Anuken:
> We need a json validator in tests

There is, it's just the same one that the game uses. Extremely permissive.

osp54:
> > We need a json validator in tests
> 
> There is, it's just the same one that the game uses. Extremely permissive.

How can I get this "Extremely Permissive Parser" in JS? I have a website, and on it the statuses of the xcore servers, and because of these syntax errors, the json is simply not processed

PR #9467, opened 2024-01-17
Minor fixes

ApsZoldat:
- Made dynamic speed status effect use unit type's speed in `tiles/sec` instead of `units/tick` (was confusing for `setprop speed` users).
- Made map locales dialog convert newlines to `\n` and vice-versa on property import/export.
- Translated 2 lines in `bundle_ru.properties`, which somehow remained untranslated from v7 release.
If this translation is fine, i will work on translating all the new features later.
- Changed outdated info about `@` placeholders in map bundles info dialog.
**However, i don't know if automatic bundle update changes this in other bundles too. If not, i will change it later.**

The 🧠🔨 commit:
- Fixed incorrect calculation of columns amount in locale bundles dialog
- Replaced 8 in speed status effect fix with `Vars.tilesize`
- Map locale bundles import/export now supports escaping `\n`


camelStyleUser:
first i would want to know what will happen if i will put a backslash before a newline

Darkness6030:
You should use Vars.tilesize instead of 8f

ApsZoldat:
Commit change logs are in the updated PR comment (description)

PR #9492, opened 2024-01-27
Added logic sensor to unit spawned wave number

BlackDeluxeCat:
for mapping purposes


JasonP01:
What? @wave and @waveNumber are already a thing???

ApsZoldat:
This allows to check in which wave was unit spawned

Still dunno why and for what

JasonP01:
> This allows to check in which wave was unit spawned

9/10 times it would be from the previous waves tho?
10/10 times if u have wait for enemies toggled

ApsZoldat:
It can still be useful when, for example, giving multiple status effects to a wave without doing more complex shenanigans. The only controversy about this feature may be that "it bloats the logic system with really niche features even more". Otherwise, in my opinion, this feature is fine

Anuken:
This is incredibly niche, and increases the size of units over the network/on the disk.

PR #9498, opened 2024-01-28
Effect spawn rectangle part + fix ArmorPlateAbility stat display

stacktrace-error:
Useful for exhausts, thrusters, sparks, etc.
Should also fix ArmorPlateAbility's multiplier formatting as 60100% instead of 160%. 

https://github.com/Anuken/Mindustry/assets/89076920/7710f5ca-f620-4434-9411-c34ad2c68b21

_Debug area drawing + absurd values_

https://github.com/Anuken/Mindustry/assets/89076920/ca1dc498-46e0-40c0-aa0a-b34fd5db95ad

_Potential usecase_

### DIFF
--- a/core/src/mindustry/entities/abilities/ArmorPlateAbility.java
+++ b/core/src/mindustry/entities/abilities/ArmorPlateAbility.java
@@ -29,7 +29,7 @@ public class ArmorPlateAbility extends Ability{
 
     @Override
     public void addStats(Table t){
-        t.add("[lightgray]" + Stat.healthMultiplier.localized() + ": [white]" + Math.round(healthMultiplier * 100f) + 100 + "%");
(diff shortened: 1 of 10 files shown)


Anuken:
Doing logic in draw() is really cursed, there must be a better way...

stacktrace-error:
didn't see it necessary to add an update to drawparts, guess I'll just do that now

stacktrace-error:
probably not the greatest solution, but it still works normally and was easy to do

Anuken:
On second thought, while spawning effects in draw() is awful, I think the overhead of having to update a bunch of parts that have nothing to do with effects is going to be very high. There's a large amount of conflicts now, so I'll just take the effect spawner part and merge that.

PR #9507, opened 2024-02-01
Allow mods to add to logic effect map

MEEPofFaith:
I guess they could also replace existing effect entries.


Anuken:
I don't think the API should use the Object... format, just add a method to add one instead. It's fine if they have to call it a bunch of times, most methods are like that, and at least it's harder to mess it up.

PR #9516, opened 2024-02-04
Fix generation of drills' booster stats

BalaM314:
If you have a beam drill (or regular drill) with multiple boosters or multiple required liquids, like this:
image
Before:
before
the boosters are determined incorrectly
After:
image
fixed

code was strange, I think it assumed there would only be one or two liquids


BalaM314:
Drill, BeamDrill, ForceProjector(maybe), MendProjector, OverdriveProjector, RegenProjector, and RepairTurret need modification to correctly handle modded blocks where the block requires a liquid/item and accepts liquids/items as a booster

Maybe combine all the StatValues.(boosters, speedBoosters, itemBoosters) into one function that takes all the boost consumers as an argument, that way we could support blocks that require an item and a liquid for boosting

BalaM314:
this is totally wrong

SomeonesShade:
wha

PR #9550, opened 2024-02-14
Extend logic `read` and `write`,support targeting `LogicBuild`

way-zer:


BasedUser:
what

way-zer:
> what

Allow processor to transfer variable to another directly, including Unit, Building...

The4codeblocks:
Problem: This allows for processors to act as memory storage, possibly greater than 64, rendering all memory storage useless

way-zer:
> Problem: This allows for processors to act as memory storage, possibly greater than 64, rendering all memory storage useless

Yeah, the processor is more powerful than memory storage, as it can store not only number but object.

But processor don't support string concat, so the `read` is not indexable. and it's more complex.

The4codeblocks:
> > Problem: This allows for processors to act as memory storage, possibly greater than 64, rendering all memory storage useless
> 
> Yeah, the processor is more powerful than memory storage, as it can store not only number but object.
> 
> But processor don't support string concat, so the `read` is not indexable. and it's more complex.
> 
> The feature of this PR is much powerful, that allow many complex logic design. (like function store processor, mulitiprocessing)

`@counter` arrays:

way-zer:
> > > Problem: This allows for processors to act as memory storage, possibly greater than 64, rendering all memory storage useless
> > 
> > 
> > Yeah, the processor is more powerful than memory storage, as it can store not only number but object.
> > But processor don't support string concat, so the `read` is not indexable. and it's more complex.
> > The feature of this PR is much powerful, that allow many complex logic design. (like function store processor, mulitiprocessing)
> 
> `@counter` arrays:

I know that, but it can't stop use without this pr as it can use inside the same one processor. And this pr is also a enhancement for many complex logic like `@counter` arrays.

The4codeblocks:
Random idea: make the textbuffer in processors usable

way-zer:
> Random idea: make the textbuffer in processors usable

I has thinked that, it sounds good. A special variable points to the `textbuffer` StringBuilder and can be used in `read` and `write`.

The4codeblocks:
> Reading from the text buffer implies making an allocation to copy the string to a variable, which is not acceptable.

welp

though that should also mean string variables would be unacceptable?

Anuken:
Strings in logic are immutable. You can reassign the variable to a difference reference of a string, but you can't modify what it points to.

The4codeblocks:
> Strings in logic are immutable. You can reassign the variable to a difference reference of a string, but you can't modify what it points to.

yeah, strings in logic are like `java.lang.String`, not `java.lang.StringBuilder`.

Yarplay1:
>possibly greater than 64, rendering all memory storage useless. Somehow, but hasganter already made a processor handle 321 variable inside, and be perfectly usable, acting like memory bank with a bit less variables but way more compact

way-zer:
Besides, I want to pr another one like this:

Allow MemoryBlock save LVar like LogicBlock.

Which one do you perfer?

core/src/mindustry/logic/LExecutor.java
```diff
@@ -580,6 +601,16 @@ public void run(LExecutor exec){
 
             if(from instanceof MemoryBuild mem && (exec.privileged || (from.team == exec.team && !mem.block.privileged)) && address >= 0 && address < mem.memory.length){
                 mem.memory[address] = value.num();
+            }else if(from instanceof LogicBuild logic && (exec.privileged || from.team == exec.team) && position.isobj && position.objval instanceof String name){
+                //limit write, need target processor set writable.
+                LVar writable = logic.executor.optionalVar("@writable");
```
Anuken, line 606: This is very strange. It's completely un-intuitive, and I don't understand why this variable has to be set.

BasedUser: memory protection - you wouldn't want to break world processor maps by accident; and sometimes you want to use world processors for API or something

really the statement below should also check for privilege

core/src/mindustry/logic/LExecutor.java
```diff
@@ -557,6 +571,13 @@ public void run(LExecutor exec){
 
             if(from instanceof MemoryBuild mem && (exec.privileged || (from.team == exec.team && !mem.block.privileged))){
                 output.setnum(address < 0 || address >= mem.memory.length ? 0 : mem.memory[address]);
+            }else if(from instanceof LogicBuild logic && (exec.privileged || from.team == exec.team) && position.isobj && position.objval instanceof String name){
+                LVar fromVar = logic.executor.optionalVar(name);
+                if(fromVar != null && !output.constant){
+                    output.objval = fromVar.objval;
```
BasedUser, line 577: this doesn't check if the executor you're reading is itself privileged, which means that same-team world processors may be readable, potentially leaking secret variables to an unprivileged context

Anuken: > this doesn't check if the executor you're reading is itself privileged

I was aware when I merged this, which is why it was fixed as the next commit.

core/src/mindustry/logic/LExecutor.java
Anuken: See other comment - this was already patched to check for privileged access.

PR #9583, opened 2024-02-22

jexjws:
In the description of core unit operations, the original translation used the phrase `（默认使用[拾取，]放下载荷）` which, although it implied the information about shortcut keys, might not be intuitive or easy to understand for players.

Therefore, I have modified it to a more explicit expression: `（默认使用快捷键“[”拾取载荷，“]”放下载荷）` Such a change is aimed at making the game's operational instructions clearer, especially for beginners.


BlackDeluxeCat:
True .The text between "[" and "]" is often thought of as the name of shortcut.

PR #9593, opened 2024-02-25
Add diagonal symmetry compat

The4codeblocks:
Adds functionality for blocks with a diagonal symmetry axis (defined by `Block.diagonalSymmetry`).
_`Block.rotation` = 0 means facing bottom-right_

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -58,3 +58,10 @@ jobs:
         cd ../Mindustry
     - name: Run unit tests
       run: ./gradlew tests:test --rerun --stacktrace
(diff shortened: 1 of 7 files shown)


Anuken:
I don't want to maintain PRs that change the API in ways that vanilla (or anyone else?) will never need to use.

This also changes method signatures, which can break mods.

The4codeblocks:
> This also changes method signatures, which can break mods.

Thanks for the critique, I fixed the issue in the branch, but because the PR is closed, it's not appearing in the commit list (as of now as I know)

PR #9599, opened 2024-02-28
Shader getShaderFi check for mod files

MEEPofFaith:
~~If the initial `Core.files.internal` doesn't find a file, try finding with `tree.get`.~~
Use `tree.get` instead of `file.internal` in `getShaderFi`. Makes custom shaders with js/java slightly less annoying as modders wouldn't need to make their own version of `LoadShader` just to load their shader files and reduces copy/paste for things like surface shaders, as a copy of the `SurfaceShader` that extends off the copied version of `LoadShader` wouldn't be needed anymore, and also allows mods with a frag with an identical name to override vanilla shaders.

### DIFF
--- a/core/src/mindustry/graphics/Shaders.java
+++ b/core/src/mindustry/graphics/Shaders.java
@@ -474,6 +474,8 @@ public class Shaders{
     }
 
     public static Fi getShaderFi(String file){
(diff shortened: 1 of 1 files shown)


Anuken:
Test this first. `Core.files.internal` can't be null.

Issue #9651, opened 2024-03-18
Poly repairs enemy phase walls


JasonP01:
### Platforms

Windows

### Build

146

### Issue

Polys can heal none same team phase walls, for whatever reason, i suspect mega might as well tho i havent tested if it does

### Steps to reproduce

Play map provided

Spawn polys
Place damaged enemy phase walls

### Mods used

vanilla

### Save file

[phaseWallPolyBug.zip](https://github.com/Anuken/Mindustry/files/14632541/phaseWallPolyBug.zip)

### Submission

- [X] I have updated to the latest release (https://github.com/Anuken/Mindustry/releases) to make sure my issue has not been fixed.
- [X] I have searched the closed and open issues to make sure that this problem has not already been reported.

mahdoosh1:
> Not a bug. The projectile is deflected, after which it heals the wall.

so phase walls change the projectile team? it would be cool if we had a turret that could change an enemy unit team to our team

PR #9662, opened 2024-03-21
Fix poly

JasonP01:
Fixed poly healing enemy phase walls #9651 
Poly bullet cant be deflected anyway, this only fixes the healing issue

### DIFF
--- a/core/src/mindustry/content/UnitTypes.java
+++ b/core/src/mindustry/content/UnitTypes.java
@@ -1318,6 +1318,7 @@ public class UnitTypes{
 
                     healPercent = 5.5f;
                     collidesTeam = true;
+                    reflectable = false;
                     backColor = Pal.heal;
                     trailColor = Pal.heal;
                 }};


BlueTheCube:
not a bug

JasonP01:
> not a bug

You cant tell me poly is MEANT to be able to heal enemy phase walls, so no, intended behaviour is still a bug whether you like it or not

MEEPofFaith:
> Poly bullet cant be deflected anyway, this only fixes the healing issue

I think what's happening is it *is* being deflected, which results in it immediately hitting the phase wall and hitting it. This is likely an issue with all bullets that can heal and should probably be solved in a different way than just `reflectable = false`.

broganeuquoixe:
Exactly, poly's bullets are just the most noticeable ones because the healing out heals the damage dealt.

https://github.com/Anuken/Mindustry/assets/161990992/9e29d161-a8b2-492a-a47e-cdfe051b17bd

JasonP01:
> Exactly, poly's bullets are just the most noticeable ones because the healing out heals the damage dealt.
> 
>  1711403903.mp4

Actually, poly is the only one as its bullet comp is the only one that can be reflected by default, quad bullet has reflectable as false, and mega is a laser bullet which is also false by default

PR #9665, opened 2024-03-22
Custom rules dialog enhancement, small fixes

ApsZoldat:
- Fixed wrong (old) info about format placeholders in all unfinished locales (it was saying to use `@` instead of `{0}` for logic `format` instruction) (that's why so many files are changed in this PR)
- Added rule search in custom map rules dialog
- Added info buttons for some rules (basically a must have thing for mods adding new rules, but even some vanilla rules should have a better explanation than just a title)

More complicated stuff: rules in separate "categories" (previously they were simply divided by accent color titles) now belong to different tables. This way, titles only appear when there's at least one rule in category during search.

Also, the basic methods for adding new rules are now public (hope it is fine), and `setup()` runs all the stuff in `Seq<Runnable> additionalSetup` before adding category tables to dialog. Combined with having separate category tables for rules, these changes make adding new rules through modding way easier.


broganeuquoixe:
The position of the info button is really awkward. Is there a reason why the info about rules couldn't just be a tool tip that appears when hovering over?

Anuken:
I'll move it into a tooltip.

ApsZoldat:
> The position of the info button is really awkward. Is there a reason why the info about rules couldn't just be a tool tip that appears when hovering over?

Won't it be inaccessible on mobile? Okay, i'll see what Anuke does and think about it later. Also originally i wanted all the info buttons be on the right side, imo this would be somewhat convenient for mobile users

PR #9692, opened 2024-03-31
Data imported exit message

SITUVNgcd:
This will make players less confusing after game's data imported and no message tell them. They will think that is a crash.


Anuken:
The reason no message is displayed is that I am concerned that cached data (say, sector information) can overwrite the existing data file while the dialog is open. This needs rigorous testing.

PR #9705, opened 2024-04-03
Para buffs

JasonP01:
Parallax is severely underpowered right now, it takes 4 to stop a flare and 17 to stop an eclipse. As well as it just generally being underused due to how bad it is. It has come to an agreement on the buffs.

Further changes to mechanics such as debuff status effects or multiple beams still need to be decided on to make it more viable, but this general stats buff is good.

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -3369,7 +3369,7 @@ public class Blocks{
             size = 2;
             force = 16f;
             scaledForce = 9f;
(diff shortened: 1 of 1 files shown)


BlueTheCube:
parallax should be used with other antiair turrets

JasonP01:
> parallax should be used with other antiair turrets

Agreed, but as a turret itself it still shouldnt be bad, it still need to be able to do something itself like every other turret

PR #9717, opened 2024-04-07
Add `Server Favourite`

Atomic-Laboratory:

- [?] I have ensured that any new features in this PR function correctly in-game, if applicable.

What it would look like if accepted.
image

## Things that may cause this PR to be not accepted:
1) If `doInit` is changed to false, so the server name bar doesn't appear if servers are offline, there is strange behavior when there are more than 1 sever in a server group.
image
Works fine if there is only 1 server, or if `doInit` is true.

2) There is no tool-tip describing what the star button does (for accessibility?)
  
  
  
_I ask anyone who is knowledgeable to help me with this PR, I'm trying my best :sob:_

### DIFF
--- a/core/src/mindustry/ui/dialogs/JoinDialog.java
+++ b/core/src/mindustry/ui/dialogs/JoinDialog.java
@@ -407,6 +407,8 @@ public class JoinDialog extends BaseDialog{
 
             if(group.prioritized){
                 addHeader(groupTable, group, hidden, false);
+            } else if (group.favorite()) {
+                addHeader(groupTable, group, hidden, true);
             }
             //table containing all groups
             for(String address : group.addresses){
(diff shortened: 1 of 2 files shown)


Atomic-Laboratory:
`doInit` _should_ be false to prevent things like this from happening
image

core/src/mindustry/ui/dialogs/JoinDialog.java
```diff
@@ -405,8 +405,11 @@ void refreshCommunity(){
 
             Table[] groupTable = {null, null};
 
+            boolean favorite = group.favorite();
             if(group.prioritized){
-                addHeader(groupTable, group, hidden, false);
+                addHeader(groupTable, group, hidden, favorite, false);
+            } else if (favorite) {
```
Anuken, line 411:
```suggestion
            }else if (favorite){
```

PR #9733, opened 2024-04-12
Allow markers to be used as light sources

Redstonneur1256:
Allows for markers to be drawn as light sources using `setmarker light <id> <true/false>`, point markers act as point lights, I'm not sure if this should be a dedicated marker or kept as is.

image


BlueTheCube (commented):
What about `Make Marker`?

Anuken:
I think this should be a different marker. Points having two "modes" is very strange.

BlueTheCube:
Block lighting when

Redstonneur1256:
> Block lighting when

Most blocks don't emit light on their own, liquids/power lasers do but would require rewriting large parts of it and risks breaking mods compatibility as it requires to change `Draw#laser`, there is also other light sources to be taken into account like bullets, fire, effects, etc...

core/assets/bundles/bundle.properties
```diff
@@ -1387,8 +1387,10 @@ rules.title.unit = Units
 rules.title.experimental = Experimental
 rules.title.environment = Environment
 rules.title.teams = Teams
+rules.title.light = Lighting
 rules.title.planet = Planet
 rules.lighting = Lighting
+rules.lighting.unitlight = Unit light
```
Anuken, line 1393:
```suggestion
rules.lighting.unitlight = Unit Lighting
```

Anuken:
image

This breaks existing saves and doesn't seem very straightforward to fix. I'm not going to be working on markers right now, so I'll be reverting it.

PR #9755, opened 2024-04-19
Disable leg damage on disarmed units

BalaM314:


TheRadioactiveBanana:
i don't think this makes sense

TheRadioactiveBanana:
disarmed should mean they don't have guns, toxopid etc should probably still be able to smash buildings with leg damage

JasonP01:
> disarmed should mean they don't have guns, toxopid etc should probably still be able to smash buildings with leg damage

Disarmed is used more by servers and players as a way to disable damage from that unit, this change makes sense as the niche if leg only damage likely was never used

Otamamori917:
Wouldn't this also count for tank trampling?

PR #9754, opened 2024-04-18
Implementing Mindustry-Suggestions#4959

S0meM:
https://github.com/Anuken/Mindustry-Suggestions/issues/4959
Note that the title beneath here is wrong. This is mindustry-suggestions#4959, not mindustry#4959.

If your pull request is **not** translation or serverlist-related, read the list of requirements below and check each box:

- [x] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [x] I have ensured that my code compiles, if applicable.
- [ ] I have ensured that any new features in this PR function correctly in-game, if applicable.

PR #9777, opened 2024-04-24
Add turret ammo capacity to the in-game help

hexagonrecursion:
- Ref: https://github.com/Anuken/Mindustry-Suggestions/issues/4959
- Ref: https://github.com/Anuken/Mindustry/pull/9754

image


Anuken:
"Max Shots" isn't a good way to describe ammo capacity - if I saw that without context, I would be confused. "Does that mean it can fire 3 shots at once sometimes? Is there are a minimum amount of shots?"

"Ammo Use" is also ambiguous - 4 ammo/shot sounds a whole lot like 4 *items* per shot, but those are very different things.

Leo-MathGuy:
Maybe "Ammo Capacity: x shots"?

hexagonrecursion:
> "Max Shots" isn't a good way to describe ammo capacity - if I saw that without context, I would be confused. "Does that mean it can fire 3 shots at once sometimes? Is there are a minimum amount of shots?"

Thanks. I implemented @Leo-MathGuy 's suggestion and updated the screenshot above.

image


> "Ammo Use" is also ambiguous - 4 ammo/shot sounds a whole lot like 4 items per shot, but those are very different things.

I only added ~"Max Shots"~ "Ammo Capacity". The "Ammo Use" stat was already there, but I can change it if you have a suggestion how it can be improved.

hexagonrecursion:
> "Ammo Use" is also ambiguous - 4 ammo/shot sounds a whole lot like 4 _items_ per shot, but those are very different things.

Possible improvement:
* 3x ammo multiplier -> One item makes: 3 ammo
* Ammo Use: 2/shot -> One shot costs: 2 ammo

1. For turets with ammo use = 1 it's not incorrect if the player reads "One item makes: 3 ammo" as "The turret can _shoot_ 3 times for the cost of one item"
2. For turets with ammo multiplier = 1 it's not incorrect if the player reads "One shot costs: 2 ammo" as "It costs 2 _items_ to shoot once"
3. For turrets with both ammo use != 1 and ammo multiplier != 1 (only one example in vanilla: Ripple) both quantities being expressed using _the same unit_ "ammo" will hopefully help the player understand that one stat should be divided by the other

Another possible improvement:
* Remove the Ammo Use stat and the ammo multiplier stat from the user interface
* Add a new stat where ammo multiplier used to be: "One shot costs: 2/3 items" (in this example ammo multiplier is 3 and ammo use is 2). For Diffuse: "One shot costs: 3 items", for Disperse "One shot costs: 1/3 items". This will require a bit of extra code in StatValues.ammo() to display this as a fraction when ammo multiplier != 1

What do you think?

Anuken:
> The "Ammo Use" stat was already there, but I can change it if you have a suggestion how it can be improved.

It's slightly less ambiguous when a "ammo capacity" stat is displayed with it, since it is different from "item capacity". That part might not need any changes after all.

Ammo Multiplier could be changed to "{0} ammo per item" instead of "x ammo multipier."

S0meM:
While you're at it, you might as well add this one:
https://github.com/Anuken/Mindustry-Suggestions/issues/4960
Here's what I did:
`bulletsPerShot = new Stat("bulletsPerShot", StatCat.function),` In Stat. java,
`stat.bulletspershot = Bullets Per Shot` In bundle.properties, and
`if(shoot.shots!=1) stats.add(Stat.bulletsPerShot, shoot.shots);` In Itemturret.java

Here's the result:

Screenshot 2024-04-25 162748

hexagonrecursion:
@Anuken Would it be better to fix Diffuse stats in a follow-up pull request?

S0meM:
Allow I suppose shots per round makes more sense than bullets per shot.

Leo-MathGuy:
At that point, just do "Bullets per Round"

hexagonrecursion:
@S0meM My main issue was actually that I expected "firing rate" to mean "how many times per second the turret shoots" (2 for diffuse) rather than "how many projectiles the turret shoots per second" (30 because each time diffuse shoots it shoots a cone of 15 bullets) because this interpretation gives a number that is meaningfully connected to "ammo capacity", "ammo use" and "ammo/item" (aka "ammo multiplier") stats.

hexagonrecursion:
@S0meM @Leo-MathGuy What is a "round"?

S0meM:
A round is just how many bullets are shot at the same time. It's not an in game stat, but it just makes more sense than shot. Although I suppose that that isn't what the dictionary thinks. It says "the amount of ammunition needed to fire one shot", which is kinda backwards from how I'm using it. That would make it make more sense to say bullets per shot. What do you guys think the best term would be?

S0meM:
> @S0meM My main issue in [Anuken/Mindustry-Suggestions#4960](https://github.com/Anuken/Mindustry-Suggestions/issues/4960) was actually that I expected "firing rate" to mean "how many times per second the turret shoots" (2 for diffuse) rather than "how many projectiles the turret shoots per second" (30 because each time diffuse shoots it shoots a cone of 15 bullets) because this interpretation gives a number that is meaningfully connected to "ammo capacity", "ammo use" and "ammo/item" (aka "ammo multiplier") stats.

So the firing rate should say bullets per second instead? Would that make more sense if we had bullet capacity, bullets per second, bullets per shot, and bullets per item?

hexagonrecursion:
> A round is just how many bullets are shot at the same time. It's not an in game stat, but it just makes more sense than shot. Although I suppose that that isn't what the dictionary thinks. It says "the amount of ammunition needed to fire one shot", which is kinda backwards from how I'm using it. That would make it make more sense to say bullets per shot. What do you guys think the best term would be?

Using "bullets per shot" would be consistent with the usage of the word "shot" in the existing "ammo use" stat as well as the new "ammo capacity" stat. I think we should avoid using multiple words to refer to the same thing - this would make the user interface more confusing. Either use "shot" everywhere or "round" everywhere. To me "shot" is less confusing.

S0meM:
That's fair. Ignore that I ever used round. It's just that we seem to be split between a shot being each bullet or a shot being a firing of a number of bullets.

PR #9778, opened 2024-04-24
Enhance json mod content load order

Leo-MathGuy:

- [x] I have ensured that any new features in this PR function correctly in-game, if applicable. (Existing mods are not affected at all)

## Mod content load order
A big problem with every single hjson mod is that the order of content is always alphabetically sorted.
This forces mod makers to make json files have indexes in the content names, like `0001_`, `00_a-`, etc. to make them ordered correctly. This is very distracting, as bundles also need to account for such things, etc.

What this PR adds is a `config.(h)json` file in a content type folder for content type specific configuration. Currently, it can have a `contentOrder` array parameter to it, which allows modders to directly specify the order of content loading (and thus, the order of appearance in the build menu and core database) of the content type. The changes I made are completely backwards compatible, with mods not having configs having the same behaviour as before.

Both the config file and the `contentOrder` parameter are optional.

An example of a such parameter in `content/blocks/config.hjson`:
```
contentOrder: [
"sealedconveyor", 
"armored-sealed-conveyor", 
"sealedrouter", 
"sealedsorter", 
"sealed-overflow", 
"sealed-unloader"
]
```

This makes it so content that is listed in the `contentOrder` will be queued in `runs` during mod content loading in the order specified.
If something is *not* listed in the `contentOrder`, then it will be loaded after the ordered content, sorted in the same way as before (alphabetically). This makes it backwards-compatible, as well the config file and `contentOrder` param optional.

## I tested:
* Multiple normal hjson mods - No effect at all to existing mods
* .hjson mod with a `contentOrder` in `content/blocks/config.hjson` - The contents were in the correct order in the core database / build menu as specified. Content that was not in `contentOrder` was placed after the ordered content, sorted alphabetically like previous behaviour.
* That same .hjson mod with the `contentOrder` configuration on latest stable build (146), loads without errors (so mods that *do* use it can be played on stable, but without the ordering)


core/src/mindustry/mod/Mods.java
```diff
@@ -731,19 +731,39 @@ public int compareTo(LoadRun l){
...
+                                FolderConfig config = json.fromJson(FolderConfig.class, Jval.read(file.readString()).toString(Jformat.plain));
+                                if(config.contentOrder != null){
+                                    contentOrder.add(config.contentOrder);
+                                }
+                                continue;
+                            }
+                            currentRun.put(file.nameWithoutExtension(), new LoadRun(type, file, mod));
+                        }
+                        Seq<String> left = currentRun.keys().toSeq();
+                        if(contentOrder.size != 0){
```
Leo-MathGuy, line 750: This check could removed I think, not sure though

Leo-MathGuy:
An alternative solution is a global `config.hjson` file in the root directory, not content type folders (not in `mod.json` because the mod list generator and the mod list view would become a lot slower). That would make the list messier, but the files more orderly. Either way, it will still be backwards compatible with current mods.

Leo-MathGuy:
After some talking in #pulls, we came to a conclusion that a root `config.json` is better for metadata not related to the mod browser, with adding the content order and excluded content to the `config.json`as well. The values moved from `mod.json` that are being moved to `config.json` will be still supported, but deprecated. Values required by the modlist generator will stay.

PR #9868, opened 2024-05-20
Update servers_v7.json: factory server

prof-milki:
New server instance, Europe, static IPv6-only (works surprisingly well), single game mode (for now).
Intended as fallback server due to other networks having recurring downtimes, and are somewhat dilatory with new maps.

 [ ] Does not currently fulfill all requirements of SERVERLIST.md - specifically no #2 automated NSFW blocking (gonna test the content verification plugin, if the need arises / unlikely to be an issue initially).
 [x] Moderation/discourse is still in the works.

### DIFF
--- a/servers_v7.json
+++ b/servers_v7.json
@@ -1,5 +1,9 @@
 [
   {
+    "name": "Redundancy Dept",
+    "address": ["min7.include-once.org:8000", "min7.include-once.org:8001"]
+  },
+  {
     "name": "TSR Network",
     "address": ["de-prem-01.hosts.optikservers.com:35526", "de-prem-01.hosts.optikservers.com:35915", "de-prem-01.hosts.optikservers.com:35250", "de-prem-01.hosts.optikservers.com:27526", "de-prem-01.hosts.optikservers.com:35376", "de-prem-01.hosts.optikservers.com:35895"]
   },


Anuken:
Server is offline.

prof-milki:
It was running along AFAICT, but it requires **IPv6** connectivity. (Which is why I thought the DNS name would be more apt than an address literal.)

PING min7.include-once.org(2a0f:5707:aaf1:7059::1) 56 data bytes
64 bytes from 2a0f:5707:aaf1:7059::1: icmp_seq=1 ttl=58 time=31.4 ms
64 bytes from 2a0f:5707:aaf1:7059::1: icmp_seq=2 ttl=58 time=27.2 ms
64 bytes from 2a0f:5707:aaf1:7059::1: icmp_seq=3 ttl=58 time=30.4 ms

But I now realize this might be a bigger discussion, if this would really be the very first IPv6 server(?)

Clearly Mindustry client and server support is working perfectly fine. And I assumed users without connectivity would simply not have the AAAA nameserver lookup resolved, and thus the server list entry just not show up (without further ping or TCP/UDP connection attempts).

TBH, if it's never been tested before (only seen one ticket now on that subject), then a **temporary** field run would be in order. I'm just strongly assuming it would simply be an inactive entry for IPv4-only clients. (Kinda what happens when I disable my own IPv6 tunnelbroker for the custom server entry.)

PR #9903, opened 2024-05-31
add ability to switch to planet by clicking on it from afar

Sivecano:

This sometimes bothered me when looking at the map. 
Just let me click on a thing to get there instead of having to figure it out via menu. 
Now I am happy.

I outlined the function `gotoPlanet` but I'm unsure if this was the correct thing to do styleguidewise.  

SCOPE EXPANDED: we're also doing smooth transitions now ig


Twcash:
interesting

Sivecano:
Tested on Desktop and Android, maybe one would want to add a highlighter to the currently hovered planet but maybe it's actually just fine as-is

Sivecano:
I'm unsure about 3 things:
* Is it a good idea to outline gotoPlanet?
* is the introduction of hoverPlanet a good idea
* if yes, should it be highlighted somehow?

Sivecano:
do not merge this for now it breaks selecting sectors to play. will fix

Sivecano:
https://github.com/Anuken/Mindustry/assets/49620143/74112e66-9a0c-4b3d-990e-032a7ced4665

Sivecano:
https://github.com/Anuken/Mindustry/assets/49620143/5b1d3a56-542a-4aed-b5ca-99443aedc146

Sivecano:
so the big issue I'm having is that I'm faced with a few choices:
* go on as before and set the camera position relative to the camera to something fixed: this does not feel very natural
* keep the same relative camera position: this feels great especially when clicking on planets directly but when choosing planets from the list it can cause you to clip backwards through a planet this is jerky as hell.
* turn the camera to face the new planet: this works pretty well especially for long zooms but is awful when selecting something like a moon or frankly anything that is not far away
what can I do: 
*  make some sort of case distinction
* choose one of the above
* figure out a more complicated approach path preferably one with a higher arc this would be epic but hard

Sivecano:
this also fixes a segfault when selecting a planet without any sectors (this happens with some misconfigured mods)

Anuken:
I'm not going to review this yet, but your style is wrong. Read the [guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md#style-guidelines).

Sivecano:
> I'm not going to review this yet, but your style is wrong. Read the [guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md#style-guidelines).

thanks, I'll look over it. I though I kept most of it in mind but apparently I messed up. my bad

Sivecano:
ran the formatter over my changes and I hope it's serviceable now. The only style thing I'm sorta unsure about is the addition of `gotoPlanet`. I wanted changing planet to have consistent behaviour but maybe that's not actually what we want to begin with.

Anuken:
I decided to implement this without any panning.

> this also fixes a segfault when selecting a planet without any sectors (this happens with some misconfigured mods)

Note that the code shown here appears to do the opposite of fixing it (it checks for empty sectors as opposed to non-empty ones)

Sivecano:
> Note that the code shown here appears to do the opposite of fixing it (it checks for empty sectors as opposed to non-empty ones)

Ah, I'm sorry 😅, but thanks for implementing this :).

PR #9920, opened 2024-06-03
Update servers_v7.json

Leo-MathGuy:
Move DontVin servers to other node as the current one is being wiped. @Ayushkiller can confirm


Leo-MathGuy:
Anuke where are you it has been a week 😭

Leo-MathGuy:
Nvm

PR #9927, opened 2024-06-07
Update servers_v7.json

Natrition:
I changed the name and IP address of the server.


rd-zf-zfy:
After pulling the code, I found that it was missing the code in the mindustry/gen directory. How did you develop and test the submission?

Natrition:
> After pulling the code, I found that it was missing the code in the mindustry/gen directory. How did you develop and test the submission?

I don't know what this is about. This text has translated

PR #9935, opened 2024-06-11
Logic `data` instruction

Redstonneur1256:
Adds logic client to server networking `data <channel key> <value>` for mods/plugins, server to client networking is already available trough `sync`, data can be received by creating a handler using `LExecutor#addDataHandler(String, Cons2<Player, Object>)`


ApsZoldat:
This c2s PR dropped so unexpectedly my jaw dropped too

Anyway, is intended to have this feature hidden due to `allowLogicData` not being available through map editor (CustomRulesDialog)? Or did i get something wrong?

core/src/mindustry/logic/LStatements.java
```diff
@@ -1911,6 +1911,42 @@ public LCategory category(){
         }
     }
 
+    @RegisterStatement("clientdata")
+    public static class ClientDataStatement extends LStatement{
+        public String channel = "\"foo\"", value = "\"bar\"", reliable = "0";
```
Anuken, line 1916:
```suggestion
        public String channel = "\"frog\"", value = "\"bar\"", reliable = "0";
```

PR #9979, opened 2024-06-23
Quad markers translation

Redstonneur1256:
* Adds default x, y, u and v positions to quad markers
* Allows for quad markers to be translated using `setmarker pos`
* Allow for quad marker vertices to be rotated around the center position using `setmarker rotation`
* Adds `@Transform` for 2D matrices in the objectives dialog
* Replacement for #9738

### DIFF
--- a/core/src/mindustry/game/MapObjectives.java
+++ b/core/src/mindustry/game/MapObjectives.java
@@ -1134,13 +1134,18 @@ public class MapObjectives implements Iterable<MapObjective>, Eachable<MapObject
     public static class QuadMarker extends ObjectiveMarker{
         public String textureName = "white";
         public @Vertices float[] vertices = new float[24];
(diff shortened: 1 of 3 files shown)


core/src/mindustry/game/MapObjectives.java
```diff
@@ -1150,6 +1155,7 @@ public void draw(float scaleFactor){
             if(fetchedRegion == null) setTexture(textureName);
 
             Draw.z(drawLayer);
+            Draw.trans(transform);
```
Anuken, line 1158: Causing a batch flush every single quad draw call is not acceptable for performance. The vertices need to be manually transformed.

JasonP01:
@Redstonneur1256 fixeth ppl want

Anuken:
#9738 should be used instead.

PR #10062, opened 2024-07-26
Added elapsed time counter

ih295:
With this update, a simple boolean-enabled elapsed time feature is added. All this does is show how much time elapses between every frame while updating it. This counter outputs that every second via logging to console. This update can and should be ignored, as it is only being done for a pull request. In that pull request, we are to detail said elapsed time as well as frame rate measured by running 3  relatively large-scale games.

I was tasked to benchmark Mindustry on games with thousands of blocks. I found three relatively large games with hundreds of nodes, potentially thousands. Each of these games were joined twice, and played for 10 minutes for accurate data. For the first map, on the first join, I got a frame rate of 20, and an elapsed time of 31830 milliseconds averaged over a minute. For the second join, it was 30 frames, with an elapsed time per frame of 32071 milliseconds averaged over a minute.

For the second map, on first join, the frame rate was 26, and its elapsed time was 26409 milliseconds. With respect to its second join, the frame rate was 19, and the elapsed time was 32806 milliseconds.

For the third and final map, after first joining it, I calculated a frame rate of 13 per second, and an elapsed time of 32269 milliseconds between frame update, averaged over a minute. After joining it a second time, the frame rate was 18, and the elapsed time was 26956 milliseconds per minute.

Per second, all of these maps and sessions had a minimum of 500 milliseconds between frames every second, and a maximum of 700 milliseconds between frames.

Again, as stated: this update can be ignored as it was only made for this pull request. I do plan on adding a separate interface that shows this information, which will be accessed via button or key press.

### DIFF
--- a/core/src/mindustry/core/Logic.java
+++ b/core/src/mindustry/core/Logic.java
@@ -30,9 +30,11 @@ import static mindustry.Vars.*;
  * This class should <i>not</i> call any outside methods to change state of modules, but instead fire events.
  */
 public class Logic implements ApplicationListener{
-
+    private long time = Time.nanos();
+    private long totalElapsedTime = 0;
+    private int frameCount = 0;
+    private boolean debug = false;
     public Logic(){
-
         Events.on(BlockDestroyEvent.class, event -> {
             //skip if rule is off
             if(!state.rules.ghostBlocks) return;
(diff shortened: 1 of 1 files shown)


JasonP01:
Yapanese, school projects for pring stuff is goofy

Anuken:
(See Discord conversation)

ih295:
So, updating this: all this code does is benchmark the performance of the update function in the Logic class. It measures how long it takes for the game to complete its update function per ms, averaged over a second.

JasonP01:
@ih295 the pr has been rejected you can stop

PR #10083, opened 2024-08-04
Allow PlaySound To Work Normally With High IPT World Processor

Mythril382:
~~polyphony~~


Anuken:
Not sure if this is a good idea. The main use case for playing multiple sounds in a frame is music, but that's often not what logic is doing. If someone is using logic sound for a frequent event (like an explosion or shot being fired), this will destroy the audio system in a way that no other sound effect does.

Mythril382:
Hmm... Maybe I could add a new field?

core/src/mindustry/logic/LStatements.java
```diff
@@ -2169,6 +2169,10 @@ void rebuild(Table table){
             }else{
                 fieldst(table, "pan", pan, str -> pan = str);
             }
+            
+            table.row();
+            
+            fieldst(table, "check frame", checkFrame, str -> checkFrame = str);
```
Anuken, line 2175: Would be nice to have this documented, with a better name. Maybe call it `limit` as in "limit to one sound per frame"? 

It could have a bundle description like this:

`playsound.limit = Limit this sound to only one playback per frame.\nOnly set this to false if absolutely necessary, e.g. for playing musical chords.`

Mythril382:
Now that I think about it, every label (that isn't already self explanatory) should have a documentation.

~~Also, it should be hinted in game that you can hold for tooltips.~~

core/assets/bundles/bundle.properties
```diff
@@ -2563,6 +2563,8 @@ unitlocate.outx = Output X coordinate.
 unitlocate.outy = Output Y coordinate.
 unitlocate.group = Building group to look for.
 
+playsound.limit = If true, disables this sound from playing if the\nsame sound has been played already in the same frame.
```
Anuken, line 2566:
```suggestion
playsound.limit = If true, prevents this sound from playing\nif it has already been played in the same frame.
```

PR #10095, opened 2024-08-06
Fix the sfx selection gui for playsound

JeanJPNM:

Fixes:
- The selected sound effect not being marked as "selected" when opening the selection menu. This happened because `.substring` was being called with 4 instead of 5 (the length of the `@sfx-` prefix).
- A crash that happened when the sfx selection button was clicked and the variable used as the effect had a name shorter than 4 characters.

### DIFF
--- a/core/src/mindustry/logic/LStatements.java
+++ b/core/src/mindustry/logic/LStatements.java
@@ -2149,7 +2149,8 @@ public class LStatements{
             table.button(b -> {
                 b.image(Icon.pencilSmall);
                 
-                b.clicked(() -> showSelect(b, GlobalVars.soundNames.toArray(String.class), id.substring(5), t -> {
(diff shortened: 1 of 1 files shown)


core/src/mindustry/logic/LStatements.java
```diff
@@ -2149,7 +2149,8 @@ void rebuild(Table table){
             table.button(b -> {
                 b.image(Icon.pencilSmall);
                 
-                b.clicked(() -> showSelect(b, GlobalVars.soundNames.toArray(String.class), id.substring(4), t -> {
+                String soundName = id.startsWith("@sfx-") ? id.substring(5) : id;
```
JeanJPNM, line 2152: Not sure if `id` should be replaced by an empty string when it's not a built-in sound effect.

PR #10110, opened 2024-08-12
Crawler balancing

Twcash:

## Changes
No longer targets under blocks (conveyors junctions ect)
HP lowered from 200 to 150 (may still be too strong)
Blast radius lowered to 5.5 blocks
blast damage reduced to 90 (50 average damage)
Screenshot 2024-08-11 215335

### DIFF
--- a/core/src/mindustry/content/UnitTypes.java
+++ b/core/src/mindustry/content/UnitTypes.java
@@ -610,13 +610,14 @@ public class UnitTypes{
 
             speed = 1f;
             hitSize = 8f;
-            health = 200;
+            health = 150;
             mechSideSway = 0.25f;
             range = 40f;
             ammoType = new ItemAmmoType(Items.coal);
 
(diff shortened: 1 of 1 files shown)


Mythril382:
Hmm... What if it didn't target units that despawned without a player?

SomeonesShade:
agreeable nerfs for reasons:

Hold Fire + Exploding on Death allows it to use its hp to get deeper into defenses and detonating at close range (maximizing its damage)
Together with the fact it deals consistent damage as it only needs to get close with boom value. While dagger drops off rapidly as defenses lowers dagger’s value even further making crawler the economical choice

So you have a really strong option that is easy to produce that can wipe out bases, nopenopenopenope

vivaan-web (commented):
Nahhhhhh, it's good as it is. Dw about it for now
Prolly just mad cause someone crawler duped your base out of existence

PR #10182, opened 2024-09-06
make shield field in database more clear

Ilya246:
currently ShieldRegenFieldAbility in core database is very misleading so this fixes that

image


MEEPofFaith:
maybe should be shield/pulse instead?

also how is 250 with a 1/sec fire rate 25 shield per pulse?

MEEPofFaith:
oh wait i see

maybe it should be "max shield" instead

Ilya246:
i made it consistent with the other strings

PR #10221, opened 2024-09-21
Update servers_v7.json

Niatheu:
Added a new Glyph project. Currently only the survival server is running, but I'm also adding an attack server which is currently offline as there are no maps available


Niatheu:
What the fuck? Why “Some checks were not successful”?

EchoDustry:
> What the fuck? Why “Some checks were not successful”?

it does that sometimes, i dont think it means anything

PR #10306, opened 2024-10-31
Percentage modifier stat for status effects

MEEPofFaith:
Before:
20241030-201229-Mindustry

After:
20241030-202150-java
(Did the default UI scale change at some point? 75% scale in the custom build is much smaller than 75% scale in 146.)

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -1151,6 +1151,7 @@ unit.minutes = mins
 unit.persecond = /sec
 unit.perminute = /min
 unit.timesspeed = x speed
+unit.multiplier = x
 unit.percent = %
 unit.shieldhealth = shield health
 unit.items = items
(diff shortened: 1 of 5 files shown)


MEEPofFaith:
??????

way-zer:
> ??????

The display mode has been changed to increase and decrease, rather than multiplier. When calculating multiple buffs, it may cause misunderstanding to those who do not understand. 

For example, with `+200%` and `+100%` will get `600%` instead `+300%`

MEEPofFaith:
So it should be a multiplier instead?
`x3` and `x2` will more accurately convey `x6`

way-zer:
> So it should be a multiplier instead? `x3` and `x2` will more accurately convey `x6`

Yeah, maybe `x[accet]300[]%` and `x[red]70[]%`

MEEPofFaith:
> > So it should be a multiplier instead? `x3` and `x2` will more accurately convey `x6`
> 
> Yeah, maybe `x[accet]300[]%` and `x[red]70[]%`

image

PR #10350, opened 2024-11-23
Add JVM argument for data directory

ahdg6:

This PR introduces a JVM argument (-Dmindustry.data.dir) to specify the data directory, reducing the dependency on the environment variable (MINDUSTRY_DATA_DIR). If the JVM argument is not provided, the fallback is still to use the environment variable. This simplifies configuration by allowing direct use of JVM arguments.

```sh
java -jar desktop.jar -Dmindustry.data.dir=./custom_dir
```

### DIFF
--- a/core/src/mindustry/ClientLauncher.java
+++ b/core/src/mindustry/ClientLauncher.java
@@ -35,7 +35,7 @@ public abstract class ClientLauncher extends ApplicationCore implements Platform
 
     @Override
     public void setup(){
-        String dataDir = OS.env("MINDUSTRY_DATA_DIR");
+        String dataDir = System.getProperty("mindustry.data.dir", OS.env("MINDUSTRY_DATA_DIR"));
         if(dataDir != null){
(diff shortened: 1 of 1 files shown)


SMOLKEYS:
mfw head repo deleted 💀 

also please lord i actually need/want this

ahdg6:
> mfw head repo deleted 💀 


The previous repository was deleted because I had forked another client, which itself was a fork of Mindustry. As a result, GitHub overwrote and deleted the original repository where the PR was created. However, I’ve now recreated the repository and resubmitted the PR.

> also please lord i actually need/want this

Regarding the PR content, using JVM arguments to set the data directory is a better approach than relying on environment variables. This is because when you make the game portable (e.g., running it from a USB drive), you often need to verify if the data directory environment variable exists. Moreover, the method for setting environment variables varies across shells and operating systems, while JVM arguments are straightforward, clear, and directly visible.

PR #10359, opened 2024-12-02
Formatting Icons

MEEPofFaith:
20241202-023526-java

This method replaces `:duo:` with the duo icon (and similarly with other content), without the need to check for character id. I originally wanted to override `Core.bundle` so that I could override `get` and `format` to automatically do this to every bundle entry; that way it'd be easier to create bundles and potentially have modded content icon support. However I could not find an easy way to do that, so for now it's just this method.

Currently applied to:
- Chat
- Message blocks
- Hints
- Objectives


MEEPofFaith:
hm what would be the best way to check for the end

MEEPofFaith:
image

MEEPofFaith:
Hm but I don't want to have to do two colons every time I want a colon if a string does have an icon in it.

MEEPofFaith:
image
image

MEEPofFaith:
Where else would it be sensible to throw `UI::formatIcons` calls?

MEEPofFaith:
oh jeez conflicts

MEEPofFaith:
maybe I shouldn't have applied this to every other bundle before merge

core/src/mindustry/core/UI.java
```diff
@@ -35,6 +35,8 @@
 public class UI implements ApplicationListener, Loadable{
     public static String billions, millions, thousands;
 
+    public static StringBuilder buffer = new StringBuilder();
```
Anuken, line 38: Should be private.

core/src/mindustry/ui/fragments/ChatFragment.java
```diff
@@ -161,14 +162,15 @@ public void draw(){
 
         float theight = offsety + spacing + getMarginBottom() + scene.marginBottom;
         for(int i = scrollPos; i < messages.size && i < messagesShown + scrollPos && (i < fadetime || shown); i++){
+            String message = UI.formatIcons(messages.get(i));
```
Anuken, line 165: Don't change the way the chat displays. On desktop, you can already type :alpha: and have it show up as you type. Inject the formatting code into the chat message sending code instead, so it is formatted right before you send it.

Anuken (changes requested):

PR #10374, opened 2024-12-15
why are noisemesh seeds planet id dependent?

coaldeficit:
really awesome installing another mod that happens to add planets only for your planet to look different and have its sector placement and landmarks mismatch

the value added to the actual seed is now a constant 7 (except for 2 colored noisemeshes in which case the color seed has 8 added instead of 7, as in the base game the seed used for the color is planet.id + seed + 1) instead of the planet's id since thats the planet id of most modded planets. this way mods that add only 1 planet dont have to make any changes to keep their planet's appearance

additionally, the makeAsteroid function now has a seed value so that gier, notva and verlius (verilus?) dont look different after this pr since they use noisemeshes

### DIFF
--- a/core/src/mindustry/content/Planets.java
+++ b/core/src/mindustry/content/Planets.java
@@ -92,7 +92,7 @@ public class Planets{
         }};
 
         //TODO names
-        gier = makeAsteroid("gier", erekir, Blocks.ferricStoneWall, Blocks.carbonWall, 0.4f, 7, 1f, gen -> {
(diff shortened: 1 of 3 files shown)


overdramaticpanromantic:
i would love this to be merged so bad

i want to make my planets have landmarks but a lot of people play with countless planet mods and that makes it difficult or outright impossible

overdramaticpanromantic:
WOOOOOOO

PR #10415, opened 2025-01-12
Add server again

Kieaer:
There were about 400 people on average per year from overseas, and all of them were griefers. They didn't do griefing on their own servers, but almost 100% of them said they would go back to their own servers after griefing because it was Korea.

Since all overseas access is blocked except for Japan, which is relatively reliable, if you want to connect to the server from a region other than that, you will need to use a VPN or provide me with the IP you are trying to connect from.

### DIFF
--- a/servers_v7.json
+++ b/servers_v7.json
@@ -310,5 +310,10 @@
       "server.mindustry-tool.com",
       "server.mindustry-tool.com:6568"
     ]
-  }
+  },
+  {
+    "name": "Korea",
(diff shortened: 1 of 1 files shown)


Kieaer:
The code will cause the server to stop every time it does a DB operation, and players will experience a brief server pause every time a new user joins the game.
This's mitigated by ExecutorService .

In order to fundamentally analyze the cause, I need logs.

PR #10422, opened 2025-01-14
Allow updateEffect spread to be customized or disabled for GenericCrafter

overdramaticpanromantic:
Adds an updateEffectSpread and updateEffectSpreadMag for customization of where the updateEffect appears on a GenericCrafter. Previously, you couldn't ensure it appears at a given location, as it would always be placed all around the sprite. By default, this does the spread and also has the same spread magnitude as before, so it's backwards-compatible.

I've read the guidelines, compiled and tested the game, and ensured that it works properly.

### DIFF
--- a/core/src/mindustry/world/blocks/production/GenericCrafter.java
+++ b/core/src/mindustry/world/blocks/production/GenericCrafter.java
@@ -40,8 +40,7 @@ public class GenericCrafter extends Block{
     public Effect craftEffect = Fx.none;
     public Effect updateEffect = Fx.none;
     public float updateEffectChance = 0.04f;
-    public boolean updateEffectSpread = true;
-    public float updateEffectSpreadMag = 4f;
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/blocks/production/GenericCrafter.java
```diff
@@ -40,6 +40,8 @@ public class GenericCrafter extends Block{
     public Effect craftEffect = Fx.none;
     public Effect updateEffect = Fx.none;
     public float updateEffectChance = 0.04f;
+    public boolean updateEffectSpread = true;
```
Anuken, line 43: This doesn't need to exist, one could make the `updateEffectSpreadMag` 0 instead.

overdramaticpanromantic: good point, i didn't think of that

PR #10454, opened 2025-02-01
fix of diode's strange behavior

Maxiron794:

While experimenting with the power diode, I found inconsistencies in its behavior.

Its purpose is to transfer power when the back power graph has a higher percentage of stored energy than the front power graph, but it doesn't actually intend to balance them that way.

For example, a fully charged large battery will transfer 50000 * (1 - 0) / 2 = 25000 energy (clamped to 4000) to a small battery. However, intuitively it should be around 3700.

This code fixes the issue. I also believe that there is no longer any need for the clamp or for dividing by 2, but it remains unchanged.

Before:
image
image
image

After:
image
image
image

### DIFF
--- a/core/src/mindustry/world/blocks/power/PowerDiode.java
+++ b/core/src/mindustry/world/blocks/power/PowerDiode.java
@@ -62,20 +62,23 @@ public class PowerDiode extends Block{
             PowerGraph frontGraph = front().power.graph;
             if(backGraph == frontGraph) return;
 
-            // 0f - 1f of battery capacity in use
-            float backStored = backGraph.getBatteryStored() / backGraph.getTotalBatteryCapacity();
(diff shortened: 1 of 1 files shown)


LieutenantH:
I think that the battery diode works on _total_ stored power, not percent.
This means the behavior is actually correct.

"Moves battery power in one direction, but only if the other side has less power stored."

Maxiron794:
> I think that the battery diode works on _total_ stored power, not percent. This means the behavior is actually correct.
> 
> "Moves battery power in one direction, but only if the other side has less power stored."

Okay, that's possible, but if batteries are switched:

image
image
image

Shouldn't it be 2k and 2k?

LieutenantH:
Did the small battery get placed first?

And yeah, it definitely is weird. I think it might not activate instantly to stop it (or something).

You could be right about it being broken.
Idk.

PR #10456, opened 2025-02-03
Translating to Bangla / Bengali

RakibRyan:


RakibRyan:
for some reason translated texts don't show up but shows blank

PR #10505, opened 2025-02-16
Minor editor/world processor fixes and tweaks

ApsZoldat:
1. Fix search input in rules dialog applying `trim()` to newly typed spaces (it made starting a new word impossible without moving cursor to the middle)
2. Make some editor-related search inputs lowercase text and remove unwanted spaces in it without affecting original user input
3. Fix info buttons in rules dialog going off the screen on mobile by enabling `scrollX` if there's not enough space
4. Make `spawnunit` able to spawn hidden units just like `setprop @payloadType` does (regarding the `!type.internal` removal: internal units still can't be spawned because they're not in global processor vars)

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -1431,7 +1431,7 @@ public class LExecutor{
 
             Team t = team.team();
 
-            if(type.obj() instanceof UnitType type && !type.internal && !type.hidden && t != null && Units.canCreate(t, type)){
+            if(type.obj() instanceof UnitType type && t != null && Units.canCreate(t, type)){
                 //random offset to prevent stacking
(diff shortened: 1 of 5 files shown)


core/src/mindustry/logic/LExecutor.java
```diff
@@ -1431,7 +1431,7 @@ public void run(LExecutor exec){
 
             Team t = team.team();
 
-            if(type.obj() instanceof UnitType type && !type.internal && !type.hidden && t != null && Units.canCreate(t, type)){
+            if(type.obj() instanceof UnitType type && t != null && Units.canCreate(t, type)){
```
Anuken, line 1434: It is very important that internal units **should be spawned**, as they could cause a crash.

ApsZoldat: I know about this, as i stated in PR description, internal units still **cannot be spawned** because such global variables just don't exist.

But now i think this change still isn't worth it, because it won't go well in case someone decides to add some global vars, so i'll revert it.

PR #10507, opened 2025-02-18
reconstructor balancing

Frolonov:
Block build cost reduced & switched raw ores to refined materials.

Unit production speed increased to prevent game being too lengthy especially in end-game attack maps.

Reduced cryo consumption to accommodate space & water limitation on map, while greatly increased power consumption to encourage the use of impact reactor & overdrive.

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -5927,14 +5927,14 @@ public class Blocks{
         }};
 
         exponentialReconstructor = new Reconstructor("exponential-reconstructor"){{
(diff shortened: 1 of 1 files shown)


jehosula:
cheapening t4 and t5 is nice but idk if you went too far with 120 (?) secs for t50

Anuken:
I will absolutely not accept changes to item types for reconstructors. **This will be break every schematic, base and map that uses them.**

JasonP01:
@Anuken the item changes were the *build* cost not the unit requirement costs. This change would not effect existing maps. Only building new ones would use the new resources in construction

~~i have not checked if the build cost is used in the unit req cost but i assume not~~

PR #10520, opened 2025-02-23
Not privileged setrate [POSTREFORMAT]

The4codeblocks:

*reposted as permitted, modified to be compatible with the refactor*


Anuken:
There's one issue I forgot about: IPT is not saved for standard logic processors. If it *were* saved, it might cause issues if the *default* IPT changes, and users load an old save.

The4codeblocks:
> There's one issue I forgot about: IPT is not saved for standard logic processors. If it _were_ saved, it might cause issues if the _default_ IPT changes, and users load an old save.

fixed saving with check for saving with max ipt:
if you have a non-max ipt when saving, it persists when loading, and is lowered to any new max if the new max ends up smaller than the saved ipt
if you have a max ipt when saving, it matches any updated base ipt value when loading

and don't worry, I did a version check

1ue999:
revision 4 is mine :3

anyways, the saving seems overcomplicated, and looks like it might run into integer overflow issues due to saving as short, but using an int.

1ue999:
i would just do
`write.i(ipt)` and `ipt=read.i()` (not clamping because im assuming it gets clamped in the instruction already)

The4codeblocks:
> revision 4 is mine :3
either of us will increment if the other gets merged 1st :)

> anyways, the saving seems overcomplicated, and looks like it might run into integer overflow issues due to saving as short, but using an int.
I simply mirrored the world processor ipt saving, I'd think that integer overflow would happen to both

The4codeblocks:
> not clamping because im assuming it gets clamped in the instruction already
ensures that if a processor were to get nerfed, no processor remains higher than normal

1ue999:
huh

1ue999:
welp, i would just make it save as an integer with clamping then ig

The4codeblocks:
hm?

The4codeblocks:
hello?

The4codeblocks:
?

The4codeblocks:
this is old
2nd oldest pr of mine by now

core/src/mindustry/world/blocks/logic/LogicBlock.java
```diff
@@ -803,6 +799,8 @@ public void write(Writes write){
 
             if(privileged){
                 write.s(Mathf.clamp(ipt, 1, maxInstructionsPerTick));
+            } else {
```
Anuken:
```suggestion
            }else{
```

core/src/mindustry/world/blocks/logic/LogicBlock.java
```diff
@@ -866,6 +864,13 @@ public void read(Reads read, byte revision){
                 ipt = Mathf.clamp(read.s(), 1, maxInstructionsPerTick);
             }
 
+            if(!privileged && revision >= 5){
+                short iptR = read.s();
+                if (iptR != 0) {
```
Anuken:
```suggestion
                if(iptR != 0){
```

1ue999:
WOOO MERGED

PR #10522, opened 2025-02-24
Added the ability for mods to use the Difficulty class

hydrophobis:

It works but it is slightly janky. The modder must call
 ```
public void mindustry.content.Difficulties.addDifficulties(Difficulty...difficulties)
```
after their own <Difficulties class>.load() to add each of their difficulties
I would have integrated it in the ModClassLoader but i wasnt sure where to put it or if it even would work in that file so I didn't do this

### DIFF
--- a/core/src/mindustry/content/Difficulties.java
+++ b/core/src/mindustry/content/Difficulties.java
@@ -1,0 +1,34 @@
+package mindustry.content;
+
+import java.util.List;
+import java.util.ArrayList;
+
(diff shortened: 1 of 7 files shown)


Anuken:
Read the [contributing guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md#use-arc-collections-and-classes-when-possible). Using Java collections is not allowed.

PR #10532, opened 2025-03-04
Fixed LogicBuild executor loop

1ue999:
This PR makes LogicBuilds stop using only half of the accumulator, thus making it more predicable.

### DIFF
--- a/core/src/mindustry/world/blocks/logic/LogicBlock.java
+++ b/core/src/mindustry/world/blocks/logic/LogicBlock.java
@@ -540,7 +540,7 @@ public class LogicBlock extends Block{
                 if(accumulator > maxInstructionScale * ipt) accumulator = maxInstructionScale * ipt;
 
                 //Previously, it was only using half of the accumulator.
-                while(accumulator > 0){
+                while(accumulator > 1){
                     executor.runOnce();
                     accumulator --;
                     if(executor.yield){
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/blocks/logic/LogicBlock.java
```diff
@@ -539,7 +539,8 @@ public void updateTile(){
 
                 if(accumulator > maxInstructionScale * ipt) accumulator = maxInstructionScale * ipt;
 
-                for(int i = 0; i < (int)accumulator; i++){
+                //Previously, it was only using half of the accumulator.
```
BasedUser, line 542:
```suggestion
                for(int i = 0; i < (int)accumulator; i++){
                    executor.runOnce();
                    if(executor.yield){
                        executor.yield = false;
                        break;
                    }
                }
                accumulator -= (float)i;
```

BasedUser (changes requested):

core/src/mindustry/world/blocks/logic/LogicBlock.java
1ue999: this will not work.

BasedUser: it will, `i` is in scope - however `i++` may need to be moved after runOnce

Anuken:
This looks wrong. I don't think the `while` loop should run unless the accumulator is at least `1`.

1ue999:
oh yea i noticed it now. ~~(omg always 1 ipt minimum feature?)~~

PR #10536, opened 2025-03-07
Kick players who send too long messages

squi2rel:
Error stacktraces are too annoying

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -268,7 +268,9 @@ public class NetClient implements ApplicationListener{
         if(message == null) return;
 
         if(message.length() > maxTextLength){
-            throw new ValidateException(player, "Player has sent a message above the text limit.");
+            player.con.kick("You have sent a message above the text limit.", 30000L);
+            Log.warn("@ &fi&lk[&lb@&fi&lk]&fb has sent a message above the text limit.", new Object[]{player.plainName(), player.uuid()});
+            return;
         }
 
         message = message.replace("\n", "");


JasonP01:
LMFAO i vote no its too funny to the server owner sending a string of 99999 "h"

squi2rel:
> 30s is not excessive but it is long, still think catching the error for no stacktrace would be better - as a player

No. It's illegal to send such a long message, it's a waste of bandwidth and storage.

JasonP01:
Its not *that* deep fam

BnDLett:
> > 30s is not excessive but it is long, still think catching the error for no stacktrace would be better - as a player
> 
> No. It's illegal to send such a long message, it's a waste of bandwidth and storage.

ong. I can't spare ~100 KB. 😔

Anuken:
How would one send a message like this?

squi2rel:
> How would one send a message like this in a vanilla client?

f8 Call.sendChatMessage

BnDLett:
> > How would one send a message like this in a vanilla client?
> 
> f8 Call.sendChatMessage

I tried reproducing this to no avail (although, in all fairness, it wasn't on BE, so that may be an issue). It'd always fail. Could you provide exact reproduction steps?

JasonP01:
> > > How would one send a message like this in a vanilla client?
> > 
> > 
> > f8 Call.sendChatMessage
> 
> I tried reproducing this to no avail (although, in all fairness, it wasn't on BE, so that may be an issue). It'd always fail. Could you provide exact reproduction steps?

`Call.sendChatMessage(*put 400 "h" here*)`
Could automate the h with a function, or yk just ctrl c + v

JasonP01:
Players will now and then run said command in their console and be annoying to server owners, thats the point of the pr. However the problem could also be solved by the owner via a plugin instead so :/

BnDLett:
> > > > How would one send a message like this in a vanilla client?
> > > 
> > > 
> > > f8 Call.sendChatMessage
> > 
> > 
> > I tried reproducing this to no avail (although, in all fairness, it wasn't on BE, so that may be an issue). It'd always fail. Could you provide exact reproduction steps?
> 
> `Call.sendChatMessage("*put 400 "h" here*")` Could automate the h with a function, or yk just ctrl c + v

I'm trying to do with with a BE client and still nothing. I'm pretty certain that you'd need a non-vanilla client. I'll give it a shot in a non-vanilla client.

BnDLett:
This is what I've set the code to
```java
    public static void sendChatMessage(Player player, String message){

        //do not receive chat messages from clients that are too young or not registered
        if(net.server() && player != null && player.con != null && (Time.timeSinceMillis(player.con.connectTime) < 500 || !player.con.hasConnected || !player.isAdded())) return;

        //detect and kick for foul play
        if(player != null && player.con != null && !player.con.chatRate.allow(2000, Config.chatSpamLimit.num())){
            player.con.kick(KickReason.kick);
            netServer.admins.blacklistDos(player.con.address);
            return;
        }

        if(message == null) return;

        //if(message.length() > maxTextLength){
        //    throw new ValidateException(player, "Player has sent a message above the text limit.");
        //}

        message = message.replace("\n", "");

        Events.fire(new PlayerChatEvent(player, message));

        //log commands before they are handled
        if(message.startsWith(netServer.clientCommands.getPrefix()) && logCommands.bool()){
            //log with brackets
            Log.info("<&fi@: @&fr>", "&lk" + player.plainName(), "&lw" + message);
        }

        //check if it's a command
        CommandResponse response = netServer.clientCommands.handleMessage(message, player);
        if(response.type == ResponseType.noCommand){ //no command to handle
            message = netServer.admins.filterMessage(player, message);
            //suppress chat message if it's filtered out
            if(message == null){
                return;
            }

            //special case; graphical server needs to see its message
            if(!headless){
                sendMessage(netServer.chatFormatter.format(player, message), message, player);
            }

            //server console logging
            Log.info("&fi@: @", "&lc" + player.plainName(), "&lw" + message);

            //invoke event for all clients but also locally
            //this is required so other clients get the correct name even if they don't know who's sending it yet
            Call.sendMessage(netServer.chatFormatter.format(player, message), message, player);
        }else{

            //a command was sent, now get the output
            if(response.type != ResponseType.valid){
                String text = netServer.invalidHandler.handle(player, response);
                if(text != null){
                    player.sendMessage(text);
                }
            }
        }
    }
```
Specifically commenting out the message length check. When I built it and tested it with the client, the server still didn't seem to receive the message. I am unable to reproduce this with the steps given.

BnDLett:
Me and @JasonP01 tried reproducing this issue together, but neither of us managed to do so. The most that we got in terms of a related stacktrace was a debug message saying that a player had sent a message over the character limit.

JasonP01:
True, i theorize your potential stacktrace may have been caused by a plugin intercepting the chat packets

squi2rel:
It's about plugins. There's a bridge from our custom listener to the original listener, so the error occurs. Maybe kick the player directly could be better?

JasonP01:
Kicking the player in vanilla cause YOUR server software is causing problems is not something that should be done, this is entirely and literally a *you* problem

squi2rel:
badly

PR #10537, opened 2025-03-07
Add a configuration to not log commands

BnDLett:
Especially given that some servers or plugins may have commands that they don't want to have logged. Such as [this plugin](https://github.com/mindustry-ddns-net/MindustryDatabase).

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -32,6 +32,7 @@ import java.util.*;
 import java.util.zip.*;
 
 import static mindustry.Vars.*;
+import static mindustry.net.Administration.Config.logCommands;
 
 public class NetClient implements ApplicationListener{
     private static final long entitySnapshotTimeout = 1000 * 20;
@@ -276,7 +277,7 @@ public class NetClient implements ApplicationListener{
         Events.fire(new PlayerChatEvent(player, message));
 
         //log commands before they are handled
-        if(message.startsWith(netServer.clientCommands.getPrefix())){
+        if(message.startsWith(netServer.clientCommands.getPrefix()) && logCommands.bool()){
             //log with brackets
             Log.info("<&fi@: @&fr>", "&lk" + player.plainName(), "&lw" + message);
         }
--- a/core/src/mindustry/net/Administration.java
+++ b/core/src/mindustry/net/Administration.java
@@ -515,7 +515,8 @@ public class Administration{
         snapshotInterval = new Config("snapshotInterval", "Client entity snapshot interval in ms.", 200),
         autoPause = new Config("autoPause", "Whether the game should pause when nobody is online.", false),
         roundExtraTime = new Config("roundExtraTime", "Time before loading a new map after the gameover, in seconds.", 12),
-        maxLogLength = new Config("maxLogLength", "The Maximum log file size, in bytes.", 1024 * 1024 * 5);
+        maxLogLength = new Config("maxLogLength", "The Maximum log file size, in bytes.", 1024 * 1024 * 5),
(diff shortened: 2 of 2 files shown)


JasonP01:
In CN we used menus with commands as those arent logged to console, very helpful with keeping account login information outside the console logs

JasonP01:
> Is there a way to get text from a menu via a server-side plugin? I didn't see a way when I was looking through `Call`.

Thats how we do it yes, check out xpdustry/imperium and our AccountCommand.kt file in the mindustry directory, i think thats where login is (you can search for it to)

JasonP01:
We use distributor framework for our menus but its possible in vanilla with some engineering

1ue999:
> > In CN we used menus with commands as those arent logged to console, very helpful with keeping account login information outside the console logs
> 
> Is there a way to get text from a menu via a server-side plugin? I didn't see a way when I was looking through `Call`.

you can use Call.textInput (or smth)

BnDLett:
That's certainly useful. Danke.

way-zer:
If it's command, you can intrecept ChatPacket before commands handle.

BnDLett:
> If it's command, you can intrecept ChatPacket before commands handle.
> 
> This configuration seems useless in most cases, and may be harder to debug.

I did something like that with this
https://github.com/mindustry-ddns-net/MindustryDatabase/blob/5b6b10b7aead36bb5b75920ea09227276f760aa9/src/net/ddns/mindustry/database/javaPlugin/commands/CommandHandler.java

It's a huge band-aid and frankensteined solution, but it worked. Granted, I'm now working towards using `Call.textInput` and converting the plugin to Kotlin.

Either way, having the configuration there is nice for convenience. Also, how would it make it harder to debug?

way-zer:
> I did something like that with this https://github.com/mindustry-ddns-net/MindustryDatabase/blob/5b6b10b7aead36bb5b75920ea09227276f760aa9/src/net/ddns/mindustry/database/javaPlugin/commands/CommandHandler.java

```kotlin
//intercept chat packet
listenPacket2Server<SendChatMessageCallPacket> { connection, t ->
    if (t.message.contains("<Picture>") && t.message.contains("http")
        || t.message.contains("<District>")
    ) {
        connection.op("forbidden message")
        return@listenPacket2Server false
    }
    return@listenPacket2Server true
}
``` 


> Either way, having the configuration there is nice for convenience. Also, how would it make it harder to debug?

In most cases, the log is useful. If disable the logging and something wrong, it will be hard to anaylze and reproduce.

BnDLett:
> In most cases, the log is useful. If disable the logging and something wrong, it will be hard to anaylze and reproduce.

The client still stores previous commands. If something did go wrong, then it wouldn't be hard to share the parameters used. Given that it is also a config, logging can be reenabled (if it ends up disabled). Logging commands is also enabled by default, so you'll need to specifically set the configuration to `false` (or have a plugin do it) for it to not log the commands.

Either way, I personally see no harm in adding it, and, at the end of it day, it's just a configuration.

way-zer:
A global configuration for one sercet input is weird. (even one player and short-term).

core/src/mindustry/core/NetClient.java
```diff
@@ -276,7 +277,7 @@ public static void sendChatMessage(Player player, String message){
         Events.fire(new PlayerChatEvent(player, message));
 
         //log commands before they are handled
-        if(message.startsWith(netServer.clientCommands.getPrefix())){
+        if(message.startsWith(netServer.clientCommands.getPrefix()) && logCommands.bool()){
```
Anuken, line 280: There's no reason to statically import this.

Anuken: Yes, it should be a non-static import.

PR #10555, opened 2025-03-15
Rework quad oct

Frolonov:
Quad becomes more of an support unit due to less damage but stronger status effect (electrified & freezing) & fire-fighting capability(a unique reason to have this unit).

Qct gets stronger shield & repair turrets, status field applies overclock. HP reduced to compensate shield buff. Fast shield regen & unit repair help friendlies survive endgame resistance including overdriven turret & unit spam, therefore friendlies could actually overcome enemy defense & end the game without relying on long range siege units.

image

image

quad1

oct1

Alternative download link of JAR file on Discord: https://discord.com/channels/391020510269669376/813425645518061678/1350512374821556245


BugTD:
As a bomber unit, the Quad's core value lies in its **drop damage** and **bombing damage**, not fire suppression or other irrelevant mechanics. The weapon's electrified/freezing effects are rendered meaningless since the Quad (unlike the Oct) does not operate on the frontlines, and its slow movement speed further negates the utility of such effects.

Frolonov:
The whole tree focus on support, which is the idea behind this rework design. Still the Quad here is not too bad at attacking but it's not designed to be better attacker compared to Antumbra (design choice / opinion only)

Similar change was made to Aegires that it no longer focus on fire power but point defense / support.

Hirarchy:
960 heal per second, and 540 shield regen per second?
...really? are your plans for serpulo t5s making them so strong they can overcome essentially any numbered sector or campaign sector with sheer stats?

Frolonov:
> 960 heal per second, and 540 shield regen per second? ...really? are your plans for serpulo t5s making them so strong they can overcome essentially any numbered sector or campaign sector with sheer stats?

That Oct has no offensive power so a strong enough support ability is needed to justify the cost. Even with that, overdriven enemy defense could do far more DPS than the heal/regen as multiple turrets firing at a single bulky target.

PR #10560, opened 2025-03-16
Rework Antumbra & Eclipse

Frolonov:
Changes make these units actually useful to beat new endgame maps.

Longer range helps compensate large hit-size of these units when used in groups, but if player choose to out-range enemy defense then units at the back could not contribute to damage output.

Both unit gets better moving speed & armor but reduced health, also projectiles not collide tiles(as they come from above) to fit air unit characteristic & encourage defender to use point defense, shield or tractor beam.

Both unit got wet immunity to better survive surge turret synergy. Also both unit got incendiary weapon to be actual threat to areas lack of fire protection.

Antumbra uses missile, as a direct upgrade from Zenith.
Eclipse gets strong cannons on both sides & slow-firing energy weapon in the middle to better fit the battle cruiser vibe.

Alternative download link of JAR file on Discord: https://discord.com/channels/391020510269669376/813425645518061678/1350853681855336573

image

image

antum1

eclipse

### DIFF
--- a/core/src/mindustry/content/UnitTypes.java
+++ b/core/src/mindustry/content/UnitTypes.java
@@ -1121,8 +1121,10 @@ public class UnitTypes{
                 splashDamageRadius = 36f;
                 splashDamage = 60f;
                 ammoMultiplier = 4f;
+                homingPower = 0.02f;
                 lifetime = 35f;
-                hitEffect = Fx.blastExplosion;
+                scaleLife = true;
+                hitEffect = Fx.flakExplosion;
                 despawnEffect = Fx.blastExplosion;
                 collidesTiles = false;
                 status = StatusEffects.blasted;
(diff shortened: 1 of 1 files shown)


MEEPofFaith:
Balancing changes should be run by and discussed in the [#balancing](https://discord.com/channels/391020510269669376/813425645518061678) channel in the official Discord before submission.

Frolonov:
> Balancing changes should be run by and discussed in the [#balancing](https://discord.com/channels/391020510269669376/813425645518061678) channel in the official Discord before submission.

This is one design proposal awaiting for playtest, not excluding any competing PR if they're actually tested & posted. Currently I don't think the channel is organized & productive enough so I post my design & hope for playtest feedback as more player could actually see it. I believe it's better to have multiple design to choose from, rather than waiting for feedback forever.

(As you may probably notice these PRs I posted early hasn't received any playtest comment. There's no competing PR & the topic is not getting tracked by the #balancing).

PR #10586, opened 2025-04-01
Fix Integer.MIN_VALUE not getting parsed correctly and add [-+]0[xb]

camelStyleUser:
Fix Integer.MIN_VALUE getting parsed as null and add +0x,-0x,+0b,-0b

### DIFF
--- a/core/src/mindustry/logic/LAssembler.java
+++ b/core/src/mindustry/logic/LAssembler.java
@@ -73,10 +73,10 @@ public class LAssembler{
         //remove spaces for non-strings
         symbol = symbol.replace(' ', '_');
 
-        int usedInvalidNum=symbol.startsWith("-")?invalidNumPositive:invalidNumNegative;//use a positive invalid number if number is negative, else use a negative number
-        double value = parseDouble(symbol, usedInvalidNum);
+        //handle edge cases of hard to detect sign(in hex or bin numbers)
+        double value = parseDouble(symbol, invalidNumNegative);
 
-        if(value == usedInvalidNum){
+        if(value == invalidNumNegative && parseDouble(symbol, invalidNumPositive) == invalidNumPositive){
             return putVar(symbol);
         }else{
             //this creates a hidden const variable with the specified value
--- a/core/src/mindustry/logic/LAssembler.java
+++ b/core/src/mindustry/logic/LAssembler.java
@@ -73,7 +73,8 @@ public class LAssembler{
         //remove spaces for non-strings
         symbol = symbol.replace(' ', '_');
 
-        int usedInvalidNum=symbol.startsWith("-")?invalidNumPositive:invalidNumNegative;//use a positive invalid number if number is negative, else use a negative number
+        //use a positive invalid number if number might be negative, else use a negative invalid number
+        int usedInvalidNum = symbol.startsWith("-") ? invalidNumPositive : invalidNumNegative;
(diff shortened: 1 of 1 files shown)


GglLfr:
The distinction between `+` and `-` is unnecessary and might be problematic IMO. Java's integers are already two's complement-signed, so e.g. `0b10000000` is already considered a negative number.

If users decide to use a binary or hexadecimal representation of an integer for some reason, chances are:
- They don't know what they're doing and are just copy-pasting scripts from somewhere else.
- They *do* know what they're doing, and know that the most significant bit is used as a sign.

camelStyleUser:
typing 32 binary digits(assuming a 32-bit value is stored) for a sign seems a little bit absurd to me.hex is less absurd but still quite a few digits are needed for a simple sign

camelStyleUser:
to make a -1 in binary you have to do `0b1111111111111111111111111111111111111111111111111111111111111111` which is 66(!) characters
you cant have less digits because you wont change the sign if you have less digits

cardillan:
> The distinction between `+` and `-` is unnecessary and might be problematic IMO. Java's integers are already two's complement-signed, so e.g. `0b10000000` is already considered a negative number.
> 
> If users decide to use a binary or hexadecimal representation of an integer for some reason, chances are:
> * They don't know what they're doing and are just copy-pasting scripts from somewhere else.
> * They _do_ know what they're doing, and know that the most significant bit is used as a sign.

This is true in general, but Mindustry Logic parses numbers quite specifically. Firstly all numbers are parsed as 64-bit integers, so `0b10000000` is positive. And even then, it won't parse a number larger than Long.MAX_VALUE, regardless of radix, so you never get a negative number without a negative sign.

Then there is the funny quirk that Mindustry Logic now doesn't accept `-0xFF`, but accepts and parses `0x-FF` as -255.

core/src/mindustry/logic/LAssembler.java
```diff
@@ -72,20 +73,25 @@ public LVar var(String symbol){
         //remove spaces for non-strings
         symbol = symbol.replace(' ', '_');
 
-        double value = parseDouble(symbol);
+        //handle edge cases of hard to detect sign(in hex or bin numbers)
+        double value = parseDouble(symbol, invalidNumNegative);
 
-        if(value == invalidNum){
+        if(value == invalidNumNegative && parseDouble(symbol, invalidNumPositive) == invalidNumPositive){
```
Anuken, line 79: Parsing numbers is (was?) a bottleneck for logic; I will not accept a solution that adds extra parsing.

BlueTheCube:
is this an april fools joke?

cardillan:
@BlueTheCube No, there's no joke. Run this in a logic processor and see for yourself:

```
print -2147483648
printflush message1
```

core/src/mindustry/logic/LAssembler.java
cardillan: The problem is only with binary and hexadecimal numbers, because they can produce a negative number without a leading minus sign - or am I mistaken? 

Anyway, is there a reason not to make the invalid num values `long` instead of `int` (i.e. Long.MAX_VALUE and Long.MIN_VALUE instead of Integer)? If there is a special, potentially unsupported value, I'd say it would be better for that value to be at the edges of the numerical range instead of somewhere in the middle.

PR #10698, opened 2025-04-20
Made unitcap variable for logic

JasonP01:
Made logic @unitcap global variable, this allows processors to access their team's unitcap amount.

Begon the awful current ucap counting methods

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -2507,6 +2507,8 @@ lglobal.@waveTime = Countdown timer for waves, in seconds
 lglobal.@mapw = Map width in tiles
 lglobal.@maph = Map height in tiles
 
+lglobal.@unitcap = The maximum number of units that can be on the map at once, based on the processor's team
+
 lglobal.sectionMap = Map
 lglobal.sectionGeneral = General
 lglobal.sectionNetwork = Network/Clientside [World Processor Only]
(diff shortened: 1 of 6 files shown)


JasonP01:
The one in LAssembler might be useless. And the bundle might need better wording. Id rather leave the bundle to the merger

Anuken:
This is very inefficient. I will not accept a solution that injects a variable into the processor and uses getVar to dynamically update it every frame for every processor. Retaining an instance to the underlying Assembler is also unacceptable.

JasonP01:
> I will not accept a solution that injects a variable into the processor and uses getVar to dynamically update it every frame for every processor.

GlobalVar has no way of updating vars. And the only other thing running updates is LogicBlock. It has to run every frame as the unitcap can change every frame. Ill try resolve getVar and the assembler but i dont see a better solution to this

JasonP01:
@Anuken 
> every processor
It has to be every processor as ucap is team specific. having it outside of every processor is going to be even less efficient as you would need 256 ucap references for every team... unless using active teams 🤔

JasonP01:
Okay yes, how in the world am i supposed to get the processor's team to feed it its correct unitcap value if i dont know the team? The only way to get the team is inject the variable to be per processor

JasonP01:
Honestly i cant do what you are asking. this is not worth the hassle if i cant have it a per proc variable. What you want is currently impossible. if you will not accept this as is you might as well close it and we wont get this needed feature. im done

*we currently have no way of getting the maps actual unitcap*

JasonP01:
Mightve overreacted. But point remains the same. I can not solve the issue, i spent 6 hours trying and could not. 

If you (anuke) wont accept the current state or fix it yourself you can freely deny this pr.

I wont try and fix the issues you stated myself.

JasonP01:
Not planned

PR #10720, opened 2025-04-25
Allow Integer.MIN_VALUE to be parsed as an unsigned bin or hex literal

cardillan:
Currently the number parsing routine uses `Integer.MIN_VALUE`/`Integer.MAX_VALUE` to indicate an invalid value when parsing numbers. That makes it impossible to parse this number in hexadecimal or binary when entered as an unsigned value: `print 0xFFFFFFFF80000000; printflush message1` outputs `null`.

This PR changes the number values used to indicate invalid numbers to `Long.MIN_VALUE` / `Long.MAX_VALUE`.

`Long.MIN_VALUE` cannot be parsed in any form anyway due to the limitations in `Strings.parseLong`. (To fix that, the arc library would have to be patched.) This PR therefore removes the limitation of not being able to parse `Integer.MIN_VALUE` without introducing any new non-parseable values.

After the fix, the following code

```
print -99
print "\n"
print 99
print "\n"
print -0xFF
print "\n"
print 0xFF
print "\n"
print -0b1111
print "\n"
print 0b1111
print "\n"
print 0xFFFFFFFF80000000
print "\n"
print 0x8000000000000000    # Long.MIN_VALUE, cannot be parsed
print "\n"
print -9223372036854775808  # Long.MIN_VALUE, cannot be parsed
printflush message1
```
produces

```
-99
99
-255
255
-15
15
-2147483648       <-- this is what this PR fixes
null
null
```

### DIFF
--- a/core/src/mindustry/logic/LAssembler.java
+++ b/core/src/mindustry/logic/LAssembler.java
@@ -74,10 +74,9 @@ public class LAssembler{
         symbol = symbol.replace(' ', '_');
 
         //use a positive invalid number if number might be negative, else use a negative invalid number
-        long usedInvalidNum = symbol.length() > 0 && symbol.charAt(0) == '-' ? invalidNumPositive : invalidNumNegative;
-        double value = parseDouble(symbol, usedInvalidNum);
+        double value = parseDouble(symbol);
 
-        if(value == usedInvalidNum){
+        if(Double.isNaN(value)){
             return putVar(symbol);
         }else{
             //this creates a hidden const variable with the specified value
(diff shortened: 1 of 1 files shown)


cardillan:
Turns out I wasn't entirely correct: after my change, values between 0x8000000000000200 and 0x8000000000000000 cannot be parsed and are invalid - when expressed as a hex or binary literal.

The reason is that, when the value gets cast to `double`, it is rounded to `Long.MAX_VALUE`.

This can be fixed without touching arc. I'm on it.

cardillan:
I've refactored the code so that the value returned from `Strings.parseLong` gets tested for validity before being cast to `double`. 
Current status:

* Values from `0x8000000000000000` to `0x800000000000000F` (inclusive) can't be parsed. This is a limitation of Strings.parseLong.
* Similar limitation would apply to binary representations.
* Decimal value `-9223372036854775808` (i.e. `Long.MIN_VALUE`) can't be parsed
* Decimal value `-9223372036854775807` **can be parsed** and after casting to `double` and back to `long` equals `-9223372036854775808`, due to precision loss in the long to double conversion.

I don't truly understand why `-9223372036854775807` can be parsed while `0x8000000000000001` can't, but I didn't dig into it very much. It might be because the decimal literal is handled by `Strings.parseDouble` first, which strips the minus sign and then uses `Strings.parseLong`.

PR #10737, opened 2025-04-28
Advanced phase factory block on Serpulo

Frolonov:
A quicker & thorium efficient phase-fabric production block, encourages player to make use of meta-glass & blast-compound, which means getting item efficiency from extra complicated production process.

On maps with limited thorium & space, players could stockpile blast-compound & meta-glass, then use this block to produce phase-fabric quickly, therefore speeding up t4 & t5 production.

This advanced production block takes design concept similar to silicon crucible (boost production by extra fuel), while using same technology as the impact reactor indicated by plasma effect during production. 
(Sprites modified from force projector & impact reactor)

JSON file on Discord: https://discord.com/channels/391020510269669376/631946157496729620/1389671151285964939

image

image


meowkitty9694e:
look dude i know you really want your changes in mindustry, but no one wants this stuff in the game

hydrophobis:
Your sprite is pretty bad for something you want included in vanilla. It has absolutely no detail and the shape is very flat with not a lot of variation.

Frolonov:
> look dude i know you really want your changes in mindustry, but no one wants this stuff in the game

Campaign players are likely to support it for extra productivity. Perhaps PvP player might against it to maintain skill gap in order to stay in comfort zone.

Frolonov:
> Your sprite is pretty bad for something you want included in vanilla. It has absolutely no detail and the shape is very flat with not a lot of variation.

well visuals takes different kind of effort. This PR would focus on function first.

Hirarchy:
We don't need any of this in serpulo. "advanced" factories only exist because their base versions are either unable to keep up with late game demand(silicon smelter v. crucible) or are very bad at their job when they don't need to be(graphite press v. multipress).
In fact, at the current moment there are no blocks in serpulo that demand an advanced version, and the phase weaver is one of them. Phase fabric is an endgame resource that is intentionally difficult to make, and it should stay that way.

Hirarchy:
And no, I'm not a pvp player.

cardillan:
> well visuals takes different kind of effort. This PR would focus on function first.

Other considerations aside, this alone makes your PR quite improbable to get accepted imo. It would mean Anuke gets stuck with bad visuals, potentially creating a problem he'd have to solve somehow if, for whatever reason, you don't provide them in a timely manner.

Perhaps you could provide a mod with this block? If I'm not mistaken, it should work in campaign and you - and whoever else would want the block - could easily use it.

Frolonov:
> We don't need any of this in serpulo. "advanced" factories only exist because their base versions are either unable to keep up with late game demand(silicon smelter v. crucible) or are very bad at their job when they don't need to be(graphite press v. multipress). In fact, at the current moment there are no blocks in serpulo that demand an advanced version, and the phase weaver is one of them. Phase fabric is an endgame resource that is intentionally difficult to make, and it should stay that way.

Phase weaver only takes raw ore, not difficult to make but inefficient. That'd not add much difficulty to maps but only slows down progress, which this T2 phase is meant to address.

SunkyMP3G:
I don't think an upgraded phase weaver is even needed. Sprite isn't good either, but can always be fixed

Frolonov:
Sprite Updated

hydrophobis:
Much better sprite but your details are 1px in some areas which is heavily advocated against because it looks odd and you've also shaded with a light source at the top which is wrong and you are specifically told not to do this in the wiki. Just suggestions though

PR #10739, opened 2025-04-28
Rebalancing for Foreshadow,Lustreand Smite

BugTD:
https://github.com/Anuken/Mindustry-Suggestions/issues/5570

Foreshadow: Damage increased to 1360
Smite: Slightly reduce warm-up time
Lustre: Increase rotation speed, reduce power consumption to 100/s
Rationale:
High-HP units become harder to eliminate when surrounded
Smite's slow warm-up hinders effectiveness against fast units
Lustre's sluggish rotation limits it to countering slow-moving targets

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -4008,7 +4008,7 @@ public class Blocks{
                     pointEffect = Fx.instTrail;
                     despawnEffect = Fx.instBomb;
                     pointEffectSpace = 20f;
-                    damage = 1350;
+                    damage = 1360;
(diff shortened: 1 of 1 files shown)


jehosula:
why are those problems? you never answered, maybe those turrets to need their downsides? we don't just buff stuff out of their weaknesses and call it a day

Anuken:
I don't agree with these changes:
- Foreshadow damage increase is less than 1%, which is completely insignificant.
- Lustre doesn't need a buff, it's already very good.
- Does every turret need to be good against fast units? I don't think so.

PR #10743, opened 2025-04-30
QoL: provide an easy way to restart a logic processor

cardillan:
This PR adds a button to restart the processor to the Edit logic dialog:

{CA08FE87-0CC4-43CD-9E0F-57EE98B5F7E8}

I've used the refresh icon for the command. It looks quite fitting to me, but I can try to draw a new one if requested (e.g. a loop with an arrow pointing to the top).

The command resets all variables and restarts the processor (by enforcing recompilation). This operation is needed when the code in the processor performs some one-time initialization, and the user wants to re-run the initialization. Currently, the following workarounds are possible:

- Making a small modification to the program to force recompilation: not always practical, potentially dangerous.
- Implement the reinitalization logic in the logic code itself, e.g. on a switch: this is often not needed outside of code development, adding unnecessary cruft to logic. At the same time adding it ad-hoc to a program when it becomes needed may be error-prone, time-consuming and just not worth it.

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -525,6 +525,7 @@ details = Details...
 edit = Edit
 variables = Vars
 logic.clear.confirm = Are you sure you want to clear all code from this processor?
+logic.restart = Restart
 logic.globals = Built-in Variables
 
 editor.name = Name:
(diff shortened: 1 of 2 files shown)


EggleEgg:
I d check out how foos implemented it since it personally prefer that implementation instead of yours. The rest is simply ctrl+c/ctrl+v
https://github.com/mindustry-antigrief/mindustry-client
image

PR #10758, opened 2025-05-03
Recompiled and refactored all modified baseparts. Added a couple new baseparts.

JianhuiDong9999:
Partial fixes have been introduced to the new unit production baseparts that reduces the instances of units spawning trapped in walls. Added a new Nucleus core schematic and a few more general defense schematics.


JianhuiDong9999:
Note that the schematic names have been refactored for better naming identification. "lp" as prefix stands for "Landing Pad" and "eg" as prefix stands for "Endgame". The naming can still use some improvements but I think this is at least an improvement from previous basepart schematic namings.

PR #10774, opened 2025-05-07
Prevent setting counter to null using write instruction

cardillan:
While the `set` instruction doesn't allow to write `null` to `@counter`, this check is missing when setting the `@counter` using `write` instruction. This PR adds this check.

(Setting `@counter` to `null` doesn't stop the processor, but it causes the value of the `@counter` to be read as `null` on when accessed from logic, leading to problems with all `@counter`-based logic.)

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -88,6 +88,7 @@ public class LExecutor{
         }
 
         if(counter.numval < instructions.length){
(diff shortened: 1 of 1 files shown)


cardillan:
Note: there are various other ways to set `@counter` to null, e.g. `op div @counter 1 0`. Maybe it would be better to simply reset the `isobj` flag for counter at the start of every instruction?? These two checks could then be removed.

PR #10795, opened 2025-05-11
Logic sensor

Redstonneur1256:
Add some logic sensors
* `building`/`buildX`/`buildY`: Wether a unit or build turret is currently building and the build coordinates (untested in multiplayer)
* `selectedBlock`: The current block selected by a player


1ue999:
I would reccomend making buildX and buildY return null instead of -1 when not building (Double.NaN instead of -1)

core/src/mindustry/entities/comp/UnitComp.java
```diff
@@ -275,6 +275,9 @@ public double sense(LAccess sensor){
             case mining -> mining() ? 1 : 0;
             case mineX -> mining() ? mineTile.x : -1;
             case mineY -> mining() ? mineTile.y : -1;
+            case building -> isBuilding() ? 1 : 0;
+            case buildX -> isBuilding() ? buildPlan().x : -1;
+            case buildY -> isBuilding() ? buildPlan().y : -1;
```
1ue999, line 280: Perhaps buildType and buildRotation could be useful?
oh i get the thing with returning -1 now

1ue999 (approved):
looks good to me overall

core/src/mindustry/entities/comp/UnitComp.java
Redstonneur1256: I was working on some way to implement the type and whether it's constructing/deconstructing for `ConstructBuild` without adding too many single use sensors.
Maybe reusing `timescale` or having `building` return some constants similarly to `controlled`.
Rotation can already be accessed by fetching the building (binding the unit and using `ucontrol getBlock`) and using the existing `@rotation`.
And for the type, I thought of having `building` return the block type from `ConstructBuild` (see above) but this might confusing having it return a boolean on units/build turrets and a block type on blocks being built, toughts?

1ue999: hmm... 
progress: 0-1, how much its built
timescale: float, how much progress will change per tick or smth
would be nice to be able to get a reference to the ConstructBuild using sensor (make building return it instead of "bool"?)

Redstonneur1256:
Would it be better to have building and breaking be exclusive and return the currently placed/broken build or keep it as is with building returning any building from a builder unit and breaking allowing to test if it's being broken?

1ue999:
no idea tbh. i would choose the variation where it uses the same thing and then reverses it

1ue999:
also the way its right now kinda feels weird

core/src/mindustry/core/NetClient.java
```diff
@@ -695,7 +696,7 @@ void sync(){
             unit == null ? 0f : unit.vel.x, unit == null ? 0f : unit.vel.y,
             dead ? null : unit.mineTile,
             player.boosting, player.shooting, ui.chatfrag.shown(), control.input.isBuilding,
-            player.isBuilder() && unit != null ? unit.plans : null,
+            control.input.block == null ? Blocks.air : control.input.block, player.isBuilder() && unit != null ? unit.plans : null,
```
Anuken, line 699: Do not use `air` as a substitute for `null` here, mark it as `@Nullable` instead. It's more consistent with the original field, and more intuitive for logic. I would rather the player be building `null` - something that returns an error when sensed - rather than `air`.

Redstonneur1256: Serialization of blocks just writes the id as a short wether it's null or not meaning the code was failing on snapshots/syncs/saves which is why I used air as a replacement, I'll modify it to support null values when I can.

PR #10798, opened 2025-05-11
Add `select` operation

Redstonneur1256:
New instruction to quickly select between two variables without using jumps
Example usage:`select result notEqual x false a b`
image

Maybe Conditional Set (`setcond`) would be a better name, `setcond` does sound like changing a condition similarly to `setprop`/`setrule`


The4codeblocks:
you have problems with `setcond`?
how about `condset`?

kJNCwBUq:
Looks like ternary operator.

PR #10805, opened 2025-05-13
Logic operation round

1ue999:

Yes, this is just 28 added chars
Should be pretty useful for most logic users
This is using Math::round instead of Mathf::round because Mathf only accepts floats


JasonP01:
Round shouldve existed long ago. No more op add value 0.5 then op floor value

PR #10823, opened 2025-05-18
Adding Kalpe Games to servers_v7.json

BobTheLouis:
Added the Kalpe Games group and it's first server.


JasonP01:
New servers wont be added this way

BobTheLouis:
@JasonP01 Thanks for clearing this up, I see now that v7 is closed.

PR #10833, opened 2025-05-20
allow memory blocks to store any object

Deric-W:
Hello,
I really like the logic system and the new features coming in the next version such as drawing text and #9550.
I was thinking of creating a pull request for quite some time after commenting on https://github.com/Anuken/Mindustry-Suggestions/issues/2313 and with the aforementioned pull request mentioning it it thought that maybe now is the time.

Since I have never contributed to Mindustry before I may have missed some technical details (like how it interacts with multiplayer), but I took the existing logic block code as inspiration and tested the changes.

- [✓] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [✓] I have ensured that my code compiles, if applicable.
- [✓] I have ensured that any new features in this PR function correctly in-game, if applicable.

### DIFF
--- a/core/assets/contributors
+++ b/core/assets/contributors
@@ -181,3 +181,4 @@ IchMagSchokolade
 MonoChronos
 RushieWashie
 ITY
+Eric Wolf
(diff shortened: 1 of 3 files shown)


Anuken:
This increases memory usage of cells 3-4x, which isn't acceptable. Maybe storing it in a parallel Object[] array with NaN as a marker for "is an object" would work, since that would only be 1.5x memory usage, but that's not what this PR does.

PR #10861, opened 2025-05-30
Logic cutscene improvements

Redstonneur1256:
* Allow to read the current cutscene status with `cutscene active <result>`
* Allow cutscenes to disable the hud with `cutscene getHud <result>` & `cutscene setHud <shown>`
* Allow screen shake using `cutscene shake <amount> <duration seconds>`


JasonP01 (changes requested):
Make the hiding of hud a graphics setting to disable. It is imperative that users not be removed of their ability to use their buttons when they do not want it. Especially in accessibility concepts.

Mobile users will have no way of leaving the map otherwise, in cases of evil use.

Redstonneur1256:
> Mobile users will have no way of leaving the map otherwise, in cases of evil use.

Mobile users can use the android back button to open the pause menu, desktop users can do so by pressing escape

JasonP01:
> Mobile users can use the android back button to open the pause menu, desktop users can do so by pressing escape

Regardless it would be nice if you added it as a setting for users to opt-out in. I already know custom clients will make this not forced, so just put it in vanilla :)

JasonP01:
> why would you want to disable cutscenes?

Some people dont like having their camera snatched and them be unable to do anything

And besides that a larger majority would not like their gui taken away from them just because they chose to join some server.

Nevertheless if you wont add the accessibility setting, i will just have to pr it in after this is merged. If it even is

JasonP01:
Most famous custom client, foos, has already had the option to disable forced cutscenes since they were introduced to the game.

All i asked from you is to add the ability for the player to opt-out of their gui being taken away. You just seem to be trying to find every reason not to add the functionality.

Redstonneur1256:
> Some people dont like having their camera snatched and them be unable to do anything

The camera is only moved temporarily and usually for a good reason, eg. showing some lore, an objective, target, boss spawning, etc..



> All i asked from you is to add the ability for the player to opt-out of their gui being taken away. You just seem to be trying to find every reason not to add the functionality.

Like mentionned above, the reason the UI can be disabled is to make the cutscene more immersive when played, if used abusively by maps/servers then that is a problem with the said map/server that must be fixed or stop being played on.



> Most famous custom client, foos, has already had the option to disable forced cutscenes since they were introduced to the game.

> Nevertheless if you wont add the accessibility setting, i will just have to pr it in after this is merged. If it even is

Completely disabling cutscenes can ruin the experience of maps designed around them if not completely break them by interacting when you aren't supposed to. It also feels disrespectful to mapmakers that spent time creating the map and the cutscenes. 
A better option would be to add a skip option for cutscenes. `cutscene skipped <result>` could be used with a skip button visible at all times but it would be on the mappers to honor it. Having it be on the client level would go back to the risk of breaking maps.

While I agree on the fact users should be given some way to skip/interact cutscenes, I disapprove a option to completely disable them.

PR #10889, opened 2025-06-03
Fix Over-Refund on Deconstruction Completion

ankando:

### DIFF
--- a/core/src/mindustry/world/blocks/ConstructBlock.java
+++ b/core/src/mindustry/world/blocks/ConstructBlock.java
@@ -354,14 +354,14 @@ public class ConstructBlock extends Block{
                     }
                 }
             }
-
+            float refundProgress = Mathf.clamp(progress, 0f, 1f); 
             progress = Mathf.clamp(progress - amount);
 
             if(progress <= current.deconstructThreshold || state.rules.infiniteResources){
                 //add any leftover items that weren't obtained due to rounding errors
                 if(core != null && !state.rules.infiniteResources){
                     for(int i = 0; i < itemsLeft.length; i++){
-                        int target = Mathf.round(requirements[i].amount * state.rules.buildCostMultiplier * state.rules.deconstructRefundMultiplier);
+                        int target = Mathf.round(requirements[i].amount * state.rules.buildCostMultiplier * refundProgress * state.rules.deconstructRefundMultiplier);
                         int remaining = target - itemsLeft[i];
 
                         if(requirements[i].item.unlockedNowHost()){


ghost:
@Anuken 

I'm sorry for not considering the case of progress = 0, so I saved it before progress was subtracted to make sure it was not 0. However, I don't know why
```java
int target = Mathf.round(requirements[i].amount * state.rules.buildCostMultiplier * state.rules.deconstructRefundMultiplier);
```
When deconstructRefundMultiplier <= 1, there is no problem, because I found that if target is not linked to progress, then target should be the total refund amount (I mean when it's fully built), which will increase the remaining amount at the end.

ghost:
> @Anuken
> 
> I'm sorry for not considering the case of progress = 0, so I saved it before progress was subtracted to make sure it was not 0. However, I don't know why
> 
> ```java
> int target = Mathf.round(requirements[i].amount * state.rules.buildCostMultiplier * state.rules.deconstructRefundMultiplier);
> ```
> 
> When deconstructRefundMultiplier <= 1, there is no problem, because I found that if target is not linked to progress, then target should be the total refund amount (I mean when it's fully built), which will increase the remaining amount at the end.

I tested it and found that even if the process is not saved in advance, the refund amount will not be reduced when it is deconstructed. This may be because I did not test it enough.

core/src/mindustry/world/blocks/ConstructBlock.java
```diff
@@ -354,14 +354,14 @@ public void deconstruct(Unit builder, @Nullable CoreBuild core, float amount){
...
-
+            float refundProgress = Mathf.clamp(progress, 0f, 1f); 
             progress = Mathf.clamp(progress - amount);
 
             if(progress <= current.deconstructThreshold || state.rules.infiniteResources){
                 //add any leftover items that weren't obtained due to rounding errors
                 if(core != null && !state.rules.infiniteResources){
                     for(int i = 0; i < itemsLeft.length; i++){
-                        int target = Mathf.round(requirements[i].amount * state.rules.buildCostMultiplier * state.rules.deconstructRefundMultiplier);
+                        int target = Mathf.round(requirements[i].amount * state.rules.buildCostMultiplier * refundProgress * state.rules.deconstructRefundMultiplier);
```
Anuken, line 364: This still doesn't look correct. I don't see why the remaining items need to be multiplied by progress at all. Even if you take the value of `progress` before it's subtracted, it's likely to be an extremely small value, making `target` 0.

Making `target` 0 fixes the issue **for your use case**, but it breaks it for everything else.

ankando: @Anuken Because that when state.rules.deconstructRefundMultiplier is greater than 1, the program will compensate the value of the entire item at the end.

So I want to make the target smaller, so that when it is multiplied by state.rules.deconstructRefundMultiplier and is greater than 1, there won't be a lot of extra materials when the construction is completed.

ankando: This is the problem. Either we calculate it separately for state.rules.construct Refund Multiplier > 1, or I don't know how to fix it better.

ghost:
I will debug it some more until I get it right, there are some things I don't understand but I will get it working.

ghost:
@Anuken 

Before I changed it, your code was almost unusable when state.rules.deconstructRefundMultiplier < 1, and would cause a bug when > 1.

I found that when I tried to use (int)totalAccumulator[i]:

Whether it is target - (int)totalAccumulator[i] or target - itemsLeft[i], it is fine to completely demolish it after it is fully built. For example, if a building requires 100 copper, I spent 100 to build it, and itemsLeft = original price - spent is equal to 0. For example, when deconstructMultiplier = 0.5, I demolished the building and it was almost finished and gave me 49. At this time, itemsLeft[i] = 49, so target - itemsLeft[i] = 100 x 0.5 - 49 is equal to 1, which is completely correct. At this time, (int)totalAccumulator[i] is also equal to 49, and the result is the same. When you build it incompletely, for example, you only spend ten, then itemsLeft = 100 - 10 = 90. When you dismantle it, for example, you are given 4 when you are almost finished, and the total should be 100 x 0.5 x 0.1, that is, 5. However, itemsLeft = 90 + 4 is equal to 94, so remaining = target - itemsLeft = 50 - 94 = -89, because the Math.clamp limit becomes 0. So is (int)totalAccumulator correct? At this time, (int)totalAccumulator = 4, which will result in target = 50 - 4 = 46, which is more serious than the consequences of itemsLeft[i], and directly gives you dozens more.

ghost:
> @Anuken
> 
> Before I changed it, your code was almost unusable when state.rules.deconstructRefundMultiplier < 1, and would cause a bug when > 1.
> 
> Whether it is target - (int)totalAccumulator[i] or target - itemsLeft[i], it is fine to completely demolish it after it is fully built. For example, if a building requires 100 copper, I spent 100 to build it, and itemsLeft = original price - spent is equal to 0. For example, when deconstructMultiplier = 0.5, I demolished the building and it was almost finished and gave me 49. At this time, itemsLeft[i] = 49, so target - itemsLeft[i] = 100 x 0.5 - 49 is equal to 1, which is completely correct. At this time, (int)totalAccumulator[i] is also equal to 49, and the result is the same. When you build it incompletely, for example, you only spend ten, then itemsLeft = 100 - 10 = 90. When you dismantle it, for example, you are given 4 when you are almost finished, and the total should be 100 x 0.5 x 0.1, that is, 5. However, itemsLeft = 90 + 4 is equal to 94, so remaining = target - itemsLeft = 50 - 94 = -89, because the Math.clamp limit becomes 0. So is (int)totalAccumulator correct? At this time, (int)totalAccumulator = 4, which will result in target = 50 - 4 = 46, which is more serious than the consequences of itemsLeft[i], and directly gives you dozens more.

So the current situation is that you should consider the process when calculating the total amount of materials that should be given, but my previous approach was not reasonable.

core/src/mindustry/world/blocks/ConstructBlock.java
ghost: > This still doesn't look correct. I don't see why the remaining items need to be multiplied by progress at all. Even if you take the value of `progress` before it's subtracted, it's likely to be an extremely small value, making `target` 0.
> 
> Making `target` 0 fixes the issue **for your use case**, but it breaks it for everything else.

Original code also has problems. After debugging, I found that when the build is not complete, it will cause the target to always be 0.

BugTD:
Your PR code is almost identical to that of itlake, who just deactivated their account two days ago due to several controversial PRs and personal attitude issues. Your account is only two days old, and your first contribution project is Mindustry—itlake. I advise you not to use such ridiculous and shameful methods to submit meaningless PRs over and over again.

PR #10890, opened 2025-06-03
Optimize server list UX

ankando:

### DIFF
--- a/core/src/mindustry/ui/dialogs/JoinDialog.java
+++ b/core/src/mindustry/ui/dialogs/JoinDialog.java
@@ -41,6 +41,9 @@ public class JoinDialog extends BaseDialog{
     String lastIp;
     int lastPort, lastColumns = -1;
     Task ping;
+    Label spinner;
+    int pendingPings = 0;
+
 
     String serverSearch = "";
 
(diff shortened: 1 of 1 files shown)


ghost:
449981664-ee3a2fbb-314f-4e11-8771-e064cfaf3935

ghost:
Optimizations:
1. Wait for the list to load before displaying it, and sort it by delay to reduce the randomness of the server location
2. Add loading text animation

Anon-deisu:
你的代码我在一份已经关闭的pr内见到过并完全一致，你有什么头绪吗ity
I've seen your code in a closed pr and it's exactly the same. Do you have any clues？

BnDLett:
"Optimize" is a bit misleading as it's ambiguous and can easily refer to performance.

For clarity, I'd recommend changing the PR title to something along the lines of "Add a reload spinner to the community server list." 

Looking at this PR on the surface-level, it doesn't seem like it does anything related to "optimizing" (other than optimizing the UX -- so you could also potentially change the title to "Optimize server list UX").

cuskiy:
@Anuken I hope you will take a look at this PR. It would be best if you add a text animation and display it in the order of ping size after it is fully loaded, and then stop loading each time until you click the search button. There is another problem now. When the network is not good, the server location may be misplaced. For example, the server of this server group runs to another server group. I have found it many times but it cannot be directly reproduced.

PR #10897, opened 2025-06-04
Fixed calculation problems when deconstruct

ankando:

I get the amount obtained in the process through 
```
refundedItems[i] += accepting;
```
Because itemsLeft[i] = totalCost[i] - usedItems[i] + refundedItems[i]
So I get usedItems[i], and then I get the total amount that should be given to you, that is, totalRefundItems[i] = usedItems[i] * deconstructMultiplier, and then I use totalRefundItems[i - refundedItems[i] to get the most accurate amount of compensation.

But even so, I found other problems. For example, if deconstructMultiplier = 2, I spent 7 coppers to build it, and refundedItems[i] = 15, it gave me one more, which means that I can't give it, but I have to reduce it by one.

### DIFF
--- a/core/src/mindustry/world/blocks/ConstructBlock.java
+++ b/core/src/mindustry/world/blocks/ConstructBlock.java
@@ -36,6 +36,7 @@ public class ConstructBlock extends Block{
     private static int pitchSeq = 0;
     private static long lastPlayed;
 
+
     public ConstructBlock(int size){
         super("build" + size);
         this.size = size;
@@ -171,6 +172,7 @@ public class ConstructBlock extends Block{
         private @Nullable float[] accumulator;
         private @Nullable float[] totalAccumulator;
         private @Nullable int[] itemsLeft;
+        private @Nullable int[] refundedItems;
 
         @Override
         public String getDisplayName(){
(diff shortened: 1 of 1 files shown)


ghost:
@Anuken The original remaining is not accurate. You should use "the total materials that should be given to you" - "the materials currently given to you" instead of "total cost x refund multiplier" - "total cost - construction cost + materials currently given to you".

core/src/mindustry/world/blocks/ConstructBlock.java
```diff
@@ -171,6 +172,7 @@ public class ConstructBuild extends Building{
         private @Nullable float[] accumulator;
         private @Nullable float[] totalAccumulator;
         private @Nullable int[] itemsLeft;
+        private @Nullable int[] refundedItems;
```
Anuken, line 175: Why add a new array? As far as I can tell, you're always incrementing refundedItems and itemsLeft by the same amount, so their values should be identical at the end.

ghost: > Why add a new array? As far as I can tell, you're always incrementing refundedItems and itemsLeft by the same amount, so their values should be identical at the end.

Besides this, itemsLeft has an initial value, so it is not equal to the amount of materials given to you during the process. It is also not equal to refundItems

ghost: @Anuken 
I found that the initial value of itemsLeft[i] is equal to total construction cost - construction cost, meaning it is 0 when you have fully built it, but not otherwise.

cuskiy:
Because the solution is not perfect, so closed

PR #10898, opened 2025-06-04
Sense number of graphics operations

cardillan:
Adds a new sensable property, `@operations`, returning a number of operations performed on/by an object. Currently only returns the number of drawflushes performed on a display. Since the property is not stored in a map file, it can be used to detect situations where the display needs to be redrawn. The counter is a 32bit value, treated as unsigned to double the range.

(I've intentionally chosen a non-specific name for the property, as perhaps it might be used to be able to sense the number of operations performed by other objects too - turrets firing, factories producing, etc.)

### DIFF
--- a/core/src/mindustry/world/blocks/logic/LogicDisplay.java
+++ b/core/src/mindustry/world/blocks/logic/LogicDisplay.java
@@ -78,7 +78,7 @@ public class LogicDisplay extends Block{
         public float stroke = 1f;
         public LongQueue commands = new LongQueue(256);
         public @Nullable Mat transform;
-        public int operations;
+        public long operations;
 
         @Override
         public void draw(){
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/blocks/logic/LogicDisplay.java
```diff
@@ -111,6 +112,7 @@ public double sense(LAccess sensor){
             return switch(sensor){
                 case displayWidth, displayHeight -> displaySize;
                 case bufferUsage -> commands.size;
+                case operations -> Integer.toUnsignedLong(operations);
```
Anuken: toUnsignedLong is Java 8 API, which is unsupported on some platforms. Just make it a `long` instead.

PR #10900, opened 2025-06-05
Add option to allow input during (logic) cutscenes

JasonP01:
You can now specify if the player can interact with the world during the cutscene. Default behavior is the same as current

Mobile users will not be able to move/rotate their unit during cutscenes due to their behavior of going to their camera.
Desktop does not have that limitation as they do not do that.

This is very useful for minigames in maps and/or other things. Also works with mods.


core/src/mindustry/logic/LStatements.java
```diff
@@ -1763,10 +1763,16 @@ void rebuild(Table table){
 
                     table.add(" speed ");
                     fields(table, p3, str -> p3 = str);
+
+                    table.add(" allow input ");
+                    fields(table, p5, str -> p5 = str);
```
Redstonneur1256: Breaks the continuity of the parameters, this should use `p4`

core/src/mindustry/logic/LStatements.java
```diff
@@ -1763,10 +1763,16 @@ void rebuild(Table table){
...
                 }
                 case zoom -> {
                     table.add(" level ");
                     fields(table, p1, str -> p1 = str);
+
+                    table.add(" allow input ");
+                    fields(table, p5, str -> p5 = str);
```
Redstonneur1256: `p2`

core/src/mindustry/logic/LStatements.java
```diff
@@ -1778,7 +1784,7 @@ public boolean privileged(){
 
         @Override
         public LInstruction build(LAssembler builder){
-            return new CutsceneI(action, builder.var(p1), builder.var(p2), builder.var(p3), builder.var(p4));
+            return new CutsceneI(action, builder.var(p1), builder.var(p2), builder.var(p3), builder.var(p4), builder.var(p5));
```
Redstonneur1256: `p4` isn't even used, adding a 5th parameter is overkill

Redstonneur1256 (changes requested):

core/src/mindustry/logic/LStatements.java
JasonP01: P4 is an already existing parameter. However it seems to be unused rn

Redstonneur1256: Unless you need more parameters for a specific sub instruction you should always use the first parameter after the last used one otherwise it will result in weird serialization (cutscene pan x y speed **0** false)

JasonP01: > Unless you need more parameters for a specific sub instruction you should always use the first parameter after the last used one otherwise it will result in weird serialization (cutscene pan x y speed **0** false)

Understood. Not a huge fan of having zoom use p2 for the boolean and action use p4 tho

Redstonneur1256: > Not a huge fan of having zoom use p2 for the boolean and action use p4 tho

That's how logic works and how it should be, otherwise the parameters won't match on the actual code

core/src/mindustry/logic/LStatements.java
```diff
@@ -1763,10 +1763,16 @@ void rebuild(Table table){
...
+
+                    table.add(" allow input ");
+                    fields(table, p4, str -> p4 = str);
                 }
                 case zoom -> {
                     table.add(" level ");
                     fields(table, p1, str -> p1 = str);
+
+                    table.add(" allow input ");
+                    fields(table, p4, str -> p4 = str);
```
Redstonneur1256: This needs to be p2 not p4

JasonP01: No. If it is p2 it messes with the y of pan. I will not be setting it to p2

Redstonneur1256: It has to be P2, that is how logic works,  and no it will not mess with the y of the pan as this is the zoom instruction

JasonP01: You are just plain wrong. I saw it changing the y. So no i will not be changing it. End of conversation

Redstonneur1256: Yes, if you switch to zoom, change the value and then change back to pan that will change the value, this is how every logic instruction works. It has to be p2 for serialization or else this will not be accepted

core/src/mindustry/logic/LStatements.java
```diff
@@ -1763,10 +1763,16 @@ void rebuild(Table table){
...
                 }
                 case zoom -> {
                     table.add(" level ");
                     fields(table, p1, str -> p1 = str);
+
+                    table.add(" allow input ");
+                    fields(table, "false", str -> p2 = str);
```
Redstonneur1256: This still needs to be p2... otherwise the correct value will never be displayed when opening the menu. If you want it to default to false you could implement some system similar to the [draw operation](https://github.com/Anuken/Mindustry/blob/47c19487fc7502a6903a825d5b1fa1c8bfa2a1c6/core/src/mindustry/logic/LStatements.java#L155-L172)

Redstonneur1256 (changes requested):

core/src/mindustry/logic/LStatements.java
JasonP01: what? this is exactly the same as other instructions. this one i will not be changing. Go actually read the files. try searching for `fields(table, "` and see other things do the same thing

Redstonneur1256: This has "false" as a constant instead of p2, so if you edit the value, close and reopen the processor it will show "false" instead of the actual value.

JasonP01: fixed

JasonP01: ik i am being unnecessarily rude but you are helping so thanks for that 😔

core/src/mindustry/logic/LStatements.java
```diff
@@ -1765,14 +1765,14 @@ void rebuild(Table table){
                     fields(table, p3, str -> p3 = str);
 
                     table.add(" allow input ");
-                    fields(table, p4, str -> p4 = str);
+                    fields(table, "false", p4, str -> p4 = str);
```
Redstonneur1256: Those are the field description ([method](https://github.com/Anuken/Mindustry/blob/47c19487fc7502a6903a825d5b1fa1c8bfa2a1c6/core/src/mindustry/logic/LStatement.java#L115)) so it would cause it to show false when hovering it, if you want the field to have `false` as a default value when changing from/to zoom/pan you need to set the field default value at that time similarly to how draw is [doing it](https://github.com/Anuken/Mindustry/blob/master/core/src/mindustry/logic/LStatements.java#L155-L172)

Redstonneur1256 (changes requested):

core/src/mindustry/logic/LStatements.java
JasonP01: ok, wish this was simpler smh

Redstonneur1256: Just because you can doesn't mean you should. There is plenty of instructions that could have such a feature but don't. In this case it's probably not needed for just a boolean

JasonP01:
Ill fix these issues tomorrow

PR #10904, opened 2025-06-06
Improve server control and logging of the DOS blacklist

BnDLett:

## Added
- A method to remove a DOS blacklisted IP.
- A command to add, remove, and list DOS blacklisted IPs.
- A line to log when a connection is being kicked for being a DOS blacklisted IP.

## Changes
- Switched the double-spaces in the subnet-ban command's listing to instead use a tab. This is better for accessibility and customization. 

## Justifications
- Some mods will accidentally trigger a DOS blacklist. The server only logs the trigger, but it doesn't log anything when the player is kicked for having an IP that is on the DOS blacklist. This can cause confusion for server owners/admins that are trying to resolve a connection issue.
- There is no built-in way to correct this issue other than by running Javascript. The Javascript that is necessary to correct this issue can be confusing to certain individuals, and running unknown Javascript code can pose a significant risk to those that aren't familiar with programming and security.
- While a lot of IDEs encourage using spaces instead of actual tabs (from my knowledge) -- this can cause issues for accessibility. Some individuals may rely on the tab character (denoted by `\t`) to be able to properly read text and information on a screen. Using spaces instead of the tab character can make it harder for an individual that relies on the tab character to be able to access all of the features of the headless Mindustry server. Additionally, the length of tabs can be customized on a local level, so having this level of customization can potentially benefit individuals that would want to customize the length of their tabs (even if they don't need it for accessibility).

### DIFF
--- a/core/src/mindustry/net/ArcNetProvider.java
+++ b/core/src/mindustry/net/ArcNetProvider.java
@@ -113,7 +113,7 @@ public class ArcNetProvider implements NetProvider{
 
                 //kill connections above the limit to prevent spam
                 if((playerLimitCache > 0 && server.getConnections().length > playerLimitCache) || netServer.admins.isDosBlacklisted(ip)){
-                    Log.info(String.format("Closing connection %s - IP marked as a potential DOS attack.", ip));
+                    Log.info("Closing connection @ - IP marked as a potential DOS attack.", ip);
 
                     connection.close(DcReason.closed);
                     return;
--- a/core/src/mindustry/net/ArcNetProvider.java
+++ b/core/src/mindustry/net/ArcNetProvider.java
@@ -113,6 +113,8 @@ public class ArcNetProvider implements NetProvider{
 
                 //kill connections above the limit to prevent spam
                 if((playerLimitCache > 0 && server.getConnections().length > playerLimitCache) || netServer.admins.isDosBlacklisted(ip)){
+                    Log.info(String.format("Closing connection %s - IP marked as a potential DOS attack.", ip));
+
                     connection.close(DcReason.closed);
                     return;
                 }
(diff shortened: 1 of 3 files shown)


core/src/mindustry/net/ArcNetProvider.java
```diff
@@ -113,6 +113,8 @@ public void connected(Connection connection){
 
                 //kill connections above the limit to prevent spam
                 if((playerLimitCache > 0 && server.getConnections().length > playerLimitCache) || netServer.admins.isDosBlacklisted(ip)){
+                    Log.info(String.format("Closing connection %s - IP marked as a potential DOS attack.", ip));
```
Redstonneur1256: This should use `Log.info(String, Object...)` instead of String.format

BnDLett: Right. Yeah. Likely something I missed while I was ironing out the code.

Redstonneur1256 (approved):
Looks good, maybe it could show the players associated with the IPs when listing them.

BnDLett:
That'd likely end up becoming a separate item of concern for the future.

server/src/mindustry/server/ServerControl.java
```diff
@@ -1054,6 +1054,35 @@ protected void registerCommands(){
             }
         });
 
+        handler.register("dos-ban", "[add/remove] [ip]", "Add or remove a DOS ban.", args -> {
+            if (args.length == 0){
```
Anuken: Incorrect formatting here, see CONTRIBUTING.md and surrounding code.

BnDLett: Force of habit -- it should be fixed now.

PR #10908, opened 2025-06-08
display StatusDuration to one decimal place

Uenhe:
casting duration to integer can be confusing. e.g. if statusDuration = 30f, it will be displayed as 0 second.

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -3437,7 +3437,6 @@ public class Blocks{
                     hitEffect = Fx.hitLancer;
                     despawnEffect = Fx.none;
                     status = StatusEffects.shocked;
-                    statusDuration = 10f;
                     hittable = false;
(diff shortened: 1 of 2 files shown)


core/src/mindustry/content/Blocks.java
```diff
@@ -3437,7 +3437,6 @@ Liquids.oil, new LiquidBulletType(Liquids.oil){{
                     hitEffect = Fx.hitLancer;
                     despawnEffect = Fx.none;
                     status = StatusEffects.shocked;
-                    statusDuration = 10f;
```
camelStyleUser, line 3440: why are you removing statusDuration of random bullets(there's 3 other places where you did that)
i would expect a PR that just makes the game display statusDuration to one decimal place to just do that and not touch Blocks.java at all

Uenhe: they are reactive status and remove them doesn't affect game at all

PR #10925, opened 2025-06-13
Bugfix on Landing Pad's "Redirect Launch Pads" UI

WhiteSpike:
- If a sector had both a launch and landing pad of the same resource, when clicking "Redirect Launch Pads" to make other sectors' launch pads to send theirs to that sector, the prompt would remain enabled due to the disable check not considering self sending that same resource, making it seem it did nothing.

Implementation details:
- Refactored some variable names and code blocks for easier reading.
- Created ``canRedirectExports`` method which determines if the provided sector's launch pads can be redirected to the player's current sector. This method is used in both the button's disabled attribute and during execution of redirecting sectors.

Testing details:
- Utilized Windows instructions for compilation and running the application;
- Used save file from Steam's "v8-beta" branch;
- Selected a sector which contained both a launch and landing pad of the same resource. Before the changes, the button would remain enabled after pressing it. After the changes, the button now remains disabled (as all other sectors were already exporting to that sector).

### DIFF
--- a/core/src/mindustry/world/blocks/campaign/LandingPad.java
+++ b/core/src/mindustry/world/blocks/campaign/LandingPad.java
@@ -383,19 +383,22 @@ public class LandingPad extends Block{
                     t.background(Styles.black6);
 
                     t.button(Icon.downOpen, Styles.clearNonei, 40f, () -> {
-                        if(config != null && state.isCampaign()){
-                            for(Sector sector : state.getPlanet().sectors){
-                                if(sector.hasBase() && sector != state.getSector() && sector.info.destination != state.getSector() && sector.info.hasExport(config)){
-                                    sector.info.destination = state.getSector();
-                                    sector.saveInfo();
-                                }
-                            }
-                            state.getSector().info.refreshImportRates(state.getPlanet());
+                        if(config == null || !state.isCampaign()) return;
+
+                        for(Sector sector : state.getPlanet().sectors){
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/blocks/campaign/LandingPad.java
```diff
@@ -383,21 +383,24 @@ public void buildConfiguration(Table table){
...
+                        if(config == null || !state.isCampaign()) return;
+
+                        for(Sector sector : state.getPlanet().sectors){
+                            if(!canRedirectExports(sector)) continue;
+                            sector.info.destination = state.getSector();
+                            sector.saveInfo();
                         }
-                    }).disabled(b -> config == null || !state.isCampaign() || (!state.getPlanet().sectors.contains(s -> s.hasBase() && s.info.hasExport(config) && s.info.destination != state.getSector())))
+                        state.getSector().info.refreshImportRates(state.getPlanet());
+                    }).disabled(button -> config == null || !state.isCampaign() || (!state.getPlanet().sectors.contains(this::canRedirectExports)))
```
Anuken, line 394: ...What is the purpose of renaming a variable that is never used?

In general, please don't refactor or rename things unless absolutely necessary. You can argue that it makes the code "more readable", but what is far more important for me is that the *diff* is readable. 

In general, I don't think using `continue` over `if` is worth a refactor. I'm not asking you to change the PR, but it doesn't look particularly more readable to me.

PR #10943, opened 2025-06-17
Little changes in 3 functions + 2 new constants for future

indexnoname:

- [yes] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [yes] I have ensured that my code compiles, if applicable.
- [yes] I have ensured that any new features in this PR function correctly in-game, if applicable.

Change in the functions.
1. copyFromCpu(). removed if statement and made 2 loops for both coordinates to avoid division and modulo.
2. rotatePlans(). Changed code a little bit to make less calculations. I had left if statements inside of the loop because compiler will remove it (at the least I think so)
3. flipPlans(). Changed code a little bit to make less calculations. I had left if statements inside of the loop because compiler will remove it (at the least I think so)

New constants.
1. public static final int halftilesize = tilesize/2;
2. public static final int doubletilesize = tilesize*2;
Introducing them to maybe potentially replace code alike tilesize/2 or tilesize*2 to reduce computations (If going to be used)

Just for a notice
copyFromCpu() function prob needs to do scanlines. I can try to do it if it is required. For right now it is just a little improvement.

If any problems or suggestions pls tell me, I will try to fix it

### DIFF
--- a/core/src/mindustry/Vars.java
+++ b/core/src/mindustry/Vars.java
@@ -119,6 +119,10 @@ public class Vars implements Loadable{
     public static final float minArmorDamage = 0.1f;
     /** size of tiles in units */
     public static final int tilesize = 8;
+    /** half size of tiles in units */
+    public static final int halftilesize = tilesize/2;
+    /** double size of tiles in units */
+    public static final int doubletilesize = tilesize*2;
     /** size of one tile payload (^2) */
     public static final float tilePayload = tilesize * tilesize;
     /** icon sizes for UI */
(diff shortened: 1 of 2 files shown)


Anuken:
Sorry, this doesn't look necessary to me. None of these functions are bottlenecks, and all of them are called very rarely. I would rather have code that I know *works* rather than (potentially) optimized code that may now be broken. 

This also doesn't fit the style guidelines.

> to reduce computations (If going to be used)

This is misguided. Java inlines constants. `Vars.tilesize * 2f` compiles to `16f` in bytecode. It does not provide any performance difference - and even if it *weren't* inlined, I doubt you would see any noticeable difference.

indexnoname:
k ty for your time

PR #10951, opened 2025-06-20
Musical sounds to be used with wlogic `playsound`

SkyeTheFoxyFox:
Adds several musical sounds that fit the style of the existing mindustry music, intended to be used with the world logic `playsound` instruction to allow map makers and logic developers to make their own music and jingles for their maps/logic.


JasonP01:
i feel like you should prefix the file names with `logic_` just so its clear the sounds are used by logic. Though they would look funny in the selection dialog, a logic folder in the sounds directory could be used but then you would need to touch the where it crawls for sounds :(

Anuken:
I assume these are taken from LMMS, so the license allows it, but these need to be relocated into a separate `instruments/` folder as per the existing sound file structure.

PR #10969, opened 2025-06-26
Mass driver bullets are destroyed like the blocks with items

Agzam4:
When a block (for example container) with items is destroyed, an `dynamicExplosion` is created depending on the items in it
Mass driver bolt too contains an items. It would be logical that when it hits an enemy, it would also creatis an `dynamicExplosion` depending on the objects inside.
I was very upset when I realized that bolt just disappears with all my items when hitting an enemy.
***
Single:

https://github.com/user-attachments/assets/5196faf0-f147-455f-93b1-2dc9002884fc

Groups:

https://github.com/user-attachments/assets/41647a9f-726a-487e-8e59-68d700479f4e

***

Considering the amount of items and the difficulty of hitting the target, I think the damage should be higher. 
But I was afraid to make it too big, given the fact that damage wasn't before.

### DIFF
--- a/core/src/mindustry/entities/bullet/MassDriverBolt.java
+++ b/core/src/mindustry/entities/bullet/MassDriverBolt.java
@@ -3,9 +3,12 @@ package mindustry.entities.bullet;
 import arc.graphics.*;
 import arc.graphics.g2d.*;
 import arc.math.*;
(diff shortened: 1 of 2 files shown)


JasonP01:
This is awesome

PR #11004, opened 2025-07-10
indexable logicBlock vars when reading and writing

1ue999:
This PR aims to make Memory cells utterly useless :3
You can now address variables by index instead of by name which is probably going to be very useful.


Anuken:
I don't like the fact that (1) this supplants the role of memory cells, and (2) the variable order is completely undefined. Writing to "variable 0" or "variable 1" doesn't make any sense. This is an extremely hacky way of implementing storage.

PR #11010, opened 2025-07-12
More escape sequences for logic

1ue999:
This PR aims to improve strings in logic, allowing pretty much any character to be in a string.

A few things i still plan to implement:
- get `"foo\" bar"` to work as a single string
- perhaps add char statements (numbers with a fancy cloak :3)

Feedback would be greatly appreciated.


core/src/mindustry/logic/LAssembler.java
```diff
@@ -67,7 +67,57 @@ public LVar var(String symbol){
 
         //string case
         if(!symbol.isEmpty() && symbol.charAt(0) == '\"' && symbol.charAt(symbol.length() - 1) == '\"'){
-            return putConst("___" + symbol, symbol.substring(1, symbol.length() - 1).replace("\\n", "\n"));
+            //Parse escape codes, loosely based on C escape codes (with some changes)
+            char[] symbolChars = symbol.substring(1, symbol.length() - 1).toCharArray();
```
Anuken: Why are you allocating a new char array just to iterate through it?

1ue999: good question...

1ue999: https://github.com/Anuken/Mindustry/pull/11010/commits/d0aac7beaea634b1ea89b33b520eb6c2e84a9a4d fixed it

1ue999:
currently contains fragments of my debugging (LogicBuild especially)

core/src/mindustry/logic/LParser.java
```diff
@@ -26,6 +26,13 @@ public class LParser{
...
     }
 
+    boolean escaped(){
+        boolean escaped = false;
+        int at = pos;
+        while(at > 0 && chars[--at] == '\\') escaped = !escaped;
+        return escaped;
+    }
+
     void comment(){
```
cardillan, line 29: I'd probably suggest maintaining a boolean flag in `string()` and `statement()` instead of this function.

1ue999: yea seems better

1ue999: should be fixed

1ue999:
my code sucks. time to fix the issues hopefully

1ue999:
Escape sequences and features implemented so far:
- `\n`
- `\ ` inside of non-strings (strings too i believe)
- `\\`
- `\0` 0 can be replaced with any octal number, up to 6 digits
- `\xF` F can be replaced with any upper or lowercase hexadecimal, up to 4 digits
- `"uwu\" owo"`
- `\;`
- `\#`
- `\<\t>` the <\t> being an unescaped \t instead of just `\t` in code
- invalid strings get parsed as normal tokens instead
- some sanitization stuff to not cause errors during parsing hopefully

1ue999:
also the sanitization should probably make bleeding into other fields impossible

cardillan:
@1ue999 In the past, Anuke expressed two concerns regarding logic (that I'm aware of):

* Niche logic features being added, making the logic-related code bigger/more complex and harder to maintain.
* Parsing performance is critical.

In light of these concerns, I'd strongly suggest to remove support for escapes outside strings. (It appears that, except `#`, the UI somehow handles special characters in tokens entered from the UI; if it wasn't a problem up to now, I'd say we don't need to solve it now). (Sorry I didn't express this earlier, I didn't realize the escape sequences outside strings before.)

Note: I don't understand the meaning or purpose of _`\<\t>` the <\t> being an unescaped \t instead of just `\t` in code_.

Inside strings, I'm not sure octal needs to be supported as well as hexadecimal. In my opinion, just hexadecimal would be enough, but if it doesn't increase source code complexity/decrease performance significantly, perhaps it could be kept.

Backwards compatibility: old code with `\` in strings probably will become broken. I though that perhaps single quote for strings supporting these new escapes could be used (this would break code using `'` at the beginning of tokens, but that seems a lesser concern to me). I'm also not sure whether/how these new strings should be supported in the UI; personally I'd be ok if they could be only entered from the clipboard (or directly written to schematic files by compilers etc), but Anuke might think differently.

1ue999:
> @1ue999 In the past, Anuke expressed two concerns regarding logic (that I'm aware of):
> 
>     * Niche logic features being added, making the logic-related code bigger/more complex and harder to maintain.
> 
>     * Parsing performance is critical.
> 
> 
> In light of these concerns, I'd strongly suggest to remove support for escapes outside strings. (It appears that, except `#`, the UI somehow handles special characters in tokens entered from the UI; if it wasn't a problem up to now, I'd say we don't need to solve it now). (Sorry I didn't express this earlier, I didn't realize the escape sequences outside strings before.)
> 
> Note: I don't understand the meaning or purpose of _`\<\t>` the <\t> being an unescaped \t instead of just `\t` in code_.
> 
> Inside strings, I'm not sure octal needs to be supported as well as hexadecimal. In my opinion, just hexadecimal would be enough, but if it doesn't increase source code complexity/decrease performance significantly, perhaps it could be kept.
> 
> Backwards compatibility: old code with `\` in strings probably will become broken. I though that perhaps single quote for strings supporting these new escapes could be used (this would break code using `'` at the beginning of tokens, but that seems a lesser concern to me). I'm also not sure whether/how these new strings should be supported in the UI; personally I'd be ok if they could be only entered from the clipboard (or directly written to schematic files by compilers etc), but Anuke might think differently.

I'll benchmark parsing and see if the escape sequences significantly impact it.
Regarding the `\t`: i believe thats just the tab character, i just wanted it to be escapable too like spaces.
octal and hexadecimal cases are quite useful for getting the ability to put anything in a string.
octal is also often used for things like `\0`
avoiding issues with backwards compatibility seems to be kind of unavoidable.

1ue999:
The changes here have no significant performance impact from my testing.

1ue999:
Making it only affect strings might make it more complex, though i haven't tried yet.

Also regarding the "no problems before"
- you couldn't use them at all (`;`, ` `)
- `#`'s delete everything behind them which might annoy some people.
and this pr fixes both of these issues...

cardillan:
> Making it only affect strings might make it more complex, though i haven't tried yet.

I thought that you could only change the code of the `string()` function and leave the rest as is, and that it would make the PR easier to review - which I also understand to be a factor affecting Anuke's willingness to accept a PR.

> Also regarding the "no problems before"

Yes, there were issues, but I mean no one considered them worth fixing. If this fixes them while making the code more maintainable overall, great!

Btw what matters is what Anuke considers a more maintainable code, and you may have more informed opinion than me in this regard - I've never created a big PR here. I'm just expressing my worries, as I'd very much like to see this merged too :)

1ue999:
> I thought that you could only change the code of the string() function and leave the rest as is, and that it would make the PR easier to review - which I also understand to be a factor affecting Anuke's willingness to accept a PR.

the string() function just grabs a token, the actual parsing of escape sequences happens in LAssembler

cardillan:
On the one hand, it would seem much cleaner conceptually to handle all escapes in LParser (especially since LParser needs to handle escaped terminating characters anyway), and have LAssembler work on the tokens as given, but on the other hand, this would probably require to change how LogicIO writes tokens to mlog. Eek! Eeek!!

Replacing empty strings with "null" in `sanitize()` is great. Further, preserving the identifiers entered by the users in the UI as they are might also be valuable. I therefore withdraw my objections - let's see what Anuke thinks.

1ue999:
applying escape codes in LParser wont happen. the issue is that the unescaped stuff would get back into the ui.

cardillan:
> applying escape codes in LParser wont happen. the issue is that the unescaped stuff would get back into the ui.

Oh! Of course.

1ue999:
pog nothing is happening here :)

Anuken:
There are some insiduous parsing bugs here:

image

The escape is double-evaluated.

image

image

This leaves both quotes unsanitized, outputting and collapsing into a single string literal:
```py
select result notEqual x " a "
print result
printflush message1
```

The current parsing code is terrible, but there are serious bugs here (probably more than I've mentioned), and I'm not interested in fixing up someone else's PR code.

PR #11012, opened 2025-07-13
Setting null to counter should be a no-op

cardillan:
In the past, setting `null` to `@counter` (or any object value for that matter) was a no-op. Recent updates changed this behavior in such a way that setting `@counter`to `null` sets `@counter` to 0. The former behavior was used by some (e.g., a calling a null function pointer resulted in a no-op).

This PR restores the original behavior.

(Another way to fix the problem would be to create a special subclass of LVar for counter, which would make it more apparent in the code for other developers in the future, but I believe LVar not having subclasses allows for better optimization by JIT at runtime. I may be mistaken, though.)

### DIFF
--- a/core/src/mindustry/logic/LVar.java
+++ b/core/src/mindustry/logic/LVar.java
@@ -103,8 +103,12 @@ public class LVar{
 
     public void set(LVar other){
         isobj = other.isobj;
-        objval = other.objval;
-        numval = other.numval;
+        // Setting a non-numeric value to @counter must preserve its numeric field
+        if(isobj){
+            objval = other.objval;
+        }else{
+            numval = other.numval;
+        }
     }
 
     public static boolean invalid(double d){


cardillan:
There was a discussion on Discord that maybe setting `@counter` to null should set it to 0 for consistency. I'd prefer what this PR does (obviously), but I don't want to steamroll the other opinions.

Anyway, the current state is not well-defined. Setting counter to a variable sets it to whatever numeric value that variable holds:

```
set i 0
op add i i 1
print i
printflush message1
set a 1
lookup item a 200
set @counter a
```

eris-webserv:
Why not just... have a dedicated no-op instruction?

cardillan:
@Erisfiregamer1 This isn't about adding a no-op per se. Anything not recognized by the parser is effectively a no-op (it gets converted to an `invalid` statement).

The issue here is two-fold: currently, assigning a non-numerical value to `@counter` is undefined, although in most cases it will lead to `@counter` being set to 0.

The other issue is that a recent commit changed how assignments to `@counter` were handled. In previous versions, assigning any non-numeric value to `@counter` was effectively ignored by the processor. This can be quite useful: when I keep a value in a separate variable and assign it to a `@counter`, a numerical value will cause a jump, while a non-numerical value will let the execution continue on the next instructions. This can be used in at least two context:

* Function pointers: calling a function via a function pointer holding a `null` is a no-op. This can be obviously quite useful -- most importantly, uninitialized function pointers won't break control flow.
* For-each loops: I implement for-each loops by storing the address of the next iteration setup in a variable. When setting up the last iteration, I store `null` there, which means the control flow will exit the loop body naturally. I use this trick when maintaining symbolic labels, as I can't get a symbolic label's absolute address in any way - I need to get to a defined distance from the target address and use actual `@counter` value to compute the final target. This takes some additional instructions, increasing code size and execution time.

For these reasons, I'd prefer the `@counter` behavior to be restored back to the previous versions (specifically, prior to https://github.com/Anuken/Mindustry/commit/b3050c299d6b0118ef8dfaa768c7bb759f3cc294)

PR #11017, opened 2025-07-15
Allowing multiple planets to have the same tech tree

DarkJune0122:
If multiple planets end-up having the same ``TechTree`` by any means - in ``ResearchDialog``, when fetching resources available on the planet, ``ResearchDialog`` will only take into account items from sectors of the **first** planet added during game/mod content loading.

As of right now, this impacts only mods.
To trigger this quirk in the base game you need to attach the same root ``TechNode`` to both the ``Planets.sun`` and ``Planets.serpulo``, for example, because ``sun`` is added to the ``content.planets()`` seq first. (e.g. both need to have the same ``Planet.techTree`` instance attached to them)

Other parts of the game is not impacted by the fix.
Also, I'm not sure who decided to use ``ObjectMap<Sector, ItemSeq> cache = new ObjectMap<>();`` to work with associated sectors, but he/she allowed to implement this change without any problems. Cheers!

### DIFF
--- a/core/src/mindustry/ui/dialogs/ResearchDialog.java
+++ b/core/src/mindustry/ui/dialogs/ResearchDialog.java
@@ -217,13 +217,13 @@ public class ResearchDialog extends BaseDialog{
             {
                 //first, find a planets associated with the current tech tree
                 rootPlanets.clear();
-                for (var planet : content.planets()){
-                    if (planet.techTree == lastNode)
+                for(var planet : content.planets()){
+                    if(planet.techTree == lastNode)
                         rootPlanets.add(planet);
                 }
 
                 //if there is no root, fall back to serpulo
-                if (rootPlanets.size == 0){
+                if(rootPlanets.size == 0){
                     rootPlanets.add(Planets.serpulo);
                 }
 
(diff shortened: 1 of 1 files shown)


DarkJune0122:
But I feel like the ultimate question should be:
Is researching from multiple planets should even be allowed, if they have the same tech tree?

I opt for yes - it feels natural. If planets are united with the same tech tree - they have the same content on them, so it's only fair to have them all contribute to researching.

But the game-play will have to be altered on the second, third and other planets - each one of them should give you a unique challenge after all. And imagine if planets will have different types of resources, unique to each of them! That's the roots of interplanetary logistics! This has to be a part of at least one mod out-there)

core/src/mindustry/ui/dialogs/ResearchDialog.java
```diff
@@ -214,21 +215,29 @@ public void rebuildItems(){
             ObjectMap<Sector, ItemSeq> cache = new ObjectMap<>();
 
             {
-                //first, find a planet associated with the current tech tree
-                Planet rootPlanet = lastNode.planet != null ? lastNode.planet : content.planets().find(p -> p.techTree == lastNode);
+                //first, find a planets associated with the current tech tree
+                rootPlanets.clear();
+                for (var planet : content.planets()){
```
Anuken: Incorrect formatting in several places - extra space before `(`

DarkJune0122: Sorry! On it.

PR #11036, opened 2025-07-22
Update bundle.properties

itflakess:


cuskiy:
This is used for discussion, no matter large or advanced, I think it should be the same as key.

coaldeficit:
literally what is the point of this

cuskiy:
> literally what is the point of this

This is more in line with "best practices". Because when I was translating, someone emphasized to me that I should use "Advanced" when discussing translation, but I and others thought that "Large" was better and in line with the key. Later, I thought that if we change them to Large or Advanced, it will be easier for you to remember when writing code.

Please reply on Discord.

PR #11052, opened 2025-07-24
Game rule for unlimited processor instructions

cardillan:
Adds a new game rule for unlimited processor instruction (default: false).

When activated, the 1000 instruction limit is not enforced at all. The primary use-case is to allow running code which, after adding debugging support, overflows the instruction limit.

Notes:
-  I'm clearing the flag when loading a map in campaign mode, because the `allowEditWorldProcessors` is being cleared there too. The flag should never be set in campaign, so it feels right. I'm not sure about other game modes.
- If no limit at all is not desirable for some reasons, the setting could just increase the limit (maybe double it).


JasonP01:
i love me 2 billion draw instructions filled with 600 character worth of strings loaded onto a single tile

JasonP01:
you should DEFINITELY limit this to wproc as aint no way this is getting merged. this WILL kill servers when the owners are, for lack of a better word, incompetent

Anuken:
Sorry, this is absolutely not being added. I'm okay with making the field non-final so that the limit can be removed by a mod, but the limit exists for a reason - processor editing UI becomes unusable, save files break, and configuration can't be sent over the network when instruction limits aren't enforced.

PR #11058, opened 2025-07-25
Update JoinDialog: Use `addChild` to add info button

way-zer:
No more empty `add()`. Better for mods to add buttons.

---


way-zer:
May we need adding some empty cell.
```java
        //keep for compatibility to old CLaj Mod
        buttons.add().width(Float.MIN_NORMAL).pad(0);
        buttons.add().width(Float.MIN_NORMAL).pad(0);
```

Anuken:
Seeing as this breaks at least mod, and mods can surgically change cells anyway, I don't see a point. Layout shouldn't be arranged in a certain way just to make certain types of mods behave better.

PR #11066, opened 2025-07-26
Support for sensing build requirements

cardillan:
Adds a possibility to sense build requirements on block types. Current game rules are taken into account.

(Inspired by a recent logic jam - makes it easier to create logic for automated base building.)

### DIFF
--- a/core/src/mindustry/world/Block.java
+++ b/core/src/mindustry/world/Block.java
@@ -1489,6 +1489,16 @@ public class Block extends UnlockableContent implements Senseable{
 
     @Override
     public double sense(Content content){
+        if(content instanceof Item item){
+            if(state.rules.infiniteResources) return 0;
+
+            for(ItemStack r : requirements){
+                if(r.item == item){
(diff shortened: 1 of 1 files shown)


core/src/mindustry/world/Block.java
```diff
@@ -1489,6 +1489,16 @@ public double sense(LAccess sensor){
 
     @Override
     public double sense(Content content){
+        if(content instanceof Item item){
```
1ue999, line 1492: that just breaks sensing block contents.

1ue999 (changes requested):

core/src/mindustry/world/Block.java
cardillan: How so? Sensing building contents is in the Building class, not Block class. The Block class represents block types, not block instances.

ETA: I've tried it in game, I can sense container content fine. If I misunderstood you, please let me know what gets broken.

1ue999: oh, i misread stuff. seems fine then.

PR #11068, opened 2025-07-26
No timeout on drop to air

cardillan:
Removes the timeout when dropping items to air. Spamming drop to air should not be a problem, because after the first one the state won't actually change. Not sure if there are other reasons for the timeout apart from a fast item transfer between nearby blocks - if there are, I guess this won't be possible to accept.

Again, inspired by the logic jam: it is difficult to use logic to mine an exact amount of items, e.g. when feeding a block using a unit. Just mining until the desired amount is reached often leads to an additional item being held, which then often incurs the timeout just to get rid of it.

A nicer alternative would be to add support for mining an exact amount of items, but that would require change to the instruction (a new parameter) and would be actually more complicated, since mining - unike itemDrop and itemTake - isn't instantaneous.

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -451,7 +451,7 @@ public class LExecutor{
                         }
                     }
                     case itemDrop -> {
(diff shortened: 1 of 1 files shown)


JasonP01:
> 

> 

>

PR #11069, opened 2025-07-26
Sensor operating on incorrect tile of a tileable display

cardillan:
The sensor instruction operates on the linked tile of a tileable display, not on the root tile, giving wrong results.

~~I've created a method returning the root display, primarily so that, in case a new sensable property gets added, there would be just one place to fix. If you prefer to handle it in TileableLogicDisplay.sense without a new method being added, it can be easily done.~~

I've just moved the rootDisplay up to LogicDisplayBuild. It seems more in line with Mindustry coding style.

### DIFF
--- a/core/src/mindustry/world/blocks/logic/LogicDisplay.java
+++ b/core/src/mindustry/world/blocks/logic/LogicDisplay.java
@@ -73,6 +73,8 @@ public class LogicDisplay extends Block{
     }
 
     public class LogicDisplayBuild extends Building{
+        //The root display (bottom left corner of display for tileable displays)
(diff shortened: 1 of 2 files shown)


cardillan:
To illustrate the problem:

- Create a processor and four tileable displays in a square.
- Paste this code into the processor:
```
sensor result display1 @operations
print result
draw clear 0 0 0 0 0 0
draw print 26 26 @center 32 a 0
drawflush display1
```
- Link different tiles to the processor as `display1`

Only when the root display tile is linked, the correct value is shown:

image

When another tile is linked, 0 is displayed.

This is also a problem when rotating schematics.

PR #11099, opened 2025-08-05
fix: add StorageBlock sync

TonyRDX:
I need to check more deeply their is no side effect, but it seems to fix a desync occuring after some seconds on all client games.
It lookis like desync are higher when latency is high.

Please let me know if any screen record or other may be helpful.

### DIFF
--- a/core/src/mindustry/world/blocks/storage/StorageBlock.java
+++ b/core/src/mindustry/world/blocks/storage/StorageBlock.java
@@ -20,13 +20,14 @@ public class StorageBlock extends Block{
         super(name);
         hasItems = true;
         solid = true;
-        update = false;
+        update = true;
         destructible = true;
         separateItemCapacity = true;
         group = BlockGroup.transportation;
         flags = EnumSet.of(BlockFlag.storage);
         allowResupply = true;
         envEnabled = Env.any;
+        sync = true;
     }
 
     @Override


JasonP01:
No more 99999999 items in a container :(

Anuken:
This seems like a band-aid fix. Why is the desync occurring in the first place?

JasonP01:
> This seems like a band-aid fix. Why is the desync occurring in the first place?

On sufficiently laggy maps, where server tps drops to say 9 and client fps dwindles to like 2 there is a chance that units can drop more items into a container than it has capacity. On extremely long maps you can find a container have 200k+ items in it with its cap being 300 items.

It takes a tremendous amount of lag on the server to have this happen and yes it sometimes does persist past a /sync.

Alternatively if only your client is extremely laggy you can experience the same phenomenon but this time /sync will briefly fix it before your lag makes it come back

JasonP01:
This has been a thing since v6, it is also very funny and never that beneficial to ever report as a bug. Considering at the stages it occurs the map is likely maxed out and a container holding extra items does not yield any beneficial or game breaking advantages

Anuken:
> yes it sometimes does persist past a /sync.

If something persists past a sync, that just sounds like a capacity bug, not a desync?

TonyRDX:
We are able to observe desync after some seconds, I think it occurs especially when destroying a belt. In my opinion, the root issue is Mindustry is not using a determinist engine or lockstep. For example, I did not checked if there is input delay or if destroy is instantly done on the client.

I will try to give you a screen record, it stunned me because it really occurs easily, and probably throwed some of our 2h survival game. It is hard for players to build more ammo if you still see plenty of ammo, and really frustrating.

/sync fixed the issue, but desync occurs really often. If you think it may leads to a huge network overhead, maybe an option would be to have more strict option.

Please note I played some games with that fix, and did not get obvious side effect.

TonyRDX:
Below i succeed to get a better description

### Setup
A host player on a computer
A client player on a second computer in the same room

### Game setup
2 belt, one horizontal, one vertical, the vertical goes on the horizontal
Have another belt, with a missing belt to join

### Trigger
add the missing belt and see the desync

### Result
client_mindustry
host_mindustry

blue arrow is client
red arrow is host

### Alternative game setup
A router instead of belt merge seems to produce a desync result

### Hypothesis
- Belt merge, router, or globally conveyor are not determinist. Insert order, or minor changes may lead to desync. For example, when the output belt change state from freeze to forward, the router will on one client takes the top belt, and on the other the same router will take the right belt.
- Network delay may lead to different setup and execution order. 
For example, if an action is executed 100ms later, maybe the belt setup will vary causing different merging. Or a mine ore will spawn on the 100ms interval

### What I can do
I can try to make some flaky but automated test to check some hypothesis. Then, we could try some fix to check if they are valuable.

### About this fix
I think the submitted fix there could be a minor turnaround and can be merge if commented properly (e.g // TODO remove sync after belt enhance)

Anuken:
Yes, the game is not remotely deterministic, and cannot be made deterministic. I am aware of this. However, there still shouldn't be massive *repeated* desync in container items like that. What kind of setup did you have in order for that to occur?

JasonP01:
There is no reproducable setup. Honestly to see it even semi reliably you will have to play lategame big survival maps with huge walls and lots of containers and units. The circumstances are not development reproducible. But the basics are really just lots of units, lag, low tps and unit logic

If you do not have units putting items in containers then you do not get the overstacking, i have NEVER seen a conveyor ever do it in my 4 years of this game

TonyRDX:
I do not want to be rude, but i will repeat myself, I reproduce it each time I make the setup on the screenshot. 

If my screenshot and step to reproduce are not explicit, please let me know what you are expecting. As shown, bug occurs on the belt just before a potential vault

Sorry, I will be a bit clumsy for some days and it is hard to me to get clean screen record

Anuken:
> I reproduce it each time I make the setup on the screenshot.

Your screenshot doesn't show any container item desync. What I was asking for is a setup reproducing **major storage block item desync** that repeats itself even after a /sync (not just a couple items being wrong).

TonyRDX:
the root issue is belt, and occurs all the game long causing more desync as the game goes, the vault was desync because of the cascade. If you really want a vault, just add a vault on my screenshot setup

PR #11131, opened 2025-08-15
Update LaunchLoadoutDialog.java

50275:
When you try launching to a sector that doesn't allow launch loadouts: 
Old Behavior: Mindustry checks if the sector you're in can launch to the destination. 
New Behavior: Mindustry checks if any sector you own can launch to the destination. 

This PR has a bug: when it changes the sector you launch from, it won't update the line between the source and destination sector -- it's better explained in the image. 
image

### DIFF
--- a/core/src/mindustry/ui/dialogs/LaunchLoadoutDialog.java
+++ b/core/src/mindustry/ui/dialogs/LaunchLoadoutDialog.java
@@ -36,11 +36,31 @@ public class LaunchLoadoutDialog extends BaseDialog{
     public void show(CoreBlock core, Sector sector, Sector destination, Runnable confirm){
         cont.clear();
         buttons.clear();
-
         buttons.defaults().size(160f, 64f);
         buttons.button("@back", Icon.left, this::hide);
-
         addCloseListener();
+
+        selected = universe.getLoadout(core);
+        if(selected == null) selected = schematics.getLoadouts().get((CoreBlock)Blocks.coreShard).first();
+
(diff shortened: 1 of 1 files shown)


50275:
Planned improvements: This PR should also take effect on Serpulo if and only if your current sector is unable to launch to the destination. In this case, it will choose the sector which can launch and has the highest value for its items; value is determined by the sum of the quantity of each item * its cost as defined in Items.java.

50275:
If I actually do add this improvement then I will likely need to introduce another function canLaunch(). In this case I will need advice as to how I can determine whether or not I can launch (most importantly my current criteria acceptable?)

50275:
This PR needs to be updated to include this change. https://github.com/Anuken/Mindustry/commit/ac937ce9100b0f2069545aa085eccbb01c251580

Anuken:
Ideally, it would be implemented [in this method](https://github.com/Anuken/Mindustry/blob/master/core/src/mindustry/maps/generators/PlanetGenerator.java#L57) - you can already see a relevant TODO there. This would also fix the issue of the launch candidate not showing up properly in the overview UI.

50275:
@Anuken that webpage doesn't exist! `https://github.com/Anuken/Mindustry/pull/url` Did you mean PlanetGenerator.findLaunchCandidate(Sector, Sector)? I'll try that right away!

Anuken:
Fixed - I didn't paste the link correctly.

PR #10762, opened 2025-05-04
Allow markers to be used as light sources

Redstonneur1256:
#9733 but fixed to support existing saves.

- [X] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [X] I have ensured that my code compiles, if applicable.
- [X] I have ensured that any new features in this PR function correctly in-game, if applicable.

PR #11142, opened 2025-08-16
Allow markers to use content/displays as textures

Redstonneur1256:
* Allow textures/quad markers to use content or logic displays as their texture.


Redstonneur1256:
Regarding a few design choices, `JsonSerializable` is implemented on the base marker rather than texture/quad markers so that it can be compatible with #10762. I'm not a big fan of the `TextureHolder` class but using `TypeIO` for that seems like the most reliable option than to add json serializers for all types that might be thrown here from logic processors (or through network calls), if there's a better option I missed please tell me. Texture selection on the objectives dialog currently only supports a string as it did previously, maybe a selector similar to the sensor operation/core database could be added as well.

The4codeblocks:
drawing from displays is a little unreliable, the displayed content isn't on the server in any usable way
consider canvases

Redstonneur1256:
Why would display contents need to be on the server? Also thanks for remiding me about canvases, I'll see what I can do when I'm on PC for this and drawing on displays as well. Please do not merge in the meantime (can't find draft option on mobile)

Redstonneur1256:
Should be good to go.

The4codeblocks:
> Why would display contents need to be on the server?

if the display isn't visible to the client, it won't work
displays don't do anything when not rendered

Redstonneur1256:
> if the display isn't visible to the client, it won't work

[Are you sure about that](https://github.com/Anuken/Mindustry/blob/774c6e90f8244c2fe04d7f623341e204e3bd3184/core/src/mindustry/game/MapObjectives.java#L1347-L1362)

The4codeblocks:
> > if the display isn't visible to the client, it won't work
> 
> [Are you sure about that](https://github.com/Anuken/Mindustry/blob/774c6e90f8244c2fe04d7f623341e204e3bd3184/core/src/mindustry/game/MapObjectives.java#L1347-L1362)

huh

core/src/mindustry/game/MapObjectives.java
```diff
@@ -1304,16 +1321,66 @@ private void setUv(int i, double u, double v){
...
+    public static class TextureHolder implements JsonSerializable{
+        private static final ReusableByteOutStream stream = new ReusableByteOutStream();
+        private static final DataOutputStream output = new DataOutputStream(stream);
+
+        public Object value = "white";
+
+        @Override
+        public void write(Json json){
+            stream.reset();
+            TypeIO.writeObject(Writes.get(output), value);
```
Anuken: This is... very, very cursed. Writing raw binary data in JSON as an array doesn't seem right. Even if it's more code, I think it would preferable to do it like so:
```java
"building": 7442931 //integer position in world only if instanceof Building
"string": "value" //string value only if instanceof String
"content": "copper" //content name only if instanceof UnlockableContent
```

Redstonneur1256: While I do agree that it is very cursed, like I explained any value that might get here will either come from a Call call which uses TypeIO or a logic processor which also uses TypeIO for serialization. While it does be cursed it's still a much safer and easier than re-implementing each type in TypeIO. It would also make it so that it doesn't require any changes in the future when TypeIO gets extended to even more types, there's also the prloblem that some types are sligthly more complex like content requiring both the name and type. And lastly the markers are already written as binary json so it shouldn't need to be human readable.

PR #11168, opened 2025-08-24
added two way ping (also script improvements)

5GameMaker:
## Stuff

- Server now pings client back
- Scripts can now be ran from any directory\*
- Scripts are now in readme

*\* for as long as bash allows it*

## Checkboxes!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ config/
 buildSrc/
 *.gif
 /tests/out
+/run
 
 /core/assets/basepartnames
 version.properties
--- a/README.md
+++ b/README.md
@@ -28,8 +28,8 @@ _Sprite Packing:_ `gradlew tools:pack`
 
 ### Linux/Mac OS
 
-_Running:_ `./gradlew desktop:run`  
-_Building:_ `./gradlew desktop:dist`  
+_Running:_ `./server/run-server -1`  
+_Building:_ `./desktop/run-client -1`  
 _Sprite Packing:_ `./gradlew tools:pack`
 
 ### Server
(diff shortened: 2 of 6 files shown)


JasonP01:
Why on earth would the server EVER need to ping the client?

JasonP01:
AND you are changing the build scripts!? This pr is NOT getting merged 😭

5GameMaker:
> Why on earth would the server EVER need to ping the client?

So I can have a better estimate on when to send manual sync than "I guess that's long enough".

I need this to prevent client desync in cases when it comes to vanilla players.

> AND you are changing the build scripts!? This pr is NOT getting merged 😭

To my understanding of them, they're meant to be ran from one specific directory anyway, so this shouldn't mess up existing workflows ([except for the part where project root won't be polluted with server garbage anymore](https://xkcd.com/1172/)).

This isn't a required part part of the PR anyway, so it can be excluded.

5GameMaker:
To think of it, I should've added rlwrap to server build script too so copypaste and history would work

Anuken:
- I don't agree with `run-server` no longer requiring a build version. It **must** be provided.
- I don't see a reason to have a run-client script, and it's not something I will maintain. Restarting infinitely doesn't make sense, either.
- You're essentially combining two different PRs that do completely unrelated things.
- The ping implementation doesn't make any sense. It adds a field, but the field is never updated. The result is logged for some reason, polluting output. Also, why is a ping high priority?

5GameMaker:
>     * The ping implementation doesn't make any sense. It adds a field, but the field is never updated. The result is logged for some reason, polluting output. Also, why is a ping high priority?

It is already high priority as far as I can tell
https://github.com/Anuken/Mindustry/blob/master/core/src/mindustry/core/NetClient.java#L325-L328

PR #11196, opened 2025-08-30
Fix a possible division by zero

eruvanos:

I just came across this potential division by zero error, hope it is worth to just fix it, if not, I am sorry for the interrupt.


Anuken:
Dividing by zero doesn't cause an error in a floating point context, and the function returns before a potential NaN/Inf value can be used.

PR #11220, opened 2025-09-13
Editor copy paste (revisited)

jakubDoka:
This does not support:
- ~~rotating the selection - I have no idea how to visualize that~~
- preserving the building config beyond rotation - just copy stuff in sendbox mode

As suggested, the sampling selection is visualized as a white rectangle (if you know how to make it prettier, let me know). It has some quirks when doing overlapping pastes, but otherwise, it would not make sense.

Let me know if controls are intuitive. I am unable to judge since I implemented them (though I tried my best).


Anuken:
Sorry, but once again, this isn't complete enough. If copy-paste it to be added, it needs to support configs and previews, like schematics do - it can probably use the same system, too. Otherwise, it's just half-baked, and users will (rightly) complain.

PR #11328, opened 2025-11-02
Thorium reactors now vaporize water, burn tar, and destroy boulders

alstocras:
This was an implementation of a feature request candidate.

[Original feature suggestion](https://github.com/Anuken/Mindustry-Suggestions/issues/5847)

If your pull request is **not** translation or serverlist-related, read the list of requirements below and check each box:

- [x] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [x] I have ensured that my code compiles, if applicable.
- [x] I have ensured that any new features in this PR function correctly in-game, if applicable.

PR #11342, opened 2025-11-04
Improved Power Generator Explosions

MEEPofFaith:
Better implementation of #11328.
- Added shake (already a feature, but was unused).
- Added scorch effect
- Added fire
  - Should I use fireball projectiles instead?

https://github.com/user-attachments/assets/81a24c5b-8f46-4f90-bd3d-8b8ee0e72426

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -2505,7 +2505,7 @@ public class Blocks{
 
             explosionShake = 1f;
             explosionScorchSize = 1;
-            explosionIgnitionChance = 0.5f;
+            explosionIgnitionChance = 0.2f;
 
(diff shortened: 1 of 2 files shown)


MEEPofFaith:
(`constructFinish` with a null unit is safe, right?)

JasonP01:
What if modder doesnt want it to break their props 🤔 probably just add `canBreakProps` or something to that breakable check.

alstocras:
I think it would look a lot better if the fireballs were projectiles, as then it would fit in with the 'outward radial explosion' effect that the Th reactor already has, and the other generators would look good with. And um… this is a lot better than mine.

JasonP01:
Projectiles could increase lag during the explosion. Not sure its necessary to begin with

PR #11356, opened 2025-11-12
scale forcefield/shield abilities with healthmultiplier gamerule

EggleEgg:
Not scaling makes units like oct or quasar really useless when it s >1 (low survivability compared to other units), and also very tanky the other way around

https://github.com/user-attachments/assets/04f27aaa-4ec2-4ae7-9bba-895d15d5e171

### DIFF
--- a/core/src/mindustry/entities/abilities/ForceFieldAbility.java
+++ b/core/src/mindustry/entities/abilities/ForceFieldAbility.java
@@ -12,7 +12,6 @@ import mindustry.*;
 import mindustry.content.*;
 import mindustry.gen.*;
 import mindustry.graphics.*;
-import mindustry.type.*;
 import mindustry.ui.*;
 
 import static mindustry.Vars.*;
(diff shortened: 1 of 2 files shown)


JasonP01:
bad  implementation. instead make a shield multiplier rule that works on all shields except for wave shields (for obvious reasons)

DX-36:
I was thinking something more like dividing the taken damage by the health multiplier the same way health multipliers affect unit health and shield when the actual unit is hit

also, force projector needs to be considered too

EggleEgg:
> I was thinking something more like dividing the taken damage by the health multiplier the same way health multipliers affect unit health and shield when the actual unit is hit
> 
> also, force projector needs to be considered too

Force projector is a building? Why should unit health multiplier apply to buildings when [building health multiplier already exists?](https://mindustrygame.github.io/docs/mindustry/game/Rules.html#blockHealthMultiplier)

DX-36:
> > I was thinking something more like dividing the taken damage by the health multiplier the same way health multipliers affect unit health and shield when the actual unit is hit
> > also, force projector needs to be considered too
> 
> Force projector is a building? Why should unit health multiplier apply to buildings when [building health multiplier already exists?](https://mindustrygame.github.io/docs/mindustry/game/Rules.html#blockHealthMultiplier)

as in, force projector shield health isn't affected by building health multipliers the same way unit force field ability shield health isn't affected by unit health multipliers

I also forgot to mention that status effects can have a health multiplier and should also affect the force field ability of units

PR #11367, opened 2025-11-15
ContentPatchs can edit the "itemFilter" of buildings.

m1cxzfw3q:
https://github.com/user-attachments/assets/b22db178-0ebd-4927-9f06-bb63783354e1

### DIFF
--- a/core/assets/contributors
+++ b/core/assets/contributors
@@ -186,3 +186,4 @@ Iniquit
 DSFdsfWxp
 Someone's Shadow
 buj
+MCxzfwq/m1cxzfw3q


m1cxzfw3q:
I really need this change because my map relies on it. Without this modification, the combustion generator in my map would be completely useless.

Anuken:
The filters use numerical IDs, which should never be exposed to the user and *will break* if the order of the items changes, or a consumer is added. This is not an acceptable solution to the issue.

SITUVNgcd:
So... You must implement a Parser that take __itemName__ instead of its ID?

Anuken:
No, I frankly don't understand why this is necessary in the first place. The filter should never be edited directly; it should be set up by the consumers.

PR #11376, opened 2025-11-18
Multiplayer Campaign: Allow Clients to Research Techs

50275:
This PR allows clients to research techs. It works by calling spend(...) on the host. Furthermore, it syncs available resources for research between client and host. 

This PR should have further improvement: 
- Transmit ItemStacks directly instead of int[]
- Code design/structure
- Update items available for research periodically instead of on certain events (opening dialog, researching item) 

### DIFF
--- a/core/src/mindustry/ui/dialogs/ResearchDialog.java
+++ b/core/src/mindustry/ui/dialogs/ResearchDialog.java
@@ -588,10 +588,10 @@ public class ResearchDialog extends BaseDialog{
             if(complete){
                 unlock(node);
             } 
-            // TODO: this has absolutely no effect
-            // show partial research to clients
             if(Structs.contains(node.finishedRequirements, s -> s.amount > 0)){
                 state.rules.partiallyResearched.put(node.content, node.finishedRequirements); 
+                // TODO might not be acceptable
+                Call.setRules(state.rules);
             }
 
             node.save();
--- a/core/src/mindustry/ui/dialogs/ResearchDialog.java
+++ b/core/src/mindustry/ui/dialogs/ResearchDialog.java
@@ -570,7 +570,7 @@ public class ResearchDialog extends BaseDialog{
                 ItemStack completed = node.finishedRequirements[i];
                 
                 //amount actually taken from inventory
-                int used = Math.max(Math.min(req.amount - completed.amount, items.get(req.item)), 0); // LINE 575
+                int used = Math.max(Math.min(req.amount - completed.amount, items.get(req.item)), 0); 
                 items.remove(req.item, used);
                 completed.amount += used;
 
@@ -594,7 +594,7 @@ public class ResearchDialog extends BaseDialog{
             //??????
             Core.scene.act();
             rebuild(shine);
-            itemDisplay.rebuild(items, usedShine); 
+            itemDisplay.rebuild(items, usedShine);
             checkMargin();
         }
 
--- a/core/src/mindustry/ui/dialogs/ResearchDialog.java
+++ b/core/src/mindustry/ui/dialogs/ResearchDialog.java
@@ -85,16 +85,28 @@ public class ResearchDialog extends BaseDialog{
             if(e.content.techNode == null) return;
             if(net.client()){
                 Core.app.post(() -> {
+                    // horrible
+//                    ItemStack[] before = state.rules.partiallyResearched.get(e.content);
+                    ItemStack[] now = new ItemStack[Vars.content.items().size];
+//                    if(before == null) before = now; // shouldn't I arrays.fill(shine, true) then?
+
                     // TODO submission failure
-                    boolean[] shine = new boolean[Vars.content.items().size];
-                    int i = 0;
-                    for(ItemStack stack : e.content.techNode.finishedRequirements){
-                        shine[i++] = e.items[stack.item.id] > stack.amount;
-                        stack.amount = e.items[stack.item.id];
+//                    boolean[] shine = new boolean[Vars.content.items().size];
+                    int index = 0;
+                    for (int i : e.items) {
+                        now[index] = new ItemStack(Vars.content.items().get(index), 0);
+//                        shine[index] = (before[index] == null ? 0 : before[index].amount) < now[index].amount;
+                        now[index++].amount = i;
                     }
+                    state.rules.partiallyResearched.put(e.content, now);
+//                    for(ItemStack stack : e.content.techNode.finishedRequirements){
+//                        shine[stack.item.id] = e.items[stack.item.id] > stack.amount;
+//                        target.finishedRequirements[stack.item.id].amount = e.items[stack.item.id];
+//                    }
+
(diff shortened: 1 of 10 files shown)


core/src/mindustry/ui/dialogs/ResearchDialog.java
```diff
@@ -441,7 +483,7 @@ public void rebuildAll(){
                 button.resizeImage(32f);
                 button.getImage().setScaling(Scaling.fit);
                 button.visible(() -> node.visible);
-                if(!net.client()){
+                if(true){
```
JasonP01, line 486:
```suggestion 
button.clicked(() -> {
```
Its always true no need to have this if statement

core/src/mindustry/ui/dialogs/ResearchDialog.java
```diff
@@ -485,10 +527,10 @@ protected void updateRelative(float percentDelta){
                 button.userObject = node.node;
                 button.setSize(nodeSize);
                 button.update(() -> {
-                    button.setDisabled(net.client() && !mobile);
+                    button.setDisabled(false && !mobile);
```
JasonP01, line 530:
```suggestion
                    button.setDisabled(!mobile)
```
I dont understand the reasoning for false here...

If you never want the button to be disabled then
```suggestion
```

core/src/mindustry/ui/dialogs/ResearchDialog.java
```diff
@@ -485,10 +527,10 @@ protected void updateRelative(float percentDelta){
...
                     float offset = (Core.graphics.getHeight() % 2) / 2f;
                     button.setPosition(node.x + panX + width / 2f, node.y + panY + height / 2f + offset, Align.center);
-                    button.getStyle().up = !locked(node.node) ? Tex.buttonOver : !selectable(node.node) || (!canSpend(node.node) && !net.client()) ? Tex.buttonRed : Tex.button;
+                    button.getStyle().up = !locked(node.node) ? Tex.buttonOver : !selectable(node.node) || (!canSpend(node.node) && !false) ? Tex.buttonRed : Tex.button;
```
JasonP01, line 533:
```suggestion
                    button.getStyle().up = !locked(node.node) ? Tex.buttonOver : !selectable(node.node) || (!canSpend(node.node)) ? Tex.buttonRed : Tex.button;
```

core/src/mindustry/ui/dialogs/ResearchDialog.java
```diff
@@ -792,4 +859,43 @@ public void drawChildren(){
...
-}
+    //Whether this player is allowed to research (for use in multiplayer campaign)
+    public boolean researchAllowed(@Nullable Player player){
+        return !Vars.net.active() || (player != null && (player.admin() || !state.rules.researchRequiresAdmin));
+    }
+    // TODO shouldn't this have a target?
+    @Remote
+    public static void partiallyResearched(Content content, int[] ids, int[] quantity){
+        if(!(content instanceof UnlockableContent u) || u.techNode == null) return;
+        if(net.client()) { // TODO isn't this always true?
```
JasonP01, line 870:
```suggestion
        if(net.server(() {
```
I presume the server is the one researching... why would you want the people playing to handle this?

core/src/mindustry/ui/dialogs/ResearchDialog.java
```diff
@@ -792,4 +859,43 @@ public void drawChildren(){
...
+        if(!ui.research.researchAllowed(player)) return;
+        int[] quantity = new int[Vars.content.items().size];
+        ui.research.rebuildItems();
+        int index = 0;
+        for (Item item : Vars.content.items()) {
+            quantity[index++] = ui.research.items.get(item);
+        }
+        Call.sendPlanetItems(player.con, quantity);
+    }
+    @Remote(targets = Loc.server, variants = Variant.one)
```
JasonP01, line 892:
```suggestion
    @Remote(targets = Loc.server)
```
You should be able to send planetItems to more than 1 client at a time

JasonP01 (changes requested):
Anuke will probably know more so feel free to ignore my comments if they turn out to be bad (reviewing on mobile was a bad choice...)

This is decent just a few weird things that im sure others can point out 👍

core/src/mindustry/ui/dialogs/ResearchDialog.java
JasonP01: Dont blindly merge these suggestions a few dont work as they need further alteration (aka this one)

core/src/mindustry/ui/dialogs/ResearchDialog.java
50275: This method for client-side simulation. It shows partial research progress on the client.

core/src/mindustry/ui/dialogs/ResearchDialog.java
50275: Maybe you want `variants = Variants.both` instead. Doesn't the client only need item data when ResearchDialog is open? It seems pointless to send it to everybody every time even if they aren't researching.

core/src/mindustry/ui/dialogs/ResearchDialog.java
50275: I mean, it is a server -> client method to show partial research progress.

core/src/mindustry/ui/dialogs/ResearchDialog.java
50275: When I was first editing the ResearchDialog, I just replaced net.client() with true everywhere. It was really lazy but I think I should now replace it with researchAllowed() instead.

core/src/mindustry/ui/dialogs/ResearchDialog.java
JasonP01: Faster updates are better, prevents desync if everyone is updated same time

50275: Why can't I update them when they open the ResearchDialog instead?

50275: They won't notice it anyway.

50275: > prevents desync if everyone is updated same time

the dialog takes time to open, it should be long enough for the item displays/partial research progress to update.

If your ping is extremely high, the packets can arrive very late. Then, the solution is to update all clients periodically (every X milliseconds or such). But I actually didn't learn arc.util.Timer and don't have a means to test high latency, so I won't be able to do this.

50275:
TODO: 
- In ResearchDialog, Events.on(PartialResearchEvent, ...) can create extra ItemStack[] objects than it needs to (if previous == null). 
- Tamper-proofing: null-check ResearchDialog.partiallyResearched int[] arrays just in case? (doesn't affect server)
- Use a Pool for boolean[] arrays? It'll only matter if the buttons are spammed however

PR #11383, opened 2025-11-20
Weather Effects Change Vision Radius of Units in Fog

LionPetr:
Implementin a feature from: https://github.com/Anuken/Mindustry-Suggestions/issues/5237

Different weathers now get a fogVisibilityMultiplier stored in their attributes. 
Then in the logic update method the attributes get combined into 1 fogMultiplier (if multiple weathers are active at once) that gets passed using a FogWeatherEvent to all the active units which calculate their new fogRadius using baseFogRadius and the multiplier. 

### DIFF
--- a/core/src/mindustry/core/Logic.java
+++ b/core/src/mindustry/core/Logic.java
@@ -18,8 +18,10 @@ import mindustry.maps.*;
 import mindustry.type.*;
 import mindustry.type.Weather.*;
 import mindustry.world.*;
+import mindustry.world.blocks.Attributes;
 import mindustry.world.blocks.storage.*;
 import mindustry.world.blocks.storage.CoreBlock.*;
+import mindustry.world.meta.Attribute;
 
 import java.util.*;
 
@@ -213,6 +215,14 @@ public class Logic implements ApplicationListener{
                 state.stats.unitsCreated++;
             }
         });
+
+        Events.on(FogWeatherEvent.class,
+                e -> {
+                    for(Unit u : Groups.unit)
+                    {
+                        u.type.updateFogRadius(e.multiplier);
+                    }
+                });
     }
 
     private void checkOverlappingPlans(Team team, Tile tile){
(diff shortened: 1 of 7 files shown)


core/src/mindustry/type/UnitType.java
```diff
@@ -144,6 +144,8 @@ public class UnitType extends UnlockableContent implements Senseable{
     softShadowScl = 1f,
     /** fog view radius in tiles. <0 for automatic radius. */
     fogRadius = -1f,
+    /** base fog radius before effected by weather */
+    baseFogRadius = -1f,
```
Mythril382, line 148: You're modifying a variable in `UnitType`, and while this works because weather is global, I'm sure there is a better way to implement this as it feels very janky, at least in my opinion.

Also, adding a `weatherFogMultiplier` could be good to allow units to have customizable changes.

core/src/mindustry/core/Logic.java
```diff
@@ -353,6 +363,19 @@ protected void updateWeather(){
...
+        for(WeatherState entry : Groups.weather){
+            float mult = entry.weather.attrs.get(Attribute.fogVisibilityMultiplier);
+            if(mult != 0f){
+                fogMultiplier *= mult;
+            }
+        }
+
+        Events.fire(new FogWeatherEvent(fogMultiplier));
+    }
+
```
Mythril382, line 378: I think it would be more visually appealing if the change was interpolated. You could probably use whatever lets weathers fade.

Or it might already exist and I'm blind.

core/src/mindustry/type/Weather.java
```diff
@@ -32,6 +32,7 @@ public class Weather extends UnlockableContent{
     public float soundVol = 0.1f, soundVolMin = 0f;
     public float soundVolOscMag = 0f, soundVolOscScl = 20f;
     public boolean hidden = false;
+    public float fogRadiusMultiplier = 1f;
```
Mythril382, line 35: You should probably use this instead of the attribute.

Mythril382:
I think this is a nice addition and is great for visuals. You could also add this for light radius.

LionPetr:
> I think this is a nice addition and is great for visuals. You could also add this for light radius.

I actually had a question about that. I still am not sure what the light radius does as I played around with it and couldn't tell the difference in game.

core/src/mindustry/type/UnitType.java
LionPetr: > You're modifying a variable in `UnitType`, and while this works because weather is global, I'm sure there is a better way to implement this as it feels very janky, at least in my opinion.
> 
> Also, adding a `weatherFogMultiplier` could be good to allow units to have customizable changes.

For this one, yeah I can see weatherFogMultiplier being a good addition for units and I can add that. I am a little confused b the first comment. I feel like since UnitType controlls fog radius I should change it there? would you rather have me change it in Unit? Either way I will need to get the attribue from UnitType so I thought doing it inside that classs would be the easiest and cleanest implementation. let me know.

PR #11407, opened 2025-11-28
add - null-safe check to block.teamRegion

DiogoPCastelos:
Compile time error using "./gradlew tools:pack".
If a block.teamRegion is null, there is no null-safe check before going into block.teamRegion.found() (which also had no null-safe check).

Fix: add a null-safe check to the "tools/src/mindustry/tools/Generators.java" so it logs to the console that the block has no region.


Twcash:
This would be incredibly helpful for modding.

Anuken:
> This would be incredibly helpful for modding.

The class in question is never used in mods.

> If a block.teamRegion is null

Under what circumstances is it null?

DiogoPCastelos:
> The class in question is never used in mods.

It is if you're locally modding it and running it as if it were to deploy. It's used in modding, not to make a mod tho.

> Under what circumstances is it null?

If there is no sprite called "{name}-team.png". This causes the command to just exit with an error instead of ignoring that block. With this fix it will cause it to log the error block's name, log the error and continue (while still ignoring the block with the missing sprite).

Anuken:
> If there is no sprite called "{name}-team.png"

That is true of a lot of blocks, though. Maybe it shouldn't be null in the first place.

DiogoPCastelos:
> Maybe it shouldn't be null in the first place.

Agreed that it should't but it is haha. But better to have a check that also logs to the console the null ones, no?

This way you can more easily find the null-value ones as well. 

I don't think all types of blocks require a teamRegion tho. No clue which ones do but "Air" was the only block logging an error.

Anuken:
> but it is haha

Then that should be fixed, instead of adding a null check. Most blocks don't *contain* team regions, but still have a placeholder region. Air is a special case, as it is skipped and should not be processed at all.

DiogoPCastelos:
> Then that should be fixed, instead of adding a null check.

Why not both? I get it most blocks don't require it but this isn't a mandatory check, it's a check on those who do have it. It doesn't break the sprites and it warns the user about the error. Not really much more you can ask from a piece of code imo.

PR #11413, opened 2025-11-29
Od nerf now or never

EggleEgg:
### Benefits:

- Making pod more useful (gets replaced by odd almost instantly or outright ignored)
- No more defenses able to take a near infinite number of t5s (t5s are weak currently, and turrets too strong). This applies to everything from menders to turrets to segments. Countering 48 toxos with some odd segments is silly.
- Less lag: upscaling delta to 150/s is terrible for performance. Buildings like unloaders or junctions outright have lower speeds under odd underperforming at a very nice 1.36x their unboosted speed at 15fps/tps. Low fps + 2.5x the calls = lower throughput and even more lag
- Allows for new strats like 2 t3 od and nerfs stupid strats like odd metaglass scatter spam to counter t3s/t4s. Currently out-ecoing scales almost exponentially thanks to early odd.
- Placing od boost will require more braincells than just getting good coverage (there s no downside using od/odd, it s dirt cheap; specially for what it does)
- Nerfed reactive strategies: with 2.5x the resources for building and firepower it s not exactly a fair trade to build a bunch of arcs, menders or wireless/player fed swarmers somewhere inside your base under odd and melt just about anything.
- Better mixtech balance (shouldn't be accounted for but I ll mention it anyways because of multiplayer)

### Downsides:

- Controversial, "gameplay will take forever". Schematic build spam and progression will be slower without 2.5x the resources (less "fun"?)
- Many maps will need changes (only the later sectors in campaign at least)
- T4s and t5s will take forever to produce (can be arguable, erekir t5s take 180s, and with this change odd t5 takes 137s)

image
image

- [x] This is extremely controversial and I doubt it will be merged ever

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1931,7 +1931,7 @@ public class Blocks{
             requirements(Category.effect, with(Items.lead, 100, Items.titanium, 75, Items.silicon, 75, Items.plastanium, 30));
             consumePower(5f);
             size = 2;
-            speedBoost = 1.25f;
+            speedBoost = 1.5f;
             speedBoostPhase = 0.50f;
             consumeItem(Items.phaseFabric, 1).boost();
         }};
@@ -1941,7 +1941,7 @@ public class Blocks{
             consumePower(20f);
             size = 3;
             range = 200f;
-            speedBoost = 1.75f;
+            speedBoost = 2f;
             useTime = 300f;
             hasBoost = false;
             consumeItems(with(Items.phaseFabric, 1, Items.silicon, 1));


jehosula:
wouldn't this make plain OD be outclassed even faster?
i'd say keep base OD the same, everything else is good

holmes-g:
> Less lag: upscaling delta to 150/s is terrible for performance. Buildings like unloaders or junctions outright have lower speeds under odd underperforming at a very nice 1.36x their unboosted speed at 15fps/tps. Low fps + 2.5x the calls = lower throughput and even more lag

Time scale doesn’t increase update frequency; it scales delta values used in progress timers and speeds up buildings in other block-dependent ways. For many buildings, that will affect performance negatively because more block updates will be heavy updates (instead of a low-computation increment-timer-and-early-out path), but I think the negative impact should be lower at low FPS, rather than greater, since at low FPS a greater proportion of updates are already heavy updates which means there is less room for time scale to increase it further.

Rovstam:
Absolutely garbage suggestion

Quit Mindustry immediately.

cheeseballin94-lang:
overdrive dome nerf is to harsh it should be at least 110 or 120 speed the 1k power use is fine the OD nerf is horrible leave my innocent OD alone :(

StalkerBaran:
image

Lol

coaldeficit:
why did you make this pr despite the 70% community disagreement with this crap 2 months ago when you first posted about it

jehosula:
> why did you make this pr despite the 70% community disagreement with this crap 2 months ago when you first posted about it

we genuinely believe it has to be done, and having it as a PR is the best shot we got to getting it added
besides, agreement seems to be in the majority this time

EggleEgg:
> why did you make this pr despite the 70% community disagreement with this crap 2 months ago when you first posted about it

I removed the increased item consumption nerfs since that was one of the major concerns

spentcud:
imma just start yapping even though im probably going to get harrassed for this

the fact that overdrive even exists most likely shows that this game has a fundamental problem with the pacing post-plast; without overdrives, the entire lategame would be slowed to a crawl compared to what we are used to

the fact that the community has become so dependent on overpowered mechanics that they will tell someone asking for a nerf to quit the game immediately shows how much of a crutch the overdrives have become; and thus why they need to be nerfed, if not outright removed, so that we can begin fixing the game's overall pacing and balancing properly

i wanna say more but im honestly not sure where to continue
if it was up to me though i would just straight up remove both the overdrives and try to fix pacing where needed to make up for it

but point is i fully support merging this PR, since an overdrive nerf would be a massive milestone towards properly balancing the wreck that currently is serpulo

and the yappings over

EggleEgg:
image

EggleEgg:
image

PR #11428, opened 2025-12-07
Unit stance no longer affects SectorDamage sum regen per second

50275:
Units capable of building contribute more to sector AFK regen calculations if they are set to the rebuild stance. I propose making them contribute to AFK calculations as if they were always set to rebuild. This would be more intuitive, as right now, the player has no way of knowing that they have to set units to rebuild before leaving. 

### DIFF
--- a/core/src/mindustry/maps/SectorDamage.java
+++ b/core/src/mindustry/maps/SectorDamage.java
@@ -338,9 +338,7 @@ public class SectorDamage{
                 sumHealth += unit.health*healthMult + unit.shield;
                 sumDps += unit.type.dpsEstimate;
                 sumRps += unit.type.weapons.sumf(w -> w.shotsPerSec() * (w.bullet.healPercent/100f * 20f + w.bullet.healAmount));
-                if(unit.controller() instanceof CommandAI ai && ai.command == UnitCommand.rebuildCommand){
-                    sumRps += unit.type.buildSpeed * 20f;
-                }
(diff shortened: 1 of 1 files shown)


EggleEgg:
If this pr gets bad approval for buffing the stupidly inconsistent and op afk mechanics I d wait for json s nerf pr first

50275:
This isn't a buff, it's an accessibility change.

JasonP01:
This is just incredibly bad balancing. It is not an accessibility change when you dont like that you need to actually tell units to rebuild for them to rebuild

50275:
The entire mechanic boils down to clicking a few buttons before you leave. It's not my fault the numbers are too high.

JasonP01:
It is your fault that pressing 2 keys, moving your mouse and left clicking is too hard for you 🥺

jehosula:
> It is your fault that pressing 2 keys, moving your mouse and left clicking is too hard for you 🥺

it's not his fault the whole afk sector mechanic is weird AF and is literally never explained anywhere in the game
how are you supposed to know setting them to rebuild matters when so little else does? (unit positionings, range, etc)
this also doesn't affect balancing at all (if you just set em to rebuild before leaving you get the same result), it's merely convenience and well, acessibility

50275:
[no longer meaningful](<https://github.com/Anuken/Mindustry/commit/1bdac35d90599df35fde3c0f00cd1e4f3fab5d6f>)

PR #11444, opened 2025-12-14
Bullet create event

5GameMaker:
Report bullet creation so if server uses `Call.createBullet` it can set its shooter correctly.

### DIFF
--- a/core/src/mindustry/game/EventType.java
+++ b/core/src/mindustry/game/EventType.java
@@ -382,6 +382,20 @@ public class EventType{
     }
 
     /**
+     * Called when a bullet has been created.
+     * WARNING! This event is special: its instance is reused! Do not cache or use with a timer.
+     * Do not modify any tiles inside listeners that use this tile.
+     * */
+    public static class BulletCreateEvent{
+        public Bullet bullet;
+
+        public BulletCreateEvent set(Bullet bullet){
+            this.bullet = bullet;
+            return this;
+        }
+    }
+
+    /**
      * Called *before* a tile has changed.
      * WARNING! This event is special: its instance is reused! Do not cache or use with a timer.
      * Do not modify any tiles inside listeners that use this tile.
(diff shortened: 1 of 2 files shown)


core/src/mindustry/game/EventType.java
```diff
@@ -381,6 +381,20 @@ public BuildDamageEvent set(Building build, Bullet source){
         }
     }
 
+    /**
```
JasonP01, line 384: What

MEEPofFaith: Creating a billion instances of the event is probably not super performant, so just reuse the same one and attach a warning.

5GameMaker: > Creating a billion instances of the event is probably not super performant, so just reuse the same one and attach a warning.

As far as I can see this is exactly what the code does. The used instance of `BulletCreateEvent` is declared as `static`, thus only one instance of it will ever be instantiated. May you point me where does the code spam instances?

5GameMaker: Also the warning is there if you look 2 lines below

MEEPofFaith: > > Creating a billion instances of the event is probably not super performant, so just reuse the same one and attach a warning.
> 
> As far as I can see this is exactly what the code does. The used instance of `BulletCreateEvent` is declared as `static`, thus only one instance of it will ever be instantiated. May you point me where does the code spam instances?

Yes that's what I said. I described exactly what the code does do the person above.

5GameMaker: > Yes that's what I said. I described exactly what the code does do the person above.

Ah, that's fair. Not that I understand what they were trying to say in the first place.

Since "What" is the only thing that was said (and it's not like this patch is too long to read), I think that can be the end of this what GitHub calls a "conversation".

PR #11448, opened 2025-12-16
per-team toggles for core no-build radius / placement range check

WMF-Industries:
Adds two team-rules for mapmakers


core/src/mindustry/game/Rules.java
```diff
@@ -285,6 +285,10 @@ public float buildSpeed(Team team){
         return buildSpeedMultiplier * teams.get(team).buildSpeedMultiplier;
     }
 
+    public boolean assignPlayers(Team team){
```
Anuken: This doesn't have much to do with storing rulesets, so I don't think it should be moved here. Keep it inlined in NetServer.

PR #11452, opened 2025-12-16
Unloader fixed stupid bug

EggleEgg:
Prioritize core/core container always regardless if factory is full. Previously it would unload from the factory if full, lowering core unloader throughput (11/s is expected).
image
image
image

As seen in these images utilizing the corner will make the unloader unload from the factory up to 50% of the time, greatly reducing core throughput. The fix is easy but you d expect unloader behaviour to be consistent (prioritize core unloading)

### Before
image
image

### After
image
image

### Map used for testing:
image
Tldr, no lag increase, it s literally a boolean on a field that is only updated once on proximity

### Also content patcher qol
image
image

And a release:
https://github.com/EggleEgg/Mindustry/releases/tag/adsda


EggleEgg:
Important to note the coreunload var is actually highly requested and needed, even if optional

image
image
image
image
image
image

PR #11472, opened 2025-12-20
Junction dont send items if next block doesnt contain/use items (such as walls or nodes)

EggleEgg:
### Why is this needed?
As seen in the images by default junction does not send items to the last junction in a chain, but **it does** if there is a different block at the end, regardless if it can use items to begin with (such as walls or nodes), thus increasing cost (items get stuck in that direction, unable to escape ever).
Of course you can make junction to not "clog" ever (see graphite junction schematic), but keeping that behaviour was important since it doesn't need that buff (you are incentivized to use inv sorter for that).

### Before:
image
image

### After:
image
image

### Build used for testing:
image


jehosula:
but why shouldn't this just be a downside of using junction? and wouldn't this make it more similar to inverted sorter?
is it just for a little more consistency?

EggleEgg:
> but why shouldn't this just be a downside of using junction? and wouldn't this make it more similar to inverted sorter? is it just for a little more consistency?

Yes, so the end doesn't store items similarly to where the junction is the final block. Often in schematics gaps next to junctions need to be empty to not increase cost. It s a niche thing for most players but not for schem designers (tldr mlog instructions got their qol merged so why not? it doesnt affect balance, break old stuff or lag the game)
See how in the images below using inv sorter wouldn't be possible because of 2 consecutive limit, and how node placement increases sili cost with +6 (6 sili stuck in a direction):


image

imageimage

EggleEgg:
Drafting because I can use proximity checks to cut that 0.3% loss

Anuken:
image

Broken for sorters (and any other block with hasItems = false, which has a few other exceptions)

EggleEgg:
Okay closing this because I m not gonna die on a hill and produce more potential bugs for such a stupidly niche thing

PR #11503, opened 2025-12-28
``getnode`` and ``setnode``

TripleToxic:
- ``getnode`` instruction using index to retrieve a power or bridge node.
- Sensor for total nodes/links for power node/bridge (cannot be combined because Phase Bridge exist).

update:

- ``setnode`` instruction to modify connections of bridge links and power nodes (has a check for ``net.client()`` so hopefully it will not become a desync machine like ``setprop velocityX/Y``
 
Edit: these instruction have been merged into ``sensor`` and ``control``


Anuken:
This should be implemented as part of `control` and `sense` with different parameters. They don't need to be separate instructions, and they should be implemented in the blocks, instead of utilizing `instanceof`.

It would take a bit of refactoring to get `sense` working with parameters, however.

TripleToxic:
> This should be implemented as part of `control` and `sense` with different parameters. They don't need to be separate instructions, and they should be implemented in the blocks, instead of utilizing `instanceof`.
> 
> It would take a bit of refactoring to get `sense` working with parameters, however.

hmmm, let me try this one out

TripleToxic:
done

EggleEgg:
I m a bit late to this but: love the pr but configuring bridge links should only be available on privileged (wproc) or have some kind of big caveat, as otherwise allowing micro procs to be even more overpowered than they already are would be catastrophic. See:
image
image
image
image
image
image

PR #11521, opened 2026-01-01
bugfix: lancer does not scale correctly under odd and booster

EggleEgg:
image

More incentives to boost your lancers instead of spamming them, otherwise you end up with half the expected dps with cryo and odd

### Before
image

### After
image

### DIFF
--- a/core/src/mindustry/world/blocks/defense/turrets/Turret.java
+++ b/core/src/mindustry/world/blocks/defense/turrets/Turret.java
@@ -742,7 +742,7 @@ public class Turret extends ReloadTurret{
                 int barrel = barrelCounter;
 
                 if(delay > 0f){
-                    Time.run(delay, () -> {
+                    Time.run(delay / timeScale, () -> {
                         //hack: make sure the barrel is the same as what it was when the bullet was queued to fire
                         int prev = barrelCounter;
                         barrelCounter = barrel;
--- a/core/src/mindustry/world/blocks/defense/turrets/Turret.java
+++ b/core/src/mindustry/world/blocks/defense/turrets/Turret.java
@@ -75,6 +75,8 @@ public class Turret extends ReloadTurret{
     public float minWarmup = 0f;
     /** If true, this turret will accurately target moving targets with respect to shoot.firstShotDelay. */
     public boolean accurateDelay = true;
+    /** Whether shoot.firstShotDelay should be affected (reduced) by timescale */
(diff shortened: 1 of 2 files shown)


Anuken:
This doubles Lancer DPS. Why does Lancer need a buff? It's overpowered enough as it is.

EggleEgg:
> This doubles Lancer DPS. Why does Lancer need a buff? It's overpowered enough as it is.

**Reactive** (placed on the spot, no liquid boost) lancer is overpowered, watered lancer is highly resource efficient and cryo lancer is barely used. This change only buffs liquid lancer under od/pod/odd, which by that point there s far better options (plast turrets)

core/src/mindustry/world/blocks/defense/turrets/Turret.java
```diff
@@ -742,7 +742,7 @@ protected void shoot(BulletType type){
                 int barrel = barrelCounter;
 
                 if(delay > 0f){
-                    Time.run(delay, () -> {
+                    Time.run(delay / timeScale, () -> {
```
MEEPofFaith: What about charging animations? Part progress for charging also needs to change dynamically with overdrive.
Mod devs may also not want overdrive to affect charging, as this has been the behavior of charge turrets since forever.

EggleEgg: Implemented the optional field but I m not interested into scaling drawpart with overdrive since it s separate from firstShotDelay and qol for modders but not a balancing concern

SomeonesShade:
> This doubles Lancer DPS. Why does Lancer need a buff? It's overpowered enough as it is.

There is some limitations, in the fact that the approach on rebalancing power turrets into being more coolant reliant (so decrease damage or reload and move it into coolant), is problematic with lancer specifically.

Making it fire slower sucks horribly in game feel, and decreasing damage, you can't divert it into coolant as there is a maxinum reload.

So having no max reload cap on this makes the balancing for it so much easier without sacrificing making lancer feel incredibly slow.

---

Speaking of this... 

I do have a dumb solution to this though, ironically thanks to the bug, that this pr: https://github.com/Anuken/Mindustry/pull/11661 fixes, lancer could fire more shots in one shot so it avoids some of the probably sound issues?


https://github.com/user-attachments/assets/81e59613-e0a7-476a-aeb0-6815a90c5775

image

This however, is incredibly jank for my taste...

PR #11038, opened 2025-07-23
Add unitSot and add to only ripple

Twcash:
If your pull request is **not** translation or serverlist-related, read the list of requirements below and check each box:

- [x] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [x] I have ensured that my code compiles, if applicable.
- [x] I have ensured that any new features in this PR function correctly in-game, if applicable.
Adds a unit sort for artillery turrets (excluding hail and titan due to balance reasons leaving only ripple) That targets the center of unit clusters for better dps efficiency.

https://github.com/user-attachments/assets/0c4f5f1c-337a-4014-a65b-310b834c26cd

PR #11537, opened 2026-01-04
Unitsort grouped to ripple and titan

EggleEgg:
Very similar to https://github.com/Anuken/Mindustry/pull/11038, but should be more performant (thank you Twcash)

Related suggestions:
- https://github.com/Anuken/Mindustry-Suggestions/issues/5509

https://github.com/user-attachments/assets/315f46ca-c5ef-4391-b583-edd131aed4b9

https://github.com/user-attachments/assets/ebec0a76-231b-4114-b8d3-dddacffe14ac

image

### vs Old:
image

And a release: https://github.com/EggleEgg/Mindustry/releases/tag/345ter


SomeonesShade:
I'll add on the side of caution that this does buff ripple a significant amount, but I feel like it should be merged to see the impact, as making ripple more genuinely aoe would be a healthy stance for any comparison against it.

Like the stats it has, it should be broken but a lot of performance is locked away from inaccuracy or sometimes overkill usually, and here improper shooting location.

Anuken:
image

The clusters aren't indexed by team, so it just aims at the location that has the most unit *of any kind*, rather than *enemy* units. (This image is a bit scuffed, but you get the idea - it should be aiming at the top left because there are more units there)

EggleEgg:
fixed
image

core/src/mindustry/entities/UnitSorts.java
```diff
@@ -17,6 +22,28 @@ public class UnitSorts{
...
 
+    public static Sortf grouped(float radius){
+        return grouped(radius, -1f);
+    }
+    /**
+     * @param distanceWeight higher values make distance less important. Set to <= 0 to ignore distance.
+     */
+    public static Sortf grouped(float radius, float distanceWeight){
+        return (u, x, y) -> {
+            if((timer += Time.delta) > 10f || clusterCount.size < 0){
```
Anuken: Before I merge: The timer incorrectly increments **per unit**, meaning its refresh speed depends on the amount of units scanned, which gets very laggy quickly.

Edit: Actually, nevermind, I still can't merge this in its current state: The cluster key depends on radius, which is sort instance-specific, but the map isn't, which means that something with a radius of 1 will be sharing keys with something that has a radius of 100, leading to bizarre results depending on who is updating first and the unit position in question.

Also, the timer and cluster counts should be private, they're implementation details.

EggleEgg: fixed I think. It should cache per radius/distanceWeight, regardless of the number of turrets
image

Anuken:
I had to fix a bunch of other issues:
- O(n) update for every single grouped sort size, every time it needs to update. This is infrequent, but still bad.
- Storage never resets if unused, so the memory just hangs around forever.
- You're taking flying units into account, which these turrets can't hit.
- The comment says, "Only updated when needed, regardless of callers", but this... isn't true, it's updated **every call**, just like before; you are incrementing the timer per-call!

I made all the clusters the same size (70, close enough, I doubt separate sizes matter) and moved the updating to be in TeamData and truly caller-independent. Also, I made the default grouped sort take distance into account, just a little - it doesn't make sense for these turrets to *completely* ignore distance. Cluster size is still the much bigger factor.

PR #11549, opened 2026-01-09
Version/revision numbers for logic

cardillan:
Adds the `@version` built-in constants for logic. Allows checking the game/processor version from the logic code, either to refuse running on incompatible processors, or to allow choosing optimal code paths for a specific version.

Notes:
- The version number is encoded as a double value of `build + version / 100.0`. This allows comparing versions for greater/smaller with a single instruction, while allowing for up to 99 revisions (an improbable value).
- I've used `@version` instead of `@build`, as `@build` might be one day needed for something related to building in-game.
- For custom build, the `@version` is set to 999,999 instead of -1, to allow meaningful comparison of version numbers even in custom builds.
- The variable was added into the "Processor" section, as they generally determine the processor's capabilities. Maybe the "General" section would be better.

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -2618,8 +2618,9 @@ lglobal.sectionLookup = Lookup
 lglobal.@this = The logic block executing the code
 lglobal.@thisx = X coordinate of block executing the code
 lglobal.@thisy = Y coordinate of block executing the code
-lglobal.@links = Total number of blocks linked to this processors
+lglobal.@links = Total number of blocks linked to this processor
 lglobal.@ipt = Execution speed of the processor in instructions per tick (60 ticks = 1 second)
+lglobal.@version = Version number of the game
 
 lglobal.@unitCount = Total number of types of unit content in the game; used with the lookup instruction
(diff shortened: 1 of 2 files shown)


Anuken:
I'm slightly concerned that this will make debugging very confusing, as logic can now have different behavior (and thus different bugs reported) depending on who runs it. "works on my machine" could become a real issue.

cardillan:
This would be a problem when debugging the logic in a build different from the original build, particularly in a custom build, which is always the build used for debugging - do I understand it right?

What I would actually prefer is for the `@version` to contain the most recent "official" release version, e.g., `154.03` as of now. That should be stable enough, if I understand the problem right. However, that information isn't readily available and would need modifications to the build scripts to obtain and parse the latest tag name. I don't know gradle and your build process well enough to even try.

The most important use case for me is to be able to detect a "future" logic version - I can detect past versions using different behaviors of logic in different game versions. My problem is that my compiler might make assumptions that can be broken in a future version (e.g., I may unroll a loop over all items, but when new items get added to the game, `@itemCount` changes and the loop unrolling is incorrect). I'd like to be able to add a check that would prevent the execution of the code on unknown logic versions and require recompilation. It's pretty niche and I can live without that.

PR #11552, opened 2026-01-10
Crucible output scales with attribute boost

EggleEgg:
The vanilla speed magma boost is notoriously bad, as it provides no real advantage, to the point placing thermal generators is objectively the better option
 
 **New:**
2025-11-2920-53-28-ezgif com-video-to-gif-converter
image
Fractional part is deterministic and accumulated over time, so no item loss. Also nice qol for modders as scaling crafter output wasnt really possible before

.


JamesBunny404:
why not make this a seperate PR? too many changes arent mentioned here

EggleEgg:
> why not make this a seperate PR? too many changes arent mentioned here

What specific changes you d want to be mentioned? I cant exactly pr half of the code and expect that to work, nor only implement crafter yield scaling without a usage

Edit*: If you mean the commit spam with balancing changes the changes are not in the actual pr, that s just remaining from when I cloned branches and reverted stuff after that

SomeonesShade:
I forgot to put a comment here on the why, specifically boost to production rate (how fast on manufacturing) isn't that helpful, since all it does is reduce the amount of silicon crucibles needed (and not every other component/building). (for example, a 2x doesn't mean a 2x to space eff or item eff)

Meanwhile, increased yield amount, allows to gain a severe advantage as with the same space and ores you are gaining more proportional output. (if its a 2x, it is a 2x to space eff and item eff).

This should overall feel like a substantial buff that people would worth consider, and it is something that the map maker can control (although if it being coupled to thermals may or may not be an issue, most likely not)

EggleEgg:
image
image
image

Does not break any vanilla factories, including AttributeCrafters. Database entries are fine too

core/src/mindustry/world/meta/StatValues.java
```diff
@@ -261,7 +264,7 @@ public static Stack stack(UnlockableContent item, int amount){
         return stack(item.uiIcon, amount, item);
     }
 
-    public static Stack stack(UnlockableContent item, int amount, boolean tooltip){
+    public static Stack stack(UnlockableContent item, float amount, boolean tooltip){
         return stack(item.uiIcon, amount, item, tooltip);
     }
```
Anuken, line 276: This (and the other float change) changes API and will break mods that use it.

core/src/mindustry/world/blocks/production/AttributeCrafter.java
```diff
@@ -75,6 +106,23 @@ public float efficiencyScale(){
             return scaleLiquidConsumption ? efficiencyMultiplier() : super.efficiencyScale();
         }
 
+        @Override
+        public float scaleOutput(float amount, boolean item, boolean accumulate){
+            scaledOutput = amount * outputMultiplier();
+
+            if(item){
+                if(accumulate){
+                    accumulator += scaledOutput;
```
Anuken: `accumulator` is global for all item output types, but it's called once *per stack of item*, so if something has multiple outputs, this breaks.

I think `scaleOutput` should just return a single a single float and not accept item/accumulate parameters; it's unintuitive and fragile. The actual accumulation should happen in an overridden `craft()` that calls super.craft(). You're mixing a method that returns *and* calculates here.

EggleEgg: wouldn't it be better to handle it inside GenericCrafterBuild.craft() directly (since it already runs through outputItems and output.amount) ?
for modders, and so that other crafter classes may use it in the future (HeatCrafter comes to mind)?

EggleEgg: okay I ll just handle it in attributecrafter like you said since there s no point currently to extend it to other classes

core/src/mindustry/world/blocks/production/GenericCrafter.java
```diff
@@ -119,6 +120,7 @@ public void load(){
     public void init(){
         if(outputItems == null && outputItem != null){
             outputItems = new ItemStack[]{outputItem};
+            outputAccumulator = new float[outputItems.length];
```
Anuken: - This doesn't re-initialize after data patches and access will crash if the output item array changes.
- If someone just specifies outputItems directly, it will crash, because this field never gets initialized.

core/src/mindustry/world/blocks/production/GenericCrafter.java
```diff
@@ -298,12 +300,22 @@ public float totalProgress(){
             return totalProgress;
         }
 
+        public float scaleOutput(float amount){
```
Anuken, line 304: Needs a comment explaining what this does.

Anuken (changes requested):

core/src/mindustry/world/blocks/production/GenericCrafter.java
EggleEgg: fixed (i think)
image

core/src/mindustry/world/meta/StatValues.java
```diff
@@ -247,7 +252,7 @@ private static Stack stack(TextureRegion region, int amount, @Nullable Unlockabl
         if(amount != 0){
             stack.add(new Table(t -> {
                 t.left().bottom();
-                t.add(amount >= 1000 ? UI.formatAmount(amount) : amount + "").name("stack amount").style(Styles.outlineLabel);
+                t.add(amount >= 1000 || amount == (int)amount ? UI.formatAmount((int)amount) : amount + "").name("stack amount").style(Styles.outlineLabel);
```
Anuken:
```suggestion
                t.add(amount >= 1000 || Mathf.equal(amount, (int)amount) ? UI.formatAmount((int)amount) : amount + "").name("stack amount").style(Styles.outlineLabel);
```

core/src/mindustry/world/blocks/production/GenericCrafter.java
```diff
@@ -25,6 +25,7 @@ public class GenericCrafter extends Block{
     public @Nullable ItemStack outputItem;
     /** Overwrites outputItem if not null. */
     public @Nullable ItemStack[] outputItems;
+    public @Nullable float[] outputAccumulator;
```
Anuken: Wait. You put this in GenericCrafter, *the block*, not the building, so it's shared across all buildings. They will all accumulate at once; very broken.

core/src/mindustry/world/meta/StatValues.java
```diff
@@ -375,6 +395,74 @@ public static StatValue blocks(Attribute attr, boolean floating, float scale, bo
...
+    
+                    if(state.isGame()){
+                        var blocks = Vars.content.blocks()
+                        .select(block -> (!checkFloors || block instanceof Floor) && indexer.isBlockPresent(block) && block.attributes.get(attr) != 0 && !((block instanceof Floor f && f.isDeep()) && !floating))
+                        .with(s -> s.sort(f -> f.attributes.get(attr)));
+    
+                        if(blocks.any()){
+                            for(var block : blocks){
+                                c.table(Styles.grayPanel, b -> {
+                                    float effiency = 1f + block.attributes.get(attr) * scaleEff;
```
Anuken: typo: `effiency`

Anuken (changes requested):

core/src/mindustry/world/blocks/production/GenericCrafter.java
```diff
@@ -298,12 +300,27 @@ public float totalProgress(){
...
+            return amount;
+        }
+
         public void craft(){
             consume();
 
             if(outputItems != null){
-                for(var output : outputItems){
-                    for(int i = 0; i < output.amount; i++){
+                if (outputAccumulator == null || outputAccumulator.size != outputItems.length) {
```
Anuken: - Why is it a FloatSeq? That's a wrapper around float[] for easily adding things + dynamically resizing, but you're not doing that here. It just makes things more clunky here.
- Improper formatting and extra ; at the } for some reason.

PR #11555, opened 2026-01-12
pierceable block variable

EggleEgg:
Not that useful currently after fragile was added but could be in the future

https://github.com/user-attachments/assets/3cf75867-bdea-48ab-95a9-e42f55dabf84

https://discord.com/channels/391020510269669376/396416151032299521/1460313084295118950
For those who didnt read the convo in development


Anuken:
- What is the reason to make nodes pierceable, if they aren't a problem for bullets right now due to `underBullets = true`? I am concerned that this makes infrastructure more fragile without good reason. It's also confusing, as it makes it seem like certain bullets have infinite pierce, when they do not.
- How does this work when bullets not normally expected to pierce (e.g. with fragmentation effects) hit a pierce-able block?

EggleEgg:
> * What is the reason to make nodes pierceable, if they aren't a problem for bullets right now due to `underBullets = true`? I am concerned that this makes infrastructure more fragile without good reason.

Some structures might not make sense to ignore bullets, or end up in a situation where both ignoring and not ignoring bullets actually benefit them unfairly (see nodes blocking pathfinding and being unhittable, but being hittable isnt an option either)

> * How does this work when bullets not normally expected to pierce (e.g. with fragmentation effects) hit a pierce-able block?

### Updated
**Before:**
image

**After:**

image



I ll remove the field from blocks.java just in case, but I wouldn't be surprised if new/old cheese strats are developed for other units that can or cannot crush spammable blocks (thus requiring a fix like this). I really dont know a better approach to solve this problem even if it s confusing; even if the spammable block had 1 hp it would still have worthwhile trades just for the pierce absortion alone.
For example: now that serpulo nodes are underbullets this allows for situations like these. I have no knowledge whether this is actually a big deal or not.

image

Anuken:
Considering it's not used in vanilla anymore... I'd rather not take the risk of breaking something just for the sake of a mechanic only mods will use.

PR #11559, opened 2026-01-13
Endgame build cost scaling

Efsfse2280:
This is needed, as plast+ tech only has up to 130% of tick value/item. (why is metaglass higher??)

Calculated by (adding all relevant item's tick value in recipe)/2. (plast not divided)

So;

- 2 for plastanium

- 4.4/2= 2.2 for phase fabric

- 8.5/2= 4.25 for surge alloy

#### draft status; it would be coupled with the general bspeed nerf.


JasonP01:
I dont see why making the game slower is beneficial in this manner, this does not benefit any player whatsoever

Efsfse2280:
By making endgame tech slower to build, https://github.com/Anuken/Mindustry-Suggestions/issues/6059 becomes more relevant

Right now, a few support T2's build speed can easily deploy an endgame setup

In order to get that, a better/bigger support unit squad should be needed

PR #11572, opened 2026-01-16
Make Jumps to self yield

1ue999:
So, since there is no way for the condition of a jump condition which jumps to itself to change within the same tick, it can safely yield. (if there is ANY way for it to change, this pr breaks some stuff prob)
This is going to make busy-waits (jump 0 lessThan @tick 123) less impactful on performance (yay, million ipt busy waits are less performance eating now :))
Very minor pr to increase performance a little bit :)

raaa i just cleaned up my master branch and now i did a commit to master again


1ue999:
okay so a concern is that the accumulator increases...
a solution would be setting it to 0 instead of yielding.
im not sure if that should be implemented though.

1ue999:
i thought about this, bad idea.

PR #11584, opened 2026-01-19
Scale build plan (construct block) with construct and deconstruct progress

EggleEgg:
No more spending 2 minutes building a very expensive block or core just for a flare to destroy it instantly because it has 10 hp

Max buildplant health is 10 by default as before, but with an optional variable to set plan max health to a % of the block s health (constructHealthMultiplier).
_Example: if a block has 5.8k hp and the multiplier is 0.5, the buildplan s health will increase with progress (0-1) up to its 2.9k maxhealth. If the multiplier is <=0 (default) it will only increase up to 10_

**Blocks with constructHealthMultiplier:**
- Impact reactor, neoplasia reactor, foreshadow, spectre, meltdown, scathe, smite, malign, multiplicative reconstructor, tetrative reconstructor -> 0.5
- Core shard, foundation, nucleus, bastion, citadel, acropolis -> 0.65

https://github.com/user-attachments/assets/61e6a99e-db68-4741-9833-e3efbe7994d0

https://github.com/user-attachments/assets/fe0e68f9-1f27-4c2c-83ca-a597d045fba7

.

And a release https://github.com/EggleEgg/Mindustry/releases/tag/34trre


core/src/mindustry/world/blocks/ConstructBlock.java
```diff
@@ -430,6 +435,14 @@ public float progress(){
             return progress;
         }
 
+        public void scaleHealth(float progress){
```
way-zer: The PR looks good, and I think it can be applied to all buildings.

However, this function seems to need some optimization:
1. Why use `tile.build` instead of `this`?
2. `previousHealth` and `damaged` are not persisted or synchronized.

way-zer: Perhaps we could implement like this: 
* old = health / maxHealth
* maxHealth = xxxx 
* health = old * maxHealth

LeoSko:
I implemented this back in v6 for Ranked server, just would like to point out for thorough testing on building, then deconstructing building mid-build, then building it a bit up again until built, etc. Especially when mid-built the ConstructBlock got some damage (then, again, it started to de/re-construct by the player).

EggleEgg:
After some basic testing done on a hosted server this pr is horribly broken and only becomes worse with lower tps. Thank you LeoSko, I ll try to fix it with my 2 braincells

redloong9527:
This feature is really great.

MinDoUser:
This is actually implementing my suggestion I made a couple months ago (Mindustry-Suggestions/6338) huh. 
Didn‘t even realize someone had attempted to do this pr because there were like >100 at that time. 
Looking forward for this.

PR #11600, opened 2026-01-25
Mixtech nano rebalance: thor reactor

EggleEgg:
- Max heat per side: 15 -> 12
- Scale thor usage with heat stored (0-1) up to 10x the consumption (1.7 thor/s)
- Reactor only produces erekir heat as long as it s producing power
- Increased health and heat increases slower, reduced coolant mult (serpulo ratios are exactly the same as old). Should make it more viable when reactor explosions &/or logic are disabled. Time to explode when full 0.833s -> 3.54s

image
image
image
image

### After:

image
image

https://github.com/user-attachments/assets/69f7f83e-5cb6-4226-92c5-3e93e3ec61e9

(reactor explosion is disabled for testing)


JasonP01:
Extirpater is the one who designed the schem in the images i shared :3

coaldeficit:
mixtech is unsupported

jehosula:
> mixtech is unsupported

maybe in the context of normal balancing
but here, this isn't hurting any of the main "supported" elements for the sake of mixtech, if mixtech is truly unsupported why would this feature even be implemented?

JasonP01:
> mixtech is unsupported

Anuke specifically designed erekir in a way so that mixtech is entirely supported. Beam nodes are the best example. Your statement is factually incorrect

coaldeficit:
> Anuke specifically designed erekir in a way so that mixtech is entirely supported. Beam nodes are the best example. Your statement is factually incorrect

designing the game so mods can use block **types** from both planets doesnt mean mixtech balancing is supported my guy

spentcud:
how many nerfs does thor reactor heating need-
like jesus christ

alecthe2nd:
Something to bear in mind with this change is that this may have some effect on Serpulo campaign. Namely, it causes Thorium Reactors to consume thorium a lot slower at full tilt when fully cooled.
The math is as follows:
heating = 0.02
itemCapacity = 30
itemDuration = 360
heatConsumeRate = 10
timeScale and delta() assumed to be 1 or close to it (as it would be at 60fps without overdrive)
reactor is assumed to be full (as that generates the most power) i.e. items.get(Items.thorium) = 30
fullness = items.get(Items.thorium) / itemCapacity = 30/30 = 1
reactor is assumed to be fully cooled, as in a standard serpulo thor reactor power setup, meaning heat is always zero by the time the update method next runs.
heat is first increased by fullness * heating * delta() = 0.02 * 1 * 1 = 0.02
next the timer is checked, looking at itemDuration /(timeScale * 1 * heat * heatConsumeRate ) = 360 / (1 * 1 * 0.02 * 10) = 360/0.2 = 1800.
I.E. this method of power gen consumes thorium 5 times slower than before. That is, an effective production time change of 6 sec_ -> 30 seconds
we go from the already good thor per sec of 0.166_ to 0.0333_
For more context, one unwatered laser drill on exactly 1 ore tile of thorium would power 3.6 thorium reactors, where before it wouldn't fully power 1.)
If this is to avoid effecting campaign and serpulo balancing, you will need to ensure that, when fully cooled, the reactor runs at the expected 1 item per 6 second pace.

EggleEgg:
image
image
no longer breaks schems, even at low fps

SomeonesShade:
I'd like some footage on what is needed for thor reactor to go boom now, making it more responsive and easier to deal with, would help allieviate some of the rtg vs thor reactor comparisons, as players still prefer the former even if its deathly expensive.

so probably how much time to notice ish and prevent

DeterMination-Wind:
The thorium reactor doesn't need that many changes at all. You just need to specify that the thorium reactor must have coolant to reduce heat—instead of decreasing when the thorium reactor is left idle. This adjustment also doesn't affect Serpulo at all. @EggleEgg

DeterMination-Wind:
BEFORE:
image
image
image
AFTER:
image
image
image

EggleEgg:
> The thorium reactor doesn't need that many changes at all. You just need to specify that the thorium reactor must have coolant to reduce heat—instead of decreasing when the thorium reactor is left idle. This adjustment also doesn't affect Serpulo at all. @EggleEgg

It is intended to not affect serpulo as much, with slower heat increase and more hp to make it less reliant on logic and noob-friendly. It does not break serpulo ratios.
I implemented thor reactor to only produce heat when actively cooled before but it was seen as a bug and anuke patched it to give thorium reactor passive cooling without cryofluid. I cannot revert sadly this as it is anuke s decision (and the one from other players)

DeterMination-Wind:
> > The thorium reactor doesn't need that many changes at all. You just need to specify that the thorium reactor must have coolant to reduce heat—instead of decreasing when the thorium reactor is left idle. This adjustment also doesn't affect Serpulo at all. @EggleEgg
> 
> It is intended to not affect serpulo as much, with slower heat increase and more hp to make it less reliant on logic and noob-friendly. It does not break serpulo ratios. I implemented thor reactor to only produce heat when actively cooled before but it was seen as a bug and anuke patched it to give thorium reactor passive cooling without cryofluid. I cannot revert sadly this as it is anuke s decision (and the one from other players)

Even if "the reactor only produces Erekir heat as long as it's producing power," the thorium reactor remains largely unaffected. For example, when the thorium reactor heats up, it almost always maintains +30 power output rather than dropping to 0.

DeterMination-Wind:
image

EggleEgg:
after:
image

now the reactor needs to be full to produce max erekir heat, which makes it extremely risky

And a release for testing. Thank you for your efforts https://github.com/EggleEgg/Mindustry/releases/tag/43re

SomeonesShade:
OHHH ITS IN! Nice

PR #11605, opened 2026-01-27
Ability to see and grab what's inside of bridge blocks

EggleEgg:
Implements 
- https://github.com/Anuken/Mindustry-Suggestions/issues/6074

image
image


JasonP01:
Anuke has taken things straight from foos, and i know buthed and foos himself dont mind.

JasonP01:
Okay that weave interval thing. How are mobile users supposed to do that

Also forced keybind is 🥀

LixieWulf:
The weave placement would remove the main niche of plastanium belts in the campaign. It does for me, at least. May be better to keep such a behaviour hidden.

JasonP01:
Weaving should not be a toggle or automatic feature in vanilla. There is a reason custom clients explicitly contain features that may be qol but dont belong in vanilla. You should not add bridge weaving as a setting to vanilla.

Anuken:
image

I don't think it's worth it. Under some configurations/zoom levels, it becomes impossible to link bridges. Mass drivers can get away with this, since they're typically placed far away, but bridges don't work that way.

EggleEgg:
> I don't think it's worth it. Under some configurations/zoom levels, it becomes impossible to link bridges. Mass drivers can get away with this, since they're typically placed far away, but bridges don't work that way.

Maybe reducing the drawing size would help? even if it makes it slightly more annoying to unclog manually it might still better than nothing/replacing the bridge

JasonP01:
There is also the option of rendering the item on the opposite side of the cursor. So it moves out of the way of the mouse, this behaves weirdly on mobile so i dont see a good way for that specifically

EggleEgg:
https://github.com/user-attachments/assets/d8ed0021-c412-497e-86dc-596c4c956759

buthed010203:
> ~~There is also the option of rendering the item on the opposite side of the cursor. So it moves out of the way of the mouse, this behaves weirdly on mobile so i dont see a good way for that specifically~~

~~The branch is called stealing from foo's but you didn't bother stealing the impl that we have that fixes this.
In foo's , we just move align to the bottom instead of top for bridges specifically. It's not a perfect solution obviously but I would argue that both solutions could be combined instead. You're going to have to swap the display on a lot of bridges manually otherwise. When merging foo's, integrating your system for "swapping" the side is quite simple:~~
image



image

Edit: Nevermind. Anuke is killing swapping and just adding this himself.

PR #11617, opened 2026-02-03
added color mappings for Items and Liquids

lowlightee:
Added colors for all Items and Liquids. This will allow prettier and more consistent ingame printing and drawing. I also suggest slightly tweaking the colors in the future for better visuals.

obraz

### DIFF
--- a/core/src/mindustry/ClientLauncher.java
+++ b/core/src/mindustry/ClientLauncher.java
@@ -11,6 +11,7 @@ import arc.math.*;
 import arc.util.*;
 import mindustry.ai.*;
 import mindustry.audio.*;
+import mindustry.content.*;
 import mindustry.core.*;
 import mindustry.ctype.*;
 import mindustry.game.EventType.*;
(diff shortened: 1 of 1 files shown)


way-zer:
Within a paragraph, icons with colored names are not more aesthetically pleasing than icons plus names. 
They only look better when listed in this way. It can also easily lead to aesthetic fatigue.

As for whether or not to include it in Colors, I reserve my opinion and hope it won't be misused.

lowlightee:
My goal was to create a standardized way of drawing/printing messages like these. Although not necessary for function, it would make the colors cleaner and more intuitive across schematics made by different people.

This idea can also be expanded by changing the colors codes to the "code" names of materials ("[@phase-fabric]" instead of "[phase]"). This would allow extracting a color from any item/liquid purely from their reference, even modded items/liquids. This would allow to get a color for any material (even modded ones, which I don't think is possible currently) without having to manually map them in-game.

image
image
image
image

Anuken:
Looking at it again, I don't think this needs to be implemented. Content isn't dynamically registered, meaning any new items/liquids would have to be put in there manually. Furthermore, anyone using item names in brackets, e.g. [copper], will now have that turn into a color code, which may be rare, but still undesirable.

PR #11623, opened 2026-02-05
Short waits for accumulating instruction quota

cardillan:
The #11568 PR was meant to allow using the wait instruction to ensure the `LogicBuild`'s accumulator will reach high enough value to execute a portion of code in a single frame update, providing a way to execute short sections of code atomically.

Unfortunately, it turns out that the `wait` instruction doesn't always guarantee the expected value of the accumulator. Failures occur even at 60 FPS when resizing or moving the game window or when switching focus away or to the game. At higher FPS, failures occur sporadically at a quite high rate. Increasing wait times (even by a full tick) doesn't solve the problem entirely, if it helps at all.

Instead of trying to fix the problem by altering the time management in `WaitI`, which might not even be possible given the complexity of the delta time implementation, this PR ensures the wait will fill the accumulator to match the value corresponding to the wait length, taking the accumulator cap into account. This also fixes a potential problem of the accumulator starting at 0, but running wait not restarting, on map load.

I've created schematics for testing, available here: https://github.com/cardillan/golem/tree/main/benchmarks/atomic. The `test-hyper.msch`, `test-logic.msch` and `test-micro.msch` contains processors that use the `wait` instruction to execute a portion of code of a specific length atomically (several different lengths are tested), verifying the correct execution by comparing `@tick` at the beginning and at the end of the atomic section. Failures are indicated by the red color on the displays. Before this fix, failures do occur quite often. After this fix, there are no failures at all.

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -1158,8 +1158,6 @@ public class LExecutor{
 
         public float curTime;
 
-        public boolean yielded;
-
         public WaitI(LVar value){
             this.value = value;
         }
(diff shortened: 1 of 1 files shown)


cardillan:
I've simplified the implementation - there's now no distinction between short and long waits and no explicit cutoff at 5 ticks, and the code is arguably even simpler this way. I've updated the description above to match the new implementation.

cardillan:
I've probably figured a better solution, consisting of just moving a single line of code. I'll create a PR for that after some testing.

PR #11627, opened 2026-02-06
New map rule to disable unit control via logic

WMF-Industries:
No more wireless everything spam (if the mapmaker doesn't want it)

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -146,8 +146,7 @@ public class LExecutor{
 
         @Override
         public void run(LExecutor exec){
-            if(!exec.privileged && !state.rules.logicUnitControl)
-                return;
+            if(!exec.privileged && !state.rules.logicUnitControl) return;
 
             if(exec.binds == null || exec.binds.length != content.units().size){
                 exec.binds = new int[content.units().size];
@@ -200,8 +199,7 @@ public class LExecutor{
 
         @Override
         public void run(LExecutor exec){
-            if(!exec.privileged && !state.rules.logicUnitControl)
-                return;
+            if(!exec.privileged && !state.rules.logicUnitControl) return;
 
             Object unitObj = exec.unit.obj();
             LogicAI ai = UnitControlI.checkLogicAI(exec, unitObj);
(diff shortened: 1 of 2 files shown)


core/src/mindustry/logic/LExecutor.java
```diff
@@ -146,6 +146,9 @@ public UnitBindI(){
 
         @Override
         public void run(LExecutor exec){
+            if(!exec.privileged && !state.rules.logicUnitControl)
+                return;
```
1ue999: i kinda dont like the return on next line...

core/src/mindustry/logic/LExecutor.java
```diff
@@ -197,6 +200,9 @@ public UnitLocateI(){
...
```
1ue999: same here

core/src/mindustry/logic/LogicDialog.java
```diff
@@ -300,7 +300,8 @@ public void showAddDialog(int position){
                     for(Prov<LStatement> prov : LogicIO.allStatements){
                         LStatement example = prov.get();
                         if(example instanceof InvalidStatement || example.hidden() || (example.privileged() && !privileged) || (example.nonPrivileged() && privileged) ||
-                            (!text.isEmpty() && !example.name().toLowerCase(Locale.ROOT).contains(text) && !example.typeName().toLowerCase(Locale.ROOT).contains(text))) continue;
+                            (!text.isEmpty() && !example.name().toLowerCase(Locale.ROOT).contains(text) && !example.typeName().toLowerCase(Locale.ROOT).contains(text)) ||
+                            (!privileged && !state.rules.logicUnitControl && example.category().equals(LCategory.unit))) continue;
```
Anuken: should use `==` instead of equals here

core/src/mindustry/logic/LExecutor.java
Anuken: Yes, return statements should be on the same line.

WMF-Industries: ~~i'd say returns on newlines make the code more readable (i use them everywhere in mods/custom gamemodes) but fair enough~~

core/src/mindustry/logic/LogicDialog.java
WMF-Industries: fair enough, wasn't sure so used the safer option

SomeonesShade:
yay, finally you can have logic w/ only the block control stuff (unloader control, md stuff, enable/disable functions, etc.)

PR #11633, opened 2026-02-08
Do not log every single connect in a DDOS

BalaM314:
This log message was added in #10904. It is useful to help players who are falsely dosblacklisted, but it is not worth the massive amounts of console spam (and possibly lag) that it causes in a real DDoS attack.

### DIFF
--- a/core/src/mindustry/net/ArcNetProvider.java
+++ b/core/src/mindustry/net/ArcNetProvider.java
@@ -113,8 +113,6 @@ public class ArcNetProvider implements NetProvider{
 
                 //kill connections above the limit to prevent spam
                 if((playerLimitCache > 0 && server.getConnections().length > playerLimitCache) || netServer.admins.isDosBlacklisted(ip)){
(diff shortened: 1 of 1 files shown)


ZetaMap:
mmm yes but this helps to know when your server(s) is being raided.

GeoNew2011:
maybe better to suppress them when there are more than 10 messages in one minute

Yaaty256:
I don't think straight up removing it is a good idea

BalaM314:
> maybe better to suppress them when there are more than 10 messages in one minute

this is a good idea

phinner:
Maybe lower the log level of those to debug. Because no log at all is just worse...

way-zer:
Perhaps it could aggregate once every minute.
There are % connections from % IPs: xxx (omitted %)

ZetaMap:
Yes but this need to cache a Ratekeeper, so it also need another thing to clean up this cache.

Anuken:
Resubmit if replaced with some sort of rate limit.

PR #11665, opened 2026-02-13
Translate:two new sector

IYAYII-hhgs:
the name and description of fallenVessel and perilousHarbor.


Anuken:
Do not delete them from global.properties; they should have no description. Do not write your own descriptions for sectors.

PR #11721, opened 2026-02-27
Save and restore accumulator on load

cardillan:
A recent commit (6ad8d06) saves the waited time for individual wait instructions. This PR adds the accumulator to the save file too, so that the accumulator will match the time waited even after map load, which is important for atomic execution of a block of code based on waits.

I haven't bumped the LogicBlock's revision number, so any map saved after 6ad8d06, but before this PR, would be invalid. Let me know if I should bump the revision number instead.

I couldn't test the PR, because the I get some build errors I'm struggling to fix. Looks like a problem connected to switching to Gradle 9. I'll try to test the feature using the GitHub build.

### DIFF
--- a/core/src/mindustry/world/blocks/logic/LogicBlock.java
+++ b/core/src/mindustry/world/blocks/logic/LogicBlock.java
@@ -764,6 +764,8 @@ public class LogicBlock extends Block{
                 write.s(waitIndices.get(i));
                 write.f(waitValues.get(i));
             }
+
+            write.f(accumulator);
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


cardillan:
I've tested the PR using the GitHub build and it works fine.

However, I was unable to produce the expected bug due to the unsaved accumulator before this PR, and I have no explanation for that. It's true that this needs a specific timing when saving the game, and perhaps I just wasn't lucky enough. I'll try to add some randomization to my test harness.

ETA: can't produce the expected bug even with randomization. Perhaps I'm misunderstanding something?

I'd try debugging, but I need to fix the build first.

PR #11739, opened 2026-03-06
The logic processor can use a sensor to obtain the "message" field of the message.

m1cxzfw3q:
image

image

image

I think this will be very useful for those who write logical processors.

### DIFF
--- a/core/src/mindustry/world/blocks/logic/MessageBlock.java
+++ b/core/src/mindustry/world/blocks/logic/MessageBlock.java
@@ -181,8 +181,15 @@ public class MessageBlock extends Block{
         public double sense(LAccess sensor){
             return switch(sensor){
                 case bufferSize -> message.length();
-                case config -> message.toString();
                 default -> super.sense(sensor);
+            };
+        }
+
+        @Override
+        public Object senseObject(LAccess sensor){
+            return switch(sensor){
+                case config -> message.toString();
+                default -> super.senseObject(sensor);
             };
         }
 


Kev-Vily:
hell nawh, i love this one

core/src/mindustry/world/blocks/logic/MessageBlock.java
```diff
@@ -185,6 +185,14 @@ public double sense(LAccess sensor){
             };
         }
 
+        @Override
+        public Object senseObject(LAccess sensor){
+            return switch(sensor){
+                case config -> message.toString();
```
camelStyleUser, line 191: CONTRIBUTING.md explicitly warns again memory allocation(which since `String` is immutable, `StringBuilder.toString()` is going to basically always allocate memory, unless the string can be found in the string pool iirc).
Since this memory allocation can be performed automatically from logic, it is probably even worse.
The final decision is up to Anuke, however I find it unlikely that the PR will be merged in the current state.

Anuken:
This is not acceptable, as it introduces a copy or mutable reference that can unexpectedly change. Read from the message block character-by-character if you need to instead.

m1cxzfw3q:
Alright.

core/src/mindustry/world/blocks/logic/MessageBlock.java
m1cxzfw3q: I will figure it out.

PR #11743, opened 2026-03-09
Add missing store localization for Brazilian Portuguese

DinoWattz:
Both Steam and Google Play seem to be currently using really old localizations and hopefully this PR addresses that.

_I know you put minecraft references in some of the steam achievements, you can't hide them from me_


DinoWattz (commented):
I fully approve of these changes!

Image

PR #6122, opened 2021-10-08
Getter method for `items`

MEEPofFaith:
I just wanna
image

If your pull request is **not** translation or serverlist-related, read the list of requirements below and check each box:

- [X] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [ ] I have ensured that my code compiles, if applicable.
- [ ] I have ensured that any new features in this PR function correctly in-game, if applicable.

PR #11768, opened 2026-03-20
Arc Changes and Adding PierceCap for Lightning

SomeonesShade:
##### Serpulo Balancing: Solo Send

**Info**
Part 2 of changes has been completely wiped from the face of earth, as I decided to move it to a different pr later down the line.

Pierce Cap will only work for Units.

---

**Implements 1 Suggestion Post:**

### [Arc Changes #6122](https://github.com/Anuken/Mindustry-Suggestions/issues/6122)
**_Summary Statement:_** Addressing the high performance to difficulty ratio of Arc, due to no item wiring requirement, by moving its performance from non-coolant to coolant (more usage depth). This allows Arc to also be used non-reactively.

#### Main Change
**Arc** **(-)**
**(-)** Damage: 20 -> 12
**(-)** Pierce Cap: inf -> 5
`The basis is mostly experimental, while [B1] Indicates similar range dps with silicon duo or hail, it's not the main restriction. It's a good starting nerf as it does control arc's performance. Water Coolant brings back some of that dps, and will be detailed on the coolant change. Also no more infinite pierce Arc! Should be easier to control and balance in the future.`

**(~)** Building Damage Multiplier: 0.25 -> 0.4
`20 * 0.75 is almost 12 * 0.4, so this is almost equalizing the single target damage. I'm not strict on this though and if there is a good reason to make this into 1x (12 dmg from 5 ish) because the pierce matters more.... sure.`

- **Coolant** **(+)**
**(+)** Coolant Boost: (120%, 145%) -> (180%, 280%)
`The main appeal [B2], non-water-cooled arc's dps is now almost the same as regular arc (170 ish) albiet with some damage dropoff from armor, at the cost of some space efficiency, and returns the cheap power consumption it had previously. It is a nerf though vs maces (B3), but the +14 damage from synergy can help counteract.`

### Community Feedback:

image

### Notes:

**_I'm currently on hiatus,_** if anyone hates or likes the pr or changes whatever (assuming if you think the changes being merged makes the game worse than what it is before, rather than not being the best solution but a good start), I'll reply on a later date or someone else does idk.

**_As always, check the suggestion post attached if you hate the changes before proceeding._**

Thx to Eggo for the code needed to make pierceCap work for Arc.


core/src/mindustry/entities/Lightning.java
```diff
@@ -78,6 +78,10 @@ private static void createLightningInternal(@Nullable Bullet hitter, BulletType
                 });
             }
 
+            if(hitter != null && hitter.type.pierceCap > 0 && hit.size >= hitter.type.pierceCap){
```
JasonP01: Are you sure this wont affect modded lightning (or any other in the game)?

SomeonesShade: theoretically, it shouldn't if the lightning has a pierceCap that isn't a valid number (from the 0 or below check). If they want the lightning to not pierce, it should be well no pierceCap


I only got this from eggo, as he sent me over the code for it, but I'll check if there is even a good reason to have a nonzero pierceCap before it was actually functional.

JasonP01: It shallnt let me close the comment so ig it stays

SomeonesShade: oh ae, ig ill do it then i think

SomeonesShade:
Out of all of the testing, I legitimately forgot to test vs buildings, give me a second...

SomeonesShade:
This has been extensive with me mentally going at it for a couple of hours lmao.

The only remaining cases to deal with is if an atrax is in top of walls, and a bullet toggle to have it jump over buildings (similar to as before) or use the new raycast (similar to how units are handled).

Once all of those are done, documentation for what changed will be shown, and retesting for any bugs... and the pr is fully functional.

On the actual balance changes, I would increase the water coolant even more, but need the functionality to have separate boosting amounts for cyro and water (yes, inconsistency, but better than cryo-Arc being absolutely broken.)

SomeonesShade:
With that, I think that's most of the potential bugs squashed

(why do i feel like theres a simpler way to do all of this, eh whatever)

I'm not adding more stuff like multi-hit detection for units nor a toggle if a multi-hit should count as one pierce consume or multiple. for sanity unless I wanna go at this even more.

I think a potential refactor is possible? but the code is functional
I'll start on the documentation and i think thats pr done

SomeonesShade:
image

There is one bug that I'm not sure I could really handle?, Basically if the unit somehow runs into 2 lightning bullets at once while moving, im just gonna throw the flag and no. no. okay no. this is already difficult to do proactive detection

SomeonesShade:
Okay I think I decided to do a reversion here. The rest of the stuff I did here is better as a different pr and I'll adress accordingly.

SomeonesShade:
The Largest Reversion I've done, I think thats it and Ill clean up the pr desc and stuff

core/src/mindustry/content/Blocks.java
```diff
@@ -3646,8 +3647,10 @@ Liquids.oil, new LiquidBulletType(Liquids.oil){{
             size = 1;
             health = 260;
             shootSound = Sounds.shootArc;
+            
             consumePower(3.3f);
             coolant = consumeCoolant(0.1f);
+            coolantMultiplier = 20f;
```
Anuken: Even with this value, I don't think anyone is going to water arcs. It's just so much easier to spam more, especially considering many don't have access to proper water distribution at this tier. I would go for 30, and even that might not be enough.

SomeonesShade: It is a start, there are some incentives into doing it (increased dps per tile, and reduced power costs)... and the values here are somewhat reserved for later adjustment. 

I wanted to see anyway first hand on how the community would react.

Anyway, thank you for commenting.

SomeonesShade:
Thank you Anuke.

PR #11769, opened 2026-03-20
Nova Buff

SomeonesShade:
##### Serpulo Balancing: Solo Send
**Implements 1 Suggestion Post:**

## Second Set of Changes
June 1, 2026

Preemptive maybe too high stats
- Nerfed Healing Rate down from 6x15 hp/s to 6x10 hp/s (sametypehalmult from 10% to 15% so no effect for other novas)
_Notation: targets x base heal rate_
- Reduced health from 240 to 200
- Nerfed DPS from 2x32.5 -> 2x26 
_Notation: pierce x base dps_

Overall, a nerf but it's to prevent nova from being completely broken on release for infiltration.

New Behavior Introduced from EggleEgg
- Overhealing is now gone as an issue! (Via smartHeal)
- RandomTimer is now removed due to this.
- Performance Increase for Healing especially under maxTargets
- Added smartHeal (only Nova has it so far)
- Added randDesync (Also nothing uses it)

#### Nova vs some Copper Duos

https://github.com/user-attachments/assets/35c83345-9a7b-4eab-b372-01383e8611b4

#### Nova Mace in Frontier (Skill Issue moment)

https://github.com/user-attachments/assets/47780af5-e1eb-471a-b889-460e6dec4f76

#### Foundation Number (16) In Salt Flats

https://github.com/user-attachments/assets/8be58709-fdcb-47a7-97e6-2c9dbfb889c5

### [Nova Buff #6130](https://github.com/Anuken/Mindustry-Suggestions/issues/6130)
**_Summary Statement:_** Due to Nova's high cost and titanium tech placement, it should be treated as T1.5 and gain buffs to emphasize its current design (a support unit on offense).

#### Main Change
**Nova** **(+)** 
**(+)** Health: 120 -> 240 (2x)
`I'd like Nova to be less squishy or volatile, and shouldn't be too bad? From testing it's so that they don't die that easily, well too easily.`
**(+)** Boost Multiplier: 1.5x -> 2x
`Makes it more fun to utilize, and fairly decent, if this buffs its homebase usage too much, may be adjusted post merge, as other aspects like its healing and build speed will be revamped in a more dedicated suggestion post/pr`

  - **Repair Field Ability** **(+)** 
  **(+)** Repair Rate: 10 hp / 4 sec -> 30 hp / 2 sec (2.5 hp/s -> 15 hp/s)
  `Very meaningful repair rate, this tita tech unit is now a better healer, have fun supporting your units.`
  **(+)** Heal Range: 7.5 tiles -> 12.5 tiles
  `Also to allow for easier and QOL usage, the range has been fairly increased to make it easier to use.`
  **(-)** Max Targets: inf -> 6
  `Don't worry on it being broken though, the amount of targets is only to 6 rather than infinite amount of them (requires PR for mechanic to be added), at roughly at a maximum of 90 health per second vs perfectly distributed AoE (big assumption!), and only 15 health per second against single targets, this isn't just a repair beam and requires depth.`
  **(-)** Same Type Heal Multiplier: 1 -> 0.1
  `Somewhat of a control as well, to prevent from nova only compositions (not good for a supporting unit), and to keep them as a high target, Novas heal way lower for its companions and the player must put in the APM (actions per minute, a measurement of effort), to actually prevent their Novas from dying. High Risk High Reward.`
  **(~)** Random Timer: 0 -> 1 second
  `Needs a pr to be implemented, but there is an issue with Nova activating its heal all at once, so to spread it out, a timer of +1 -1 is applied so that they don't end up synchronizing.`

  - **Weapon** **(+)** 
  **(+)** Pierce (Units and Buildings): 1x -> 2x
  `Have fun with this, it's to setup nova to present the nova tree as the piercing laser weapons, and be ready for survival, and making it somewhat distinct from dagger and makes it more dangerous.`

### Community Feedback:

image

### Notes:

Oh right uh, there isn't a display yet for random timer, I'll get back to it when once the following has ended....

**_I'm currently on hiatus,_** but wanted to get the changes out since I think people like this stuff. It's also pretty difficult to avoid this game when you see your feed recommendation and seeing how people enjoy the game. I think it reminded me why I go through hell for these changes sometimes as it feels meaningful. I do want players to have multiple options even if it's at the risk of being imbalanced initially for some fresh air.

**_As always, check the suggestion post attached if you want more details._**

If this ends up being broken, we can nerf it back (Like what I'm planning for graph salvo to be tentatively nerfed if testing proves adequate with a 1/1.1 x reduc)

https://github.com/user-attachments/assets/3a658b41-1f28-4c94-9d55-9cb50fe02bf0


SomeonesShade:
Okay I think I can add in the UI... just give me a sec...

SomeonesShade:
Okay so

randomTimer doesn't work due to desync issues (thx Lixie for pointing it out)
and now there is a chance maybe a new class is needed? since the repair part is now being changed from discrete to continuous. Which I think is deserving of something new.

The reason why this is necessary to prevent novas from stacking the heals at once to prevent weird gaps? This is potentially a non-problem in some persepectives, but it makes nova a bit too unstable

SomeonesShade:
Current plans:

Set same type heal mult to 100%, since the extra complexity isnt needed, and mass novas were nerfed anyway via the single target + they still are very weak and you do not want them their heals being redirected...

Next remove random thing and get eggo's changes merged here for better ai/control

Then yeah

SomeonesShade:
@EggleEgg pinging just incase if any quetsions and stuff on this

General gist i got is that the code is now performant, 
smartHeal and randomDesync are added (although I turned them off as the thing I really need is that)
basically no more overhealing issues

SomeonesShade:
Final stuff so far is final evaluations

Potentially sameTypeHealMult may be set to 1 in favor of reducing coreDatabase stuff? (just scream at suggestions or at serpulo-balancing if this makes nova somehow very broken)

Potential preemptive nerfing of the heal / second due to infiltration stage (where the unit is travelling internally) really giving the most value. as it can allow for some decent sustain

---

Edit: Hell No for removing sameTypeHealMult. I'm not encouraging solo nova infiltrations for what is meant to be for healing offensive pushes. Hard no

PR #10499, opened 2025-02-12
@printbuffer

1ue999:
This adds a reference to the print buffer to processors (`@printbuffer`), and also changes some behaviours to allow stringbuilders to be used in logic.

If your pull request is **not** translation or serverlist-related, read the list of requirements below and check each box:

- [x] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [x] I have ensured that my code compiles, if applicable.
- [x] I have ensured that any new features in this PR function correctly in-game, if applicable.

PR #11838, opened 2026-04-05
Allow reading links from a processor

cardillan:
This PR implements this feature request: https://github.com/Anuken/Mindustry-Suggestions/issues/6160 - i.e., reading links from other processors using the `read` instruction, as if the link name was a variable name. Works even for links that aren't referenced from the code of the processor containing the link. Useful for more complex schematics, where many processors need to access the same, large set of blocks (typically, memory banks or cells).

Edit: added the ability to access linked block by index.

### DIFF
--- a/core/src/mindustry/world/blocks/logic/LogicBlock.java
+++ b/core/src/mindustry/world/blocks/logic/LogicBlock.java
@@ -608,6 +608,9 @@ public class LogicBlock extends Block{
                 }
                 if(output.constant) return;
                 output.set(ret);
+            }else{
+                int index = position.numi();
+                output.setobj(index >= 0 && index < executor.links.length ? executor.links[index] : null);
             }
         }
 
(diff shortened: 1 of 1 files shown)


1ue999:
horray, another reason to procrastinate on finally making #10499 work :)
looks good!

1ue999:
how is reading `@links` going to work?
it might be easier to just make variable reading also allow constants, which would also include links... making `sense` work with it might be a huge pain...

cardillan:
I figured you don't necessarily need `@links`. You can go from zero and once you get back a `null`, you know you're at the end of the linked block list.

Edit again: it's not that easy. `@links` is a constant value, so it is not stored with the rest of variables inside the executor, which is where the variables for remote read/write are taken from. The entire list of variables and constants is only used while compiling the code and then discarded (the `@links` variable actually resides as an `LVar` instance in individual instructions).

I could either handle `@links` specifically, which is quite ugly, but local, or I'd have to store the full list of processor's variables _and_ constants somewhere, increasing the complexity. Also, this would only provide access to processor's constants, not global constants. Should global constants be accessible via `read`?

PR #11845, opened 2026-04-07
Add server max chat message length config

Yaaty256:
Allows servers to configure the max chat message length.


Anuken:
The idea itself is fine, but until I come up with a good system of syncing server-side configs like this, I'll shelve this. Doing it with Call for every single config type is not great design.

PR #11849, opened 2026-04-08
Ability to Duplicate objective blocks in the Editor

### LINKED ISSUE #5243: Ability to Duplicate objective blocks in the Editor
### Describe the content or mechanics you are proposing.

I have no idea how to put this straight, but I'll try.
Screenshot_2024-09-28-21-39-31-282_io anuke mindustry
When making something like this, you need to add each block individually. But it would be nice if we could copy the blocks so we don't have to place each blocks, then writing it's contents again.

### Describe how you think this content will improve the game. If you're proposing new content, mention how it may add more gameplay options or how it will fill a new niche.

Make Map making faster. Ability to make Animations with Objectives faster. And improve Modding Quality with less understanding of complicated languages like Java, instead they can spend 1 hours tinkering with Objectives.

### Before making this issue, check the boxes below to confirm that you have acknowledged them.

- [X] I have checked the [Trello](https://trello.com/b/aE2tcUwF/mindustry-trello) to make sure my suggestion isn't planned or implemented in a development version.
- [X] I am familiar with all the content already in the game or have glanced at the wiki to make sure my suggestion doesn't exist in the game yet.
- [X] I have read [README.md](https://github.com/Anuken/Mindustry-Suggestions/blob/master/README.md) to make sure my idea is not listed under the "A few things you shouldn't suggest" category.

Adleanogare:
Allow the user to copy paste an objective (and all of it's content) in the map editor. 
An error message is shown if there is no previously copied objective. 
For now and for simplicity reasons, the objective is copied on the same tile as the copied objective even if it was moved.
The current implentation tracks the position of the objective to copy it, meaning you can move around one block of code after copying it, doing a modification and then pasting the code with a modification without having to click each time on the copy button. (Do not worry the method used do not create reference issue, a new objective is created each time)

The current modification : 

image

If no copied objective : 

image

Resolves Anuken/Mindustry-Suggestions/issues/5243

---


Anuken:
On mobile, the paste button shouldn't show up at all if there is nothing to copy.

On desktop, it should use Ctrl+C and Ctrl+V instead of using a button (no buttons should appear at all.)

Adleanogare:
The change were tested on pc and tests were launched again.

In case of any error, it's now "Invalid copy."
The ctrl-c copy the objective under the mouse.
The ctrl-v paste is now only pasting if the position under the mouse is valid.

I am still struggling to test on android even with the guide, I will continue to try. Thus I am still unsure about the button placement for the copy.

Adleanogare:
Everything was tested both on pc and android. I am just unsure about the UI for mobile but I implemented it in a similar way than before with the same UI choice as for creating a new block for the visibility of the button as asked.

lu69159:
It works well on PC❤️. And I have a small suggestion: why not add a box selection copy function? It would allow copying objectives within a specific area, instead of copying everything or copying one by one.

lu69159:
> It works well on PC❤️. And I have a small suggestion: why not add a box selection copy function? It would allow copying objectives within a specific area, instead of copying everything or copying one by one.

Just like holding F in-game to copy blocks.

Adleanogare:
> > It works well on PC❤️. And I have a small suggestion: why not add a box selection copy function? It would allow copying objectives within a specific area, instead of copying everything or copying one by one.
> 
> Just like holding F in-game to copy blocks.

I will check what I can do. It will take a bit of time to get the UI right after but on pc I think it can be done and I have no idea how to adapt it for the mobile version.

Anuken:
I had to remove mobile support, as it was *incredibly* broken. As soon as I started merging I ran into very serious code issues with this PR (bizarre clipboard handling, improper paste tracking, duplicates at the same position, unclear error messages...) so most of it had to be thrown out.

PR #11851, opened 2026-04-08
Improve Korean localization wording and localize hardcoded UI text

bluehyena:
## Summary
- improve awkward Korean wording, particles, and spacing in bundle_ko.properties
- translate previously untranslated Korean bundle entries
- move several hardcoded UI strings into bundle keys so they can be localized
- add the corresponding English bundle keys and Korean translations for those UI strings

## Verification
- ./gradlew.bat core:compileJava

Closes #11850


core/src/mindustry/editor/MapEditorDialog.java
```diff
@@ -336,7 +336,7 @@ private void editInGame(){
             state.rules.sector = null;
             state.rules.fog = false;
             state.map = new Map(StringMap.of(
-                "name", "Editor Playtesting",
+                "name", Core.bundle.get("editor.playtesting"),
```
Anuken: This shouldn't be edited, it should not depend on locale.

core/src/mindustry/editor/MapEditorDialog.java
```diff
@@ -188,14 +188,14 @@ public MapEditorDialog(){
         //this is gated behind a property, because it's (1) not useful to most people, (2) confusing and (3) may crash or otherwise bug out
         if(OS.hasProp("mindustry.editor.simulate.button")){
```
Anuken, line 189: This dialog is not visible to players and may be reworked in the future, it should not be localized.

core/src/mindustry/ui/dialogs/PausedDialog.java
```diff
@@ -63,7 +63,7 @@ void rebuild(){
             boolean showObjective = state.rules.sector != null && state.rules.sector.preset != null && state.rules.sector.preset.description != null;
 
             if(showObjective){
-                cont.button("@objective", Icon.info, () -> ui.fullText.show("@objective", state.rules.sector != null && state.rules.sector.preset != null ? state.rules.sector.preset.description : "oh dear")).padTop(-60f);
+                cont.button("@objective", Icon.info, () -> ui.fullText.show("@objective", state.rules.sector != null && state.rules.sector.preset != null ? state.rules.sector.preset.description : Core.bundle.get("objective.missing"))).padTop(-60f);
```
Anuken: This is an error message that should never show up; if it does, it's a bug. Don't localize it.

Anuken (changes requested):

bluehyena:
Thank you for the review!

Addressed the review points

reverted localization for the editor-only parts.

PR #11867, opened 2026-04-10
Fix Inconsistency between Mobile and Desktop for configuring messages

1ue999:
This fixes a small inconsistency between desktop and mobile, and so i went with the desktop behavior due to it allowing to configure a message without changing any text (for example with a proc constantly changing the text) and still having the config be done if there is a mismatch

Unfortunately, i didn't figure out how to compile for mobile (and sadly Input::getTextInput doesnt work on desktop) so i cannot check the last check (compiles for desktop which is my reason to check the 2nd check)


JasonP01:
Follow the android sdk steps, then you shall have to send the apk to your phone

1ue999:
i followed the steps...and couldnt figure out why it doesnt build for mobile some part of the way there

PR #11906, opened 2026-04-16
Translate bundle_ja.properties

wancor1:
I translated the contents of V8!
I checked it, so it should be correct.


Anuken:
How does this compare to #11907?

wancor1:
There are discrepancies in nuance, rough translation, and even completely incorrect proper nouns.
This version also has some parts that were lost from the original text. (Currently being revised)

wancor1 (commented):
done
Anyway, this one is better.

PR #11924, opened 2026-04-17
Fix formatting in RU bundle

vovan2055:
Fixed a malformed string in ru bundle


vovan2055 (commented):
Translated more strings

PR #12036, opened 2026-05-04
A Controversial Problem

LixieWulf:
It's been three, four years of this mechanic.
The wrecking mechanic has been difficult to work with, in-game and trying to balance stats. Air units have become the unquestioned meta of destroying enemies on Serpulo, with the only other option being sieges. It turns attack into a "throw at enemy and win" breeze, and entirely overshadows many other combat options or unit gameplay design.
A small squad of zeniths, or two quads are capable of instantly destroying a nucleus core when filled with blast compound. Averts are capable of diving disperses for heavy damage. Octs are used in extremes solely because their durability can get their nuclear item payload through enemy fire. The issue extends to both planets, with varying effects on the game.

This PR entirely removes crash damage and item explosiveness from all units. Extreme, yes, but it might as well be tried. So far the suggested other methods involve decently complex multipliers that attempt to balance wrecking. Personally, I have my doubts on how well it even potentially could be balanced.

I love the wrecking aesthetic - it's a great visual add, and why I only targeted damage dealt here. This PR also adds the ability to modify item effects for datapatch and modding uses. If this makes it into the game, maps and servers can revert it via creating a data patch. Also, these multipliers can be easily changed for balancing wrecks properly, if it becomes viable to do.

I suspect there may be... a spicy response to this PR. I did attempt to mitigate it with the added fields, but this is inherently out of sheer exhaustion. I'm going to post it and wait for close or merged status.

### DIFF
--- a/core/src/mindustry/type/UnitType.java
+++ b/core/src/mindustry/type/UnitType.java
@@ -97,7 +97,7 @@ public class UnitType extends UnlockableContent implements Senseable{
     /** radius for circleTarget, if true */
     circleTargetRadius = 80f,
     /** multiplier for damage this (flying) unit deals when crashing on enemy things */
-    crashDamageMultiplier = 1f,
+    crashDamageMultiplier = 0f,
     /** multiplier for health that this flying unit has for its wreck, based on its max health. */
     wreckHealthMultiplier = 0.25f,
     /** a VERY ROUGH estimate of unit DPS; initialized in init() */
@@ -376,6 +376,8 @@ public class UnitType extends UnlockableContent implements Senseable{
 
     /** amount of items this unit can carry; <0 to determine based on hitSize. */
     public int itemCapacity = -1;
+    /** multiplier for explosion damage dealt by carried items when killed */
+    public float itemExplosiveness = 0.001f;
     /** @deprecated only kept for compatibility with some turrets that query this field! Remove this from your code immediately! */
     @Deprecated
     public int ammoCapacity = 1;
--- a/core/src/mindustry/type/UnitType.java
+++ b/core/src/mindustry/type/UnitType.java
@@ -376,8 +376,8 @@ public class UnitType extends UnlockableContent implements Senseable{
 
     /** amount of items this unit can carry; <0 to determine based on hitSize. */
     public int itemCapacity = -1;
-    /** multiplier for explosion damage dealt by carried items when killed */
-    public float itemExplosiveness = 0.001f;
+    /** multiplier for effects dealt by carried items when killed */
+    public float itemExplosiveness = 0.001f, itemFlammability = 1f, itemCharge = 0.5f;
     /** @deprecated only kept for compatibility with some turrets that query this field! Remove this from your code immediately! */
     @Deprecated
     public int ammoCapacity = 1;
--- a/core/src/mindustry/entities/comp/UnitComp.java
+++ b/core/src/mindustry/entities/comp/UnitComp.java
@@ -860,8 +860,8 @@ abstract class UnitComp implements Healthc, Physicsc, Hitboxc, Statusc, Teamc, I
         if(!isAdded() || !killable()) return;
 
         float explosiveness = (2f + item().explosiveness * stack().amount * 1.53f) * type.itemExplosiveness;
-        float flammability = item().flammability * stack().amount / 1.9f;
-        float power = item().charge * Mathf.pow(stack().amount, 1.11f) * 160f;
+        float flammability = (item().flammability * stack().amount / 1.9f) * type.itemFlammability;
+        float power = (item().charge * Mathf.pow(stack().amount, 1.11f) * 160f) * type.itemCharge;
 
         if(!spawnedByCore){
             Damage.dynamicExplosion(x, y, flammability, explosiveness, power, (bounds() + type.legLength/1.7f) / 2f, state.rules.damageExplosions && state.rules.unitCrashDamage(team) > 0, item().flammability > 1, team, type.deathExplosionEffect, 0f);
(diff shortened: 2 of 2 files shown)


coaldeficit:
this ***NEEDS*** to get merged regardless of community reaction. item bombing is an overpowered exploit that should have never existed to begin with, no matter what the people who rely on it think

SomeonesShade:
Only comment for me is I wish there is some damage, but very insignicant surge wall lightning levels (crashMult = 0.001 to 0.0001 type levels).


Edit:
okay from sepulo balancing a bit.

Lixie's current intentions is to just start fresh and rebuff it for certain units if needed. And thats probably enough for me to support it.

Xandertron:
ah yes the undying urge to balance a sandbox game

coaldeficit:
> ah yes the undying urge to balance a sandbox game

because the presence of a sandbox mode for testing and messing around purposes clearly means the game is a 100% sandbox game and has no balance issues whatsoever

SomeonesShade:
> ah yes the undying urge to balance a sandbox game

umm...
do you not want like slag wave to be better, maybe ripple a bit better, something a bit weaker? Maybe Pyra Swarmer being fun to use or maybe full sends of T3-T4 a bit better than siege tactics...

The main point of balance changes is to make options more meaningful. It's really not just a sandbox game

It's not like mindustry is just minecraft creative mode (hell even in the sandbox game minecraft, elytras are overdominant compared to rails, the enchanment system and anvil/repair is a bit outdated etc.).

Inevitably, whenever you have choices, or experiences to manage, balancing is the heart of making sure it works.

From the overal game level with how content is balanced in relation to each other, to the map level with wave design and base placements trying to create unique, interesting, and meaningful experiences.

jehosula:
> ah yes the undying urge to balance a sandbox game

in all my dealing with bad arguments against balance changes
i have never seen an argument THIS bad, this argues against EVERY balance change ever, 
AND this completely misunderstands what sandbox label really means, it means the player has lots of freedom in what they chose to do, 
not that the game is minecraft creative mode
for freedom of choice, it's better to have every choice be meaningful and good, rather than the feeling you are rewarded for using only specific options rather than how you used them or punished for using anything else, not because you used it at the wrong time, but because you used it at all

50275:
I support removing crash damage. I do not, however, support making the numbers specific for each unit. Please make `type.itemExplosiveness`, `type.itemFlammability` constants, ideally 0. Also, couldn't you have named this PR something more descriptive, like "Remove Crash Damage"?

jehosula:
> I support removing crash damage. I do not, however, support making the numbers specific for each unit. Please make `type.itemExplosiveness`, `type.itemFlammability` constants, ideally 0. Also, couldn't you have named this PR something more descriptive, like "Remove Crash Damage"?

the idea is so that data patch can bring it back for the units the players want
but ye the PR should've been named that

Icexuegao:
Obviously, this is a breaking change and won't be accepted unless it's targeting the next major version.

RE-AST:
> this _**NEEDS**_ to get merged regardless of community reaction. item bombing is an overpowered exploit that should have never existed to begin with, no matter what the people who rely on it think

No, players still humans, also would like to say, you do not want to wast lot of time on pvp/attack because of disabled bombing and crash damage, too radically, also you already can nerf bombing with patches, right point in pr but not right solution

core/src/mindustry/type/UnitType.java
```diff
@@ -376,6 +376,8 @@ public class UnitType extends UnlockableContent implements Senseable{
 
     /** amount of items this unit can carry; <0 to determine based on hitSize. */
     public int itemCapacity = -1;
+    /** multiplier for effects dealt by carried items when killed */
+    public float itemExplosiveness = 0.001f, itemFlammability = 1f, itemCharge = 0.5f;
```
BalaM314, line 380: Why is the default item explosiveness 0.001 (or 0.1%, one thousandth) for all units? This makes it annoying to get any significant amount of item bombing damage back with content patches as you have to patch every single unit type. This should default to something more reasonable like 0.75 or 1. The PR already disables item explosiveness by default (by crashdamagemult=0). This only frustrates users that changed settings to explicitly enable crash damage.

BalaM314 (changes requested):

Anuken:
Sorry, this isn't happening. Removing *all* damage is not an acceptable solution.

SomeonesShade:
That is completely fair, means if any adjustment has to happen, it's with tested stats now...

1092384:
here are some of my thoughts on crash damage/item bombing 

right now it seems that air unit crash damage kinda works like crawler where the whole point is to suicide the unit. wouldn't it be possible then to theoretically balance this mechanic, assuming crawlers are also not an "unsavable mechanic that needs to be removed"? it's also not that suicide units (crawler) cannot also be standard attacking units, as other hybrid units such as vela being really good at both offense and support can exist without significant balancing issues; and having more options for attacking bases isn't a bad thing.

when compared with crawlers, air units have significantly more vitaility and explosive power when used for suicide strategies. since reducing vitality probably isn't the right solution, the only option is to reduce explosive power. here's where i don't understand, why is it "too complex" to just implement a few multipliers that significantly reduce the amount of explosive power that crash damage/item bombing outputs? many complex solutions were implemented recently to address issues arguably more niche compared to crash damage: armor weakness were introduced to balance turrets such as lancer and copper ammo salvo, unit healing caps for support units such as nova were also implemented, and item deposit cooldown was added to some turrets. i genuinely don't think that all crash damage/item bombing damage needs to be removed, and even a simple 0.05x multiplier would lessen the issue significantly (don't comment on that specific part please i have not tested 0.05x at all) while hopefully still making it a potential strategy.

PR #12061, opened 2026-05-11
Add code: Localize key names in keyboard setting window

redloong9527:

#### Hello, this is my first time creating a pull request for source code.
I wrote code for localize the key names in the keyboard key mapping window, and then added relevant translations to all the bundles (of course, I do not know so many languages, most are translated by AI, so I kindly ask native speakers to help me check them).
I have independently compiled and verified the feasibility of the code, and ensured that it is free of bugs.

_My English isn't very good, so it might sound a bit awkward._


Yaaty256:
Why did you translate all the bundles? They're incorrect, at least in my language. You should've just left them unchanged, translators can translate them themselves later.

Anuken:
These should not be translated.

redloong9527:
> These should not be translated.

Why? If it's something like tab, shift, or Ctrl that's unnecessary, then there are words like the middle mouse button, which should be fine, right?

redloong9527:
> Why did you translate all the bundles? They're incorrect, at least in my language. You should've just left them unchanged, translators can translate them themselves later.

I added a translation API to this pull request; adding translations to the bundle was just a side effect. If you're not satisfied, I can delete them. However, Anuke directly closed this pull request, which I really don't understand.

PR #12063, opened 2026-05-12
decrease disperse's oxide build cost from 150 to 50

spentcud:
disperse is already weak enough imo, it does _not_ need to cost over 100 oxide
might increase to 75 if 50 is too little


1092384:
as mystery player i agree

jehosula:
i was initially surprised to find out that it was 150 oxide and not 50, it should be 50

PR #12088, opened 2026-05-17
Serpulo Non-Artillery Scaled AoE Rework

Efsfse2280:
Currently;
Area Of Effect (splash) damage values of non-artillery serpulo turrets and units don't display their true values in the database info. This discrepancy is misleading the player into overestimating the potential ceiling of those turrets.

The non-scaled damage formula of AoE takes unit `HitSize/2` into consideration; 
- It's making bigger units gain ludicrous amounts of dmg reduction and even immunity
- It's reducing _everything_, meaning that the database AoE numbers are... 0% accurate. Yep. The 75 AoE dmg of surge cyclone never occurs.

### Example 
Swarmer with pyratite ammo, for example, returns 0 splash dmg to Antumbra because `ammo radius (20)<HitSize of Antumbra/2 (23)`

This rework uses the `scaledSplashDamage` field to fix this issue; instead of using the unit's **center point**, it uses the unit's **edge** when calculating damage; this means...

- a significant accuracy increase over non-scaled damage.
- Less falloff from distance with nearby units, which greatly increase its crowd control ability.

Artillery isn't counted here, because unit center calculations are actually representative of damage thresholds. (shell landing on edge of unit shouldn't do full dmg)

### Example
One pyratite Salvo bullet (25/15) will do `(16/6)` instead of (`16/0)` point blank damage against Antumbra.

### Proposed solutions;

Scatter -  Equal effective DPS AoE adjustments. (reference; Flare)*
Plast tier Turrets - Overall Equal DPS AoE adjustments. (reference; Antumbra)
Pyra Salvo - Equal effective DPS AoE adjustments. (reference; Zenith)
Pyra Spectre - Unchanged (means a buff)

Minke - Equal effective DPS AoE adjustments. (reference; Flare)
Everything else non-artillery has been unchanged. (means a buff)

#### Before;

https://github.com/user-attachments/assets/138facab-b8ec-4f85-9a40-0ff8a5c7a96b

#### After;

https://github.com/user-attachments/assets/76f62479-751a-45e1-a9c4-270595657a30

ammo
flare

*: Air dynamics are broken. Effectiveness of generalist turrets is rock-bottom due to their absurd speed values, which makes scatter, the dedicated AA, super difficult to balance with scaled AoE and T4+ air. The damage nullification from current aoe calculations served as a band-aid, so removing it required some... _counter-measures..._

Flare/Horizon had a -50% vitality decrease in the form of -33% speed and -25% HP. Poly, 50% HP.
Zenith, being in a transition phase and less overpowered from speed, only had ~30% speed and ~20% HP nerfs.

With this rework, splash damage in the database will become way more reliable while also making it more significant against crowds. 
_Another alternative would be to literally add the damage formula in the ammo/weapon database stats, but it would add unnecessary complexity in the game and code._


jehosula:
i wanna be clear on armor weakness being a temporary thing that would be removed next PR yes (with rebalancing)?

Efsfse2280:
I hope so
Air dynamics is a whole can of worms on its own

SomeonesShade:
PR is well made, proud of it.

Only thing missing is video before-after if the stats are finalized, or maybe a ttk table idk, any documentation will do.

Efsfse2280:
> Only thing missing is video before-after if the stats are finalized, or maybe a ttk table idk, any documentation will do.

Added a rudimentary before-after on scatter to show ttk is equal with flare

Anuken:
Sorry, no. I agree that splash damage numbers should be fixed, but making the only 'fast' units in the game almost as slow as everything else is not acceptable. It looks incredibly lame and destroys their feel completely.

PR #12128, opened 2026-05-26
The number of each row of the item display UI is increased

no9527dada:
The number of each row of the item display UI is increased
With more and more items in the game and the increase of MOD.
The original display mode can no longer be satisfied
The mobile device can also use the top resource display (landscape screen only)


Anuken:
No, this looks terrible.

PR #12147, opened 2026-05-29
update bundle id (final) 

Kev-Vily:
:speaking_head:


Kev-Vily:
i notice strings lines changes on that one branch, I'll wait for it, so this can be merge without conflict

PR #12152, opened 2026-05-30
drawNoise/drawNoiseLayers as an effect so that json can use them

EggleEgg:
**Example with a datapatch:**

https://github.com/user-attachments/assets/b29e9e9e-ca1a-4c01-b990-c68a1d52b868


Kev-Vily:
YO, I'VE AN IDEA

core/src/mindustry/entities/effect/NoiseEffect.java
```diff
@@ -0,0 +1,65 @@
...
+    public float windY = 0f;
+    public int layers = 4;
+    public float layerSpeedMul = -1.3f;
+    public float layerAlphaMul = 0.7f;
+    public float layerSclMul = 0.8f;
+    public float layerColorMul = 0.9f;
+
+    @Override
+    public void render(EffectContainer e){
+        Texture tex = Core.assets.get(noisePath, Texture.class);
```
Anuken: If you look at the source for `get`, you'll see that libGDX did something really funny by making it **crash instead of return null** when the file isn't found. This is obviously a bad idea for effects. I'm adding a `getOrNull` method to fix this, but this PR will have to wait until that's merged in.

Also, it should be cached so that it only needs to call `getOrNull` the first time instead of calling it every render.

core/src/mindustry/entities/effect/NoiseEffect.java
```diff
@@ -0,0 +1,64 @@
...
+    public float baseSpeed = 0.4f;
+    public float intensity = 1f;
+    public float windX = 1f;
+    public float windY = 0f;
+    public int layers = 4;
+    public float layerSpeedMul = -1.3f;
+    public float layerAlphaMul = 0.7f;
+    public float layerSclMul = 0.8f;
+    public float layerColorMul = 0.9f;
+    public Texture tex = Core.assets.getOrNull(noisePath, Texture.class);
```
Anuken: This is run at constructor time, which means that reassigning noisePath doesn't do anything. This may also cause issues on servers. How about `if(tex == null) tex = Core.assets.getOrNull(noisePath, Texture.class);` in render() instead?

PR #12153, opened 2026-05-30
Make all research prioritize resources outside of the currently active sector

EggleEgg:
Implements
- https://github.com/Anuken/Mindustry-Suggestions/issues/6336

https://github.com/user-attachments/assets/842add71-cd3f-478e-aba6-aca57d692840

### DIFF
--- a/core/src/mindustry/ui/dialogs/ResearchDialog.java
+++ b/core/src/mindustry/ui/dialogs/ResearchDialog.java
@@ -26,7 +26,6 @@ import mindustry.input.*;
 import mindustry.type.*;
 import mindustry.ui.*;
 import mindustry.ui.layout.*;
-import mindustry.ui.layout.BranchTreeLayout.TreeLocation;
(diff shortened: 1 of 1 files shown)


core/src/mindustry/ui/dialogs/ResearchDialog.java
```diff
@@ -269,6 +271,19 @@ public void add(Item item, int amount){
                         counter[0] -= toRemove;
                     });
 
+                    //only consume from the active sector last
+                    if(counter[0] > 0 && curSector != null){
+                        var seq = cache.get(curSector);
+                        if(seq != null){
+                            int toRemove = Math.min((int)Math.ceil(percentage * seq.get(item)), counter[0]);
```
Anuken: I don't think having a percentage here makes sense. If you've taken from every sector *and you still need to take more*, that means that the percentage you needed to take was 100%, so this calculation can be simplified, no?

PR #12169, opened 2026-06-01
Extend `TypeIO` with a Customizable Type Interface for Efficient Sync of Mod Data Classes

woodjecket:
## Motivation

Currently `TypeIO` only supports limited built‑in types (primitives, `IntSeq`, `Content` – the latter being a utility for vanilla block configuration). For a mod’s custom data class (e.g.complex config object), the developer is forced to fall back to ambiguous `IntSeq` or `byte[]`.

## Changes Overview

Introduce a new interface `CustomData`. Any mod custom data class can implement `CustomData`, provide its own serialization/deserialization logic, and then be handled by `TypeIO` just like other types.

## Usage

``` java
package mindustry.io;

import arc.util.io.*;

public class CustomDataExample implements CustomData{
    public int x;
    public int y;

    public CustomDataExample set(int x, int y){
        this.x = x;
        this.y = y;
        return this;
    }

    static{
        CustomData.register(CustomDataExample.class, reads -> new CustomDataExample().set(reads.i(), reads.i()));
    }

    @Override
    public void dataWrite(Writes write){
        write.i(x);
        write.i(y);
    }
}

```

Then it can be simply used in building configuration:

``` java
config(CustomData.class, (Building build, CustomDataExample data -> {
    //...
});
```

Then it can be synced correctly.


Moreperbullet:
is it too complicated to be added?

woodjecket:
> is it too complicated to be added?

As is showed in the source code, `TypeIO` is used in two occasions, sending packets about tile configuration and storing build's config in the schematics. I had just thought of the first usage so I have made a big mistake.

After comtemplating the idea, I have found that there must be a compromise between compability, stability and save size:

- At first, I want to allow modders to assign the id themselves, but this will have a great risk to break compabilities between mods.
- After that, I had thought that the `writeObject()` is only used in sync, which means there is no chance that the registration will changed, so I made it as is presented now.
- Later, I found that if keeping the current design, schematics will simply crashed when adding/removing mod or updating registration sequence, which is more unacceptable, so I closed the pull.

However, a friend of mine suggested me writing the fully qualified name instead of id to prevent corruption or unstability, (As for renamning a custom data class, we can expose a method that allows register with a `String` rather than `Class`), which will extend the id from 4 bytes to 40-80 bytes. Considering that there will not be a lot of the custom data, I wonder if this is accepted?

Maybe it will look like this:
``` java
package mindustry.io;

import arc.util.io.*;

public class CustomDataExample implements CustomData{
    public int x;
    public int y;

    public CustomDataExample set(int x, int y){
        this.x = x;
        this.y = y;
        return this;
    }

    static{
        CustomData.register("mindustry.io.CustomDataExample", reads -> new CustomDataExample().set(reads.i(), reads.i()));
    }

    @Override
    public void dataWrite(Writes write){
        write.i(x);
        write.i(y);
    }
}

```

PR #12180, opened 2026-06-03
Distributor buff/rework + db entry qol

EggleEgg:
In its current state distributors are near useless. Not only they share the **exact same properties and speed as routers** (_unchained = throughput same as fps, chained = 7/s_) but their bigger size and higher cost often makes them a terrible choice; and specially compared to proper distribution blocks such as gates and sorters. 
Why bother wasting space when you can have infinite throughput with gates?

.

Now distributors act somewhat to a 2x2 gate: they cannot be chained but share many of the benefits of them:

image

.

This allows for interesting implementations and designs where they can aid gates, unloaders and sorters, rather than being a waste of space:

image

.

### Db entries added/modified
**Gates and sorters**
image
**Unloader**
image

.

image
image
image
image
image
image
image
image
image
image

.

And a release for testing
https://github.com/EggleEgg/Mindustry/releases/tag/2342r

.


jehosula:
this does hurt exactly one use though
distributor chain in pvp for earlier first item

spentcud:
mostly unrelated but does this change the gates and sorter's entries too

SomeonesShade:
Theres a concern from coal deficit and I mostly summarized it for here to send... maybe something to address

image

MinDoUser:
Why do you call the bundle entry <code>bar.unabletoclog</code> ? This is not used for any bar at all… Also, adding „None.“ before „Unable to clog“ seems a bit wierd too…
What about instead of saying:
Item Capacity: None. Unable to clog.
do something like this:
Item Capacity: 0
Unable to clog
basically making „Unable to clog“ its own stat?

I like the general idea of some router that can not clog tho

Shawak:
Are these interesting implementations in a room with us

Anuken:
I'm concerned that many (bad, admittedly) bases with chained distributors will suddenly stop working for no clear reason. The kinds of people who chain these aren't likely to read descriptions.

SomeonesShade:
I think the idea on distributors being able to hold multiple item types might be best here....

As the throttle for chaining still exists to limit it, while making it a bit more useful

EggleEgg:
Reverted distributor field changes for now. The db entry and buildcomp qol are very useful however. Currently distributor works as expected from my testing

faretek1:
Distributors are super useless as of right now, which is sad. These changes would be interesting 

> I'm concerned that many (bad, admittedly) bases with chained distributors will suddenly stop working for no clear reason. The kinds of people who chain these aren't likely to read descriptions.

Legacy distributor & new distributor?

EggleEgg:
> I'm concerned that many (bad, admittedly) bases with chained distributors will suddenly stop working for no clear reason. The kinds of people who chain these aren't likely to read descriptions.

Currently considering dynamic capacity changes on proximity updates (+ avoid voiding items when chained). Maybe for the future

PR #12184, opened 2026-06-05
faster core proximity updates

WMF-Industries:
Removes the need to iterate all of the team's cores upon one updating its proximity
_Should_ work just fine, I've tested all the cases that came to my mind (adding storage, removing storage, cores and storage dying, changing teams, etc)

### DIFF
--- a/core/src/mindustry/core/Logic.java
+++ b/core/src/mindustry/core/Logic.java
@@ -282,7 +282,7 @@ public class Logic implements ApplicationListener{
 
                     for(ItemStack stack : state.rules.loadout){
                         //make sure to cap storage
-                        entity.items.add(stack.item, Math.min(stack.amount, entity.storageCapacity - entity.items.get(stack.item)));
+                        entity.items.add(stack.item, Math.min(stack.amount, entity.capacity() - entity.items.get(stack.item)));
                     }
                 }
             }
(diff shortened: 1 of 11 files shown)


Abdullah-Ajeebi (approved):
I tested the following changes and the lag is little to none.

WMF-Industries:
found a few more edge cases that i have to fix before this can be merged

WMF-Industries:
i did quite a bit more testing, seems to be working correctly with only a minor bug in the in-game editor, if you enter it for the first time, everything's fine, but if you quit and re-enter, the item capacity for every team is 0 (editor only)
i'll try to get that fixed soon

WMF-Industries:
okay, should be all good now

Anuken:
What does this fix?

WMF-Industries:
in most scenarios, pretty much nothing

however if someone plays a map that either lets you place or has a lot of cores placed around (by a lot i mean a **lot**), and one of them happens to be destroyed, the game will no longer freeze for 37 minutes trying to update them all

I've tested it mostly against a [map](https://discord.com/channels/391020510269669376/416719902641225732/1512349625623777495) with unrealistic amounts of cores, though there are a few custom gamemodes and maps which can replicate this scenario while being something people may actually play

Anuken:
...Core storage code has historically been very buggy, and this makes it *even more* fragile by maintaining the count instead of scanning each time. Unless this comes with extensive unit tests covering ever conceivable scenario (multiple touching cores have a container placed next to them, container dropped via payload, container picked up via payload, container destroyed, etc...), I am not going to risk it.

PR #12191, opened 2026-06-07
Sensing heat

cardillan:
My attempt at implementing https://github.com/Anuken/Mindustry-Suggestions/issues/6353.

As there isn't a `heat` field common to all blocks, I've added the ability to sense `@heat` to blocks that define the field, returning the value of the field as-is. Apparently, different blocks use different ranges for heat, so maybe adding a `@heatCapacity` or `@maxHeat` property might be useful, but I didn't want to go there as I don't understand the mechanism all that much. It should still be useful for map authors using world processors as it is.

### DIFF
--- a/core/src/mindustry/world/blocks/heat/HeatConductor.java
+++ b/core/src/mindustry/world/blocks/heat/HeatConductor.java
@@ -8,6 +8,7 @@ import mindustry.*;
 import mindustry.entities.units.*;
 import mindustry.gen.*;
 import mindustry.graphics.*;
+import mindustry.logic.*;
 import mindustry.ui.*;
 import mindustry.world.*;
 import mindustry.world.draw.*;
@@ -86,6 +87,12 @@ public class HeatConductor extends Block{
 
             lastHeatUpdate = Vars.state.updateId;
             heat = calculateHeat(sideHeat, cameFrom);
+        }
+
+        @Override
+        public double sense(LAccess sensor){
+            if (sensor == LAccess.heat) return heat;
+            return super.sense(sensor);
         }
 
         @Override
--- a/core/src/mindustry/world/blocks/heat/HeatConductor.java
+++ b/core/src/mindustry/world/blocks/heat/HeatConductor.java
@@ -91,7 +91,7 @@ public class HeatConductor extends Block{
 
         @Override
         public double sense(LAccess sensor){
-            if (sensor == LAccess.heat) return heat;
+            if(sensor == LAccess.heat) return heat;
             return super.sense(sensor);
         }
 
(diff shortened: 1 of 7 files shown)


EggleEgg:
how does this interact with nuclear reactor heat? it is primordial it does not reach values >1 to not break serpulo logic

cardillan:
> how does this interact with nuclear reactor heat? it is primordial it does not reach values >1 to not break serpulo logic

It doesn't change the `heat` values at all anywhere, I've only added the ability to sense them.

EggleEgg:
I see now, HeatBlock is unchanged (nuclear reactor implements it) so sensing is (or should be) exactly the same. My concern was sensor heat returning values >1, as heatFrac()/heatOutput can reach up to 15. This would ve broken quite a lot of basic serpulo logic

core/src/mindustry/world/blocks/power/VariableReactor.java
```diff
@@ -151,6 +152,12 @@ public float heatRequirement(){
             return maxHeat;
         }
 
+        @Override
+        public double sense(LAccess sensor){
+            if (sensor == LAccess.heat) return heat;
```
Anuken:
```suggestion
            if(sensor == LAccess.heat) return heat;
```

core/src/mindustry/world/blocks/heat/HeatConductor.java
```diff
@@ -88,6 +89,12 @@ public void updateHeat(){
             heat = calculateHeat(sideHeat, cameFrom);
         }
 
+        @Override
+        public double sense(LAccess sensor){
+            if (sensor == LAccess.heat) return heat;
```
Anuken:
```suggestion
            if(sensor == LAccess.heat) return heat;
```

core/src/mindustry/world/blocks/heat/HeatProducer.java
```diff
@@ -60,6 +61,12 @@ public float heat(){
             return heat;
         }
 
+        @Override
+        public double sense(LAccess sensor){
+            if (sensor == LAccess.heat) return heat;
```
Anuken:
```suggestion
            if(sensor == LAccess.heat) return heat;
```

core/src/mindustry/world/blocks/power/HeaterGenerator.java
```diff
@@ -62,6 +63,12 @@ public float heat(){
             return heat;
         }
 
+        @Override
+        public double sense(LAccess sensor){
+            if (sensor == LAccess.heat) return heat;
```
Anuken:
```suggestion
            if(sensor == LAccess.heat) return heat;
```

core/src/mindustry/world/blocks/production/HeatCrafter.java
```diff
@@ -71,6 +72,12 @@ public float warmupTarget(){
             return Mathf.clamp(heat / heatRequirement);
         }
 
+        @Override
+        public double sense(LAccess sensor){
+            if (sensor == LAccess.heat) return heat;
```
Anuken:
```suggestion
            if(sensor == LAccess.heat) return heat;
```

PR #12206, opened 2026-06-11
scale heat with timescale and allow overdrive on its producers + consumers

EggleEgg:
Warning: read and write heatOutScaled are needed to avoid differences between heat and heatOutput on load (visual flashing), so no backporting

image

image

https://github.com/user-attachments/assets/049c34f7-851e-4256-8400-02ea3f63e10a

.

image

.


core/src/mindustry/world/blocks/heat/HeatBlock.java
```diff
@@ -1,8 +1,20 @@
...
+
+    /** @return heatOutput scaled by timeScale */
+    static float approachHeatOutput(float heatOutput, boolean scaleHeat, float timeScale){
+        return heatOutput * (scaleHeat ? timeScale : 1f);
+    }
+
+    /** Smoothly approaches heat at a rate regardless of efficiency */
+    static float approachHeat(float current, float target, float rate){
+        return target < current ? target : Mathf.approach(current, target, Math.min(rate, Math.abs(target - current)));
+    }
```
EggleEgg, line 19: Awful interface methods with input fields that really should take them from (currently inaccessible I think?) block class parents for convenience

Anuken:
I've thought about this for a bit, and I don't think this should be merged.
- Overdrive isn't a good mechanic and the less if affects, the better
- This *only* affects mixtech, which I don't care about anyway, and 'maybe a mod might use it' is a weak argument
- Heat doesn't have any logical link to overdrive. They are different mechanics and should not stack.
- It's quite invasive and changes a lot of IO code, which I just don't want to deal with, sorry.

PR #12232, opened 2026-06-16
Finish processor link changes optimization / avoid unwanted 'wait' resets

cardillan:
Commit https://github.com/Anuken/Mindustry/commit/d13afc1439622457e5759513b8e5360f3075a10b contains changes to optimize processor link changes, but the job wasn't finished. The new code only runs when a linked block is deleted. When links to existing blocks are added/removed via GUI, the code is still recompiled from scratch.

I've updated the code which manages adding/removing links to use the new system. For this to work, the variables representing linked blocks must exist in the executor (so that they're available via `optionalVar()`), so I'm copying them from the assembler to the executor as well - built-ins and literals are still excluded to keep the array small. 

As a side effect, it is now possible to read linked blocks which also exist as a variable in the code via the `read` instruction. This is a bit inconsistent, as blocks that are linked but not referenced from the code can't be read that way. There's the #11838 PR for that; if you accept this PR, I'll update that one to be compatible.


core/src/mindustry/logic/LExecutor.java
```diff
@@ -124,7 +124,7 @@ public void runOnce(){
     public void load(LAssembler builder){
         stop = false;
         nameMap = null;
-        vars = builder.vars.values().toSeq().retainAll(var -> !var.constant).toArray(LVar.class);
+        vars = builder.vars.values().toSeq().retainAll(var -> !var.constant || var.name.charAt(0) != '_' && var.name.charAt(0) != '@').toArray(LVar.class);
```
1ue999, line 127: why?

cardillan: I want to add only the constants representing the linked variables here, to keep the array small. Before, all constants were excluded. Now I'm excluding the constants representing mlog literals (triple underscore) and builtin variables (`@this`, `@links`, etc.). What remains should be the constants representing linked blocks.

I could include the builtins as well, and they would be remotely readable, but the only really useful one is `@links`, and since links can't be read remotely anyway, its utility is limited. I'd probably add them in #11838.

Anuken:
> but the job wasn't finished. The new code only runs when a linked block is deleted. When links to existing blocks are added/removed via GUI, the code is still recompiled from scratch.

The purpose of the new code was to fix exponential lag machines caused by spamming processors with hundreds of links - one block change could cause hundreds of processor re-links per second. As far as I'm aware, there is no way to do the same by manually configuring blocks, due to the config rate limit. 

Did you run into some performance issues with the current code?

cardillan:
In that case it was a misunderstanding on my part. I assumed the intent was to have a single way to manage links, and perhaps to allow managing links from logic one day.

cardillan:
There's something I've just discovered: when the code is recompiled as a result of a linked/unlinked block, any running `wait` instruction gets reset. Times spent waiting are saved and restored on map save/load, but not on link changes, if I'm not missing something. 

Avoiding the compilation would solve that. I'd suggest to reconsider.

ETA: linking and unlinking a block turns the corresponding variable into a constant. Perhaps that was acceptable when the block was destroyed and might not be acceptable when it is a user action. It's easy to fix and I'll do it later today.

PR #12253, opened 2026-06-22
Public/Private key security for unspoofable, cross-server security on servers

OnlyGergo:

Hi all, creating a draft pr for feedback to add a better UUID/USID

# Problem
UUID is spoofable
USID is per host:port, not good for networks
IP + UUID/USID isn't ideal

# Solution

The idea came to me while trying to sleep, this basically is equivalent to UUID, except for using public/private key, and I remembered the attempt in #6402 

It basically works by:
1) client generating a keypair
2) server asks for client to sign a nonce (random bytes, yes this is the real term)
3) client signs nonce, and also adds what it thinks is the host
4) server verifies the signed nonce + claimed host against learnt or stored player public key (tied to uuid for backwards compatibility)

## Advantages:
- Can't be replaid, the mitm can't sign the random nonce server sends
- Can't be relayed, as the client claimed host won't match configured allowed hosts (similar to bungeecord, or nginx/apache trusted proxies config)
- Forms a stable identifier, same cardinality as uuid, except can't be stolen

## Disadvantages:
- To increase speed, I used Ed25519, which is very secure, but not quantum resistant
- Adds complexity? Not sure. Tried to add a rasonable amount of comments, but can't get rid of the extra steps

## Feedback...
With all those being said, I'd like feedback.

Maybe there is a reason I'm missing as to why this doesn't exist already?
Maybe I used the wrong classes? It works currently, and is backwards compatible (in terms of netcode). It doesn't run for non-headless, but is always on.
I'm not sure of the implications of storing extra data inside PlayerInfo, but it works, haven't tested backwards compatiblity yet, but works on my local. I'm not sure if I did packets correctly, and I don't know if it would be better in Call generated classes as this is key to the functioning of the connection

## Why its a draft

Still one major todo: allowed hosts. The plan is to add a config command, sort of like startupCommands where server hosts can supply their host, I'm also not sure where to reliably get host:port combo, hence the draft. Currently I just hardcoded the if statement that checks this to true, so can still be tested

Maybe its too big of a feature, maybe I should've asked before, but it didn't take too long to make so not worried about wasting my time

If anyone has any feedback I'd appreciate this?


ZetaMap:
why not but what info are you using to only identify one client, not a full ip?
because this design more to validate a server than validate a specific client. moreover a server cannot keep a list of all possible created clients, so sending a challenge to a client without knowing the exact answer is useless, unless you admit that the first connection is always right.

OnlyGergo:
@ZetaMap Yes, it is based on the assumption of trust on first sight, sort of like openssh, client confirms their identity via they private key which only they have, the server IP is just to bind the signed message to a specific channel that can be intercepted, but can't be replayed or used in a relay attack (mitm can take it, nothing valuable is inside)

Your concern is valid, it can't really be verified for certain, but I imagine it's unlikely someone steals registers a public key, before the player being stolen has logged in?
Regardless, you have a point I can add a server command to unlearn a public key, I will implement it once I get an official response over inline completions use


The public key for every player does get stored in Player info (for verifying), just like every player IP, stored as a byte[] it's 32bytes always (not insignificant, open to suggestions on how to store), but for comparison the string representation of an IP (storing an infinite amount, forever) is 7-15 bytes each (assuming UTF8 encoding, can't recall what java Strings are), so I personally don't think it's major, and don't see a way around. Maybe I could store inside a map like other things are, UUID (or usid) -> key?

I'm not sure what you mean by verifying server though, I think you mean the host inside response packet? it contains what client thinks is server so it can't be replayed , as host would be the mitm server, not the attacked server in case of attack

TheRadioactiveBanana:
If you can do it without making anything else super cancer for devs, I am all for it.

5GameMaker:
As someone with a working [key auth implementation](https://github.com/Darkdustry-Coders/MindurkaCompat/blob/master/src/main/java/mindurka/Protocol.java), may I suggest storing the keys outside of settings.bin so they're easier to delete from exports?

I've been meming about bug tracker being the craziest UUID/USID leak for quite some time

ZetaMap:
Yes but for moment mindustry can only handle one settings file at time, due to arc.Settings not designed to be instantiated multiple times.

OnlyGergo:
I've thought about it for a few days but no solution seems quite right.
I haven't looked at the exact arc implementation, so going from memory

The only thing I can think of that is most viable is another PR that adds some sort of exclude, or a separate .bin file for things that need security (like UUID, USIDs or any other "player specific" settings) but I think this is a change for Arc to maybe add a secure flag. This is a really big change though, and may not be worth the complexity in some scenarios 

Other than that I didn't know about any existing implementations so it's nice to hear someone else has the same idea

I will leave this here for a bit longer to get some more opinions, then I will probably ask Anuke if this is worth taking further, it's starting to seem like a big change now


I don't think this will make it cancer for Devs, the idea behind the verified flag is for this to be optional and for code to potentially behave differently for verified/unverified (haven't added the config flag yet however) where public key is sent just not verified, it only runs on headless currently, so shouldn't cause issues on steam / Lan either, I will be testing all this if/when I finish everything.
It is also possible for Devs to implement their own logic, like overriding the Connect packet handler to only verify admins or something

I'm open to any ideas for where to store keys, but I think it's best leaving it inside normal settings (for now)

5GameMaker:
> Yes but for moment mindustry can only handle one settings file at time, due to arc.Settings not designed to be instantiated multiple times.

You can just put them as separate files, or load them separately

Anuken:
Not the worst idea, but code has a lot of inconsistent style/formatting, obvious AI comments, conflicts, etc. I also don't like the idea of slowing down connections by adding another handshake step. If this is ever to be implemented, it needs careful planning and discussion beforehand.

PR #12259, opened 2026-06-25
fix: randomize hosts order in JoinDialog

5GameMaker:

### DIFF
--- a/core/src/mindustry/ui/dialogs/JoinDialog.java
+++ b/core/src/mindustry/ui/dialogs/JoinDialog.java
@@ -406,6 +406,7 @@ public class JoinDialog extends BaseDialog{
         //if the servers have been fetched, use the fetched list
         //otherwise use the cached list + the extra servers that may have been included by mods
         var servers = fetchedServers ? defaultServers : tmpServers.clear().addAll(cachedServers).addAll(defaultServers);
+        servers.shuffle();
 
         for(int i = 0; i < servers.size; i ++){
             ServerGroup group = servers.get((i + servers.size/2) % servers.size);


5GameMaker:
(was tested by opening the game and spamming refreshes for a bit, and it seems to work)

1ue999:
why though?

5GameMaker:
> why though?

It prevents the order of server categories from affecting their positions in server list, so the server new players go to is less influenced by whoever optimized serverlist json file better. Something that apparently is a thing.

WMF-Industries:
aren't servers sorted based on ping anyway?

5GameMaker:
> aren't servers sorted based on ping anyway?

Order still has an effect as IO itself takes some time

WMF-Industries:
a few millis yeah

5GameMaker:
> a few millis yeah

And the ping race is indeed at milliseconds scale

Yaaty256:
Mindustry Server List Optimization is so real.

Server groups are currently sorted by their group name hash code.
https://github.com/Anuken/Mindustry/blob/d22a2995a32b1f0d9e68af208d57d5c4c15739e2/core/src/mindustry/ui/dialogs/JoinDialog.java#L730

And despite ping playing a role in their ordering, the group ordering is still so impactful, that if you're at the very end of the order, you're going to be at the end of the server list, despite having a lower ping.

As an example, I compared the Esco server group with their currently high ranking
image
and it was placed as the 4th server in the server list
image

However, if you change that group name to something with a really low ranking
image
it'll be with the very bottom servers.
image


And you can also see that the Esco server group has 64ms ping, while other nearby server groups have 180ms ping and even 300ms ping.

core/src/mindustry/ui/dialogs/JoinDialog.java
```diff
@@ -406,6 +406,7 @@ void refreshCommunity(){
         //if the servers have been fetched, use the fetched list
         //otherwise use the cached list + the extra servers that may have been included by mods
         var servers = fetchedServers ? defaultServers : tmpServers.clear().addAll(cachedServers).addAll(defaultServers);
+        servers.shuffle();
```
Yaaty256, line 409: Wouldn't it be better to change the code at https://github.com/Anuken/Mindustry/blob/d22a2995a32b1f0d9e68af208d57d5c4c15739e2/core/src/mindustry/ui/dialogs/JoinDialog.java#L730
to be shuffled, instead of changing the order again?

Anuken:
This was already proposed, and it doesn't fix the real issue; servers constantly changing orders on the user end is very annoying.

Anuken:
...After some testing, I'll just go with the approach above of shuffling them *at fetch time* instead of every refresh, which should be relatively fair while not being annoying. I didn't see any significant differences in order on my end either way.

PR #12260, opened 2026-06-25
No unnecessary map editor menu hiding

gtvrurker:

Basically when you pick almost any option in the map editor menu or the map info menu it hides , so you when you go back you go straight to the map editor itself and sometimes its very annoying
I removed all lines of code that unnecessairly do that and added where needed
Also the save/export buttons dont hide the menu so i think it wont hurt if all buttons didn't

- [V] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [V] I have ensured that my code compiles, if applicable.
- [V] I have ensured that any new features in this PR function correctly in-game, if applicable.

### DIFF
--- a/core/src/mindustry/editor/MapEditorDialog.java
+++ b/core/src/mindustry/editor/MapEditorDialog.java
@@ -70,23 +70,23 @@ public class MapEditorDialog extends Dialog implements Disposable{
         menu.cont.table(t -> {
             t.defaults().size(swidth, 60f).padBottom(5).padRight(5).padLeft(5);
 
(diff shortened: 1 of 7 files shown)


Anuken:
No, it's actually very annoying when you just want to select the submenu and it doesn't hide the original menu selector. There are also some **glaring** bugs, like quitting not hiding the menu, which is obviously wrong.

gtvrurker:
> There are also some **glaring** bugs, like quitting not hiding the menu, which is obviously wrong

I dont have this issue, i even tested it right now and the menu doesnt persist after quitting

PR #12274, opened 2026-07-01
Blacklist UDP address along with TCP address

BalaM314:
I've heard rumors that the recent wave of bots is using a TCP-only proxy and is using different IPs for UDP and TCP. Currently Mindustry only blacklists the TCP address. They have fewer UDP addresses so blacklisting those should result in better mitigation.

(You may instead want to edit arc netcode to require TCP and UDP addresses have the same host)


BalaM314:
I am currently unable to build mindustry because jitpack said no so I'm using the CI runners to test it

PR #12277, opened 2026-07-01
update ja cruxscape name

wancor1:
I interpreted `cruxscape` as referring to something like `Enemy Central Stronghold` or `Enemy Main Stronghold` in Japanese, so I translated it accordingly.
If that's not the intended meaning, please let me know.


acutescv:
by it's name you can interpret it as, something like crux's capital city 
unfortunately the map creator left awhile ago so i cant fully know the true meaning

wancor1:
oh...

Anuken:
Actually, this name never shows up in the game, so it's unnecessary. I should remove it from the bundle.

wancor1:
oh.................

PR #12285, opened 2026-07-04
Entity serialization optimization, part 1

BalaM314:
Currently, entity snapshots are written once for each player, because some entities can be hidden. In vanilla this is only used for fog of war, but it may also be used by mods.

In lategame attack maps, this requires serializing thousands of entities for up to 20 players, which is inefficient.

Previously:

image

This PR changes entity snapshot serialization to use one of three options:
a) Always send all entities, serializing only once. This is used when FoW is off.
b) Serialize entities once for each team, and only call isSyncHidden for the first player in each team. This is used when FoW is off.
c) Serialize entities once per player and call isSyncHidden once per player. This is used when a flag is set, and is required for mods or plugins that use isSyncHidden to show entities only to some players.

After the optimization was applied:

image

This has been tested and works correctly with (a). This has not yet been tested with fog of war (b) or the flag disabled (c).

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -537,7 +537,7 @@ public class NetClient implements ApplicationListener{
         }
     }
 
-    @Remote(variants = Variant.one, priority = PacketPriority.low, unreliable = true)
+    @Remote(priority = PacketPriority.low, unreliable = true)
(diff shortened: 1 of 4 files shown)


BalaM314:
The code is working correctly with fog of war. Disabling the flag does not blow it up, I haven't tested it with plugins that override isSyncHidden.

PR #12308, opened 2026-07-07
Improve sound selector UI to support tabs

### LINKED ISSUE #6034: Change the sound effect picker for `playsound`
### Describe what you would like changed, and why.

The current sound effect picker is too simple for how many sounds are now in the game, their chaotic order also doesn't help. I end up finding myself looking at the sound files in github because it's easier to find specific sounds which isn't an ideal experience for a player.

### Describe the changes you want to propose. Include possible alternatives.

Change the sound effect picker in `playsound` into a menu with tabs based on sound category, and a button to preview the sound.

Abdullah-Ajeebi:

Fixes Anuken/Mindustry-Suggestions#6034

### DIFF
--- a/core/src/mindustry/logic/LStatements.java
+++ b/core/src/mindustry/logic/LStatements.java
@@ -2,7 +2,6 @@ package mindustry.logic;
 
 import arc.*;
 import arc.audio.*;
-import arc.files.*;
 import arc.func.*;
 import arc.graphics.*;
 import arc.scene.style.*;
@@ -2446,9 +2445,8 @@ public class LStatements{
                 Sound sound = entry.value;
                 if(sound == Sounds.none || sound.file == null) continue;
 
-                String name = sound.file.nameWithoutExtension();
-                Fi parent = sound.file.parent();
-                String category = parent == null ? "other" : parent.name();
+                String name = Strings.getFileNameWithoutExtension(entry.key);
+                String category = soundCategory(entry.key);
 
                 SoundChoice choice = new SoundChoice(category, name, sound);
                 choices.add(choice);
@@ -2516,8 +2514,8 @@ public class LStatements{
                                 id = "@sfx-" + choice.name;
                                 build(table);
                                 hide.run();
-                            }).growX().left().checked(choice.name.equals(current)).padRight(4f).minWidth(540f);
-                        }).growX().padBottom(3f).minWidth(620f).row();
+                            }).growX().height(40f).left().checked(choice.name.equals(current)).padRight(4f).minWidth(540f);
+                        }).growX().height(40f).padBottom(3f).minWidth(620f).row();
                     }
                 };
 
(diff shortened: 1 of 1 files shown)


Anuken:
image

This incorrect displays data patch sound based on their filename; it should use their asset entry name instead (as it is currently).

image


The height of the selector button doesn't fill the vertical space.

Abdullah-Ajeebi:
Latest commit should fix the incorrect name issue and the buttons not filling all of the space

core/src/mindustry/logic/LStatements.java
```diff
@@ -2436,6 +2436,130 @@ public LInstruction build(LAssembler builder){
...
+        private static String soundCategory(String entryName){
+            String normalized = entryName.replace('\\', '/');
+            int end = normalized.lastIndexOf('/');
+            if(end < 0) return "other";
+
+            int start = normalized.lastIndexOf('/', end - 1);
+            return normalized.substring(start + 1, end);
+        }
+
+        private static void previewSound(Sound sound){
```
Anuken, line 2569: A few issues here:
- Most sounds don't actually play first try. Try going into the shoot sound tab, and see for yourself. You will need to use the new Sound method that accepts a bus parameter instead of reassigning it.
- The sound keeps playing even after the dialog is closed, which is very bad for long sounds.
- There's no indicator for when a sound is playing.
- (minor) The dialog visibly flickers when switching tabs.

Abdullah-Ajeebi:
(says co-authored by copilot because for some reason it wouldn't push on my own)

core/src/mindustry/logic/LStatements.java
```diff
@@ -2436,6 +2437,170 @@ public LInstruction build(LAssembler builder){
...
+            if(lastPreview != null && lastPreview.countPlaying() > 0){
+                lastPreview.stop();
+            }
+
+            lastPreview = sound;
+
+            // set the sound's bus to the UI bus while starting playback, then restore after a short delay
+            var oldBus = sound.bus;
+            try{
+                sound.setBus(control.sound.uiBus);
```
Anuken: As per last comment: **You will need to use the new Sound#play method that accepts a bus parameter instead of reassigning it.** setBus is reassigning it. This is not acceptable.

Abdullah-Ajeebi: `sound.bus = control.sound.uiBus; sound.play();` also sets the bus, right?

Abdullah-Ajeebi: In hindsight, asking that question was useless.
Also, I'm not sure if the new play method was made after my PR was made and thus I need to update.

Abdullah-Ajeebi: After searching through my local Arc and the latest Arc version, I found zero instances of play accept an AudioBus parameter. I'm not sure if you knew this or not.

Abdullah-Ajeebi: I do not know why, but GitHub managed to hide it away from me, which is somehow both impressive and quite bad.
Updating to latest Arc, and I will push with the correct fix (make play use AudioBus)

Abdullah-Ajeebi:
I hate intellij but I have infinite disk space and must fill it up

Anuken:
(There were still many bugs so I had to go fix it myself)

redloong9527:
It need to turn off the game's background music when playing sound effects, otherwise it will be very noisy.

PR #11661, opened 2026-02-12
Fixes the multiple reload shots ( https://github.com/Anuken/Mindustry/issues/11636 )

SomeonesShade:
Fixes https://github.com/Anuken/Mindustry/issues/11636 , without reversion.

#### Broken Behavior (caused by charging or clicking multiple times)
https://github.com/user-attachments/assets/67c847d3-8e2c-4b76-a50e-41c1c4212b89

#### Fixed Behavior

https://github.com/user-attachments/assets/84f1ea59-80a5-4577-a658-c97c45950521

### So... what happened?

- Basically, there is a failsafe system where the turret will fire multiple shots, 
- if the excessReload generated would've been extremely high, (this was mostly done out of just in case, which is questionable),
- and reload Shots tracks the number of shots it should fire immediately when this happens.

```
        protected void capReload(){
            //cap reload for visual reasons, need to store the excess reload to keep the firerate consistent
            if(canReload() && reloadCounter >= reload){
                reloadShots += (int)(reloadCounter / reload);
                excessReload += reloadCounter % reload;
            }
            reloadCounter = Math.min(reloadCounter, reload);
            reloadShots = Math.min(reloadShots, 5);

            //if it isn't constantly firing, do not keep excess reload
            if(!wasShooting || shootWarmup < minWarmup){
                reloadShots = 0;
                excessReload = 0;
            }
```
This updated line `!wasShooting || shootWarmup < minWarmup` didn't consider that the turret can also not fire if the turret is currently warming up (my bad).

It now does, and it should work.

### Notes

- I'm still thinking if the implementation I did before was okay, maintable, etc., 
- I guess it was merged and it did have this only bug... 
  - Theoretically, `wasShooting` tracks already if the turret doesn't constantly fire, 
  - the `!charging()` already prevents turrets that are charging like lancer to pile up reload, 
  - and the `shootWarmup < minWarmup` now prevents warmup conditions... 
- completing the list in `if(reloadShots > 0 && !charging() && shootWarmup >= minWarmup){` at `updateShooting()`

If your pull request is **not** translation or serverlist-related, read the list of requirements below and check each box:

- [x] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [x] I have ensured that my code compiles, if applicable.
- [n/a] I have ensured that any new features in this PR function correctly in-game, if applicable.

PR #11245, opened 2025-09-27
Reactivity and Assorted Changes

SomeonesShade:
##### _Updated: December 31, 2025_

# Reactivity Balance
## Section Notes

- The general summary is that turrets main disadvantage is the **upfront cost** the defender has to foot in, in exchange for great ammo cost to killed unit cost _(otherwise you would be using units for defense)_. Reactive mechanics essentially **bypasses** the upfront cost; this is notorious for lancer/arc, or shockmines against Ground T1-T3.
- Problematic mechanics that can't be managed with stat balancing... i.e. no wiring power turrets and item filling... needs to be managed. 
- This is the purpose of this pr, to allow future balance to be possible under the new stat framework given.
-  ...I wouldn't be able to get the numbers right the first time.
- A shockmines fix hasn't been done yet, since there is a stat based solution for it, I'm thinking of (spikeweed shockmines or iirc see Blackberry's mega pr)
<br></br><br></br>

## I. Activation Time
**[Status: Ready to Merge]**
- Implemented Activation Time, which is when Turrets need some time before they activate.
  - Only includes setBars(), and a setStats(), visuals arent implemented yet
- So far due to feedback, saving capability has been reverted, visual work has been reverted (although will be done at a later point), and arc/lancer no longer has it. It is only datapatch accesible

#### Feedback
- Should activationTime be in the Block level instead (nice to have shockmines access it, although not sure if basal blocks need the mechanic to begin with unless you are doing a cool intro/anim for it)
<br></br><br></br>

## II. Building Specific Deposit Cooldown
**[Status: Ready to Merge]**
- Implemented Deposit Cooldowns on a per building basis (similar to itemDepositCooldown).
  -  i.e. fuses may inflict a 1 second cooldown, while salvos being 2.0 seconds.

#### Caveats and Disclaimers
- Deposit Cooldown needs to be DP'd out as it is opt out now... I wanted to make the setStats() update depending on the current rule set... _(see needed feedback)_.

#### Feedback
- Should there be any values to start w/ for the deposit Cooldown or should I yeet the ones I have?
- I know that the feature could be better communicated as it doesn't have a setStats() entry, its tricky if this needs to be made clear.
<br></br><br></br>

---

# Assorted/Utility Balance Changes
## III. Reload and Coolant Fix
**[Status: Ready to Merge, can be changed from feedback]**
- Fixed turret overdrive boost and coolants getting chopped off from Math.min(), making multiplicative boosts not fall short. 
[...]

PR #12336, opened 2026-07-11
A functional not so buggy reload fix

SomeonesShade:
This is a better version of https://github.com/Anuken/Mindustry/pull/11661 that shouldn't break anything.

This handles or fixes the case where the cooling reload doesn't go through due to updateReload blocking updateCoolant from being accounted for.
I removed the unused canReload() since... use handleReload() instead. And the reloadCounter < reload condition for updateCoolant is only added into the one non-Turret.java class manually... (PointDefenseTurret.java)

If any subclass adds like a updateItemCoolant() as an example, or if the if condition should be altered, handleReload() should help w/o requiring to mess up updateTile()

#### [A1] Firerate differences w/ Cyclone... all correct.
https://github.com/user-attachments/assets/496325de-3d3b-450d-838c-8f4474744039

**Q: (Shouldn't the changes be counternerfed due to the buffs?)**
_A: Turns out the changes in Blocks.java weren't reverted in the original reversion... so I don't need to... https://github.com/Anuken/Mindustry/commit/d90c830b2770556094951ada0eee4b1abc6a5968_
Or by checking the diff and seeing Blocks.java, and seeing the current stats and find that they weren't reverted... https://github.com/Anuken/Mindustry/pull/11245/changes#diff-5fcb58432c6c68d32771609241d663343e35cbd86baa58c85923fdb9890748be

**Q: (You shouldn't buff cyclone/swarmer!)**
_A: I already did the counternerf required months ago, do a seperate pr/suggestion post. I'm not dealing with any form of official balancing at all, but I do want this weird behavior altered._

**Q: (What happens when the reloadCounter goes past 2x the reload? The modulo induces a loss)**
_A: Learned my lesson to not handle what if cases that doesn't appear... if it does happen a seperate suggestion post or pr is preffered. It made the original pr a bit... more complicated than it needed to be due to me trying to handle this scenario._

### DIFF
--- a/core/src/mindustry/world/blocks/defense/turrets/Turret.java
+++ b/core/src/mindustry/world/blocks/defense/turrets/Turret.java
@@ -544,11 +544,7 @@ public class Turret extends ReloadTurret{
                 ((Building)this).rotation = Mathf.mod(Mathf.round(rotation / 90f), 4);
             }
 
-            //turret always reloads regardless of whether it's targeting something
-            if(reloadWhileCharging || !charging()){
-                updateReload();
-                updateCooling();
-            }
+            handleReload();
 
             if(state.rules.fog){
                 float newRange = hasAmmo() ? peekAmmo().rangeChange : 0f;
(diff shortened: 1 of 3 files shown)


core/src/mindustry/world/blocks/defense/turrets/Turret.java
```diff
@@ -716,8 +712,17 @@ public boolean charging(){
             return queuedBullets > 0 && shoot.firstShotDelay > 0;
         }
 
+        protected void handleReload(){
+            //turret always reloads regardless of whether it's targeting something
+            if(reloadWhileCharging || !charging() && reloadCounter < reload){
```
Anuken: Shouldn't it be 
```suggestion
            if((reloadWhileCharging || !charging()) && reloadCounter < reload){
```

SomeonesShade: Ah... whoops...

SomeonesShade:
Okay... I should probably do another footage thing (will take like 5 mins), to see alterations... but it theoretically shouldn't

SomeonesShade:
https://github.com/user-attachments/assets/55abc873-9363-4cca-8065-dc8b329eece3

Yep, all values are clear. Should be good to go... (can't believe I missed that)

2,5 for odd diff, 2.35 and 1.6 for coolant and 5.875 for odd and cyro

SomeonesShade:
Thank you Anuke, deleting branch now...

PR #12348, opened 2026-07-12
Improved Basin missile and fixed markers

SunkyMP3G:
People wait to see an actual missile after "missile launch" is announced, but instead core just explodes out of nowhere.

How this fix works:
A Sharded team scathe missile is spawned after camera pans on core.
Cuz it's a sharded missile, it won't collide with player builds.
When missile reaches the core (now checks the X pos of missile instead of 3 second wait), explosion happens which kills the missile as well.

https://github.com/user-attachments/assets/e0190c0d-e2fa-405e-8d18-a6ec6f0d32b6

A failsafe is there too. If missile dies by any means (It would happen if curious players somehow locked enemy units in east side of core, which is pretty much impossible), explosion happens anyway.

https://github.com/user-attachments/assets/cb1f71d0-ef2d-4cfe-81ae-952b1eb3a54d

Also I fixed pulsing marker being in wrong place and also made all 3 markers show up in minimap.
изображение

There's a chance I might have accidentally misclicked somewhere, so please check for that just in case.


SunkyMP3G:
I can make a more logical failsafe by making missile destroy anything that it can collide with in front of it, but it's might be inconsistent

PR #12351, opened 2026-07-13
Fix issue where afterReadAll in Building is never called

Icexuegao:
This PR fixes a bug where the afterReadAll lifecycle hook in the Building class is never invoked during the read process.
image
image


Anuken:
This isn't a bug, that method is only supposed to be used in units/world entities, and is never even overridden for buildings in vanilla. If you're using it for mods, though...

PR #12369, opened 2026-07-15
Plastanium Conveyor Buffs/Stat Fix (Post-OD Nerf)

Efsfse2280:
Plastanium conveyor is in a tight spot;

- it has the singular _standard_ highest max input speed.
- Can be very efficient in small distances without demand in better output.

_and yet..._
- Mass Driver beats it easily with distance.
- As soon as phase fabric is obtained, phase conveyors are a much more space-efficient upgrade.
- In-game throughput at 60fps is closer to 38.5/s than 40, which means that it'll lose efficiency in a harsher way than the endgame options under low frames and TPS; that best _standard_ max input speed advantage vs MD shrinks `(36 vs 38.5)`*
- Oh yeah. Bridgeweave beats it in max throughput `(44 vs 40)`.This transport technique is less available to players since wiring cost is highest. _especially in multiplayer_

For compensation in performance falloff, I increased its _speed_ by 4%. (`4->4.16`)

In order to keep it viable as a standard conveyor block (wiring cost >MD or phase), I buffed its throughput by +20% towards item capacity `(40->48 i/s)`. The buff's focus on that stat is to further solidify Plastanium conveyors as a **bulk**-oriented transport block.
plast
### Verdict

Overall, a ~20% buff for it will enhance mid- and endgame transport logistics at larger scale.

*: Due to this discrepancy, I've taken the opportunity to edit the database info type for stack conveyors. Otherwise, the displayed ~50/s throughput would've been misleading. Surge conveyors also got displayed at `48/s` to line up with the same inaccuracy threshold as plastanium's.
surge


SomeonesShade:
> For compensation in performance falloff, I increased its speed by 4%. (4->4.16)
> In order to keep it viable as a standard conveyor block (wiring cost >MD or phase), I buffed its throughput by +20% towards item capacity. The buff's focus on that stat is to further solidify Plastanium conveyors as a bulk-oriented transport block.

What's the new throughoutput? 48 i/s?

Edit:
On other notes, I feel like the actual speed of plastanium stuff could be made accurate, rather than just using display speed? But I know it takes a bit of work

I'm pretty okay with the titanium conveyor buff, ill be honest out of the transports pre plast, it should be the highest.

I'm not sure exactly on what to feel about bridge weave, or if it should be a point of comparison... I may need some time to think about it...
Plast buff is fine, 
the limit is at 60 i/s anyway where the question on if phase-md or plast should be more "lategame" or how they should diversify... which probably needs more paragraphs and design docs...

Efsfse2280:
Endgame output is still the best with those two indeed

Anuken:
I will not accept any buffs to these conveyors until overdrive is correspondingly nerfed. Movement speed under ODD under current stats is insane enough as-is.

SomeonesShade:
Oh, this is... very interesting...

DeterMination-Wind:
I have no issues with the changes to Plastanium conveyors; they are fine. However, based on Anuke's perspective and my vote in #development, I think the buff to Titanium conveyors should focus on speed rather than throughput capacity (you know, changing capacity affects everything significantly, like blueprints, etc.). Titanium conveyors are already a must-use conveyor. When basic conveyors get buffed, and Titanium conveyors should follow suit, they should become relevant as early as thorium mining (which requires Titanium conveyors to be faster than router belts).

Anuken:
> When basic conveyors get buffed, and Titanium conveyors should follow suit

I fundamentally disagree with this premise.

There are two options here:

1. Copper conveyors were buffed **because they were disproportionally bad**, which has nothing to do with titanium or any other transportation.

or,

2. Copper conveyors were buffed as some kind of half-baked attempt at making *all* transportation faster, which means the buff should be reverted and reconsidered.

---

Looking at the blocks visually now, I think the copper conveyor buff was misplaced and should at least partially be reverted to something like 5/sec. Titanium conveyors don't need to be faster at all.

Efsfse2280:
titanium conveyor buff reverted - ~10/s has been the standard for years and never has a demand for it to be faster appeared until normal conveyors suddenly got buffed
_i shouldn't have been influenced by #serp-balancing..._

PR #12373, opened 2026-07-15
implement https://github.com/Anuken/Mindustry-Suggestions/issues/6426

1ue999:
implement https://github.com/Anuken/Mindustry-Suggestions/issues/6426
This also adds status effects to GlobalVars. Should `status` get changed to use those? Perhaps setprop could have the same effect as `status`?
checks if it has the effect to make it maybe a little bit cheaper to compute

### DIFF
--- a/core/src/mindustry/entities/comp/StatusComp.java
+++ b/core/src/mindustry/entities/comp/StatusComp.java
@@ -34,7 +34,25 @@ abstract class StatusComp implements Posc{
 
     /** Adds a status effect to this unit. */
     public void apply(StatusEffect effect, float duration){
+        applyStatus(effect, duration, false);
+    }
+
+    public float getDuration(StatusEffect effect){
+        var entry = statuses.find(e -> e.effect == effect);
+        return entry == null ? 0 : entry.time;
+    }
+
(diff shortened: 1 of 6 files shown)


Anuken:
> Perhaps setprop could have the same effect as status?

That would be a good idea.

1ue999:
the UI for `status` is broken right now, thus draft

1ue999:
desktop: image

mobile: image

image

image

1ue999:
the category icon is kinda not what i want tho

cardillan:
@1ue999 Just curious: why does the `status` instruction allow more states to be set than the `setprop` instruction? (`setprop` excludes locked and hidden statuses, `status` doesn't).

1ue999:
i uhh

have no idea i was just modifying setprop code and status code

cardillan:
Maybe it would also be a good idea to add the status category to the `sensor` instruction in the UI. I can try to do it some other day.

cardillan:
When I try to set the `boss` status, I can't choose the duration (I guess it is meant to be permanent), but sensing the status still reports a decreasing time. Is this a bug?

PR #12378, opened 2026-07-16
Unset key bindings

cardillan:
Implements https://github.com/Anuken/Mindustry-Suggestions/issues/6428: Allow players to set their Keybinds to "unset".

This PR requires an Arc PR: https://github.com/Anuken/Arc/pull/205. Not sure what the proper way to do organize this is (the CI build fails because of this).

### DIFF
--- a/core/assets/bundles/bundle.properties
+++ b/core/assets/bundles/bundle.properties
@@ -1098,6 +1098,7 @@ settings.data = Game Data
 settings.reset = Reset to Defaults
 settings.rebind = Rebind
 settings.resetKey = Reset
+settings.unbindKey = Unbind
 settings.controls = Controls
 settings.game = Game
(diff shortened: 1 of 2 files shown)


jazztickets:
This is definitely nice to have and the PR works well. Though with three columns of buttons the UI starts to feel pretty busy. I like how Project Zomboid does their rebind popup, since you have all the buttons there with a cancel option which is currently missing:

pz

Of course that's a lot more effort to rework all that. Just throwing it out there.

cardillan:
That's a good point. I'll see whether I'd be able to do something like that.

Anuken:
I'll shelve this until the UI remake, it's a decent idea but the UI is indeed getting very crowded. A different approach is needed.

PR #12408, opened 2026-07-24
Fix fortress and collaris bullets dying before reaching their intended position

KaVoshnik:
Fixes https://github.com/Anuken/Mindustry/issues/12407

- fortress: maxRange 240 → 213 (2 * 106.5 = 213, matches actual lifetime)
- collaris: rangeOverride 385 → 330 (5.5 * 60 = 330, matches actual lifetime)

If your pull request is **not** translation or serverlist-related, read the list of requirements below and check each box:

- [x] I have read the [contribution guidelines](https://github.com/Anuken/Mindustry/blob/master/CONTRIBUTING.md).
- [x] I have ensured that my code compiles, if applicable.
- [x] I have ensured that any new features in this PR function correctly in-game, if applicable.

PR #12414, opened 2026-07-24
Revert "Fix fortress and collaris bullets dying before reaching their intended position"

EggleEgg:
Reverts Anuken/Mindustry#12408 . See conversation for more info

image


Efsfse2280:
Unless maxRange is for the main bullet of Collaris and the max potential AoE dmg fortress's shell does, #12408 is inaccurate.

PR #12418, opened 2026-07-25
Force display update when copying

cardillan:
This PR ensures a display is forced to update (process its commands) even when off-screen when used as a source in `draw image`. This is similar to what textures do when using a display as a texture. The processing needs to be done outside the normal command processing loop, since the drawing code is not reentrant. Also, checks are in place to avoid infinite recursion in case two displays are copied on each other in the same frame (this has been tested; code copying two displays onto each other works nicely).

This might perhaps cause performance problems in case a lot of off-screen displays are copied, but this only happens with specific setups forcing lots of off-screen displays to update, and should not happen in normal scenarios. Also, the update order of displays which refer to another display may change as a result of this PR. However, since the update strongly benefits users who want to copy images between displays, I think they'd be willing to adapt their code to the new behavior.


cardillan:
I've moved the rootDisplay-related bugfixes into a separate PR (#12514), so that it could be merged without the forced display update functionality.

PR #12436, opened 2026-07-31
fix NaN

JasonP01:
"Fixes" units becoming invalid if their position somehow gets set to NaN by removing it, this is mostly caused by mods ie Time Control. Ensuring it cant occur in vanilla to softlock sectors is a good approach to prevent more bug reports


EggleEgg:
Getting so many bug reports in main from the same mod to the point there s a pr for it is crazy. I m not agaisnt this (actually the opposite, hopefully this fixes it) but basic reading comprehension be dammed

JasonP01:
the real fix would be trying to patch it becoming NaN in the first place, which is practically impossible due to mods changing delta messing up too much math

PR #12459, opened 2026-08-07
allow memory blocks to store any object

Deric-W:
This is an improved version of #10833 which reduces the additional memory usage introduced in the mentioned pull request.

This is achieved by keeping two arrays (one for `double` numbers and one for `Object` instances), with the array to be read from / written to being determined by storing an internal (static) sentinel object in the object array.
Its presence marks a slot as being stored in the numbers array.

NaN is not being used as a sentinel value to prevent unwanted surprises should they be valid numbers somewhere in the future. 

### DIFF
--- a/core/src/mindustry/io/TypeIO.java
+++ b/core/src/mindustry/io/TypeIO.java
@@ -165,33 +165,11 @@ public class TypeIO{
     /** Read the {@code ObjectType} of the next object. */
     public static ObjectType readObjectType(Reads read){
         byte type = read.b();
-        return switch (type){
-            case 0 -> ObjectType.Null;
-            case 1 -> ObjectType.Integer;
-            case 2 -> ObjectType.Long;
-            case 3 -> ObjectType.Float;
-            case 4 -> ObjectType.String;
-            case 5 -> ObjectType.Content;
-            case 6 -> ObjectType.IntSeq;
-            case 7 -> ObjectType.Point2;
-            case 8 -> ObjectType.PointArray;
-            case 9 -> ObjectType.TechNode;
-            case 10 -> ObjectType.Boolean;
-            case 11 -> ObjectType.Double;
-            case 12 -> ObjectType.Building;
-            case 13 -> ObjectType.LAccess;
-            case 14 -> ObjectType.ByteArray;
-            case 15 -> ObjectType.Legacy;
-            case 16 -> ObjectType.BooleanArray;
-            case 17 -> ObjectType.Unit;
-            case 18 -> ObjectType.Vec2Array;
-            case 19 -> ObjectType.Vec2;
-            case 20 -> ObjectType.Team;
-            case 21 -> ObjectType.IntArray;
-            case 22 -> ObjectType.ObjectArray;
-            case 23 -> ObjectType.UnitCommand;
-            default -> throw new IllegalArgumentException("Unknown object type: " + type);
-        };
+        if(0 <= type && type < ObjectType.all.length){
+            return ObjectType.all[type];
+        }
+
+        throw new IllegalArgumentException("Unknown object type: " + type);
     }
 
     public static @Nullable Object readObject(Reads read){
@@ -1436,6 +1414,8 @@ public class TypeIO{
         IntArray        ((byte)21),
         ObjectArray     ((byte)22),
         UnitCommand     ((byte)23);
+
+        static final ObjectType[] all = values();
 
         final byte identifier;
 
--- a/core/src/mindustry/io/TypeIO.java
+++ b/core/src/mindustry/io/TypeIO.java
@@ -41,119 +41,157 @@ import static mindustry.Vars.*;
 public class TypeIO{
     private static final int maxArraySize = 1000, maxByteArraySize = 40_000, maxSyncedPlans = 20;
 
+    /** Write an object, prepending its {@code ObjectType}. */
     public static void writeObject(Writes write, Object object){
         if(object == null){
-            write.b((byte)0);
+            write.b(ObjectType.Null.identifier);
         }else if(object instanceof Integer i){
-            write.b((byte)1);
+            write.b(ObjectType.Integer.identifier);
             write.i(i);
         }else if(object instanceof Long l){
-            write.b((byte)2);
+            write.b(ObjectType.Long.identifier);
             write.l(l);
         }else if(object instanceof Float f){
-            write.b((byte)3);
+            write.b(ObjectType.Float.identifier);
             write.f(f);
         }else if(object instanceof String s){
-            write.b((byte)4);
+            write.b(ObjectType.String.identifier);
             writeString(write, s);
         }else if(object instanceof Content map){
-            write.b((byte)5);
+            write.b(ObjectType.Content.identifier);
             write.b((byte)map.getContentType().ordinal());
             write.s(map.id);
         }else if(object instanceof IntSeq arr){
             if(arr.size > maxArraySize) throw new ArcRuntimeException("Array size too large: " + arr.size);
-            write.b((byte)6);
+            write.b(ObjectType.IntSeq.identifier);
             write.s((short)arr.size);
             for(int i = 0; i < arr.size; i++){
                 write.i(arr.items[i]);
             }
         }else if(object instanceof Point2 p){
-            write.b((byte)7);
+            write.b(ObjectType.Point2.identifier);
             write.i(p.x);
             write.i(p.y);
         }else if(object instanceof Point2[] p){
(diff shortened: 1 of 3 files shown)


WMF-Industries:
why don't you only have the Object[] array and store doubles there?

cardillan:
> why don't you only have the Object[] array and store doubles there?

Having a `double[]` array prevents lots of `Double` instances creation when writing numbers into the memory. This could potentially be a problem for mlog programs intensely writing to the banks, as creating lots of new instances needs to be avoided in Mindustry not to strain the garbage collector.

Deric-W:
> Having a `double[]` array prevents lots of `Double` instances creation when writing numbers into the memory. This could potentially be a problem for mlog programs intensely writing to the banks, as creating lots of new instances needs to be avoided in Mindustry not to strain the garbage collector.

Exactly, as written in the contribution guidelines:
> ## Do not allocate anything if possible.
> 
> Never allocate new objects in the main loop. If you absolutely require new objects, use Pools to obtain and free object > instances. Otherwise, use the Tmp variables for things like vector/shape operations, or create static variables for re-use. If using > a list, make it a static variable and clear it every time it is used. Re-use as much as possible.

Storing Objects and doubles separately is already being done to store processor variables, probably why Anuken himself suggested a similar approach in the feed back to the previous version of this pull request.
While it is true that the second array is theoretically unnecessary its alternatives require something like a Union from C to work efficiently.

Deric-W:
To solve a merge conflict I rebased onto the current master, which broke the tests.
I will try again when the master branch has working tests.

core/src/mindustry/world/blocks/logic/MemoryBlock.java
```diff
@@ -94,13 +131,23 @@ public void damage(float damage){
...
             super.write(write);
 
-            write.i(memory.length);
-            for(double v : memory){
-                write.d(v);
+            write.i(objectMemory.length);
+            for(int i = 0; i < objectMemory.length; i++){
+                Object value = objectMemory[i];
+                if(value == sentinel){
+                    value = numberMemory[i];
```
Anuken: This allocates a new number for every single number write (most common case), which is very bad. It should essentially re-implement the double writing code in read/write itself as a special case to avoid boxing.

Deric-W: To allow reuse of the reading logic in `TypeIO` I implemented a method which allows to read the type of object separately, in turn allowing for the caller to handle some types himself without boxing.
The object type is represented by an enum, with its members not being camelCase to prevent syntax errors.

I can also solve this without modifying `TypeIO`, but that would require writing an additional byte per object since `Reads` does not support "unreading" a byte.
Let me know what you think.

core/src/mindustry/io/TypeIO.java
```diff
@@ -156,6 +156,44 @@ public static void writeObject(Writes write, Object object){
...
     }
 
+    /** Overload of {@code writeObject(Writes, Object)} which prevents boxing. */
+    public static void writeObject(Writes write, double object){
+        write.b(ObjectType.Double.identifier);
+        write.d(object);
+    }
+
+    /** Read the {@code ObjectType} of the next object. */
+    public static ObjectType readObjectType(Reads read){
```
1ue999, line 166: This defeats the point of having an enum. `ObjectType.all[id]` is better in every way (declared as `public static final ObjectType[] all = values();` prob, as done in other enums)

core/src/mindustry/world/blocks/logic/MemoryBlock.java
```diff
@@ -46,7 +52,19 @@ public boolean canBreak(Tile tile){
     }
 
     public class MemoryBuild extends Building implements LReadable, LWritable{
-        public double[] memory = new double[memoryCapacity];
+        /** Marks a memory slot as being stored in {@code numberMemory} (insted of {@code objectMemory}) */
+        private static final Object sentinel = new Object();
```
1ue999, line 56: Why not use NaN in number memory as a sentinel?

1ue999 (changes requested):
I honestly see no point in having memoryblocks able to store objects. If you want to store objects, just use direct reads/writes from processors. There are multiple ways to get a nice indexable storage out of processors.

core/src/mindustry/world/blocks/logic/MemoryBlock.java
1ue999: also its very unlikely that they will be valid numbers in the future.

core/src/mindustry/io/TypeIO.java
Deric-W: Do you mean the switch statement or the whole method?
I agree with replacing the switch statement with a lookup table, but having a dedicated method prevents the error handling logic from being duplicated all across the project.

core/src/mindustry/world/blocks/logic/MemoryBlock.java
Deric-W: Because I think it is less surprising that way, also it makes it impossible to forget to clear the existing object reference when storing a number.

Deric-W:
> I honestly see no point in having memoryblocks able to store objects.

> There are multiple ways to get a nice indexable storage out of processors.

While you can use processors for this purpose (requiring additional setup and being less efficient since every variable has its own `LVar` object) I think it is wrong that the buildings specifically designed for the purpose of storing data are "worse" at it than processors.

core/src/mindustry/io/TypeIO.java
1ue999: ye this works

1ue999:
> > I honestly see no point in having memoryblocks able to store objects.
> 
> > There are multiple ways to get a nice indexable storage out of processors.
> 
> While you can use processors for this purpose (requiring additional setup and being less efficient since every variable has its own `LVar` object) I think it is wrong that the buildings specifically designed for the purpose of storing data are "worse" at it than processors.

ok thats a very good point

core/src/mindustry/io/TypeIO.java
```diff
@@ -1368,6 +1388,42 @@ public static String readStringData(DataInput buffer) throws IOException{
         }
     }
 
+    /** Type of a written object. */
+    public static enum ObjectType{
```
Anuken: Please don't add any new classes for this, it's so much more ceremony and makes adding new types more annoying. Make them constants if you have to.

core/src/mindustry/world/blocks/logic/MemoryBlock.java
```diff
@@ -84,8 +111,14 @@ public boolean writable(LExecutor exec){
         @Override
         public void write(LVar position, LVar value){
             int address = position.numi();
-            if(address < 0 || address >= memory.length) return;
-            memory[address] = value.num();
+            if(address < 0 || address >= objectMemory.length) return;
+
+            if(value.isobj) {
```
Anuken:
```suggestion
            if(value.isobj){
```

core/src/mindustry/world/blocks/logic/MemoryBlock.java
```diff
@@ -46,7 +51,19 @@ public boolean canBreak(Tile tile){
     }
 
     public class MemoryBuild extends Building implements LReadable, LWritable{
-        public double[] memory = new double[memoryCapacity];
+        /** Marks a memory slot as being stored in {@code numberMemory} (insted of {@code objectMemory}) */
```
Anuken:
```suggestion
        /** Marks a memory slot as being stored in {@code numberMemory} (instead of {@code objectMemory}) */
```

core/src/mindustry/world/blocks/logic/MemoryBlock.java
```diff
@@ -18,7 +18,6 @@ public class MemoryBlock extends Block{
 
     public MemoryBlock(String name){
         super(name);
-        update = true;
```
Deric-W, line 21: This line is important to ensure that `afterReadAll` is called, without it there is no unboxing taking place.

I tested the current master (after the merge) and objects in memory cells are not unboxed, maybe revert 76388ec3edeb1613afaac24e5c8f661e44703079?

Anuken: That's a bug in afterReadAll, then, because you shouldn't need to set update to true just for that method to run. Every memory cell updating despite not actually having to update is wasteful.

Deric-W: Makes sense, thanks for fixing it on master, with 6ef175cb29997c3e239cb775cee20590232e5499 the objects are unboxed.

Thanks also for putting up with my inexperience with Java.

PR #12480, opened 2026-08-12
Update bundle_ja.properties - server.outdated

wancor1:
Other languages had changed, so I thought it might be better to update this as well.
translated by ai.


SunkyMP3G:
Uhh there's no reason to mark that as "ai translated" if all it did is replace the symbol and color.

PR #12482, opened 2026-08-13
Implement image streaming API

buthed010203:
Took me a while to finally get around to cleaning it up but it's finally done.
Adds support for dynamically sending clients images via `Vars.netServer.sendTexture(...)`.
Dynamic images must be manually cleaned up with `Call.removeTexture(...)` to avoid garbage piling up during a game.

Streaming considerations:
- Since the images are streamed, they are not bound to typical packet size limits
  - It may be worth adding some way to limit the size of a stream as well as the number of them incoming as currently there are no such limitations. Only the dimensions of the image are checked and even so it is once the stream is fully received. A malicious server could abuse this to OOM clients with high bandwidth connections.
- Asynchronous streaming means that sending many textures should be minimally disruptive on low bandwidth connections
- **If attempting to `Call.removeTexture` before the texture is fully received on the client, it could be processed out of order resulting in dangling textures.**
  - Fixing this one would require rearchitecting some of the streaming stuff or adding a timestamp to the stream start, maintaining a list of removed textures and their timestamps to see if it was removed after streaming commenced.
- Some changes were made under the hood to the menu API to accommodate streamed images
  - Image, Button, ImageButton now have an optional placeholder that will be used if the primary TextureRegion can't be located
  - When a texture stream is finished, any Image objects in open menus will be updated to point to the texture instead of the fallback.
    - Currently, if the texture is subsequently removed again the image will just go blank rather than returning to the placeholder. This should probably be changed.

Other notes:
- I don't know when the proper time to forcefully evict all of the server streamed textures is so I'm just doing it on DataImagePacker.unload() which seems to work fine but it's potentially not ideal (for example, wouldn't that mean that servers would have to resend images every map if they want to have them between maps?)
- There was an unused class in DataImagePacker (likely from before a refactor) so I nuked it
- I have tested this on my own server but there are likely still issues with it considering the scope

Edit: Man this almost reads like one of those wall of text ai slop PRs but it'd be impossible to interpret my weird writing style as that

### DIFF
--- a/core/src/mindustry/core/NetClient.java
+++ b/core/src/mindustry/core/NetClient.java
@@ -165,6 +165,16 @@ public class NetClient implements ApplicationListener{
             Call.requestAssets(missing.toArray());
         });
 
+        net.handleClient(TextureStream.class, data -> {
+            try(DataInputStream in = new DataInputStream(data.stream)){
+                String name = in.readUTF();
+                byte[] pngData = in.readAllBytes();
+                if(!headless) state.data.addTexture(name, pngData);
+            }catch(IOException e){
+                Log.err("Failed to read server texture stream", e);
+            }
+        });
+
         net.handleClient(StreamBegin.class, data -> {
             boolean isWorld = data.type == Net.packetIdWorldStream, isAssets = data.type == Net.packetIdAssetStream;
 
--- a/core/src/mindustry/core/NetServer.java
+++ b/core/src/mindustry/core/NetServer.java
@@ -552,6 +552,38 @@ public class NetServer implements ApplicationListener{
         debug("Packed @ of world data to @ (@ / @)", Strings.formatByteCount(stream.size()), player.name, player.con.address, player.uuid());
     }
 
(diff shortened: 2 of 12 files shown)


JasonP01:
> (for example, wouldn't that mean that servers would have to resend images every map if they want to have them between maps?)

Perhaps make it on the disconnect event, as i could see many servers using the images regularly throughout different maps (such as for events).

As for size limitations, i dont think an image should be bigger than 5mb, having a borderline infinite amount of images to send to a client will massively increase the cache used by the game that then needs to be deleted in a short amount of time during disconnect, if this isnt a real concern just ignore it

buthed010203:
Even a single megabyte for an image is more than enough for every reasonable use case of this api.
On disconnect would make sense but honestly, I'm not sure how to best do that at this point and I'm somewhat content with the state of this PR as is outside of the remove before receive race condition that can occur. I would appreciate if someone else would fix that one though as I don't particularly want to do so right now.

PR #12487, opened 2026-08-15
Fixing force projectors don't have the right break effects

Martian238:
It's still 6-sided without rotation in 159.7 so I tried to fix it.


Martian238:
image
these names are automatically generated tho

Issue #8897, opened 2023-08-05
Unit Commandable After Unbinding


strangeprogrammer:
### Platforms

Android

### Build

fdroid build 145.1

### Issue

Logic controller seems to keep stale data of a previously bound unit?

NOTE: I am having trouble uploading the save file as a 'zip' to GitHub, so it is instead encoded as a base64 string within the text field.  I will provide information on how to decode said file if requested.

### Steps to reproduce

Made a logic controller bind a unit.
Made the same logic controller unbind said unit.
Used the "Unit Control" command to move said unit to the logic controller's coordinates.

The unit moves when it shouldn't.

### Mods used

N/A

### Save file

eJylfVtsZEd2WPX73Xw/59FUz/SMRiJnSA7nqV2tdqTVSrsjr8zV2ruWpd0m2SR71eymu9kz4mpl
b2wjhhOTJgcG8uOfAA6QnwBOjDgO4Lx+AsN5GMhvgAD5cID8BQgQ5CfZ1Knzqrp9ZyQjAjR9b92q
U6fO+5yqe/ned7/+S8aYnDHFvBk3+cNO8/iofdAy2fX11bW7ayb/tPmk5Vpyaxsbq6s3V01mYJt2
TGXt7oO19Tvrq+sPHmzcMbkn7dbTw97AzLy8trq2dnPj7oP7y+sP7t68c291ff2GyXV7271+a2Ay
u83OoGXSR+3tT8zEg7v3Hty88+Dexvrtu/fW7967a8EfNY8G5rc+2xq2Ozvt7t7gkb04enhnWRre
am33uoOj/nD7qLXz8PayxXq7tfOo09v+5M3esHv08LOD9na/t3LY7223BoNe/+H68mHzuNNr7qwM
esP+duvh2vLgaftoe99eHNguzT3b9PnysNs+GrzZbzUB7trnJnfQPOw27eLHvt7dbre6R0tvNjs7
rX7TVC0GR/1ep9Pa+eD4sGUyzc7hftNkHI4mtWYpkj7o7QxM8sOPTOZpe+do36TW79w12f1We2//
CG8y/WGnNcgPP2seHfXbW8Oj1uDhZ58vt7u7bYtJa7OF2A4e2rW2lgH/3vDo4Yef2WcHD7d7h4et
[...]

PR #12499, opened 2026-08-16
Rename 'unit control unbind' with 'unit control decontrol'

A4-Tacks:
The name 'unbind' has led to the creation of many foolish programs, such as:

```gas
ubind @flare
ucontrol unbind 0 0 0 0 0
```

But the behavior of `ucontrol unbind` is to ~~control and~~ release control, but because of the name `unbind`, it guides people to use it in pairs with `ubind` (unit bind).
This has resulted in poor outcomes: some programs that do not require control have established unexpected ~~controls and~~ immediately released them, inadvertently disrupting the programs of many ongoing control units

So this PR renamed 'unbind' to 'decontrol', so that people no longer confuse control and binding

Since only the display of logic editor has been changed, there will be no compatibility problems with exporting/importing, and users who write code outside of logic editor are usually not misled by this confuse naming

**CURRENT**:
According to the comments, changing the internal naming helps with consistency, and we are waiting for your opinion on which commit to use

**NOTE**: This PR is missing translation updates, but I am unable to commit update patches as it has been closed

---

This is different from the fixes in https://github.com/Anuken/Mindustry/issues/8897#issuecomment-1683302765 (I noticed that was reverted), this PR establishes the correct naming to prevent it from being confused

### DIFF
--- a/core/src/mindustry/logic/LUnitControl.java
+++ b/core/src/mindustry/logic/LUnitControl.java
@@ -21,19 +21,12 @@ public enum LUnitControl{
     deconstruct("x", "y"),
     getBlock("x", "y", "type", "building", "floor"),
     within("x", "y", "radius", "result"),
-    unbind;
+    decontrol;
 
     public final String[] params;
     public static final LUnitControl[] all = values();
 
     LUnitControl(String... params){
         this.params = params;
-    }
-
-    public String symbol() {
-        return switch (this) {
-            case unbind -> "decontrol";
-            default -> this.name();
-        };
     }
 }
--- a/core/src/mindustry/logic/LUnitControl.java
+++ b/core/src/mindustry/logic/LUnitControl.java
@@ -29,4 +29,11 @@ public enum LUnitControl{
     LUnitControl(String... params){
         this.params = params;
     }
+
+    public String symbol() {
+        return switch (this) {
+            case unbind -> "decontrol";
+            default -> this.name();
+        };
+    }
 }
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -344,7 +344,7 @@ public class LExecutor{
             if(!exec.privileged && !state.rules.logicUnitControl) return;
 
             Object unitObj = exec.unit.obj();
-            boolean control = type != LUnitControl.unbind && type != LUnitControl.within;
+            boolean control = type != LUnitControl.decontrol && type != LUnitControl.within;
             LogicAI ai = checkLogicAI(exec, unitObj, control);
 
             //only control standard AI units
@@ -370,7 +370,7 @@ public class LExecutor{
                             unit.clearBuilding();
                         }
                     }
-                    case unbind -> {
+                    case decontrol -> {
                         if(unit.controller() instanceof LogicAI){
                             unit.resetController();
                         }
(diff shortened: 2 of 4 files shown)


core/src/mindustry/logic/LUnitControl.java
1ue999: you didn't rename this?

1ue999 (commented):
if this is not meant to be only cosmetic, it wont work.

since text mlog will still be
`ucontrol unbind`

A4-Tacks:
This maintains compatibility in a simple enough way, and see:

> and users who write code outside of logic editor are usually not misled by this confuse naming

I don't want any more magic in LParser

cardillan:
Just fyi, since https://github.com/Anuken/Mindustry/commit/e5e417265d1192952d62be7837229f8e1eb8cdf6, `ucontrol unbind` no longer controls a unit that was previously not controlled.

A4-Tacks:
The key is that it releases control rather than binding, whether it controls first or not, it leads to misuse and accidentally releases units controlled by other programs

cardillan:
What about modifying `ucontrol unbind` to also set `@unit` to `null`? It would look quite useful to me - certainly more useful than the current implementation. Then the current naming would be correct.

Also:

> This maintains compatibility in a simple enough way, and see:
> 
> > and users who write code outside of logic editor are usually not misled by this confuse naming
> 
> I don't want any more magic in LParser

On the other hand, people most probable to advice newbies are expert mlog users, and each difference between a text mlog and UI mlog is an opportunity for misunderstanding - code is often shared in text form on discord, and advanced mlog users think in terms of the text mlog, not UI mlog. I'd consider a small amount of magic in LParser reasonable for keeping the nomenclature as consistent as possible.

A4-Tacks:
> What about modifying `ucontrol unbind` to also set `@unit` to `null`? It would look quite useful to me - certainly more useful than the current implementation. Then the current naming would be correct.

This used to exist, but it has been reverted:
> https://github.com/Anuken/Mindustry/issues/8897#issuecomment-1683302765 (I noticed that was reverted)

> code is often shared in text form on discord

That sounds good too

cardillan:
> This used to exist, but it has been reverted:

Oh, sorry, I missed this one. The message on the reverting commit (https://github.com/Anuken/Mindustry/commit/4b6152ffe9827760b3113aab6f64e1b25f184792) only says "Removed pointless statement", which doesn't really explain the reason for the revert. Hm.

Anuken:
No, I'm not renaming it.

A4-Tacks:
Will renaming internal or renaming display not merge?

This has indeed caused long-term inconsistency and extensive misuse, and I would like to know the reasons for not merging

Anuken:
I don't see how changing the name will change people's logic or use of it.

> This used to exist, but it has been reverted:

it was reverted because it did, in fact, break people's logic. I'm not going down that path again. Things will stay as they are.

A4-Tacks:
> I'm not going down that path again.

Hey, my PR provides two different solutions than 'reverted', in two commits

A4-Tacks:
Solution 1: No breaking, only changed the display name
Solution 2: Change the internal name, export from new version, and breaking when importing to lower versions

A4-Tacks:
> I don't see how changing the name will change people's logic or use of it.

The old name is inconsistent with its behavior, which is misleading

Newbies relies heavily on names to understand the meaning of statements, especially when it is difficult to detect hover descriptions on mobile devices

If you want to use the least break 'Solution 1', I will revert back to the first commit

A4-Tacks:
@Anuken From the description, it appears that this is indeed eliminating inconsistencies:

```diff
-lenum.unbind = Completely disable logic control.\nResume standard AI.
+lenum.decontrol = Completely disable logic control.\nResume standard AI.
```

PR #12500, opened 2026-08-16
Fix setblock placement for even-sized multiblock buildings

### LINKED ISSUE #12493: Logic setblock instruction places 2x2 buildings incorrectly
### Platforms

Android

### Build

159.7

### Issue

On older versions the 2x2 block was placed in the exact same position that sensor returns, on the current version the block is placed 1 tile higher and to the right

### Steps to reproduce

Attempt to replace a 2x2 block using logic on a v7 build and v8 build 159

### Mods used

_No response_

### Save file

_No response_

### (Crash) logs

_No response_

### Submission

- [x] I have updated to the latest release (https://github.com/Anuken/Mindustry/releases) to make sure my issue has not been fixed.
- [x] I have searched the closed and open issues to make sure that this problem has not already been reported.
- [x] I am not using Foo's Client, and have made sure the bug is not caused by mods I have installed.

atirna:
## Description

`setblock` was using `Mathf.round` on logic coordinates directly, which lands one tile off when placing even-sized blocks from sensor `@x`/`@y` (those coords include the block draw offset, so a 2x2 anchor at tile 5 reads as 5.5).

I changed block-layer tile lookup to match manual placement and `ucontrol build`: subtract the target block's offset, then `World.toTile`. Floor/ore layers keep the old rounding path.

Fixes #12493

## Why

Replacing a 2x2 building via logic using sensed coordinates should anchor at the same tile the sensor came from, not one tile up/right.

## Verification

- before on `master`: `./gradlew tests:test --tests ApplicationTests.setBlockMultiblockSensorCoords` fails (`expected: <container> but was: <air>` on `setblock block @container 5.5 5.5`)
- after: same test passes (also checks 5.6 coords and a 3x3 vault case)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -1584,8 +1584,16 @@ public class LExecutor{
         public void run(LExecutor exec){
             if(net.client()) return;
 
-            Tile tile = world.tile(Mathf.round(x.numf()), Mathf.round(y.numf()));
-            if(tile != null && block.obj() instanceof Block b){
+            if(!(block.obj() instanceof Block b)) return;
+
+            Tile tile;
+            if(layer == TileLayer.block){
+                tile = world.tile(World.toTile(World.unconv(x.numf()) - b.offset), World.toTile(World.unconv(y.numf()) - b.offset));
+            }else{
+                tile = world.tile(Mathf.round(x.numf()), Mathf.round(y.numf()));
+            }
+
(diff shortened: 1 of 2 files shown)


7amithorn:
... did you write this, or did you have an AI write this?

atirna:
I wrote the code & made the submission, but had ai review and draft the pr, it supposedly added the made with Cursor tag in the draft.

PR #12519, opened 2026-08-21
Replace random+CRC32 UUID with proper time-based UUIDv1

dsdsfadsfsadf:
## Summary
Player identifiers were previously 8 random bytes padded with a CRC32 checksum of those same bytes (unchecked, unused) to reach the 16-byte wire size - giving only 64 bits of real entropy. This replaces that scheme with a proper RFC 4122 version 1 (time-based) UUID, using the machine's MAC address as the node ID.

## Security
- **MAC-based node ID** makes the identifier tied to actual hardware, making it harder to spin up new identities to dodge bans compared to a purely random value.
- **Removed the Steam-account-derived UUID** (`Rand(steamAccountId).nextBytes(...)`). Steam account IDs are often publicly visible, so this was a deterministic, guessable mapping from a public SteamID to a player's in-game uuid - anyone could compute another player's identifier from their SteamID. Steam clients now use the same random UUIDv1 generation as everyone else.
- Dropped the unchecked CRC32 padding, which added wire bytes without any real verification.

## Compatibility
Wire format size is unchanged (still 16 bytes in `ConnectPacket`), so client and server from this fork remain fully compatible with each other.

### DIFF
--- a/core/src/mindustry/core/Platform.java
+++ b/core/src/mindustry/core/Platform.java
@@ -2,7 +2,6 @@ package mindustry.core;
 
 import arc.*;
 import arc.files.*;
-import arc.math.*;
 import arc.struct.*;
 import arc.util.serialization.*;
 import mindustry.mod.*;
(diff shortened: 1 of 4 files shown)


Anuken:
This is clearly AI slop. However, I'll comment on it anyway:

- This changes existing UUIDs for everyone. Not acceptable.
- MAC address-based UUIDs were tried before, and rejected, because many are the same/contain garbage data.

> MAC-based node ID makes the identifier tied to actual hardware, making it harder to spin up new identities to dodge bans compared to a purely random value.

What on earth is the point of using the MAC address 'to tie it to actual hardware' if you're mixing `Time.millis()` into the result? That removes any real hardware ties, unless you expect servers to perform some kind of bit surgery to extract the (useless, possibly duplicate) MAC address from UUIDs and implement their own ban system, which is absurd.

PR #12529, opened 2026-08-25
remove pulsar and quasar's ability to mine (pr this time)

spentcud:
original suggestion, also made by me: https://github.com/Anuken/Mindustry-Suggestions/issues/6496

the nova line's ability to mine is such a random ability for them to have; their main role (or at least, their _intended_ main role) is supporting units at the front line, and being able to mine has absolutely nothing to do with that
doesnt help that umining happens to be one of the most broken mechanics in the game, so these units almost always end up as mining slaves instead of offensive support units

this change will _not_ drastically affect game balancing up until you hit the unit cap since i didnt change the mono tree at all, so you can just fall back on poly and mega if youre going for umining

### DIFF
--- a/core/src/mindustry/content/UnitTypes.java
+++ b/core/src/mindustry/content/UnitTypes.java
@@ -394,8 +394,6 @@ public class UnitTypes{
             armor = 4f;
             riseSpeed = descentSpeed = 0.07f;
 
-            mineTier = 2;
-            mineSpeed = 3f;
 
             abilities.add(new ShieldRegenFieldAbility(20f, 40f, 60f * 5, 60f));
 
@@ -440,7 +438,6 @@ public class UnitTypes{
         }};
 
         quasar = new UnitType("quasar"){{
-            mineTier = 3;
             boostMultiplier = 2f;
             health = 640f;
             buildSpeed = 1.1f;
@@ -457,7 +454,6 @@ public class UnitTypes{
             speed = 0.5f;
             hitSize = 13f;
 
-            mineSpeed = 4f;
             drawShields = false;
 
             abilities.add(new ForceFieldAbility(60f, 0.4f, 500f, 60f * 6));


1ue999:
we love not checking the checks

1ue999:
also, this will likely annoy a *lot* of people, since many use the nova tree to mine.

bill88t:
Green units mine. Taking mining away from Vela already was kinda egregious.
It's not like they're free. T3s are relatively expensive at a scale, and if you can afford them, they should have a use even after capturing the sector.

Also, with how many tight sectors there are in V8, umining is more important than ever.

50275:
All this does is eliminate potential, not balance. Who does this benefit? It's not the player's fault pulsar and quasar are bad supporting units compared to poly and mega, especially not after their nerf: https://github.com/Anuken/Mindustry/commit/f5a0b5ec1a988348cb40177d60cfa7f8c72a027a. With this, they'll be completely obsolete. This change only targets the fraction of players who actually choose to opt-in to unit mining, especially when unit caps and available resources are low, such as in singleplayer.

MinDoUser:
While I do get your point, I don‘t think straight up removing this mechanic is the right way to go about this. Mindustry is a sandbox game afterall, where anyone can basically do what they want and see how far it goes. If your choice is to not use nova tree for mining, that is a valid choice of yours, but this doesn‘t mean it is the best choice for everyone.
However, if unit mining is actually that OP, there should probably be some balancing happening, so that no matter what you do, whether it be using nova tree to mine, support or defend, it should be equally good. (since nova tree is <i>meant</i> to be used as supportive units, they should probably be a bit better at that tho)

Anuken:
As much as I don't like how they're used now, a complete removal of the mechanic would just cause backlash and make the game less interesting. A nerf with a corresponding combat buff might work.

50275:
No nerf is necessary. They're already much worse miners and builders. No combat buff to pulsar and quasar will replace the support role of the mono tree.

PR #12533, opened 2026-08-26
Add nuclear power server

ziyikong:
Add nuclear power server (47.103.45.169) to the server list.


JasonP01:
SERVERLIST.md....

camelStyleUser:
> # Note: The v7 server list is frozen. No new servers will be accepted. All v8 server PRs should be made [here](https://github.com/Anuken/MindustryServerList).

from SERVERLIST.md
nonetheless the server is not even v7 so adding it to servers_v7 makes no sense at all

PR #12534, opened 2026-08-26
Multiple liquid boosters for drills

EggleEgg:
### DO NOT MERGE UNTIL CODE REVISION PLEASE

Implements multiple optional liquid boosters for drills. While it works, I really dont like this implementation as it is not really a refactor but adding new code to something that _really_  needs refactoring. It is just about the furthest thing away from:
- https://github.com/Anuken/Mindustry-Suggestions/issues/5989

.

I truly have no idea how to do this better (thus the revision needed, some hints at what should be refactored, etc)

- https://github.com/Anuken/Mindustry-Suggestions/issues/6478

image

.

image
image


MinDoUser:
As I already said on the original suggestion, having cryofluid boost drills even more than water seems a little too OP… but besides this, I like the idea

EggleEgg:
@MinDoUser the balancing for this will come in a separate pr (or removed its implementation in blocks.java if anuke or more users wish so). This is not done and never will be done in the same pr to avoid having large amounts of internal logic be denied because of gameplay or balancing concerns like it has happened in the past

jehosula:
> As I already said on the original suggestion, having cryofluid boost drills even more than water seems a little too OP… but besides this, I like the idea

the numbers are for demonstration purposes only
the goal is to add the possibility to include multiple liquid boosters for drills so that airblast and possibly laser drills can use cryofluid for a more difficult but more rewarding alternative to water over the current place a water extractor and call it a day

PR #12538, opened 2026-08-27
Add integer configs to unit build control

BensonLing0925:

Problem: Currently ucontrol build instruction does not support Integer config even if some of the blocks already have Integer configuration handlers.  
Changes: Default ucontrol build config value set to "null" instead of "0". Numeric p5 values are read using p5.numi() and forwarded as Integer configs.  
Compatibility: Old 4-operand ucontrol build instruction is compatible, Old 5-operand with numeric config may observe different behavior on blocks with Integer configuration handlers.  
Testing: Tested with Phase Conveyor, Power Node, Mass Driver, Unit Factory and Illuminator. Also verified existing Content and Building configs still work.

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -456,8 +456,11 @@ public class LExecutor{
                             }
 
                             var conf = p5.obj();
+                            if(!p5.isobj){
+                                conf = p5.numi();
+                            }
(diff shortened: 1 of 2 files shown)


Anuken:
Integer configs **should not** be supported. Those are packed positions, which don't have a good representation in logic. It's also inconsistent with `control config`.

PR #12561, opened 2026-08-31
Fix the sending of unnecessary UDP packets before the world has fully loaded.

hack-banana:

## Description
Changed `NetServer` snapshot sending to send packets directly to individual `NetConnection`s instead of using global `Call` broadcasts.

This prevents snapshot packets from being sent to clients that have not loaded the world yet.


WMF-Industries:
the quality of this 'fix' is very low...
spamming enhanced for loops that iterate over Sequences is a bad practice since it involves allocating an iterator each time the loop runs
the formatting is also quite bad

Anuken:
Looks like AI slop. Solves a problem that doesn't exist, and makes server perf worse by skipping the bulk-serialization system.

PR #12564, opened 2026-08-31
Sandbox target dummy

EggleEgg:
Allows players to easily try out their turrets/units without having to rely on wproc logic, mods, spreadsheets or other external tools

From https://github.com/MEEPofFaith/extra-sand-redux with meep s permission

image

image

.


core/src/mindustry/world/blocks/defense/TargetDummy.java
```diff
@@ -0,0 +1,299 @@
...
+            tile.resetTime = v.y;
+        });
+        config(Float.class, (TargetDummyBuild tile, Float f) -> tile.dummySize = f);
+        config(int[].class, (TargetDummyBuild tile, int[] config) -> {
+            tile.unitTeam = tile.team;
+            if(config[0] == 1) tile.unitTeam = Team.get(tile.dummyTeam());
+            tile.boosting = config[1] == 1;
+            tile.unitArmor = Float.intBitsToFloat(config[2]);
+            tile.resetTime = Float.intBitsToFloat(config[3]);
+            tile.dummySize = Float.intBitsToFloat(config[4]);
```
JasonP01: We love an IndexOutOfBoundsException if this ever doesnt contain those indexs for some reason

core/src/mindustry/world/blocks/defense/TargetDummy.java
```diff
@@ -0,0 +1,299 @@
...
+            hits++;
+        }
+
+        @Override
+        public void damage(float damage){
+            //just in case
+        }
+
+        @Override
+        public void kill(){
```
JasonP01: Few plugins/scripts use kill() to remove a block as its easier, nothing calls kill() in vanilla if it cant be damage(), id re-enable kill() for easier removal of the block if someone is trying to mass clear some blocks with a script

JasonP01: This is like very not needed if you dont want it sincs `Tile.setNet(Blocks.air)` exists

core/src/mindustry/world/blocks/defense/TargetDummy.java
```diff
@@ -0,0 +1,299 @@
...
+                t.add(Core.bundle.get("stat.hitsize"));
+                t.field("" + dummySize, TextFieldFilter.floatsOnly, s -> configure(Strings.parseFloat(s))).padLeft(8f).growX();
+                t.add(StatUnit.worldUnits.localized()).padLeft(8);
+            }).top().grow().margin(8f);
+        }
+
+        public int dummyTeam(){
+            if(unitTeam != team) return team.id; //Return to own team
+
+            if(team == state.rules.defaultTeam) return state.rules.waveTeam.id; //Set to wave team if player team
```
JasonP01:
```suggestion
            else if(team == state.rules.defaultTeam) return state.rules.waveTeam.id; //Set to wave team if player team
```
Why cant it be the same team as the player? What if i want it to be my team

Okay scratch that, all calls to this function are fucked. There are no else branches it will continue going back and forth between 20 different options each call, the only call you need is the first one, the unit team should match the block's team

JasonP01 (changes requested):

core/src/mindustry/world/blocks/defense/TargetDummy.java
EggleEgg: you still need 3 different returns (1 unitTeam state, then 2 depending if block team is sharded or not)
actually wave team can be the same as player team.... hmm, then meep's fallbacks would be correct (kinda)

then you need 5 different returns

JasonP01: fixed in mah pr, resolve this

Anuken:
Merged in b3e8aa9294e90954ada9983fff6dda1640e40b00

MEEPofFaith:
completely forgot about this lmao

PR #12567, opened 2026-09-01
Add configPos unit control

BensonLing0925:

Adds ucontrol configPos x y tx ty, allowing units to configure nearby buildings with a target position using normal logic coordinates.

This is a different approach to #12538. No internal information exposed to user. The target coordinates are converted internally to the existing Point2 configuration used by compatible blocks. Blocks without a Point2 configuration handler are left unaffected.

Tested in-game with phase conveyors/conduits, bridge conveyors/conduits, and mass drivers. Also tested with a power node to verify that unsupported configuration types are not affected.


Anuken:
Absolutely not. Player logic should never be able to configure positional blocks under any circumstances; this is a massive buff to logic with the tiny cost of *any* unit being parked next to a schematic.

PR #12576, opened 2026-09-02
Before calculating the amount of liquid to transfer, first check if input of that liquid is allowed.

NeilGreenFly:

Before calculating the amount of liquid to transfer, first check if input of that liquid is allowed. The acceptLiquid check from transferLiquid has been moved into dumpLiquid. Since transferLiquid is hardly used elsewhere, this usually doesn't have side effects.


Anuken:
Why?

NeilGreenFly:
Because when I’m personally writing a mod, there’s a building that needs to use multiple liquid models, but I found that when a liquid source outputs liquid, it first calculates the amount to transfer and then checks if it can transfer. At that point, any preprocessing I do in acceptLiquid can’t affect the final transfer amount, so I hope to move this check earlier.

NeilGreenFly:
This is just my personal need, so if it's really unnecessary, we can skip this PR.

Anuken:
This might break mods that expect transferLiquid to do acceptLiquid checks. You should override the methods if you need this.

NeilGreenFly:
Okay

PR #12584, opened 2026-09-03
Update bundle_zh_CN

redloong9527:
And...
Anuke, why are "animatedshields" not included in the bundle? I don't find it.


Anuken:
Animated shields aren't a valid option anymore.

PR #12589, opened 2026-09-03
Allow selecting a version when reinstalling a mod

cardillan:
When reinstalling a loaded mod, the latest version is always installed. To downgrade a mod, it needs to be located in the mod browser, which then allows to select the desired version to install. That's a bit cumbersome.

This PR changes the process to offer a list of versions to select from even when reinstalling a loaded mod (Settings -> Mods -> click on a loaded mod -> Reinstall).


core/src/mindustry/ui/dialogs/ModsDialog.java
```diff
@@ -523,6 +529,62 @@ public void showModError(Throwable error){
         }
     }
 
+    public void viewReleases(String repo, boolean isJava) {
```
Anuken: This will not work for mods that have no releases, as is the case for many JSON mods.

cardillan: Is it true that releases are compulsory for all Java mods and optional for all JSON mods?

If it is the case, perhaps I could add an entry at the top of the list of releases for JSON mods that would say "Latest", and this would use the current method of mod reinstall. Would this be acceptable?

redloong9527:
I don't know if the specific code implementation is good or not, but if the effect is really as described, it seems pretty good.

core/src/mindustry/ui/dialogs/ModsDialog.java
DeterMination-Wind: why don't you just compare the version number , and what about the areas that can't access github?

cardillan: The goal is to enable the user to select a specific version to install, e.g. to downgrade a mod. The game already supports downloading and displaying the list of releases and installing a specific one, but it is only available from the mod browser. I'm trying to make it also available when viewing the details of an installed mod.

This entire functionality is about downloading a version of the mod from github. If you can't access github, you can still install a mod from a local file, no change here.

Anuken: Releases are mandatory for Java/JVM mods, and not required for JSON/JS mods.

If a mod has no releases and is a JSON/JS mod, it should not show any releases at all, and just reinstall the whole thing from the latest commit as before.

cardillan: When the mod has no releases and it is a JSON mod, it will reinstall normally. When it is a Java mod, an error will be shown, as in the mod browser. No attempt at installing the mod will be made.

Screen for a JSON mod with a release:
image

core/src/mindustry/ui/dialogs/ModsDialog.java
```diff
@@ -523,6 +529,79 @@ public void showModError(Throwable error){
...
+                                }).right();
+                            }).margin(5f).growX().pad(5f);
+
+                            if(j < releases.size - 1) p.row();
+                        }
+                    }).width(500f).scrollX(false).fillY();
+                    downloads.buttons.button("@back", Icon.left, () -> {
+                        downloads.clear();
+                        downloads.hide();
+                        //sel.show();
```
Anuken: Why are these (and other lines) commented out instead of deleted?

cardillan: That's an oversight, sorry. Shouldn't be there. I need to review it once more to see whether the changes affected the mod browser itself somehow.

cardillan: That was code which temporarily hides the mod details screen while fetching and showing the list of releases. On my system, this hiding and showing of the screen caused a bit of a flicker, which wasn't very pleasant. I've removed it.

In the mod browser, the mod description may be quite prominent and colorful. Perhaps that was the reason to hide the screen, but I haven't found it too distracting when the release list was displayed on top of that.

Also, in the mod detail screen, the mod description isn't word wrapped, so only a part of the first line is actually visible. It would take me some time to figure out how to fix that, so I'll try to do it some other time. THis should be now ready to merge.

PR #12590, opened 2026-09-03
Write to multiple adjacent elements at once

cardillan:
This PR adds a `length` parameter to the `write` instruction. It specifies the number of elements to write to. The primary goal is to allow clearing memory cells faster, e.g. `write 0 cell1 0 64` clears the entire memory cell in a single instruction. The default value of `length` is `1`, meaning that old code works the same as before.

The `LWritable.write()` method has been updated to accept the `length` parameter. The `CanvasBlock` and `MemoryBlock` implementations limit the length to the maximum supported value, so that providing an absurd number, such as `1e20`, doesn't hang the processor in a loop. The `LogicBlock` implementation ignores the parameter entirely, as the `address` is a string and `length` doesn't make sense in that context.

`MemoryBlock` uses `Arrays.fill()` to set the arrays. The `CanvasBlock` uses loops; it could be optimized and I'll do it if required, but I didn't want to risk introducing bugs if not necessary.

### DIFF
--- a/core/src/mindustry/logic/LExecutor.java
+++ b/core/src/mindustry/logic/LExecutor.java
@@ -645,12 +645,13 @@ public class LExecutor{
     }
 
     public static class WriteI implements LInstruction{
-        public LVar target, position, value;
+        public LVar target, position, value, length;
 
-        public WriteI(LVar target, LVar position, LVar value){
+        public WriteI(LVar target, LVar position, LVar value, LVar length){
             this.target = target;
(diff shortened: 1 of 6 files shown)


1ue999:
when memcpy

Anuken:
Sorry, no, this is not what `write` is supposed to do, and not something logic should be doing. It's a (potentially very large) O(n) operation for something doesn't need to be (unlike printing).

PR #12607, opened 2026-09-06
update Allure to an archive public repo

JasonP01:
Since the Allure mod has been archived, there is no good json test. This replaces the fetch with an "archived" version of allure before it went private. Permission to do this has been gotten from LixieWulf, and an invite as collaborator to the repository has been given to Anuken for him to do as he wishes
image

This archive is solely so the test can run, i will not be maintaining it, and nobody should attempt to play this mod as it has been removed of all assets as per request. Check the repo https://github.com/JasonP01/AllureMod README.md for slightly more information regarding that


JasonP01:
The test error has been fixed by porting the fix from an updated version of allure

PR #12645, opened 2026-09-12
Silicon crucible duplicated enhancements issue

Cry0flu1d:
likely caused by #11552 
silicon crucible's efficiency was affected by both output rate and production time, making it (167.5%)^2 more effective instead of 167.5% in 159.7 does.(take hot rock tiles for example)
it might be an unexpected buff for silicon crucibles, making it extremely op in hot places.

### DIFF
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1092,7 +1092,6 @@ public class Blocks{
             hasPower = true;
             hasLiquids = false;
             itemCapacity = 30;
-            boostScale = 0.15f;
             outputScale = 0.15f;
             drawer = new DrawMulti(new DrawDefault(), new DrawFlame(Color.valueOf("ffef99")));
             ambientSound = Sounds.loopSmelter;
--- a/core/src/mindustry/content/Blocks.java
+++ b/core/src/mindustry/content/Blocks.java
@@ -1092,6 +1092,7 @@ public class Blocks{
             hasPower = true;
             hasLiquids = false;
             itemCapacity = 30;
+            boostScale = 0f;
             outputScale = 0.15f;
             drawer = new DrawMulti(new DrawDefault(), new DrawFlame(Color.valueOf("ffef99")));
             ambientSound = Sounds.loopSmelter;


SunkyMP3G:
Intentional

> The vanilla speed magma boost is notoriously bad, as it provides no real advantage, to the point placing thermal generators is objectively the better option

If it's like so broken to the point of being a win condition, probably should just decrease the values

coaldeficit:
this pr removes the boostScale from the silicon crucible, leading it to be set to the default value of 1 instead of 0.15, which actually significantly buffs the silicon crucible

it is very clear the person who made this doesnt know what they're doing and just seems to have an issue with silicon crucible boosting being usable now

Cry0flu1d:
> this pr removes the boostScale from the silicon crucible, leading it to be set to the default value of 1 instead of 0.15, which actually significantly buffs the silicon crucible
> 
> it is very clear the person who made this doesnt know what they're doing and just seems to have an issue with silicon crucible boosting being usable now

i didnt notice the default value sorry

AngryHungryGoose:
Efficiency is not affected by output rate. Silicon Crucible is 167.5% more efficient and 167.5% faster on Hot Rock - these are two separate metrics, which in total give ~280.6% output. The provided reasoning is fundamentaly flawed, however, I do not see an issue with the commit itself.

The efficiency bonus is extremely impactful by itself (since it's essentially a free resource multiplier on top of the existing advantage of the Silicon Crucible over the Silicon Smelter) and a static consumption rate would make ratios consistent (they can currently get very weird with varied tiles underneath).

SomeonesShade:
> Describe what you would like changed, and why
> Describe the changes you want to propose. Include possible alternatives.

I may as well treat this as a suggestion post... (I'll be honest, this type of stuff should've been a s.p. then a p.r. with proof of proper support... but whatever)

Guessing for the first part... You want to make it less strong because it's too strong or op? (It's a bit vague, I would like details on why its problematic...)
For the second... it seems like the attempted change is to remove the speed boost...

I'd like at least an explanation on why its problematic, so that the appropriate nerf **_if proven to be necessary_** can be deduced...

I would give more appropriate questions once a proper reasoning is provided...

- Shade

SunkyMP3G:
You should go to #serpulo-balancing on discord server. There will be a lot more people to discuss with

PR #12658, opened 2026-09-14
Fix cluster array dimension based on costTypes size

sora54:
### Description

Fix an incorrect array dimension in the cluster initialization.

The array was previously initialized using `Team.all.length`, but the first dimension should match the number of cost types.

### Changes

Changed:

```java
dim1 = clusters[team] = new Cluster[Team.all.length][];
```

To:

```java
dim1 = clusters[team] = new Cluster[costTypes.size][];
```

This ensures the cluster array is initialized with the correct number of cost types.
This is a minimal fix with no other changes.


EggleEgg:
🤖🫵

PR #12667, opened 2026-09-15
Deduplicate requested assets during handshake

acts-1631:
Client asset requests allowed the same valid asset ID to appear multiple times. The server appended each occurrence to its response and reread the asset from disk, allowing a client to multiply memory and serialization work during the multiplayer handshake.

Track requested IDs while validating the request so each asset is serialized at most once. Normal requests for multiple distinct assets are unchanged.


WMF-Industries:
the **IntSet** could probably be a **Bits** collection instead

PR #12678, opened 2026-09-17
Improve hex and binary literal parses

cardillan:
This PR requires https://github.com/Anuken/Arc/pull/217 (so it won't compile now).

It allows **all** possible long values to be parsed in binary or hex format (`0b`/`0x`) - it handles zeroes separately and uses it to indicate an invalid value later on - and also refuses invalid numbers which currently do parse (such as `0x+abcd` or `0b-0101`). A small optimization to refuse parsing strings starting with `@` (logic built-ins) was also added.


cardillan:
I've moved the unsigned long parsing to `LAssembler`. This avoids the hassle with inventing a value for invalid numbers, as I can use `Double.NaN` directly at that point.

Improved the detection of obvious non-numbers, so that well-formed identifiers and logic constants skip the parsing entirely.

